PR #198, opened 2019-11-18
Show tickets

umbernhard:
Displays consistent sampler ticket numbers in reports.


umbernhard:
Note that I reconfigured our data model. I realized that the output of the consistent sampler actually lends itself much better to putting the ballot into the database each time it is sampled, so we don't need a list of the ticket numbers for the SampledBallot db entries. Sorry for making @eventualbuddha do more work than necessary :grimacing:

tests/test_app.py
```diff
@@ -618,7 +618,8 @@ def test_multi_round_audit(client):
     # round 2 retrieval list should be ready
     rv = client.get('{}/jurisdiction/{}/2/retrieval-list'.format(url_prefix, jurisdiction_id))
     lines = rv.data.decode('utf-8').splitlines()
-    num_ballots = sum([int(line.split(",")[4]) for line in lines[1:] if line!=""])
+    #num_ballots = sum([int(line.split(",")[4]) for line in lines[1:] if line!=""])
```
eventualbuddha:
```suggestion
```

sampler.py
```diff
@@ -36,7 +36,6 @@ def __init__(self, audit_type, seed, risk_limit, contests):
         Outputs:
         """
         self.seed = seed
-        self.prng = SHA256(seed)
```
eventualbuddha: The import for `SHA256` isn't used now either, so delete it?

models.py
```diff
@@ -129,15 +129,16 @@ class SampledBallot(db.Model):
     comment = db.Column(db.Text, nullable=True)
 
     # comma-separated ticket numbers
-    ticket_numbers = db.Column(db.Text, nullable=True)
+    # ticket_numbers = db.Column(db.Text, nullable=True)
```
eventualbuddha:
```suggestion
```

models.py
```diff
@@ -129,15 +129,16 @@ class SampledBallot(db.Model):
...
+   # @property
+   # def tickets(self):
+   #     return self.ticket_numbers.split(',') if self.ticket_numbers else []
 
-    @tickets.setter
-    def tickets(self, tickets: Union[List[str], None]):
-        self.ticket_numbers = ','.join(tickets) if tickets else None
+   # @tickets.setter
+   # def tickets(self, tickets: Union[List[str], None]):
+   #     self.ticket_numbers = ','.join(tickets) if tickets else None
```
eventualbuddha: Delete commented out code?

models.py
```diff
@@ -120,7 +120,7 @@ class SampledBallot(db.Model):
     ballot_position = db.Column(db.Integer, nullable=False)
     
     __table_args__ = (
-        db.PrimaryKeyConstraint('round_id', 'jurisdiction_id', 'batch_id', 'ballot_position'),
+        db.PrimaryKeyConstraint('round_id', 'jurisdiction_id', 'batch_id', 'ballot_position', 'ticket_number'),
```
eventualbuddha: Do you anticipate that there will be `SampledBallot` records with identical `round_id`, `jurisdiction_id`, `batch_id`, and `ballot_position` that only vary by `ticket_number`? If so, I think we'll need to update this endpoint: `/election/<election_id>/jurisdiction/<jurisdiction_id>/batch/<batch_id>/round/<round_id>/ballot/<ballot_position>`.

umbernhard: Yeah, I think that is quite likely.

eventualbuddha: @MorganLove do you have a sense of how this would need to change?

mcchilders: @benadida can you weigh in on the above? we will definitely need to support selected ballots that differ only by ticket number

benadida: Ah, I see `API.md` is not up to date. It would be good if we kept it up to date.

As best as I can tell,  `../round/<round_id>/ballot/<ballot_position>` is only for setting a ballot, not fo fetching one, which I think is exactly the right thing, since I don't see any reason to fetch ballots other than by round with `../ballot-list`. The ballot list API call should return all the sampled ballots, with each entry representing a particular ballot position, each time it's sampled, each with its ticket number. When you set the ballot, you only to do that once for that batch ID and position. So I don't think the API needs to change.

I *would* suggest changing the API for setting the ballot from

`POST ../batch/<batch_id>/round/<round_id>/ballot/<ballot_position>`

to

`POST ../batch/<batch_id>/ballot/<ballot_position>`

since the content of a ballot does not depend on the round in which it's selected.

app.py
```diff
@@ -180,23 +180,22 @@ def sample_ballots(election, round):
...
-        if lookup in seen_ballot_positions:
-            batches_to_ballots[batch_id].append(ballot_position)
-            continue
 
-        seen_ballot_positions.add(lookup)
         if batch_id in batch_sizes:
-            batch_sizes[batch_id] += 1
-            batches_to_ballots[batch_id].append(ballot_position)
+            if times_sampled == 1: # since we've already seen it, it doesn't affect batch size
+                batch_sizes[batch_id] += 1
```
benadida: since we're now recording each time the ballot is sampled as separate row, shouldn't we kill the `times_sampled` field?

umbernhard: This is just for bookkeeping for balanced retrieval lists. I could modify that to compute the batch sizes, but it'll take a little more re-engineering.

benadida: I'm worried the SQL data model can become inconsistent, e.g. what if `times_sampled` says 3, but there are actually 4 matching `SampledBallot` rows? Also, are we updating all the `times_sampled` fields for all the matching sampled ballots?

models.py
eventualbuddha: > Ah, I see API.md is not up to date. It would be good if we kept it up to date.

*Gazes longingly at tools that generate such documentation automatically.*

app.py
umbernhard: This field is never used outside of this function, so I'm not really sure how inconsistency would factor into it? This is just a back-of-the-envelope to figure out how many pieces of paper are in each batch so we can balance workload. All we need to know is that the ballot has been already sampled.

umbernhard: @benadida if you'd rather me read a value out of the DB for that, then I can, but that seems like a lot of overhead for a thing the sample is already giving us anyway.

models.py
```diff
@@ -118,26 +118,18 @@ class SampledBallot(db.Model):
...
     ballot_position = db.Column(db.Integer, nullable=False)
+
+    ticket_number = db.Column(db.String(200), nullable=False)
     
     __table_args__ = (
-        db.PrimaryKeyConstraint('round_id', 'jurisdiction_id', 'batch_id', 'ballot_position'),
+        db.PrimaryKeyConstraint('round_id', 'jurisdiction_id', 'batch_id', 'ballot_position', 'ticket_number'),
     )
     
     times_sampled = db.Column(db.Integer, nullable=False)
```
benadida: @umbernhard this is the field I'm worried about. Do we still need it in the DB now that we have a row per sample?

umbernhard: oooohhhh. No, I think we can remove it. Right @eventualbuddha?

umbernhard: Oh I see what's happening. We need to keep track of the total number of times a ballot was sampled so we can tell the Sampler where to pick up its sample. So removing this field breaks that.

benadida: right, so that's where the inconsistency could sneak in. I think we're in agreement that, with this new model, the total number of times a ballot was sampled is the number of rows with a given `(jurisdiction_id, batch_id, position)`?

umbernhard: Yeah, that's right.

benadida: @umbernhard so it seems to me that we should remove this field and make sure that, when we need it, we select it from the database. I can take a look if you prefer?

umbernhard: Yes please! Sorry, I thought we’d already talked about it but I must be misremembering.

mcchilders:
@benadida we need to make sure that the retrieval list only shows each ballot one time in the retrieval list (if a ballot is sampled multiple times in the audit round, the list should include the multiple ticket numbers comma-separated in the column, rather than creating a separate row for each time the ballot is sampled.)

benadida:
This PR got a bit stale, so I rebased and cleaned up in #252

PR #202, opened 2019-11-20
refactor(db): exclusively use postgres

### LINKED ISSUE #199: Remove sqlite in favor of only postgres
They don't act entirely the same way. Let's just standardize on one.

eventualbuddha:
Using sqlite for development and testing is easy, but it differs from postgres in ways that make troubleshooting staging and production issues difficult. This commit switches everything from sqlite to postgres in all environments.

The difficulty I had in making this change was primarily around ensuring we're able to easily provide the application with the right configuration in whatever environment it happens to be running in. For example, we need to continue to respect the `DATABASE_URL` environment variable because that's what heroku is using. However, asking people to set environment variables in their shell or IDE to run Arlo properly is not great.

To address this, I set up a system that vaguely resembles Rails with a `config/database.cfg` that is ignored by git but can have per-`FLASK_ENV` database configuration.

Finally, because tests now run against a persisted database we have the possibility of primary key collisions. Since there were lots of primary keys hardcoded in the tests, this happened quite a bit. I fixed this by ensuring that primary keys are always generated randomly, but really we should be using something like https://github.com/jeancochrane/pytest-flask-sqlalchemy to automatically roll back sessions for each test. This was more difficult to set up than I expected, so I'm leaving it as a future task.

Closes #199

### DIFF
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -1,0 +1,27 @@
+import configparser
+import os
+import sys
+
+DEFAULT_DATABASE_URL = 'postgres://postgres@localhost:5432/arlo'
+
+def read_database_url_config() -> str:
+  environment_database_url = os.environ.get('DATABASE_URL', None)
+
+  if environment_database_url:
+    return environment_database_url
+
+  database_cfg_path = os.path.normpath(os.path.join(__file__, '..', 'database.cfg'))
+  database_config = configparser.ConfigParser()
+  database_config.read(database_cfg_path)
+
+  flask_env = os.environ.get('FLASK_ENV', 'development')
+  result = database_config.get(flask_env, 'database_url', fallback=None)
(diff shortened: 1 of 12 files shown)


eventualbuddha:
Build failure is on master, fixed by #203.

config/__init__.py
```diff
@@ -0,0 +1,28 @@
+from typing import Optional
+import configparser
+import os
+import sys
+
+DEFAULT_DATABASE_URL = 'postgres://postgres@localhost:5432/arlo'
+
+def read_database_url_config() -> Optional[str]:
```
benadida: look at those type annotations!

eventualbuddha: And they’re wrong! Need to fix …

PR #204, opened 2019-11-20
Add Docker infrastructure

eventualbuddha:
This commit adds the ability to develop Arlo using Docker as an alternative approach. This does not replace the existing approach, just makes a new one available.

The main advantages are reproducibility and isolation of development environment, ease of installing dependencies regardless of the host system, and ease of orchestrating development processes. The main disadvantages are adding indirection to development systems and adding reliance on yet more software (Docker).

If we're not opposed to this in principle, I'd appreciate if you could try it out and let me know what you think.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.git
+node_modules
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM python:3.7.3
+RUN pip install --upgrade pip
+RUN pip install pipenv
+RUN mkdir /code
+WORKDIR /code
+COPY Pipfile Pipfile.lock /code/
+RUN pipenv install --dev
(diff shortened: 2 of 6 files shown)


benadida:
I've gotten some errors on first try (no Arlo database), which left the system in a bad state, but on second `docker-compose up`, it seems to work. I suspect it's because on subsequent tries, the database is maintained.

Development seems to work well. Looks good!

eventualbuddha:
Anyone actually get a chance to try this out?

benadida:
> Anyone actually get a chance to try this out?

Did you see my note above? I tried it, and it worked well, except errors out the first `docker-compose up` time around.

eventualbuddha:
Thanks @benadida. I found some issues that should now be fixed on this branch. It should now work even if you `git clean -dfx` at the root of the repo and `docker-compose down --volumes` to destroy everything first.

MorganLove:
@eventualbuddha Trying to run this, and it works up until it tries to build `client-tests`. It gives me the error `ERROR: Cannot locate specified Dockerfile: Dockerfile`.

eventualbuddha:
> @eventualbuddha Trying to run this, and it works up until it tries to build `client-tests`. It gives me the error `ERROR: Cannot locate specified Dockerfile: Dockerfile`.

Whoops, good catch @MorganLove. Should be fixed by https://github.com/votingworks/arlo/pull/225.

PR #207, opened 2019-11-21
feat: add "Votes" field for contests

### LINKED ISSUE #189: Add a "vote for" parameter to the contest data model to facilitate a correct validation of the total number of ballot cards cast.

eventualbuddha:
This is not really a complete feature in and of itself. This commit allows targeted contests to have a number of votes allowed but does not do anything with that number. It is included alongside "Number of Winners" in the audit report, but none of the actual audit math changes based on it.

The changes to actually use this data will have to come from someone else, likely @umbernhard.

Closes #189


app.py
```diff
@@ -755,6 +758,7 @@ def audit_report(election_id):
     
     report_writer.writerow(["Contest Name", contest.name])
     report_writer.writerow(["Number of Winners", contest.num_winners])
+    report_writer.writerow(["Number of Votes", contest.votes_allowed])
```
umbernhard: I think we should change the CSV header to also be "Votes Allowed", for consistency.

eventualbuddha:
```suggestion
    report_writer.writerow(["Votes Allowed", contest.votes_allowed])
```

umbernhard:
Also @eventualbuddha, the only reason we need this feature right now is so we can do frontend validation of ballots cast correctly for vote-for-more-than-one, as stated in #189. 

Right now, the validation is essentially `assert totalBallots > sum(votes_for_candidates)`, but it needs to be `assert totalBallots*voteAllowed > sum(votes_for_candidates)`. 

Pardon the python speak, I'm not sure how to express that in typescript :sweat_smile:

eventualbuddha:
> Right now, the validation is essentially `assert totalBallots > sum(votes_for_candidates)`, but it needs to be `assert totalBallots*voteAllowed > sum(votes_for_candidates)`.

Done.

umbernhard:
@eventualbuddha does the deploy work for you? It hangs when I hit “start audit”

PR #209, opened 2019-11-25
fix PG deploys

benadida:


create.py
```diff
@@ -7,9 +7,13 @@
...
     print(f'database: {engine.url}')
 
-    if not database_exists(engine.url):
-        print('creating database…')
-        create_database(engine.url)
+    try:
+        if not database_exists(engine.url):
+            print('creating database…')
+            create_database(engine.url)
+    except:
```
umbernhard: @benadida won't this cause major problems later on if the DB doesn't exist? Do we have a guarantee that it will?

benadida: The idea here is that this only happens if you're not able to check that the database exists, which only happens if your setup is really weird, or if you're on Heroku, in which case the database has already been created for us.

PR #210, opened 2019-11-25
Fixing build failures on staging

MorganLove:
**DESCRIPTION**

- Adding type declarations back in, since `$` isn't recognized by TS on build.


MorganLove:
@eventualbuddha not sure why this started failing on staging, but it looks like #163 wasn't working there. I put the type declarations back in explicitly and moved the relevant type packages from dev in the package.json, and it's working now. Something to explore later perhaps.

PR #212, opened 2019-11-27
Improving error handling

MorganLove:
**DESCRIPTION**

- Capturing errors returned from the server and displaying them in react-toastify

**New Tests**

- Updating old tests involved with the api


arlo-client/src/components/AuditForms/EstimateSampleSize.test.tsx
```diff
@@ -670,4 +676,43 @@ describe('EstimateSampleSize', () => {
       expect(updateAuditMock).toBeCalledTimes(0)
     })
   })
+
+  it('handles errors from the server side', async () => {
+    apiMock.mockReset()
+    // apiMock.mockImplementation(() =>
```
eventualbuddha: Is this code supposed to be commented out? It's not really clear just looking at the diff what expectations in this test are expecting an error.

arlo-client/src/components/AuditForms/EstimateSampleSize.tsx
```diff
@@ -193,13 +193,20 @@ const EstimateSampleSize: React.FC<IProps> = ({
...
+        `/election/${electionId}/audit/basic`,
+        {
+          method: 'POST',
+          body: JSON.stringify(data),
+          headers: {
+            'Content-Type': 'application/json',
+          },
+        }
+      )
+      if (toaster(response)) {
```
eventualbuddha: I agree with @benadida on the naming of this. It's really not clear from the name `toaster` (and the fact it's in the `utilities` module) that this has a UI side effect since it looks here like it's just checking for something in the response.

arlo-client/src/components/AuditForms/ResetButton.test.tsx
```diff
@@ -39,7 +52,53 @@ describe('ResetButton', () => {
...
+    await wait(() => {
+      expect(apiMock).toBeCalledTimes(1)
+      expect(toasterMock).toBeCalledTimes(1)
+      expect(updateAuditMock).toBeCalledTimes(0)
+      expect(toastSpy).toBeCalledTimes(0)
+    })
+  })
+
+  it('handles 404 errors', async () => {
+    toasterMock.mockImplementationOnce(() => true)
```
eventualbuddha: I'm not sure whether you're trying to standardize on just using `mockImplementation(Once)`, but if not you can use the other helpers like `mockReturnValueOnce(true)` or `mockRejectedValueOnce(new Error('404'))`.

arlo-client/src/components/utilities.ts
```diff
@@ -1,14 +1,21 @@
...
-  if (!res.ok) {
-    throw new Error(res.statusText)
+  try {
+    const res = await fetch(endpoint, options)
+    if (!res.ok) {
+      throw res
+    }
+    return res.json() as Promise<T>
+  } catch (err) {
+    const error = await err.json()
```
eventualbuddha: Should this be more conservative about assuming that error responses are JSON?

arlo-client/src/components/utilities.ts
```diff
@@ -1,14 +1,21 @@
...
+    return Promise.reject(error)
```
eventualbuddha:
```suggestion
    throw error
```

Rather than mixing `async`/`await` with explicit `Promise`.

arlo-client/src/components/utilities.ts
```diff
@@ -104,3 +111,16 @@ export const openQR = (id: string, name: string) => {
     )
   }
 }
+
+export const toaster = (response: IErrorResponse): boolean => {
+  const { errors } = response
+  if (errors) {
+    toast.error(
+      'There was a server error regarding: ' +
+        errors.map((v: { message: string }) => v.message).join(', ')
```
eventualbuddha:
```suggestion
        errors.map(({ message }) => message).join(', ')
```

arlo-client/src/components/utilities.ts
```diff
@@ -104,3 +111,16 @@ export const openQR = (id: string, name: string) => {
     )
   }
 }
+
+export const toaster = (response: IErrorResponse): boolean => {
```
eventualbuddha: Rather than requiring `response` to be `IErrorResponse`, this seems to be answering the question "is `response` an error response?". So from a type system perspective, this might be better:

```ts
const getErrorsFromResponse = (
  response: unknown
): { message: string }[] | undefined => {
  if (typeof response !== 'object' || !response) {
    return
  }

  const errors = (response as { [key: string]: unknown })['errors']

  if (!Array.isArray(errors)) {
    return
  }

  return errors
}

export const toaster = (response: object): response is IErrorResponse => {
  const errors = getErrorsFromResponse(response)
  if (errors) {
    toast.error(
      'There was a server error regarding: ' +
        errors.map(({ message }) => message).join(', ')
    )
    return true
  } else {
    return false
  }
}
```

arlo-client/src/components/CreateAudit.tsx
```diff
@@ -23,11 +24,21 @@ const Wrapper = styled.div`
 const CreateAudit = ({ history }: RouteComponentProps<ICreateAuditParams>) => {
   const [loading, setLoading] = useState(false)
   const onClick = async () => {
-    setLoading(true)
-    const { electionId } = await api('/election/new', {
-      method: 'POST',
-    })
-    history.push(`/election/${electionId}`)
+    try {
```
eventualbuddha: If you were to update the type of `toaster` I as suggested, you could do this here:

```ts
    try {
      setLoading(true)
      const response = await api<{ electionId: string }>('/election/new', {
        method: 'POST',
      })
      if (toaster(response)) {
        return
      }
      const { electionId } = response
      history.push(`/election/${electionId}`)
    } catch (err) {
      toast.error(err.message)
    }
```

arlo-client/src/components/utilities.test.ts
```diff
@@ -93,4 +99,15 @@ describe('utilities.ts', () => {
...
   })
+
+  describe('toaster', () => {
+    it('toasts errors', () => {
+      expect(toaster({ errors: [{ message: 'error' }] })).toBeTruthy()
+      expect(toastSpy).toBeCalledTimes(1)
+    })
+
+    it('returns false without errors', () => {
+      expect(toaster({})).toBeFalsy()
```
eventualbuddha: And `expect(toastSpy).toBeCalledTimes(0)`?

arlo-client/src/types.ts
```diff
@@ -119,4 +125,5 @@ export interface IAudit {
   contests: IContest[]
   jurisdictions: IJurisdiction[]
   rounds: IRound[]
+  errors?: { message: string }[]
```
eventualbuddha: This isn't actually part of an `IAudit`. This would be more properly modeled as `IAudit | IErrorResponse` where you're currently using just `IAudit`.

arlo-client/src/components/AuditForms/SelectBallotsToAudit.tsx
```diff
@@ -151,35 +152,45 @@ const SelectBallotsToAudit: React.FC<IProps> = ({
         const body = {
           size, // until multiple contests are supported
         }
-        await api(`/election/${electionId}/audit/sample-size`, {
+        const response: IErrorResponse = await api(
```
eventualbuddha: Casting this as an `IErrorResponse` is confusing. If you follow my other suggestions, I think you could just make `IErrorResponse` have a non-optional `errors` field and use it as a distinct type from valid responses.

arlo-client/src/components/utilities.ts
MorganLove: @eventualbuddha I agree, but I need to sort out a better way of handling it.

MorganLove:
@eventualbuddha @benadida Updated things along the lines of your suggestions, along with covering some more api responses I'd missed earlier.

PR #216, opened 2019-11-30
Round probability values

### LINKED ISSUE #213: Change ASN "completion in 1 round" estimate to an integer
Arlo currently displays the chance that a sample size equal to the the ASN will complete in one round of auditing as a decimal (see screenshot) - we should round down to the nearest integer instead.

Example: Change "58.23424445537" to "58"

Current behavior:

umbernhard:
Closes #213.


umbernhard:
Ah, this appears to be a frontend problem, not a backend problem. Disregard.

PR #219, opened 2019-12-04
fix readme voting.works link

vitordino:
self explanatory, using the `[content](link)` syntax without the `https://` was creating an internal link


eventualbuddha:
Thanks @vitordino! I know it seems silly for something this small, but mind signing the CLA? Or, if you have and the comment is out of date, let us know?

vitordino:
just did @eventualbuddha

PR #223, opened 2019-12-05
Macro

umbernhard:
Implements MACRO for batch comparison audits on the backend. See [here](https://papers.ssrn.com/sol3/papers.cfm?abstract_id=1443314) for more. We'll need some significant frontend/middle-end changes too, for instance the ability to ingest store batch-level results and samples, and the ability to select this kind of audit on the frontend.

### DIFF
--- a/audits/macro.py
+++ b/audits/macro.py
@@ -28,6 +28,7 @@ class MACRO(RiskLimitingAudit):
                     v_wp = reported_results[contest][winner]
                     v_lp = reported_results[contest][loser]
 
+
                     a_wp = sampled_results[contest][winner]
                     a_lp = sampled_results[contest][loser]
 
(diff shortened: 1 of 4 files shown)


tests/test_macro.py
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+
+    yield Sampler('MACRO', seed, risk_limit, contests, batches)
+
+
+
+def test_max_error(sampler):
+
+    # this is kind of a hacky way to do this but ¯\_(ツ)_/¯
```
eventualbuddha: What's hacky with this? What would I do to make it not hacky?

tests/test_macro.py
```diff
@@ -0,0 +1,186 @@
...
+    computed_p, result = sampler.audit.compute_risk(sampler.contests, sampler.margins, sampler.batch_results, sample)
+
+    U = sampler.audit.compute_U(sampler.contests, sampler.margins, sampler.batch_results)
+    expected_p =  0.247688222
+
+    delta = abs(expected_p - computed_p)
+
+    assert delta < 10**-4, 'Incorrect p-value: Got {}, expected {}'.format(computed_p, expected_p)
+
+    assert result, 'Audit did not terminat but should have'
```
eventualbuddha:
```suggestion
    assert result, 'Audit did not terminate but should have'
```

audits/macro.py
```diff
@@ -0,0 +1,194 @@
+import math
+from scipy import stats
+from audits.audit import RiskLimitingAudit
+
+class MACRO(RiskLimitingAudit):
+
+    def __init__(self, risk_limit):
+        super().__init__(risk_limit)
```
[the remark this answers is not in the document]
umbernhard: Ah, I didn't realize python init'd the super class automatically. Will fix.

tests/test_macro.py
umbernhard: I think that comment was before I settled on a testing strategy. seems like there might have been a better way to set up the test data from Stark's paper, but this works. I can take the comment out if you think this is fine.

PR #230, opened 2019-12-13
[WIP] enable multiple contests

### LINKED ISSUE #222: When multiple targeted contests with the same universe, use largest sample size among them
User story:
As an audit administrator I want to target multiple contests so that I can audit more of the ballot in the same workflow, without having to run separate audits.

Description:
We already handle the flow where a single contest is targeted for audit, the diluted margin is calculated for that contest, and then that diluted margin is used to determine the appropriate number of ballots to audit. To support multiple contests being targeted, we need to calculate the diluted margin for *each* contest, then choose the smallest of those margins to continue the process. 

Conditions of acceptance:
- When multiple contest objects appear in the audit definition API call from the front-end, the sample size for each of the contests is calculated independently on the back end using the appropriate data from that contest
- Once all the sample sizea have been calculated, the largest sample size among them is used to continue the process (select the samples etc.)
- If the contest universes for the targeted contests are not the same, throw an error

z142:
First batch of changes to enable multiple targeted contests (fixes #222).

This commit only allows multiple contests to be saved, calculates the sample sizes correctly (hopefully) per #222, and allow the sample size to be chosen per round.

Manifest and result submissions weren't touched -- those will break the unit tests & the UI.

### DIFF
--- a/app.py
+++ b/app.py
@@ -80,7 +80,7 @@ def compute_sample_sizes(round_contest):
     sampler = get_sampler(election)
 
     # format the options properly
-    raw_sample_size_options = sampler.get_sample_sizes(sample_results(election))[election.contests[0].id]
+    raw_sample_size_options = sampler.get_sample_sizes(sample_results(election))[round_contest.contest_id]
     sample_size_options = []
     sample_size_90 = None
     sample_size_backup = None
@@ -121,11 +121,24 @@ def compute_sample_sizes(round_contest):
         round_contest.sample_size = sample_size_90        
         sample_ballots(election, the_round)
 
+    # update the parent round's sample size options
+    if the_round.sample_size_options is None:
+        the_round.sample_size_options = round_contest.sample_size_options
+
+    else:
+        round_options = json.loads(the_round.sample_size_options)
+
+        # TODO: this assumes that the lists are always in the same order
+        for i, option in enumerate(sample_size_options):
+            if round_options[i]["size"] < option["size"]:
+                round_options[i] = option
+
+        the_round.sample_size_options = json.dumps(round_options)
+
     db.session.commit()
-        
+
+
 def setup_next_round(election):
-    if len(election.contests) > 1:
-        raise Exception("only supports one contest for now")
 
     rounds = Round.query.filter_by(election_id = election.id).order_by('round_num').all()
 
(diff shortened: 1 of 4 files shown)


app.py
```diff
@@ -80,7 +80,7 @@ def compute_sample_sizes(round_contest):
     sampler = get_sampler(election)
 
     # format the options properly
-    raw_sample_size_options = sampler.get_sample_sizes(sample_results(election))[election.contests[0].id]
+    raw_sample_size_options = sampler.get_sample_sizes(sample_results(election))[round_contest.contest_id]
```
z142: For some reason `bgcompute` iterates over contests and calls `compute_sample_sizes` here. 
This is kind of redundant: `get_sample_sizes` has all the info we need for a round, could easily refactor this method and `bgcompute` to be a lot more simple.

app.py
```diff
@@ -121,11 +121,24 @@ def compute_sample_sizes(round_contest):
...
 
+    # update the parent round's sample size options
+    if the_round.sample_size_options is None:
+        the_round.sample_size_options = round_contest.sample_size_options
+
+    else:
+        round_options = json.loads(the_round.sample_size_options)
+
+        # TODO: this assumes that the lists are always in the same order
+        for i, option in enumerate(sample_size_options):
```
z142: At one point this was a monster single-line lambda expression but it was so ugly that I just went with a simple loop instead. Could revert to a one-liner to "flex" if someone finds it entertaining. ;)

z142: Relying on the list order is not great, couldn't find a good way to "id" the options (some have types, some don't). Probably an easy fix, didn't have the time to figure out the most elegant way.

app.py
```diff
@@ -121,11 +121,24 @@ def compute_sample_sizes(round_contest):
...
+                round_options[i] = option
+
+        the_round.sample_size_options = json.dumps(round_options)
+
     db.session.commit()
-        
+
+
 def setup_next_round(election):
-    if len(election.contests) > 1:
```
z142: A lot of dominoes fell when I removed this one.

models.py
```diff
@@ -106,6 +106,9 @@ class Round(db.Model):
     started_at = db.Column(db.DateTime, nullable=False)
     ended_at = db.Column(db.DateTime, nullable=True)
 
+    sample_size_options = db.Column(db.String(1000), nullable=True)
```
z142: I was wondering why we didn't use the JSON data type. SQLAlchemy supports it. 
There's an awful lot of `json.dumps` happening around the code to keep this a string.

tests/test_app.py
```diff
@@ -56,9 +57,12 @@ def test_whole_audit_flow(client):
     assert result2["riskLimit"] == 5
 
 def setup_whole_audit(client, election_id, name, risk_limit, random_seed):
-    contest_id = str(uuid.uuid4())
+    contest_id_1 = str(uuid.uuid4())
```
z142: I did the bare minimum to run a test with the multiple contests. 
Stopped fixing tests (which now break) and making code changes when I got to manifests & results.

z142:
I have a few notes we could discuss IRL. 
Due to the time constraint I chose to avoid any kind of refactoring or major changes.

Issue #127, opened 2019-09-11
Use QR codes for audit board login/auth


mcchilders:
**User Story:** As a user I would like a simplified way to log each audit board in to their respective data entry interface, so that if I have a lot of boards (or limited time) I can do so efficiently while still getting the boards to the correct place.

**Background:** Once we have a data entry interface for each audit board, we will need a way to authenticate users and direct them to the correct interface for their particular audit board. Persistent login schemes are less than ideal since users are one-time volunteers and co-located with the election official, and we'd like to improve on CORLAs model of making the election official responsible for logging each audit board on to their device (cumbersome, makes less sense if we start seeing boards that want to enter data on their personal phones.) Providing QR codes that the election official can print and distribute still gives us the necessary security while minimizing overhead/upkeep that traditional login schemes require.

**Requirements:**
- Add an additional button below the "Download retrieval list" button in the UI, labeled "Download Audit Board Credentials for Data Entry"
- When clicked, the button should generate a PDF containing 1 page per audit board, labeled with:
-- Audit board name
-- Basic instructions
-- QR code that contains both the website URL of that board's data entry interface and whatever hash/code (if any) they need to authenticate
-- The human-friendly version of that URL
(see screenshot)
 **NOTE: Audit board credentials should be audit-specific, not round-specific, so they'll only need to be shown as we create Round 1.

**Open questions:** 
- The human-friendly URL in the mockup is an idea based on what we've discussed, but I'm not tied to it. If there's a good reason to do something other than "random series of words" for that URL, let's discuss!
- Sizing of the QR code (and really all the QR code details are up to the implementer - the screenshots are meant to give you an idea, not be prescriptive.

<img width="777" alt="Screen Shot 2019-09-11 at 9 52 34 AM" src="https://user-images.githubusercontent.com/50123300/64717312-0e324500-d492-11e9-8b39-c9a9e86da9a1.png">
<img width="657" alt="Screen Shot 2019-09-11 at 10 07 11 AM" src="https://user-images.githubusercontent.com/50123300/64717313-0e324500-d492-11e9-91a7-742bd64d0d1a.png">

mcchilders:
Ok @MorganLove, did some testing and  it looks like what's on staging still has a couple pieces missing. Can we please: 
- remove the onscreen QR codes display (great for testing tho!)
- implement the downloadable PDF pages w/the QR code/instructions/etc for audit board login, 1 per audit board, as described above
[...]

PR #232, opened 2019-12-19
Updating QR code generation

MorganLove:
**DESCRIPTION**

- Removed display of QR codes from form two
- Changed method of printing QR codes to PDF
- Added button to the first round to print the QR codes
- Added more content to the QR code sheet

Solves some of the recent updates to votingworks/arlo#127 

**New Tests**

- Updated snapshots and tests

**Checklist**

*Function*

[*] Code is fully functional and ready for review (or is tagged WIP)

[*] No errors/warnings in the browser console ( ﾟヮﾟ)/

[*] package.json is committed if I added or removed any dependencies

[*] Tests added as needed to ensure code base still has 100% passing coverage

[*] Tests added as needed to cover new functionality

*Style*

[*] Passes linter without errors

[*] Code is easily understandable and self-documenting (using naming conventions in keeping with the rest of the codebase)

[*] Interfaces all have 'I' before their names (as in `IAudit`)

[*] Comments are only included as needed to explain the 'why' instead of the 'what'

[*] Common standards and coding practices are adhered to

[*] DRY: Don't Repeat Yourself. Code is reasonably structured to avoid repetition

[*] Code is intuitively structured following a Separation of Concerns

[*] No class components; only functional components (using Hooks as needed)


MorganLove:
Rebased on #241 and incorporated it into the QR page downloads.

arlo-client/src/components/AuditForms/CalculateRiskMeasurement.tsx
```diff
@@ -177,6 +202,48 @@ const CalculateRiskMeasurement: React.FC<IProps> = ({
     }
   }
 
+  const downloadDataEntry = async (): Promise<void> => {
```
eventualbuddha: Nothing in this function actually requires it be `async`. Is there a missing `await` somewhere?

arlo-client/src/components/AuditForms/CalculateRiskMeasurement.tsx
```diff
@@ -177,6 +202,48 @@ const CalculateRiskMeasurement: React.FC<IProps> = ({
...
+          auditBoards.text(
+            auditBoards.splitTextToSize(
+              'If you are not able to scan the QR code, you may also type the following URL into a web browser to access the data entry portal.',
+              180
+            ),
+            20,
+            120
+          )
+          auditBoards.text(
+            `arlo.votingworks.app/auditboard/${board.passphrase}`,
```
eventualbuddha:
```suggestion
            `${window.location.origin}/auditboard/${board.passphrase}`,
```

arlo-client/src/types.ts
```diff
@@ -63,6 +63,7 @@ export interface IAuditBoard {
   id: string
   name: string
   members: IAuditBoardMember[]
+  passphrase?: string
```
eventualbuddha: What's the advantage of marking this as optional? We assume the passphrase is set in `CalculateRiskMeasurement` anyway.

MorganLove: @eventualbuddha It is set on the server side, so before we get it back from the `/status` endpoint, it won't exist in the data tree. I expect it would throw errors in `SelectBallotsToAudit` when trying to submit the jurisdictions data with audit boards without passphrases unless they're marked as optional.

arlo-client/src/components/AuditForms/CalculateRiskMeasurement.tsx
MorganLove: @eventualbuddha Good catch, I was copying over from the other pdf generation functions, and those used the `await getBallots()` function, but this doesn't.

PR #234, opened 2019-12-20
Implements ballots comparison math

umbernhard:
Implements back-end math for ballot comparison, according to [Super-Simple Simultaneous Single-Ballot Risk-Limiting Audits](https://www.usenix.org/legacy/events/evtwote10/tech/full_papers/Stark.pdf).

I went ahead and self-decided on a CVR data format for the back end, but we should maybe discuss this further before merging. We can either pick a new format and implement it throughout, or write a translator somewhere in the middle. Think this latter option is preferable, since we will have to handle many different CVR types anyways, but I'm open to other ideas.


audits/supersimple.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def __init__(self, risk_limit):
+        self.l = 0.5
+        self.gamma = 1.1
+        super().__init__(risk_limit)
+
+
+    def compute_diluted_margin(self, contests, margins, total_ballots):
+        """
+        Computes the diluted margin across all contests
```
nealmcb: You do not want to compute a single diluted margin across all the contests, which at first glance seems to be what you're doing here. I know that's what super-simple allows you to do, but the HUGE downside is that if you go to a full hand count, you need to hand count ALL the contests being audited. No state would ever want to do that.

Instead, as is done in ColoradoRLA (after we talked with Philip about this), you want to compute a diluted margin and sample size for each contest, using super-simple just across all the winners and losers in EACH contest as necessary, to handle more than two candidates.  Then, for each round, you take the maximum sample size across all the contests in a given "universe" (county, district or other sampling frame) and pick that many ballots from the manifests for that universe.

umbernhard: Hrm. We don't currently have a notion of "universe", since we don't support multi-jurisdiction right now. So right now we just have one universe and are computing the margins across that. This might be famous last words, but I don't think it will be too bad to add multiple universes on top of what we're doing here.

umbernhard: Also @nealmcb since the comments are all put into a single conversation on the PR, would you mind removing your mainline comment ([here](https://github.com/votingworks/arlo/pull/234#pullrequestreview-335504162) just so it's easier to follow the thread? Thanks!

nealmcb: I know the multi-universe changes were pretty extensive in ColoradoRLA, so I do encourage you to keep that case in mind.

But even with a single universe, any time you're supporting audits of multiple contests, officials will want to audit them separately, and in particular, escalate to a full hand count separately. So the idea of computing a diluted margin across multiple contests is still problematic.

audits/supersimple.py
```diff
@@ -0,0 +1,175 @@
...
+        U = 2*self.gamma/diluted_margin
+
+        lowest_p = 1
+        result = False
+        for ballot in sample_cvr:
+            e_r = 0
+
+            for contest in contests:
+                if contest not in sample_cvr[ballot]:
+                    # TODO I think this is right
```
PlauditsForAudits: At least conceptually, yes -- if the contest isn't on the ballot, there's no effect on the p value.

audits/supersimple.py
```diff
@@ -0,0 +1,175 @@
...
+                        V_wl = contests[contest][winner] - contests[contest][loser]
+
+                        e = ((v_w - a_w) - (v_l - a_l))/V_wl
+                        if e > e_r:
+                            e_r = e
+
+            p_b = (1 - 1/U)/(1 - (e_r/((2*self.gamma)/V)))
+            p *= p_b
+
+            if p < lowest_p:
```
PlauditsForAudits: This is mathematically correct if the CVRs are guaranteed to be in unsorted sample order -- but as a matter of policy, I would be predisposed to use all available audit data. It's uncomfortable if attaining the risk limit depends on whether an observed overstatement is or isn't included in the calculation.

audits/supersimple.py
```diff
@@ -0,0 +1,175 @@
...
+                lowest_p = p
+
+            if p < self.risk_limit:
```
PlauditsForAudits: In effect, this repeatedly compares lowest_p to the risk limit. That only needs to be done once (or not at all, if the entire sample is used).

audits/supersimple.py
```diff
@@ -0,0 +1,175 @@
+import math
+from scipy import stats
+from audits.audit import RiskLimitingAudit
+
+class SuperSimple(RiskLimitingAudit):
+
+    def __init__(self, risk_limit):
+        self.l = 0.5
```
PlauditsForAudits: I recommend against using the "lambda" approach (in which the tolerable rate of 1-vote overstatements equals lambda times the diluted margin); if you do use it, I'd use a much smaller value of lambda, but it's probably better to estimate "expected" 1-vote overstatements somewhat akin to https://www.stat.berkeley.edu/~stark/Vote/auditTools.htm.

audits/supersimple.py
umbernhard: Ah, I had to add this in because of the way the SuperSimple paper presented its data. I'm going to rework it all with numbers generated from [Stark's tool](https://www.stat.berkeley.edu/~stark/Vote/auditTools.htm) and go from there.

umbernhard:
Because you were both oh so right @PlauditsForAudits and @nealmcb, I refactored the code to operate on a per-contest basis. Please let me know if you see any outstanding issues.

jonahkagan:
@umbernhard what's the status on this PR?

umbernhard:
@jonahkagan it's ready to merge, it just needs rebasing/refactoring to the current codebase.

jonahkagan:
Awesome! Let me know when it's ready

umbernhard:
@jonahkagan okay we're good now. It's not at 100% coverage but that'll be my next task.

umbernhard:
well, we were :sweat_smile: will whip back into shape tomorrow.

app.json
```diff
@@ -1,29 +0,0 @@
...
-      "quantity": 1,
-      "size": "standard-1x"
-    }
-  },
-  "name": "arlo",
-  "scripts": {
-    "postdeploy": "heroku pg:backups:restore vx-arlo-staging:: -a $HEROKU_APP_NAME --confirm $HEROKU_APP_NAME"
-  },
-  "stack": "heroku-18"
-}
```
jonahkagan, line 29: we need this file!

server/audit_math/supersimple.py
```diff
@@ -0,0 +1,172 @@
...
+# This sets the expected number of one-vote misstatements at 1 in 1000
+o1 = 0.001
+u1 = 0.001
+
+# This sets the expected two-vote misstatements at 1 in 10000
+o2 = 0.0001
+u2 = 0.0001
+
+
+def nMin(risk_limit, contest, o1, o2, u1, u2):
```
jonahkagan, line 16: do you want to add types to this file?

server/audit_math/supersimple.py
```diff
@@ -0,0 +1,172 @@
...
+
+    return nMin(risk_limit, contest, r1, r2, s1, s2)
+
+
+def compute_risk(risk_limit, contest, cvrs, sample_cvr):
+    """
+    Computes the risk-value of <sample_results> based on results in <contest>.
+
+    Inputs:
+        contests       - the contests and results being audited
```
jonahkagan, line 110: contests or contest?

server/audit_math/supersimple.py
```diff
@@ -0,0 +1,172 @@
...
+                    e_r = e
+
+        denom = (2 * gamma) / V
+        p_b = (1 - 1 / U) / (1 - (e_r / denom))
+        p *= p_b
+
+    if 0 < p < risk_limit:
+        result = True
+
+    return p, result
```
jonahkagan, line 172: curious why this is called `supersimple` lol. is there a more complex method coming? wondering if it would be helpful to have "ballot comparison" in the name

app.json
umbernhard: ack I'll put it back :grimacing:

server/audit_math/supersimple.py
umbernhard: Yeah, I can do that.

server/audit_math/supersimple.py
umbernhard: Contest, will fix.

server/audit_math/supersimple.py
umbernhard: It's called that because that's what Stark called it. "Ballot comparison" is fairly vague and can refer to many different things, unfortunately. FWIW, we don't use "ballot polling" or "batch comparison" in our other modules for this reason too...

umbernhard:
okay @jonahkagan requested changes have been made

PR #238, opened 2019-12-30
Dedupes ballots in retrieval list, puts in list of ticket numbers

umbernhard:
Addresses @mcchilders comment about deduplicating the retrieval list on #198. https://github.com/votingworks/arlo/pull/198#issuecomment-569736536

### DIFF
--- a/app.py
+++ b/app.py
@@ -728,17 +728,6 @@ def jurisdiction_retrieval_list(election_id, jurisdiction_id, round_num):
     jurisdiction = Jurisdiction.query.filter_by(election_id = election.id, id = jurisdiction_id).one()
     round = Round.query.filter_by(election_id = election.id, round_num = round_num).one()
 
-    '''
-    ballots = SampledBallotDraw.query.filter_by(round_id = round.id) \
-                    .join(SampledBallotDraw.batch).filter_by(jurisdiction_id = jurisdiction_id)  \
-                    .join(SampledBallotDraw.sampled_ballot).join(SampledBallot.audit_board) \
-                    .add_entity(Batch).add_entity(AuditBoard) \
-                    .group_by(Batch.name)\
-                    .group_by(SampledBallotDraw.ballot_position) \
-                    .order_by(AuditBoard.name, Batch.name, SampledBallotDraw.ballot_position, SampledBallotDraw.ticket_number) \
-                    .all()
-    '''
-
     # Get deduped sampled ballots
     ballots = SampledBallotDraw.query.filter_by(round_id = round.id) \
                     .join(SampledBallotDraw.batch).filter_by(jurisdiction_id = jurisdiction_id)  \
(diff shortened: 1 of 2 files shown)


tests/test_app.py
```diff
@@ -204,11 +204,12 @@ def setup_whole_audit(client, election_id, name, risk_limit, random_seed):
     # get the retrieval list for round 1
     rv = client.get('{}/jurisdiction/{}/1/retrieval-list'.format(url_prefix, jurisdiction_id))
     lines = rv.data.decode('utf-8').splitlines()
-    assert lines[0] == "Batch Name,Ballot Number,Storage Location,Tabulator,Ticket Number,Audit Board"
+    assert lines[0] == "Batch Name,Ballot Number,Storage Location,Tabulator,Ticket Numbers,Audit Board"
     assert len(lines) > 5
     assert 'attachment' in rv.headers['content-disposition']
 
-    num_ballots = len([line for line in lines[1:] if line!=""])
+    lines = csv.DictReader(io.StringIO(rv.data.decode('utf-8')))
```
benadida: can we package this up as a utility function since we're reusing it in a couple of places?

app.py
```diff
@@ -734,8 +734,33 @@ def jurisdiction_retrieval_list(election_id, jurisdiction_id, round_num):
...
 
+    ballot_to_sampled = {}
     for ballot_draw, batch, audit_board in ballots:
-        retrieval_list_writer.writerow([batch.name, ballot_draw.ballot_position, batch.storage_location, batch.tabulator, ballot_draw.ticket_number, audit_board.name])
+
+        index = (batch.name, ballot_draw.ballot_position)
+        if index in ballot_to_sampled:
+            ballot_to_sampled[index]['ticket_numbers'].append(ballot_draw.ticket_number)
+        else:
+            ballot_to_sampled[index] = {
```
benadida: too much code here, why stash these in a dictionary that we then immediately pull out? How about just storing the array of `ballot_draw` objects and then pulling them out with attributes, not keyed dictionary entires? Would shrink this code by 10-15 lines.

tests/test_app.py
umbernhard: yeah, that's a nice idea.

app.py
umbernhard: I did it for logical simplicity (i.e. laziness :sweat_smile:). I'll take a look at refactoring.

app.py
```diff
@@ -721,21 +721,39 @@ def jurisdiction_retrieval_list(election_id, jurisdiction_id, round_num):
...
+                    .add_entity(Batch).add_entity(AuditBoard) \
+                    .group_by(Batch.name, Batch.id, Batch.storage_location, Batch.tabulator, AuditBoard.name)\
+                    .group_by(SampledBallotDraw.ballot_position) \
+                    .values(Batch.id, SampledBallotDraw.ballot_position, Batch.name, Batch.storage_location, Batch.tabulator, AuditBoard.name)
+
+    rows = []
+    for batch_id, position, batch_name, storage_location, tabulator, audit_board in ballots:
+
+        # Get ticket_numbers
+        ticket_numbers = ','.join([b.ticket_number for b in SampledBallotDraw.query.filter_by(batch_id = batch_id, ballot_position=position, round_id = round.id).all()])
```
benadida: ah, so this will issue another query within the loop, which means now we're hammering the database a lot. I don't recommend this, it will come back to bite us the moment we have any significant load.

What were you going for with the change to the outer query btw?

PR #240, opened 2019-12-31
simple HTTP auth for /admin to list running audits

benadida:


benadida:
> Looks good to me. Do we want to update the documentation as well to mention that a password must be set, or is this too temporary for that?

Good question.. Ultimately I think it's too temporary. I'm going to add a ticket to make sure we remove this later though with #243.

PR #241, opened 2019-12-31
added passphrase to audit-board based on english words, URL slug redirection

benadida:


app.py
```diff
@@ -468,7 +471,8 @@ def jurisdictions_set(election_id):
             audit_board_obj = AuditBoard(
                 id = audit_board["id"],
                 name = audit_board["name"],
-                jurisdiction_id = jurisdiction_obj.id
+                jurisdiction_id = jurisdiction_obj.id,
+                passphrase = xp.generate_xkcdpassword(WORDS, numwords=4, delimiter="-")
```
eventualbuddha: Just to clarify, we don't care that these will be different given the same seed, right?

MorganLove:
#232 has the frontend corollary to this by including the passphrase link in the downloadable PDF.

models.py
```diff
@@ -96,6 +96,7 @@ class AuditBoard(db.Model):
     member_1_affiliation = db.Column(db.String(200), nullable=True)
     member_2 = db.Column(db.String(200), nullable=True)
     member_2_affiliation = db.Column(db.String(200), nullable=True)
+    passphrase = db.Column(db.String(1000), nullable=True)
```
eventualbuddha: Should there be a unique index on this?

benadida: yeah, there probably should, and should also check for error conditions, will add those.

app.py
benadida: correct, the consistency we care about is for the audit data itself, not for authentication.

benadida:
added the unique constraint. I don't think we need any additional error checking for now. On the offchance we hit a duplication, the jurisdiction set call will fail and roll back the transaction. If we hit it for real, we can make the error a bit nicer.

benadida:
Ah, we've gotten ourselves in a bit of a merge mess here. @MorganLove let's sync up and see if we can disentangle this. I think the printed credential URLs are the old URLs, not the passphrase ones, and I'm having trouble disentangling the commits.

MorganLove:
> Ah, we've gotten ourselves in a bit of a merge mess here. @MorganLove let's sync up and see if we can disentangle this. I think the printed credential URLs are the old URLs, not the passphrase ones, and I'm having trouble disentangling the commits.

@benadida I resolved the merge conflict in the end to end test file, and double checked to make sure the printed url is the one with the passphrase, [and it is](https://github.com/votingworks/arlo/blob/0ee75dd772c0be87e843da1265f311384c594505/arlo-client/src/components/AuditForms/CalculateRiskMeasurement.tsx#L236). We should be good to go then. 🙂

eventualbuddha:
As for rebasing & merging issues, I rebased this branch onto master: `generate-audit-board-slugs-rebased`. It seems to be identical to this branch in that `git diff` is empty. Feel free to force-push that branch to this one if you want to use it @benadida or @MorganLove.

MorganLove:
> As for rebasing & merging issues, I rebased this branch onto master: `generate-audit-board-slugs-rebased`. It seems to be identical to this branch in that `git diff` is empty. Feel free to force-push that branch to this one if you want to use it @benadida or @MorganLove.

@eventualbuddha It was the commit history that was mangled up, so I went back in and did an interactive rebase to clean it up and force pushed that. Thanks though!

@benadida Good catch on the git history. Should we set that up as a standard part of the flow, so that we always remember to do an interactive rebase to clean up the commits before merging?

PR #245, opened 2020-01-03
chore: forward Arlo web port

eventualbuddha:
This makes it so you don't have to manually configure it in VS Code.

### DIFF
--- a/arlo.code-workspace
+++ b/arlo.code-workspace
@@ -7,5 +7,13 @@
 			"path": "arlo-client"
 		}
 	],
-	"settings": {}
+	"settings": {
+		"remote.SSH.defaultForwardedPorts": [
+			{
+				"localPort": 3000,
+				"name": "Arlo Web",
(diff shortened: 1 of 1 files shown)


arlo.code-workspace
```diff
@@ -7,5 +7,13 @@
 			"path": "arlo-client"
 		}
 	],
-	"settings": {}
+	"settings": {
+		"remote.SSH.defaultForwardedPorts": [
```
benadida: I'm not versed in docker well enough, but I'm assuming this means ssh tunneling? Is that the standard way to do it?

eventualbuddha: > I'm assuming this means ssh tunneling?

Yes, it's SSH port forwarding / tunneling. This actually has nothing to do with Docker. This is for when you're doing [remote development with SSH](https://code.visualstudio.com/docs/remote/ssh), which is what I use.

PR #253, opened 2020-01-20
More makefile

benadida:
make it easier to dev in Linux without Docker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,23 @@ Like any open-source software, Arlo welcomes suggested changes in the form of pu
 
 ### Setting up the dev environment
 
+#### Linux
+
+We recommend Ubuntu 18.0.4.
+
+* Install Node10. See https://joshtronic.com/2018/05/08/how-to-install-nodejs-10-on-ubuntu-1804-lts/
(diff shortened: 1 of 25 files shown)


README.md
```diff
@@ -92,6 +92,25 @@ Like any open-source software, Arlo welcomes suggested changes in the form of pu
 
 ### Setting up the dev environment
 
+#### Linux
+
+We recommend Ubuntu 18.0.4.
+
+* Install Node10. See https://joshtronic.com/2018/05/08/how-to-install-nodejs-10-on-ubuntu-1804-lts/
+* `sudo make deps`
```
eventualbuddha: These instructions don't make use of the `dev-environment` task. Also, it implies that you need `sudo` to run `deps`, which `dev-environment` doesn't do since `deps` is run as a dependency.

benadida:
@eventualbuddha fixed that up :)

benadida:
@eventualbuddha I'm hoping it's okay if I merge this, as I made the change you requested, and things are backed up on this small PR now (we merged another PR into here.)

PR #254, opened 2020-01-20
fix winners --> numWinners

benadida:
master was broken somehow. fixing field name.


benadida:
@MorganLove Given that master is red, I'm going to solo-merge this small fix as soon as it turns the tree back to green. Would love an after-the-fact review to make sure I didn't mess anything up.

PR #257, opened 2020-02-05
Timeout fix

MorganLove:
**DESCRIPTION**

- Restructured some of the hooks and conditions in the audit data entry flow index so that it's not using stale and blank data at unexpected times.
- This solves the randomly appearing timeout error toast popup

**New Tests**

- Made testing for when there is and isn't ballots more stable and wrapped more `render` calls in `async act` blocks.

**Checklist**

*Function*

[*] Code is fully functional and ready for review (or is tagged WIP)

[*] No errors/warnings in the browser console ( ﾟヮﾟ)/

[*] package.json is committed if I added or removed any dependencies

[*] Tests added as needed to ensure code base still has 100% passing coverage

[*] Tests added as needed to cover new functionality

*Style*

[*] Passes linter without errors

[*] Code is easily understandable and self-documenting (using naming conventions in keeping with the rest of the codebase)

[*] Interfaces all have 'I' before their names (as in `IAudit`)

[*] Comments are only included as needed to explain the 'why' instead of the 'what'

[*] Common standards and coding practices are adhered to

[*] DRY: Don't Repeat Yourself. Code is reasonably structured to avoid repetition

[*] Code is intuitively structured following a Separation of Concerns

[*] No class components; only functional components (using Hooks as needed)

*Workflow*

[*] Do an interactive rebase or a rebase

### DIFF
--- a/arlo-client/package.json
+++ b/arlo-client/package.json
@@ -138,7 +138,7 @@
     "@wdio/local-runner": "^5.14.4",
     "@wdio/spec-reporter": "^5.16.11",
     "babel-jest": "^24.9.0",
-    "chromedriver": "78.0.1",
+    "chromedriver": "80.0.0",
     "husky": "^2.3.0",
     "is-ci-cli": "^1.1.1",
     "jest-canvas-mock": "^2.1.1",
--- a/arlo-client/src/components/AuditFlow/BoardTable.tsx
+++ b/arlo-client/src/components/AuditFlow/BoardTable.tsx
@@ -138,18 +138,20 @@ const BoardTable: React.FC<IProps> = ({
       <RightWrapper>
         {roundComplete ? (
           <Button intent="primary">Review Complete - Finish Round</Button>
-        ) : ballots.length && unauditedBallot ? (
-          <Link
-            to={
-              url +
-              `/round/1/batch/${unauditedBallot.batch.id}/ballot/${unauditedBallot.position}`
-            }
-            className="bp3-button bp3-intent-primary"
-          >
-            Start Auditing
-          </Link>
-        ) : null // eslint-disable-line no-null/no-null
-        }
+        ) : (
+          ballots.length > 0 &&
+          unauditedBallot && (
+            <Link
+              to={
+                url +
+                `/round/1/batch/${unauditedBallot.batch.id}/ballot/${unauditedBallot.position}`
+              }
+              className="bp3-button bp3-intent-primary"
+            >
+              Start Auditing
+            </Link>
+          )
+        )}
       </RightWrapper>
       {/* <ActionWrapper> // commented out until feature is added
         {!roundComplete && (
--- a/arlo-client/src/components/AuditFlow/BoardTable.tsx
+++ b/arlo-client/src/components/AuditFlow/BoardTable.tsx
@@ -121,7 +121,8 @@ const BoardTable: React.FC<IProps> = ({
     setCols(columnWidths())
   }, [ballots])
 
-  const roundComplete = ballots && ballots.every(b => b.status === 'AUDITED')
+  const roundComplete =
+    ballots.length && ballots.every(b => b.status === 'AUDITED')
 
   const unauditedBallot = ballots.find(b => !b.status)
 
@@ -137,20 +138,18 @@ const BoardTable: React.FC<IProps> = ({
       <RightWrapper>
         {roundComplete ? (
           <Button intent="primary">Review Complete - Finish Round</Button>
-        ) : (
-          ballots &&
-          unauditedBallot && (
-            <Link
-              to={
-                url +
-                `/round/1/batch/${unauditedBallot.batch.id}/ballot/${unauditedBallot.position}`
-              }
-              className="bp3-button bp3-intent-primary"
-            >
-              Start Auditing
-            </Link>
-          )
-        )}
+        ) : ballots.length && unauditedBallot ? (
+          <Link
+            to={
+              url +
+              `/round/1/batch/${unauditedBallot.batch.id}/ballot/${unauditedBallot.position}`
+            }
+            className="bp3-button bp3-intent-primary"
+          >
+            Start Auditing
+          </Link>
+        ) : null // eslint-disable-line no-null/no-null
+        }
       </RightWrapper>
       {/* <ActionWrapper> // commented out until feature is added
         {!roundComplete && (
--- a/arlo-client/src/components/AuditFlow/index.test.tsx
+++ b/arlo-client/src/components/AuditFlow/index.test.tsx
@@ -1,5 +1,11 @@
 import React from 'react'
-import { render, wait, fireEvent } from '@testing-library/react'
+import {
+  render,
+  wait,
+  fireEvent,
+  act,
+  RenderResult,
+} from '@testing-library/react'
 import { StaticRouter } from 'react-router-dom'
 import { routerTestProps } from '../testUtilities'
 import AuditFlow from './index'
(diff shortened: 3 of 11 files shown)


arlo-client/package.json
```diff
@@ -138,7 +138,7 @@
     "@wdio/local-runner": "^5.14.4",
     "@wdio/spec-reporter": "^5.16.11",
     "babel-jest": "^24.9.0",
-    "chromedriver": "78.0.1",
+    "chromedriver": "80.0.0",
```
eventualbuddha: Related, or just included in this PR?

arlo-client/src/components/AuditFlow/BoardTable.tsx
```diff
@@ -121,7 +121,8 @@ const BoardTable: React.FC<IProps> = ({
     setCols(columnWidths())
   }, [ballots])
 
-  const roundComplete = ballots && ballots.every(b => b.status === 'AUDITED')
+  const roundComplete =
+    ballots.length && ballots.every(b => b.status === 'AUDITED')
```
eventualbuddha:
```suggestion
    ballots.some(b => b.status !== 'AUDITED')
```

arlo-client/src/components/AuditFlow/BoardTable.tsx
```diff
@@ -137,20 +138,18 @@ const BoardTable: React.FC<IProps> = ({
...
-                url +
-                `/round/1/batch/${unauditedBallot.batch.id}/ballot/${unauditedBallot.position}`
-              }
-              className="bp3-button bp3-intent-primary"
-            >
-              Start Auditing
-            </Link>
-          )
-        )}
+        ) : ballots.length && unauditedBallot ? (
```
eventualbuddha:
```suggestion
        ) : ballots.length > 0 && unauditedBallot && (
```
This allows you to remove the `: null` bit at the end, I believe.

arlo-client/src/components/AuditFlow/index.test.tsx
```diff
@@ -49,15 +55,30 @@ describe('AuditFlow ballot interaction', () => {
...
   it('renders board table with no ballots', async () => {
-    const { queryByText, container } = render(
-      <StaticRouter {...staticRouteProps}>
-        <AuditFlow {...routeProps} />
-      </StaticRouter>
+    apiMock.mockImplementation(
+      async (endpoint: string): Promise<IAudit | { ballots: IBallot[] }> => {
+        switch (endpoint) {
+          case '/election/1/audit/status':
+            memberDummy.jurisdictions[0].auditBoards = [dummyBoard[1]]
```
eventualbuddha: This test now has side effects that could affect other tests. Would something like this be better?

```ts
return {
  ...memberDummy,
  jurisdictions: [
    {
      ...memberDummy.jurisdictions[0],
      auditBoards: [dummyBoard[1]]
    },
    ...memberDummy.jurisdictions.slice(1)
  ]
}
```

arlo-client/src/components/AuditFlow/index.tsx
```diff
@@ -12,8 +12,8 @@ import {
   IErrorResponse,
 } from '../../types'
 import { api, poll, checkAndToast } from '../utilities'
-import { statusStates } from '../AuditForms/_mocks'
-import { dummyBallots } from './_mocks'
+// import { statusStates } from '../AuditForms/_mocks'
+// import { dummyBallots } from './_mocks'
```
eventualbuddha: Delete?

arlo-client/src/components/AuditFlow/index.tsx
```diff
@@ -75,18 +60,19 @@ const AuditFlow: React.FC<IProps> = ({
     updateAudit()
   }, [updateAudit])
 
-  const board:
-    | IAuditBoard
-    | undefined = audit.jurisdictions[0].auditBoards.find(
-    (v: IAuditBoard) => v.id === token
-  )
+  const board: IAuditBoard | undefined = audit
```
eventualbuddha: The type annotations in this statement are not required. TypeScript can figure them out. But leave them in if you like!

arlo-client/src/components/AuditFlow/index.tsx
```diff
@@ -75,18 +60,19 @@ const AuditFlow: React.FC<IProps> = ({
...
+    ? audit.jurisdictions[0].auditBoards.find(
```
eventualbuddha: Also, were you on TypeScript 3.7, this whole thing could be `const board = audit?.jurisdictions[0].auditBoards.find(v => v.id === token)`.

arlo-client/src/components/AuditFlow/index.tsx
```diff
@@ -75,18 +60,19 @@ const AuditFlow: React.FC<IProps> = ({
...
+        (v: IAuditBoard) => v.id === token
+      )
+    : undefined
 
-  const round = audit.rounds[audit.rounds.length - 1]
+  const round = audit ? audit.rounds[audit.rounds.length - 1] : undefined
```
eventualbuddha: Another easier TS 3.7 bit: `const round = audit?.rounds[audit.rounds.length - 1]`.

arlo-client/src/components/AuditFlow/index_members.test.tsx
```diff
@@ -40,33 +40,45 @@ describe('AuditFlow initial load', () => {
...
-    )
+    let utils: RenderResult
+    await act(async () => {
+      utils = render(
+        <Router {...routeProps}>
+          <AuditFlow {...routeProps} />
+        </Router>
+      )
+    })
+    const { container, getByText } = utils!
```
eventualbuddha: Worth wrapping all this up in a helper?

```ts
const { container, getByText } = await renderAtomic(<Router>…</Router>)
```

Could be defined something like this:

```ts
async function renderAtomic(component: React.Component): Promise<RenderResult> {
  let result: RenderResult

  // You have this callback as `async`, but I don't think it needs to be.
  await act(() => {
    result = render(component)
  })

  return result!
}
```

arlo-client/package.json
MorganLove: @eventualbuddha Just included in this PR because it's required to pass CI. We need to customize the image I think to pin the chrome version to match chromedriver.

arlo-client/src/components/AuditFlow/BoardTable.tsx
MorganLove: Nice touch!

arlo-client/src/components/AuditFlow/index_members.test.tsx
MorganLove: @eventualbuddha Something like that could work. [It does actually have to be `async`](https://reactjs.org/docs/testing-recipes.html#act) according to the new pattern, though.

arlo-client/src/components/AuditFlow/BoardTable.tsx
MorganLove: @eventualbuddha The `ballots.length` is necessary to test that it's not an empty array, in which case it's not complete -- it's not even started yet.

`ballots.some(b => b.status !== 'AUDITED')` would return `true` if any of the ballots are incomplete, right? What we want is to test that *none* of them are incomplete. All of them have to be audited for the round to be done, which is what this variable measures.

arlo-client/src/components/AuditFlow/index.tsx
MorganLove: @eventualbuddha Typescrypt 3.7 looks awesome! A separate refactor PR that upgrades to it would be awesome.

arlo-client/src/components/AuditFlow/BoardTable.tsx
eventualbuddha: `Array#some` will always return `false` for an empty array, so that covers the `ballots.length` part. And `ballots.every(b => b.status === 'AUDITED')` checks that _all_ ballots have status `AUDITED`. `ballots.some(b => b.status !== 'AUDITED')` checks that _any_ ballot has a status other than `AUDITED`. Those are equivalent statements, right?

arlo-client/src/components/AuditFlow/index.tsx
eventualbuddha: For sure not part of this. Just wanted to nudge for later 😄

arlo-client/src/components/AuditFlow/index_members.test.tsx
eventualbuddha: My example helper is async. Did you mean "does not actually have to be `async`"? Anyway, this was just a suggestion, and I do tend to have a higher tolerance for repeated code in tests so I think it's fine as-is.

arlo-client/src/components/AuditFlow/index_members.test.tsx
```diff
@@ -41,44 +42,32 @@ describe('AuditFlow initial load', () => {
...
-    let utils: RenderResult
-    await act(async () => {
-      utils = render(
-        <Router {...routeProps}>
-          <AuditFlow {...routeProps} />
-        </Router>
-      )
-    })
-    const { container } = utils!
+    const { container } = await utilities.asyncActRender(
```
eventualbuddha: Oh, I see you did take my suggestion. Cool 😄

PR #264, opened 2020-02-10
Moving asyncActRender from utilities to testUtilities

MorganLove:
**DESCRIPTION**

- Moving `asyncActRender` from utilities to testUtilities so that `@testing-library/react` can stay in devDependencies.

EDIT: Still moving it to testUtilities, but evidently it needs to be in regular dependencies regardless. Oh well.

**New Tests**

- N/A

**Checklist**

*Function*

*Style*

*Workflow*

### DIFF
--- a/arlo-client/package.json
+++ b/arlo-client/package.json
@@ -7,6 +7,7 @@
     "@blueprintjs/core": "^3.20.0",
     "@blueprintjs/table": "^3.8.1",
     "@testing-library/jest-dom": "^4.0.0",
+    "@testing-library/react": "^9.3.2",
     "@types/jest": "^24.0.15",
     "@types/jsdom": "^12.2.4",
     "@types/jspdf": "^1.3.1",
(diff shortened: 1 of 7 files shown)


arlo-client/package.json
```diff
@@ -7,6 +7,7 @@
     "@blueprintjs/core": "^3.20.0",
     "@blueprintjs/table": "^3.8.1",
     "@testing-library/jest-dom": "^4.0.0",
+    "@testing-library/react": "^9.3.2",
```
eventualbuddha: The description says:

> Moving `asyncActRender` from utilities to testUtilities so that `@testing-library/react` can stay in devDependencies.

But this is moving it to `dependencies`.

MorganLove: @eventualbuddha Yeah, I initially thought that would resolve it on its own, but evidently the build still required it in the regular dependencies. I updated the comment to reflect that.

PR #266, opened 2020-02-14
Docs update

MorganLove:
**DESCRIPTION**

- Cleaning up outdated documentation in /arlo-client
- Documenting api interaction for the data entry flow and the separation of offline and online auditing

**New Tests**

N/A

**Checklist**

*Function*

*Style*

*Workflow*


arlo-client/docs/api_sequence.md
```diff
@@ -764,6 +731,370 @@ happened on the first round, would look like this:
...
+- `src/components/AuditForms/CalculateRiskMeasurement.tsx` >
+  `CalculateRiskMeasurement` > `calculateRiskMeasurement()` > `api()`
+
+```
+{
+	"contests": [
+		{
+			id: "contest-1",
+   			results: {
+				"candidate-1": 0, // The backend doesn't care about these values, since they are supplied separately
```
eventualbuddha: That seems odd. Why are we requiring and documenting them if they're ignored? Or is this the response?

arlo-client/docs/api_sequence.md
```diff
@@ -764,6 +731,370 @@ happened on the first round, would look like this:
...
+### Initial data load
+
+Upon loading the data entry portal for an audit board, the data for the audit
+and the complete list of ballots are both fetched.
+
+GET the updated status:
+
+- `GET /election/{electionId}/audit/status`
+
+- `src/components/AuditFlow/index.tsx` > `AuditFlow` > `getSTatus()` > `api()`
```
eventualbuddha:
```suggestion
- `src/components/AuditFlow/index.tsx` > `AuditFlow` > `getStatus()` > `api()`
```

arlo-client/docs/api_sequence.md
```diff
@@ -764,6 +731,370 @@ happened on the first round, would look like this:
...
+  ```
+
+  Once this is accepted and the above endpoints are pinged again to update the
+  data.
+
+  GET the updated status:
+
+- `GET /election/{electionId}/audit/status`
+
+- `src/components/AuditFlow/index.tsx` > `AuditFlow` > `getSTatus()` > `api()`
```
eventualbuddha:
```suggestion
- `src/components/AuditFlow/index.tsx` > `AuditFlow` > `getStatus()` > `api()`
```

arlo-client/docs/api_sequence.md
```diff
@@ -764,6 +731,370 @@ happened on the first round, would look like this:
...
+          id: 'audit-board-1',
+          name: 'Audit Board #1',
+          members: [ // The members array isn't empty anymore!
+            {
+              name: "Member A",
+              affiliation: "IND",
+            },
+            {
+              name: "Member B",
+              affiliant: ""
```
eventualbuddha: Do we care about single vs. double quotes in these? I'd propose we use the `jsonc` language (JSON with comments) for these docs so that they're syntax highlighted correctly. That'd require quoting the keys, but I think having prettier format them automatically would be worth it.

arlo-client/docs/api_sequence.md
```diff
@@ -200,7 +198,7 @@ Form two has three separate POST calls.
 
 ```
 {
-	'size': 578,
+	'size': '578',
```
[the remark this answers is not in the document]
MorganLove: @eventualbuddha Yeah, it's parsed from string to int on the backend. It's easy enough to convert it to a number on the frontend like we do in other places, and I think we did it that way here at one point, but changed to this for some reason. I might be remembering wrong, though, or the reason may no longer be valid. But the current behavior is to send it back as a string, hence updating the docs to reflect that.

arlo-client/docs/api_sequence.md
```diff
@@ -37,22 +36,23 @@
...
-- `src/components/AuditForms/EstimateSampleSize.tsx` > `EstimateSampleSize` >
-  `handlePost()` > `api()`
-
 - `src/components/AuditForms/EstimateSampleSize.tsx` > `EstimateSampleSize` >
   `handlePost()` > `api()`
 
 ```
 {
 	name: "Primary 2019",
+  online: true,
```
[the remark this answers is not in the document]
MorganLove: @eventualbuddha Agreed. I use two spaced soft tabs myself, but I think there are still some tabs running around. I did a quick search and replace in this file after you commented and found a bunch. Definitely should standardize with prettier on this.

arlo-client/docs/api_sequence.md
MorganLove: @eventualbuddha I switched these to single quotes in here, but we should definitely decide on a prettier standard, yeah.

arlo-client/docs/api_sequence.md
MorganLove: @eventualbuddha This is what gets sent and what prompts the calculation for the next round. It definitely is odd, yes. It's leftover from before we separated the offline and online data entry flows, and should be updated to something that makes more sense. I didn't even realize this was still done this way until I was walking through and documenting, haha.

arlo-client/docs/api_sequence.md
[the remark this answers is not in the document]
eventualbuddha: https://github.com/votingworks/arlo/pull/269

arlo-client/docs/api_sequence.md
eventualbuddha: https://github.com/votingworks/arlo/pull/269

arlo-client/docs/api_sequence.md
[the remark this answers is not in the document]
eventualbuddha: https://github.com/votingworks/arlo/issues/270

arlo-client/docs/api_sequence.md
eventualbuddha: https://github.com/votingworks/arlo/issues/271

PR #248, opened 2020-01-05
Authentication proof of concept [wip]

MorganLove:
**DESCRIPTION**

- Adding in integration with Auth0

**New Tests**

- None yet

**Checklist**

*Function*

[] Code is fully functional and ready for review (or is tagged WIP)

[] No errors/warnings in the browser console ( ﾟヮﾟ)/

[] package.json is committed if I added or removed any dependencies

[] Tests added as needed to ensure code base still has 100% passing coverage

[] Tests added as needed to cover new functionality

*Style*

[] Passes linter without errors

[] Code is easily understandable and self-documenting (using naming conventions in keeping with the rest of the codebase)

[] Interfaces all have 'I' before their names (as in `IAudit`)

[] Comments are only included as needed to explain the 'why' instead of the 'what'

[] Common standards and coding practices are adhered to

[] DRY: Don't Repeat Yourself. Code is reasonably structured to avoid repetition

[] Code is intuitively structured following a Separation of Concerns

[] No class components; only functional components (using Hooks as needed)

PR #267, opened 2020-02-15
List audits on create audit view [wip]

MorganLove:
**DESCRIPTION**

- Adds a simple (temporary) route for fetching all the elections and lists them on the Create Audit view so you can select one and view it instead of creating a new audit.
- Adds `created_at` to the Elections model so we can sort them by creation time.

Notes:

This is an initial setup. When we have users linked to elections on #248 we can filter these elections by user.

**New Tests**

- TODO

**Checklist**

*Function*

*Style*

*Workflow*

### DIFF
--- a/app.py
+++ b/app.py
@@ -299,7 +299,7 @@ if ADMIN_PASSWORD:
 
 @app.route('/elections', methods=["GET"])
 def audits_list():
-    elections_list = Election.query.all()
+    elections_list = Election.query.order_by(Election.created_at).all()
 
     return jsonify(
         elections = [
--- a/app.py
+++ b/app.py
@@ -297,6 +297,21 @@ if ADMIN_PASSWORD:
         result = "\n".join(["%s - %s" % (e.id, e.name) for e in elections])
         return Response(result, content_type='text/plain')
 
+@app.route('/elections', methods=["GET"])
+def audits_list():
+    elections_list = Election.query.all()
+
+    return jsonify(
+        elections = [
+            {
+                "id": e.id,
+                "name": e.name,
+                "date": e.created_at
+            }
+            for e in elections_list
+        ]
+    )
+
 @app.route('/election/new', methods=["POST"])
 def election_new():
     election_id = create_election()
--- a/arlo-client/src/components/CreateAudit.tsx
+++ b/arlo-client/src/components/CreateAudit.tsx
@@ -1,7 +1,9 @@
 import React, { useState, useEffect, useCallback } from 'react'
 import styled from 'styled-components'
+import moment from 'moment'
 import { toast } from 'react-toastify'
-import { RouteComponentProps } from 'react-router-dom'
+import { RouteComponentProps, Link } from 'react-router-dom'
+import { RadioGroup, Radio } from '@blueprintjs/core'
 import FormButton from './Form/FormButton'
 import { api, checkAndToast } from './utilities'
 import { ICreateAuditParams, IErrorResponse } from '../types'
@@ -32,6 +34,8 @@ interface IElections {
 
 const CreateAudit = ({ history }: RouteComponentProps<ICreateAuditParams>) => {
   const [loading, setLoading] = useState(false)
+  const [electionId, setElectionId] = useState('')
+
   const onClick = async () => {
     try {
       setLoading(true)
@@ -66,6 +70,7 @@ const CreateAudit = ({ history }: RouteComponentProps<ICreateAuditParams>) => {
     const list = await getStatus()
     setLoading(true)
     setElections(list)
+    setElectionId(list.elections[0].id)
     setLoading(false)
   }, [getStatus])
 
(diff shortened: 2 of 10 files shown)


MorganLove:
It is passing locally on wdio, but not on heroku for some reason. Not immediately critical, though, since it's a WIP anyway and further test refinement will need to be done when it's integrated with the authentication.

arlo-client/src/components/CreateAudit.tsx
```diff
@@ -41,6 +54,35 @@ const CreateAudit = ({ history }: RouteComponentProps<ICreateAuditParams>) => {
...
+  const updateElections = useCallback(async () => {
+    const list = await getStatus()
+    setLoading(true)
+    setElections(list)
+    setElectionId(list.elections.length ? list.elections[0].id : '')
+    setLoading(false)
+  }, [getStatus])
+
+  useEffect(() => {
+    updateElections()
```
eventualbuddha: This setup is prone to race conditions. See this section on [Traditional Approaches vs. Suspense](http://reactjs.org/docs/concurrent-mode-suspense.html#traditional-approaches-vs-suspense) for an illustration of why. In _this particular case_ it probably won't be an issue, but it's an approach prone to error and should be avoided.

arlo-client/src/components/CreateAudit.tsx
```diff
@@ -41,6 +54,35 @@ const CreateAudit = ({ history }: RouteComponentProps<ICreateAuditParams>) => {
...
+    }
+    return { elections: [] }
+  }, [])
+
+  const updateElections = useCallback(async () => {
+    const list = await getStatus()
+    setLoading(true)
+    setElections(list)
+    setElectionId(list.elections.length ? list.elections[0].id : '')
+    setLoading(false)
```
eventualbuddha: This `setLoading` call will negate the `setLoading` call above, right? I'm reasonably sure React will not re-render for each of these state changes but will coalesce them.

arlo-client/src/components/CreateAudit.tsx
```diff
@@ -55,6 +97,32 @@ const CreateAudit = ({ history }: RouteComponentProps<ICreateAuditParams>) => {
...
+          <Link
+            to={`/election/${electionId}`}
+            className="bp3-button bp3-intent-primary"
+          >
+            View Audit
+          </Link>
+        </>
+      ) : (
+        <p>
+          You do not have any audits currently associated with your account.
```
eventualbuddha: Do we want to distinguish "you have no audits" from "there was an error fetching audits"?

arlo-client/src/components/CreateAudit.tsx
```diff
@@ -55,6 +97,32 @@ const CreateAudit = ({ history }: RouteComponentProps<ICreateAuditParams>) => {
...
+        <>
+          <RadioGroup
+            label="Select an existing audit:"
+            selectedValue={electionId}
+            onChange={e => setElectionId(e.currentTarget.value)}
+          >
+            {elections.elections.map(e => (
+              <Radio key={e.id} value={e.id}>
+                {e.name || <i>Not named yet</i>} (Created&nbsp;
+                {moment(e.date).format('MM/DD/YYYY')})
```
eventualbuddha: No need for `moment`, and we get better localization support:

```suggestion
                {new Intl.DateTimeFormat().format(e.date)})
```

arlo-client/src/components/CreateAudit.tsx
```diff
@@ -55,6 +97,32 @@ const CreateAudit = ({ history }: RouteComponentProps<ICreateAuditParams>) => {
       >
         Create a New Audit
       </Button>
+      {elections.elections.length > 0 ? (
+        <>
+          <RadioGroup
```
eventualbuddha: Why radio buttons instead of just a series of links?

arlo-client/src/components/CreateAudit.tsx
```diff
@@ -21,8 +23,19 @@ const Wrapper = styled.div`
   }
 `
 
+interface IElection {
+  id: string
+  name: string
+  date: string
```
eventualbuddha: This is a bit ambiguous. Is it the date of the election? The date the record was created? It's a bit hard to tell from this definition.

app.py
```diff
@@ -297,6 +297,21 @@ def admin():
         result = "\n".join(["%s - %s" % (e.id, e.name) for e in elections])
         return Response(result, content_type='text/plain')
 
+@app.route('/elections', methods=["GET"])
```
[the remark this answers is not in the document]
MorganLove: @eventualbuddha  That's likely how we will do it, yes. The current plan as I understand it is to return all of the audits that the currently signed-in user has access to view.

arlo-client/src/components/CreateAudit.tsx
MorganLove: @eventualbuddha That one's up to @mcchilders . This was how she had it in her mockup. :slightly_smiling_face:

benadida:
after discussion with @MorganLove:
- let's first land the backend API that lists the audits for the logged-in user
- then this PR can be rebased and use that API call to populate its list of audits.

PR #285, opened 2020-03-02
refactor: separate app into multiple files

eventualbuddha:
Separating the `app` object creation, `db` object creation, and route creation will make it easier to import the parts we need when we start putting things into more files. As-is, it's super easy to accidentally create import cycles that are annoying and difficult to break.


config/__init__.py
```diff
@@ -26,4 +26,4 @@ def read_database_url_config() -> str:
 
 DATABASE_URL = read_database_url_config()
 
-STATIC_FOLDER = 'arlo-client/public/' if os.environ.get('FLASK_ENV') == 'test' else 'arlo-client/build/'
+STATIC_FOLDER = os.path.normpath(os.path.join(__file__, '..', '..', 'arlo-client', 'public' if os.environ.get('FLASK_ENV') == 'test' else 'build'))
```
jonahkagan: Was this fixing a specific issue or just preventative? Just curious

Also this line is getting a bit confusing to read for me - do we have any formatting standards that might help either break it up onto multiple lines with indentation or break it up into multiple statements

eventualbuddha: This was actually required, and I don't have a very clear sense why. However, I think it accidentally worked before because the docs say it should be an absolute path. So now it is.

Formatting is one of the issues I'd like to tackle at some point. There are a few different competing options for python formatting, and I don't have a good sense which to use.

arlo-client/package.json
```diff
@@ -62,7 +62,7 @@
     "test:ci": "CI=true tsc && react-scripts test --maxWorkers 2 --env=jsdom --coverage && yarn test:wdio",
     "test:update": "react-scripts test -u  --watchAll=false",
     "test:coverage": "react-scripts test --coverage --watchAll=false",
-    "test:wdio": "pipenv run python ../app.py & pipenv run python ../bgcompute.py & react-scripts start & wdio wdio.conf.js",
+    "test:wdio": "pipenv run python ../app/__init__.py & pipenv run python ../bgcompute.py & react-scripts start & wdio wdio.conf.js",
```
eventualbuddha: I don't love this. WDYT, @benadida?

benadida: I'm not an expert at python idioms, but my understanding is that one usually wants to have a `.py` file parallel with `./app` that loads the stuff and runs it, so you're never digging into `./app` manually

config/__init__.py
jonahkagan: I don't know the formatting options too well. The last codebase I was working on used one and had an autoformatter and wow did that save some mental energy! A whole aspect of coding I never had to think about again. So I'm excited about that now. I'd be happy to pick one randomly or just let someone else who knows more decide

eventualbuddha:
> my only nit is figuring out how to never run `__init__.py` directly, typically by having a high-level file that imports the whole `app/` package.

Alright, this should be addressed.

PR #294, opened 2020-03-03
feat: enable sessions

eventualbuddha:
This provides a cookie-based application session with a secret key from `ARLO_SESSION_SECRET`. If not provided in development or test, a fixed one will be used instead.

### DIFF
--- a/arlo_server/base.py
+++ b/arlo_server/base.py
@@ -1,4 +1,5 @@
 from flask import Flask
-from config import STATIC_FOLDER
+from config import STATIC_FOLDER, SESSION_SECRET
 
 app = Flask(__name__, static_folder=STATIC_FOLDER)
+app.secret_key = SESSION_SECRET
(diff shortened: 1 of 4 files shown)


arlo_server/base.py
```diff
@@ -1,4 +1,5 @@
 from flask import Flask
-from config import STATIC_FOLDER
+from config import STATIC_FOLDER, SESSION_SECRET
 
 app = Flask(__name__, static_folder=STATIC_FOLDER)
+app.secret_key = SESSION_SECRET
```
benadida: does this work? It's not `app.config['SECRET_KEY']` ?

eventualbuddha: They're equivalent, apparently: https://flask.palletsprojects.com/en/1.1.x/api/#flask.Flask.secret_key.

eventualbuddha:
> might be worth adding a test to see if you can hit a URL that uses `session` without crashing. Would test for the question I have about whether `.secret_key` works.

I checked in my endpoint I was using to test it manually and wrote a test using it.

PR #297, opened 2020-03-03
Fix production build

benadida:
flask needs a trailing slash on static_folder for some reason.

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -41,6 +41,11 @@ def test_index(client):
     assert b'Arlo (by VotingWorks)' in rv.data
 
 
+def test_static_logo(client):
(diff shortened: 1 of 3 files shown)


tests/test_app.py
```diff
@@ -41,6 +41,13 @@ def test_index(client):
     assert b'Arlo (by VotingWorks)' in rv.data
 
 
+def test_static_logo(client):
+    rv = client.get('/arlo.png')
+    import config
+    print(config.STATIC_FOLDER)
```
jonahkagan: did you mean to leave this `print`?

benadida: nope, thanks for catching it :)

PR #300, opened 2020-03-04
Status indicator box

MorganLove:
**DESCRIPTION**

- Adding a status indicator box at the top of the view

Note: For now, it only handles the relevant statuses for our current flow, and is viewable by anyone who is managing the audit, since we don't have separate user authorization and multiple jurisdictions yet. But It will be easy to add onto it as we implement those features.

**New Tests**

- Snapshots and downloading reports from top button 

**Checklist**

*Function*

*Style*

*Workflow*

### DIFF
--- a/arlo-client/src/components/AuditForms/StatusBox.tsx
+++ b/arlo-client/src/components/AuditForms/StatusBox.tsx
@@ -1,0 +1,96 @@
+import React from 'react'
+import styled from 'styled-components'
+import { Callout, H4 } from '@blueprintjs/core'
+import { IAudit } from '../../types'
+import FormButton from '../Form/FormButton'
+
+const Wrapper = styled(Callout)`
+  .details {
+    width: 50%;
+  }
+  .wrapper {
+    margin: 10px 0 10px 0;
+    text-align: right;
+  }
+`
+
+const generateStatuses = (
+  audit: IAudit,
+  launched: boolean,
+  started: boolean,
+  complete: boolean
+): [string, string | undefined] => {
+  const STATUSES = {
+    audit: [
+      'The audit has not launched.',
+      'The audit has not started.',
+      `Round ${audit.rounds.length} of the audit is in progress.`,
+      'The audit is complete.',
+    ],
+    setup: [
+      'Audit setup is not complete.',
+      'Audit setup is complete.',
+      undefined,
+    ],
+  }
+
+  if (!launched) {
+    // during setup before launch
+    return [STATUSES.audit[0], STATUSES.setup[0]]
+  } else if (launched && !started) {
+    // after setup before first round
+    return [STATUSES.audit[1], STATUSES.setup[1]]
+  } else if (launched && started) {
+    // rounds have started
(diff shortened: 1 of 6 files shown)


arlo-client/src/components/AuditForms/StatusBox.tsx
```diff
@@ -0,0 +1,94 @@
...
+    setup: [
+      'Audit setup is not complete.',
+      'Audit setup is complete.',
+      undefined,
+    ],
+  }
+
+  if (complete) {
+    // all jurisdictions have completed the audit
+    return [STATUSES.audit[3], STATUSES.setup[2]]
```
benadida: using status indexes here is fragile and hard to follow, what if you add one? I'd recommend using objects, e.g. `STATUSES.audit.in_progress`.

arlo-client/src/components/AuditForms/StatusBox.tsx
```diff
@@ -0,0 +1,94 @@
...
+    complete
+  )
+
+  return (
+    <Wrapper>
+      <H4>{auditStatus}</H4>
+      <div className="details">{setupStatus && <p>{setupStatus}</p>}</div>
+      <div className="wrapper">
+        <FormButton
+          disabled={!complete}
```
benadida: should we allow the report to be downloaded even if the audit isn't done?

arlo-client/src/components/AuditForms/StatusBox.tsx
eventualbuddha: Definitely agree with @benadida here.

eventualbuddha: Or just remove the indirection altogether:

```ts
const generateStatuses = (
  audit: IAudit,
  launched: boolean,
  started: boolean,
  complete: boolean
): [string, string | undefined] => {
  if (complete) {
    // all jurisdictions have completed the audit
    return ['The audit is complete.', undefined]
  }
  if (launched && started) {
    // rounds have started
    return [
      `Round ${audit.rounds.length} of the audit is in progress.`,
      undefined,
    ]
  }
  if (launched && !started) {
    // after setup before first round
    return ['The audit has not started.', 'Audit setup is complete.']
  }
  /* istanbul ignore next */
  return ['The audit has not launched.', 'Audit setup is not complete.']
}
```

MorganLove: @eventualbuddha and @benadida I agree. I initially set it up as the array structure to help me visualize the flow, but you are right that it's better to do it without in the final product.

arlo-client/src/components/AuditForms/StatusBox.tsx
MorganLove: @mcchilders Thoughts?

Would the scripts for generating the report work at earlier stages?

MorganLove:
@mcchilders I'm thinking of redoing this to incorporate in the new changes when I rebase it. Just checking on how high priority that is and where in this week's timeline it would be best to slot that.

Issue #301, opened 2020-03-05
ability to freeze audit settings before starting rounds


benadida:
As per discussion with @mcchilders, we're going to add a `frozen_at` timestamp on the overall audit data structure, and an explicit `/freeze` API call that locks in all the data and creates the first `round`, without starting that round just yet.

nikhilb4a:
Closing as completed. 

Claude analysis: 

Issue #301 is closable — it was implemented the same day it was filed, and then the whole concept was consciously superseded. The postable version:

This was implemented essentially immediately: commit 3f1ba1a9 (March 4, 2020 — the same evening the issue was filed) added the frozen_at flag and "teased apart the submission of basic audit data from the freezing/creation of the first round," exactly as described. The issue just never got closed.

The freeze concept was later deliberately removed along with the single-jurisdiction flow it belonged to (PR #815; the migration dropping the column is [b5fcf654c681](https://github.com/votingworks/arlo/blob/main/server/migrations/versions/b5fcf654c681_remove_single_jurisdiction_flow.py#L27)). The multi-jurisdiction architecture that replaced it covers both halves of the original intent through different mechanisms:

Locking in audit data: launching the audit (starting round 1) is now the atomic freeze point — settings, contests, and jurisdictions all reject changes with a 409 once rounds exist (e.g. [election_settings.py#L75](https://github.com/votingworks/arlo/blob/main/server/api/election_settings.py#L75), [contests.py#L243](https://github.com/votingworks/arlo/blob/main/server/api/contests.py#L243)) — with "Undo audit launch" (PR #1231) as the reversal.
Reviewing before committing (the reason to freeze without starting): the sample preview API, POST/GET /election/<id>/sample-preview ([sample_preview.py#L99](https://github.com/votingworks/arlo/blob/main/server/api/sample_preview.py#L99), PR #1824), lets an admin see the sample that would be drawn before launching.

PR #304, opened 2020-03-05
Freeze flag and API

benadida:
addresses #301, see that ticket for motivation.

@MorganLove this tweaks the API calls so that `/audit/basic` no longer starts the first round. It just updates the data. You have to explicitly call `/audit/freeze` to mark the audit frozen, and then the `rounds` property appears and the sample set sizes are computed.

### DIFF
--- a/arlo-client/src/components/AuditForms/EstimateSampleSize.tsx
+++ b/arlo-client/src/components/AuditForms/EstimateSampleSize.tsx
@@ -215,6 +215,18 @@ const EstimateSampleSize: React.FC<IProps> = ({
         setIsLoading(false)
         return
       }
+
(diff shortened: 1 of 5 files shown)


arlo-client/src/components/AuditForms/EstimateSampleSize.tsx
```diff
@@ -215,6 +215,18 @@ const EstimateSampleSize: React.FC<IProps> = ({
         setIsLoading(false)
         return
       }
+
+      const freezeResponse: IErrorResponse = await api(
```
benadida: @MorganLove in the case of this single-page flow, `freeze` is called immediately. In the multi-jurisdiction flow, it would be called later in the flow.

PR #305, opened 2020-03-05
test coverage checking

benadida:
We should check python test coverage and enforce it. This is a PR to get there.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,3 +42,6 @@ test-client:
 # To run a specific test: TEST=<test name> make test-server
 test-server:
 	FLASK_ENV=test python3 -m pipenv run python -m pytest -k '${TEST}' --ignore=arlo-client
+
+test-server-coverage:
+	python3 -m pipenv run python -m pytest --cov=arlo_server/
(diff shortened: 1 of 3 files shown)


eventualbuddha:
This doesn't actually enforce anything yet, right?

Makefile
```diff
@@ -42,3 +42,6 @@ test-client:
 # To run a specific test: TEST=<test name> make test-server
 test-server:
 	FLASK_ENV=test python3 -m pipenv run python -m pytest -k '${TEST}' --ignore=arlo-client
+
+test-server-coverage:
+	python3 -m pipenv run python -m pytest --cov=arlo_server/
```
eventualbuddha: Do we need `FLASK_ENV=test`? what about `--ignore=arlo-client`? Do we want `-k '${TEST}'`?

benadida: I don't think we should do coverage with paring down to one test, but yeah I think we need to tweak some other settings before we merge. I should have marked this as draft. Not planning on merging until we at least figure out how to cover everything properly.

jonahkagan:
Closing in favor of #451

PR #310, opened 2020-03-06
add elections in /auth/me for audit admins

benadida:


arlo_server/__init__.py
```diff
@@ -964,8 +964,14 @@ def me():
...
                        organizations=[{
-                           "id": org.id,
-                           "name": org.name
+                           "id":
+                           org.id,
+                           "name":
+                           org.name,
+                           "elections": [{
+                               "id": e.id,
+                               "name": e.name
```
MorganLove: @benadida Can we also get the `created_at` value as well, since @mcchilders wants them sorted by that and have the dates displayed as well on the list?

benadida: do we want `created_at` or do we want the stated date of the election and/or audit? We don't currently have the `created_at`.

benadida:
I've added some more data and standardized how the election data structures are returned, Let me know if that works @MorganLove or if you'd like more fields!

arlo_server/__init__.py
MorganLove: @benadida I added that field in my other PR for listing audits as an auto-populated field when an audit is created. Should we add that here?

https://github.com/votingworks/arlo/pull/267/files#diff-f219bfe69e6ed201e4bdfdb371dc0c9bR17

PR #311, opened 2020-03-06
Add audit board members to audit report

jonahkagan:
**DESCRIPTION**

Task: #237 
Adds a row to the audit board report with each audit board member's name and party affiliation, as well as the name of the audit board they were on.

**New Tests**

Adds test_report.py with a simple test that also covers the rest of the report content.

**Checklist**

*Function*

*Style*

*Workflow*

### DIFF
--- a/arlo_server/__init__.py
+++ b/arlo_server/__init__.py
@@ -841,6 +841,16 @@ def audit_report(election_id):
     report_writer.writerow(["Risk Limit", "{:d}%".format(election.risk_limit)])
     report_writer.writerow(["Random Seed", election.random_seed])
 
+    for audit_board in jurisdiction.audit_boards:
+        report_writer.writerow([
+            audit_board.name, audit_board.member_1,
+            pretty_affiliation(audit_board.member_1_affiliation)
+        ])
+        report_writer.writerow([
+            audit_board.name, audit_board.member_2,
+            pretty_affiliation(audit_board.member_2_affiliation)
+        ])
+
     for round in election.rounds:
         round_contest = round.round_contests[0]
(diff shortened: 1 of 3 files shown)


arlo_server/__init__.py
```diff
@@ -841,6 +841,16 @@ def audit_report(election_id):
     report_writer.writerow(["Risk Limit", "{:d}%".format(election.risk_limit)])
     report_writer.writerow(["Random Seed", election.random_seed])
 
+    for audit_board in jurisdiction.audit_boards:
+        report_writer.writerow([
+            audit_board.name, audit_board.member_1,
```
benadida: this highlights that an audit board name might not be the nice "Audit Board #1", and coud bite us later. OK for now, I don't have a better solution.

tests/test_app.py
```diff
@@ -77,6 +77,28 @@ def test_whole_audit_flow(client):
     assert result2["riskLimit"] == 5
 
 
+def setup_audit_board(client, election_id, jurisdiction_id, audit_board_id):
```
benadida: nice, thanks for adding this.

arlo_server/__init__.py
jonahkagan: I'm not sure what you mean - I was just going off the data model, which has a name field. What might bite us?

benadida: I mean that if the audit board has a name of "Foobar", then the row in the CSV won't be clear that it's an audit board cause it has no title. Right now the audit boards have nice names, so it works okay.

For example, the previous line in the report says "Random Seed" explicitly.

Issue #303, opened 2020-03-05
API call to upload set of jurisdictions and admins


benadida:
- create/replace the jurisdictions
- create the users and mappings to jurisdictions

Likely a

`POST /election/<election_id>/jurisdictions_file`

or something like it.

PR #313, opened 2020-03-06
feat: add function to bulk-update election jurisdictions & admins

eventualbuddha:
Progress toward #303

### DIFF
--- a/models.py
+++ b/models.py
@@ -12,6 +12,8 @@ class Organization(db.Model):
     id = db.Column(db.String(200), primary_key=True)
     name = db.Column(db.String(200), nullable=False)
 
+    elections = relationship('Election', backref='organization', passive_deletes=True)
+
 
 class Election(db.Model):
     id = db.Column(db.String(200), primary_key=True)
@@ -67,7 +69,7 @@ class Jurisdiction(db.Model):
 class User(db.Model):
     id = db.Column(db.String(200), primary_key=True)
     email = db.Column(db.String(200), unique=True, nullable=False)
-    external_id = db.Column(db.String(200), unique=True, nullable=False)
+    external_id = db.Column(db.String(200), unique=True, nullable=True)
 
     organizations = relationship("Organization", secondary="audit_administration")
     jurisdictions = relationship("Jurisdiction", secondary="jurisdiction_administration")
--- a/tests/test_jurisdiction_bulk_update.py
+++ b/tests/test_jurisdiction_bulk_update.py
@@ -1,0 +1,66 @@
+from util.jurisdiction_bulk_update import bulk_update_jurisdictions
+from models import Election, Jurisdiction, JurisdictionAdministration, Organization, User  # type: ignore
+import arlo_server
+import pytest
+import uuid
+
+
+@pytest.fixture
+def db():
+    with arlo_server.app.app_context():
+        arlo_server.db.drop_all()
+        arlo_server.db.create_all()
+
+    yield arlo_server.db
+
+    arlo_server.db.session.commit()
+
+
+def test_first_update(db):
+    org = Organization(id=str(uuid.uuid4()), name='Test Org')
+    election = Election(id=str(uuid.uuid4()), organization=org)
+    new_admins = bulk_update_jurisdictions(db.session, election,
+                                           [('Jurisdiction #1', 'bob.harris@ca.gov')])
+    db.session.commit()
+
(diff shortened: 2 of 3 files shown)


models.py
```diff
@@ -12,6 +12,8 @@ class Organization(db.Model):
     id = db.Column(db.String(200), primary_key=True)
     name = db.Column(db.String(200), nullable=False)
 
+    elections = relationship('Election', backref='organization', passive_deletes=True)
```
benadida: lol, I have this too in my PR :)

models.py
```diff
@@ -67,7 +69,6 @@ class Jurisdiction(db.Model):
 class User(db.Model):
     id = db.Column(db.String(200), primary_key=True)
     email = db.Column(db.String(200), unique=True, nullable=False)
-    external_id = db.Column(db.String(200), unique=True, nullable=False)
```
benadida: I think I like this removal, but I wonder if it might be useful to eventually store the auth0 user id.

tests/test_jurisdiction_bulk_update.py
```diff
@@ -0,0 +1,64 @@
...
+
+@pytest.fixture
+def db():
+    with arlo_server.app.app_context():
+        arlo_server.init_db()
+
+    yield arlo_server.db
+
+    arlo_server.db.session.commit()
+    arlo_server.db.drop_all()
```
benadida: I wonder if it makes more sense to do a `.drop_all()` and `.create_all()` at the beginning of every test... just to make sure the DB always starts in the right state? Or does `init_db()` do that?

tests/test_jurisdiction_bulk_update.py
```diff
@@ -0,0 +1,64 @@
...
+    assert db.session.query(User).one() == user
+    assert db.session.query(Jurisdiction).one() == jurisdiction
+
+
+def test_remove_outdated_jurisdictions(db):
+    org = Organization(id=str(uuid.uuid4()), name='Test Org')
+    election = Election(id=str(uuid.uuid4()), organization=org)
+
+    # Add jurisdictions.
+    bulk_update_jurisdictions(db.session, election, [('Jurisdiction #1', 'bob.harris@ca.gov')])
```
benadida: I'm assuming you purposely decided to not `commit` in `bulk_update_jurisdictions` so that transactional control can remain outside of the function call. That's cool. You might want to add a comment about that by the method definition.

Adding a comment here to ask: do you want to maybe `.commit()` after this call before finishing your test? Or are you explicitly trying to test things within the current transaction? I could see it either way, but for some reason am leaning towards committing before querying and asserting.

models.py
eventualbuddha: Yes. But I didn't have an obvious reason to have it, and I wasn't sure what to set it to. I could make it nullable instead?

benadida: Yes good idea, make it nullable.

tests/test_jurisdiction_bulk_update.py
eventualbuddha: I was following the pattern from `test_app.py`, but I could do that instead.

tests/test_jurisdiction_bulk_update.py
eventualbuddha: > I'm assuming you purposely decided to not `commit` in `bulk_update_jurisdictions` so that transactional control can remain outside of the function call. That's cool. You might want to add a comment about that by the method definition.

That's true, though I realize now that I don't have a strong sense of how these sessions work. Is `db.session` always the default implicit session? Is `db.session` always the one used if you e.g. `User.query.count()`? Either way, my assertions always use the same session given to `bulk_update_jurisdictions` so it shouldn't require commiting before asserting. In my testing, probably the best reason to explicitly commit in the test before asserting is to trigger the SQL calls at a known point rather than implicitly elsewhere.

tests/test_jurisdiction_bulk_update.py
benadida: Ah, your call :)

eventualbuddha: Changed.

models.py
eventualbuddha: Done.

tests/test_jurisdiction_bulk_update.py
eventualbuddha: Added a comment to the docstring and added `commit` calls in the tests.

PR #314, opened 2020-03-06
Front end login and auth context [WIP]

MorganLove:
**DESCRIPTION**

- Adds authentication context for the front end
- Adds login buttons for the AA and JA flows

**New Tests**

- none yet

**Checklist**

*Function*

*Style*

*Workflow*


MorganLove:
@benadida It looks to me like it should work, but when I click the login buttons it redirects back to the home page and the backend logs the requests as 302s. (I am using `window.location.replace` instead of `Link` or an `a` element or `history.push()` because all of those were just returning a 404, for some reason). Any ideas where the problem might be?

arlo-client/src/components/CreateAudit.tsx
```diff
@@ -44,17 +47,48 @@ const CreateAudit = ({ history }: RouteComponentProps<ICreateAuditParams>) => {
...
+          <Button
+            type="button"
+            intent="primary"
+            fill
+            large
+            onClick={onClick}
+            loading={loading}
+            disabled={loading}
+          >
+            Create a New Audit
```
benadida: as per a discussion with @mcchilders, let's let a non-authenticated user create a new audit, too. So the create audit should be functional in both cases, not-logged-in and logged-in.

MorganLove: @benadida Gotcha! Can do.

MorganLove: @benadida Should audits created by a non-authenticated user be publicly accessible, like the old flow?

PR #315, opened 2020-03-06
scripts to create orgs and admins

benadida:


MorganLove:
@benadida This looks good, but it seems to be failing a BE typecheck.

PR #316, opened 2020-03-06
Backend refactor

umbernhard:
Refactors backend to be more functional. Adds a Contest data structure, and a few more tests.

### DIFF
--- a/arlo_server/__init__.py
+++ b/arlo_server/__init__.py
@@ -5,7 +5,7 @@ from flask import Flask, jsonify, request, Response, redirect, session
 from flask_httpauth import HTTPBasicAuth
 
 import sampler
-from audits import bravo as bravo
+from audits import bravo
 from werkzeug.exceptions import InternalServerError
 from xkcdpass import xkcd_password as xp
 
@@ -75,9 +75,9 @@ def contest_status(election):
     for contest in election.contests:
         name = contest.id
         info_dict = {
-            'ballots': contest.total_ballots_cast,
-            'numWinners': contest.num_winners,
-            'votesAllowed': contest.votes_allowed,
+            "ballots": contest.total_ballots_cast,
+            "numWinners": contest.num_winners,
+            "votesAllowed": contest.votes_allowed,
         }
 
         for choice in contest.choices:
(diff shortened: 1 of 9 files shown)


arlo_server/__init__.py
```diff
@@ -4,7 +4,8 @@
 from flask import Flask, jsonify, request, Response, redirect, session
 from flask_httpauth import HTTPBasicAuth
 
-from sampler import Sampler
+import sampler as sampler
```
eventualbuddha: Isn't this equivalent to `import sampler`?

umbernhard: yep! Will fix.

arlo_server/__init__.py
```diff
@@ -246,11 +258,14 @@ def check_round(election, jurisdiction_id, round_id):
...
     current_sample_results = sample_results(election)
 
-    risk, is_complete = sampler.compute_risk(round_contest.contest_id,
-                                             current_sample_results[round_contest.contest_id])
+    risk, is_complete = bravo.compute_risk(election.risk_limit / 100,
+                                           contest_status(election)[round_contest.contest_id],
+                                           current_sample_results[round_contest.contest_id])
+
+    print('Risk values: {}'.format(risk))
+    print('Audit is complete? {}'.format(is_complete))
```
eventualbuddha:
```suggestion
```

arlo_server/__init__.py
```diff
@@ -71,12 +73,17 @@ def contest_status(election):
...
-        contests[contest.id]["votesAllowed"] = contest.votes_allowed
+        name = contest.id
+        info_dict = {
+            "ballots": contest.total_ballots_cast,
+            "numWinners": contest.num_winners,
+            "votesAllowed": contest.votes_allowed,
+        }
+
+        for choice in contest.choices:
+            info_dict[choice.id] = choice.num_votes
```
jonahkagan: Why do we make the candidates top level fields? Didn't that lead to that bug before where the code thought `votesAllowed` was a candidate?

Any reason not to just pass in the original `contest` object to the `Contest` constructor?

umbernhard: Hmm, you're right re: candidates. That was a quickly written part of the code, so we didn't anticipate these kinds of problems. 

We could pass in the models.contest object, I just didn't /shrug

Since I merged this, that could be another PR.

PR #317, opened 2020-03-07
Home list of audits

MorganLove:
**Description**

- lists the audits currently associated with the authenticated user on the home page as link buttons
- POSTs the organization ID of an authenticated AA user to the `/election/new` endpoint so it can be associated with it

**New Tests**

- Testing of new endpoint usage
- Updated snapshots

**Notes**

- When new audits are created, they aren't associated with the authenticated user, so that needs to be done on this PR for it to be functional. @benadida can you do that? EDIT: this will take place on the backend in another PR to match up with this one.

-As it is currently, I am injecting dummy data in to demonstrate the styling. That will be removed before this is merged. EDIT: now cleaned up.

### DIFF
--- a/arlo-client/src/components/CreateAudit.test.tsx
+++ b/arlo-client/src/components/CreateAudit.test.tsx
@@ -93,7 +93,8 @@ describe('CreateAudit', () => {
       expect(apiMock).toBeCalledTimes(2)
       expect(apiMock).toHaveBeenNthCalledWith(2, '/election/new', {
         method: 'POST',
-        body: JSON.stringify({ organization: 'org-id' }),
+        // eslint-disable-next-line @typescript-eslint/camelcase
+        body: JSON.stringify({ organization_id: 'org-id' }),
       })
       expect(historySpy).toBeCalledTimes(1)
       expect(historySpy).toHaveBeenNthCalledWith(1, '/election/1')
--- a/arlo-client/src/components/CreateAudit.test.tsx
+++ b/arlo-client/src/components/CreateAudit.test.tsx
@@ -1,11 +1,12 @@
 import React from 'react'
 import { render, fireEvent, wait } from '@testing-library/react'
 import { toast } from 'react-toastify'
-import { RouteComponentProps } from 'react-router-dom'
+import { RouteComponentProps, BrowserRouter as Router } from 'react-router-dom'
 import CreateAudit from './CreateAudit'
 import { ICreateAuditParams } from '../types'
 import { routerTestProps } from './testUtilities'
 import * as utilities from './utilities'
+import AuthDataProvider from './UserContext'
 
 const apiMock: jest.SpyInstance<
   ReturnType<typeof utilities.api>,
@@ -42,7 +43,7 @@ describe('CreateAudit', () => {
     expect(container).toMatchSnapshot()
   })
 
-  it('calls the /election/new endpoint', async () => {
+  it('calls the /election/new endpoint for nonauthenticated user', async () => {
     apiMock.mockImplementation(async () => ({ electionId: '1' }))
     const { getByText } = render(<CreateAudit {...routeProps} />)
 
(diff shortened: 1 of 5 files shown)


benadida:
@MorganLove good way to scale the API would be for you to send an org id in the election creation call as a post parameter. Backend will check that the user is an admin for that org and associate the election accordingly. Will add that in a small separate PR.

MorganLove:
> @MorganLove good way to scale the API would be for you to send an org id in the election creation call as a post parameter. Backend will check that the user is an admin for that org and associate the election accordingly. Will add that in a small separate PR.

@benadida Since it's an array of organizations, should I just send back the id of the first one in the array? Or do you think we might need to have separate Create Audit buttons for each organization in the authenticated user's array?

benadida:
> > @MorganLove good way to scale the API would be for you to send an org id in the election creation call as a post parameter. Backend will check that the user is an admin for that org and associate the election accordingly. Will add that in a small separate PR.
> 
> @benadida Since it's an array of organizations, should I just send back the id of the first one in the array? Or do you think we might need to have separate Create Audit buttons for each organization in the authenticated user's array?

In almost every case, there's only going to be one. So let's start by making it the first of the array, and if we need to have an interstitial screen at some point to pick the org, we can do it then.

arlo-client/src/components/CreateAudit.tsx
```diff
@@ -22,17 +22,30 @@ const Wrapper = styled.div`
...
 const CreateAudit = ({ history }: RouteComponentProps<ICreateAuditParams>) => {
-  const { isAuthenticated } = useAuthDataContext()
+  const { isAuthenticated, meta } = useAuthDataContext()
 
   const [loading, setLoading] = useState(false)
   const onClick = async () => {
     try {
       setLoading(true)
+      const data = isAuthenticated
+        ? { organization: meta!.organizations[0].id }
```
MorganLove: @benadida Here you go! :slightly_smiling_face: Let me know if there's anything else you need on this endpoint.

arlo-client/src/components/CreateAudit.test.tsx
```diff
@@ -42,17 +43,152 @@ describe('CreateAudit', () => {
...
+      expect(apiMock).toHaveBeenNthCalledWith(1, '/auth/me')
+    })
+
+    fireEvent.click(getByText('Create a New Audit'), { bubbles: true })
+
+    await wait(() => {
+      expect(apiMock).toBeCalledTimes(2)
+      expect(apiMock).toHaveBeenNthCalledWith(2, '/election/new', {
+        method: 'POST',
+        body: JSON.stringify({ organization: 'org-id' }),
```
benadida: this should probably be `organization_id`

MorganLove: @benadida Done!

arlo-client/src/components/CreateAudit.tsx
```diff
@@ -89,6 +103,33 @@ const CreateAudit = ({ history }: RouteComponentProps<ICreateAuditParams>) => {
...
+              ))
+            )}
+          {meta!.jurisdictions.length > 0 &&
+            meta!.jurisdictions.map(({ election }) => (
+              <AuditLink
+                to={`/election/${election.id}`}
+                key={election.id}
+                className="bp3-button bp3-intent-primary"
+              >
+                {election.name || 'Not named yet'}
```
eventualbuddha:
```suggestion
                {election.name ?? 'Not named yet'}
```

arlo-client/src/components/CreateAudit.tsx
```diff
@@ -89,6 +103,33 @@ const CreateAudit = ({ history }: RouteComponentProps<ICreateAuditParams>) => {
...
+        <>
+          {meta!.organizations.length > 0 &&
+            meta!.organizations.map(o =>
+              o.elections.map(election => (
+                <AuditLink
+                  to={`/election/${election.id}`}
+                  key={election.id}
+                  className="bp3-button bp3-intent-primary"
+                >
+                  {election.name || 'Not named yet'}
```
eventualbuddha:
```suggestion
                  {election.name ?? 'Not named yet'}
```

arlo-client/src/components/CreateAudit.tsx
MorganLove: @eventualbuddha I see nullish coalescing coming in TS 3.7, but we're still on 3.5 afaik. Also, I don't think this is a good use of it, specifically, since we would rather have `'Not named yet'` than an empty string, for example. Unless I'm missing something here.

eventualbuddha: You're right, I'd misremembered that I'd updated Arlo to 3.7. That should be pretty low-hanging fruit now that `react-scripts` is updated. And in this particular case where we expect `election.name` might be an empty string then `||` makes more sense. In general I'd strive to use `??` instead and ensure the LHS, if defined, has a value we want to use.

PR #318, opened 2020-03-08
Toggle online/offline repositioning

MorganLove:
**Description**

- Moves the form element from the first form to the second form
- Sends `true` as the default value from the first form so the backend doesn't freak out
- Sends an empty array for `contests` when it POSTs to the `/audit/basic` endpoint a second time, so that the backend doesn't freak out
- Adds an if condition to the backend so that it doesn't delete the contests if no new ones are passed back

**New Tests**

- Updates snapshots
- Adjusts all the api mock tests to account for a new api call in the second form
- Tests for server and api errors on the new api call in the second form
- Updates the end to end tests to reflect the change as well

**Notes**

I have no idea if the stuff changed on the backend will affect the backend tests, or if they should. I also am not sure if my backend tweaks were done properly. It works with my manual testing, but someone more familiar back there should triple check everything, haha.

### DIFF
--- a/arlo_server/__init__.py
+++ b/arlo_server/__init__.py
@@ -398,7 +398,8 @@ def audit_basic_update(election_id):
     election.online = info['online']
 
     errors = []
-    db.session.query(TargetedContest).filter_by(election_id=election.id).delete()
+    if info['contests']:
+        db.session.query(TargetedContest).filter_by(election_id=election.id).delete()
 
     for contest in info['contests']:
         total_allowed_votes_in_contest = contest['totalBallotsCast'] * contest['votesAllowed']
--- a/arlo-client/src/components/AuditForms/EstimateSampleSize.test.tsx
+++ b/arlo-client/src/components/AuditForms/EstimateSampleSize.test.tsx
@@ -342,41 +342,6 @@ describe('EstimateSampleSize', () => {
     expect(container).toMatchSnapshot()
   })
 
-  it('changes audits to be offline and online', () => {
-    const { getByLabelText } = render(
-      <EstimateSampleSize
-        audit={statusStates[0]}
-        isLoading={false}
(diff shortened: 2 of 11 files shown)


arlo_server/__init__.py
```diff
@@ -398,7 +398,8 @@ def audit_basic_update(election_id):
     election.online = info['online']
 
     errors = []
-    db.session.query(TargetedContest).filter_by(election_id=election.id).delete()
+    if info['contests']:
+        db.session.query(TargetedContest).filter_by(election_id=election.id).delete()
```
eventualbuddha: There should be a backend test that verifies that this endpoint can be called without `contests` and that it just updates the `Election`.

MorganLove: @eventualbuddha Agreed, definitely. Should that be added into this PR by someone with more backend experience?

MorganLove:
> feels like this is screaming for a separate API call to set the contests, but in the interest of moving forward quickly, I think this approach is good. Can you file a followup ticket to separate the contests into a separate API call?

https://github.com/votingworks/arlo/issues/326 There you go!

MorganLove:
@jonahkagan @eventualbuddha @benadida How much of this should just be redone from scratch in light of the endpoint updates?

benadida:
> @jonahkagan @eventualbuddha @benadida How much of this should just be redone from scratch in light of the endpoint updates?

Since this is the single-page flow using older APIs, I don't think this needs to be redone in light of the endpoint updates, you can keep using the existing endpoints. I can see some rebase issues with frontend code, so that part is up to you, but no need to adapt these to the new API calls.

Please give a shout if you want help on the test for the backend.

jonahkagan:
Do we need this PR anymore? Or can we close?

PR #320, opened 2020-03-09
[WIP] make audit boards round-specific

benadida:
This is a medium-to-big lift, so right now it's a WIP. See #299 

Done so far:
- new APIs for audit-boards that are round-specific
- beginning of updating tests <-- this is non-trivial

What still needs to be done:
- fully updating all backend tests to match new APIs
- including splitting up when the audit boards are created, since that's now the jurisdiction's job at each round, and not part of the creation of the jurisdiction.
- including considering when `setup_next_round` and `sample_ballots` should be called in the server logic. I *think* `setup_next_round` shouldn't set the start time of the round, and setting the sample size should set the start time and call `sample_ballots`, but I'm not 100% sure, so check me.
- update the single-page flow front-end to match these new calls.


benadida:
choosing to go a simpler route in #341

PR #321, opened 2020-03-09
feat: add endpoint to handle an election jurisdictions file

### LINKED ISSUE #303: API call to upload set of jurisdictions and admins

eventualbuddha:
~This required a bit of a refactor in order to avoid cycles. Mainly, the application routes have been moved to a file called `routes.py` and a bunch of utility functions were moved to their own files.~

I managed to simply change the order of imports to avoid the cycles. That makes this PR more focused. Will follow up with a refactor later.

Closes #303

### DIFF
--- a/arlo_server/__init__.py
+++ b/arlo_server/__init__.py
@@ -13,8 +13,6 @@ from sqlalchemy.dialects.postgresql import aggregate_order_by
 
 from authlib.flask.client import OAuth
 
-from util.binpacking import Bucket, BalancedBucketList
-
 from arlo_server.base import app
 from arlo_server.db import db
 from models import *
@@ -22,6 +20,9 @@ from models import *
 from config import HTTP_ORIGIN
 from config import AUDITADMIN_AUTH0_BASE_URL, AUDITADMIN_AUTH0_CLIENT_ID, AUDITADMIN_AUTH0_CLIENT_SECRET
 from config import JURISDICTIONADMIN_AUTH0_BASE_URL, JURISDICTIONADMIN_AUTH0_CLIENT_ID, JURISDICTIONADMIN_AUTH0_CLIENT_SECRET
+
+from util.binpacking import Bucket, BalancedBucketList
+from util.jurisdiction_bulk_update import bulk_update_jurisdictions
 
 AUDIT_BOARD_MEMBER_COUNT = 2
 WORDS = xp.generate_wordlist(wordfile=xp.locate_wordfile())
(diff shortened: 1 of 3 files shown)


eventualbuddha:
@MorganLove can you take a look at this from an API perspective? Seem alright to you?

tests/test_app_upload_jurisdictions_file.py
```diff
@@ -0,0 +1,179 @@
...
+        })
+    assert rv.status_code == 200
+    assert json.loads(rv.data) == {'status': 'ok'}
+
+    election = Election.query.filter_by(id=election_id).one()
+    assert [j.name for j in election.jurisdictions] == ['J1', 'J2']
+
+    for jurisdiction in election.jurisdictions:
+        assert [a.user.email
+                for a in jurisdiction.jurisdiction_administrations] == ['a1@example.com']
```
jonahkagan: This is officially the best tested feature in the codebase.

tests/test_app_upload_jurisdictions_file.py
```diff
@@ -0,0 +1,179 @@
...
+    app.config['TESTING'] = True
+    client = app.test_client()
+
+    with app.app_context():
+        db.drop_all()
+        db.create_all()
+
+    yield client
+
+    db.session.commit()
```
jonahkagan: are `post_json` and `client` copy-pasted from `test_app.py`? I imported them in `test_report.py`. Do you think one way or the other is better? Or should we pull them out into a helper module?

arlo_server/__init__.py
```diff
@@ -315,6 +316,55 @@ def election_new():
     return jsonify(electionId=election_id)
 
 
+@app.route('/election/<election_id>/jurisdictions_file', methods=["GET"])
```
jonahkagan: I'm still wrapping my head around the API design a bit. Wondering if this should be part of the `/election/{id}/jurisdictions` API or if it's a separate API like you have it here.

tests/test_app_upload_jurisdictions_file.py
eventualbuddha: They were just copied, yeah. We could put them in `tests/utils.py` or something.

tests/test_app_upload_jurisdictions_file.py
eventualbuddha: 😂

arlo_server/__init__.py
eventualbuddha: I mostly just followed @benadida's suggestion from #303, though I think having it in `…/jurisdictions` could be a little confusing. We've been discussing with @mcchilders that elections officials have a mental model that includes "files" for things like the list of jurisdictions being the source of truth, and so we're trying to model that here.

eventualbuddha:
Alright, I'm merging so we can move ahead. @MorganLove, I'm happy to make changes in a follow up PR if it makes client integration easier.

PR #322, opened 2020-03-09
fix: remove references to undefined variable

eventualbuddha:
This was likely a copy-paste error.


eventualbuddha:
> Do we need some sort of enforcement of the linter to stop things like this from happening?

Yes. I started work on integrating pylint, and I'll continue it. However, there are a _lot_ of errors so I need to come up with a good way to deal with that.

PR #327, opened 2020-03-11
Format python with black

jonahkagan:
**Description**

Switches from `yapf` to `black` for Python formatting. `black` is waaay faster and also doesn't have some of the weird line-breaking quirks that `yapf` had.

I separated the configuration changes from the actual reformatting of all the code into separate commits, so may be easier to review commit by commit.

**Testing**

N/A

**Progress**

Ready for review

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
       "prettier --write"
     ],
     "*.py": [
-      "python3 -m pipenv run yapf --in-place"
+      "black"
     ],
     "package.json": [
       "sort-package-json"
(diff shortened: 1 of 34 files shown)


package.json
```diff
@@ -15,7 +15,7 @@
       "prettier --write"
     ],
     "*.py": [
-      "python3 -m pipenv run yapf --in-place"
+      "make format-python"
```
eventualbuddha: I think the way this works is that it passes the file names for the files changed after it. That wouldn’t work so well with `make` semantics.

jonahkagan: Oh, gotcha. Do you suggest changing it back to an explicit `black` command?

jonahkagan: I went ahead and fixed it

PR #328, opened 2020-03-11
Switch all python3 commands to python3.7

jonahkagan:
**Description**

I discovered that in my Ubuntu dev environment, `python3` is by default
`3.6.9`. To be safe, let's use `python3.7` explicitly.

**Testing**

N/A

**Progress**

Ready for review.
Note: needs to be merged after #327 because it assumes we've switched to black for `make format-python`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 
 deps:
 	sudo apt install python3.7 python3-pip nodejs libpython3.7-dev libpq-dev
-	python3 -m pip install pipenv
+	python3.7 -m pip install pipenv
 	sudo npm install -g yarn
 	sudo apt install postgresql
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -11,34 +11,34 @@ initdevdb:
...
 
 dev-environment: deps initdevdb install-development resetdb
 
 typecheck:
-	python3 -m pipenv run mypy .
+	python3.7 -m pipenv run mypy .
 
 format-python:
-	python3 -m pipenv run yapf --in-place --recursive .
+	python3.7 -m pipenv run black .
```
eventualbuddha: Unrelated change to `black`.

eventualbuddha: Ah, I see you addressed that in the description. The way I usually do that is to stack PRs, so you’d make this a PR based on that branch that targets that branch. Then you merge the first one, re-target this one, and merge it.

jonahkagan: Ah ok, that makes sense. I'll do that next time

Issue #239, opened 2019-12-30
Add ballot interpretations to audit report


mcchilders:
**User story:** As an audit admin I want a record of how each audited ballot was interpreted by the audit board, so that I can determine why any discrepancies may have occurred (e.g. data entry error on the part of the auditors, different interpretations of a ballot, etc.)

**CoAs:**
- When the audit admin selects "online" audit board data entry, the audit report should contain:
-- A row for each ballot that was sampled for audit (the sampled ballot data already appears in a single row, comma-separated, in the current report so that data identifying the sampled ballots can be repurposed here)
-- Which round(s) the ballot was sampled in (at least one, possibly more than 1) and the ticket numbers for each time it was sampled (e.g. something like - "Round 1: 0.0123, 0.0234" would indicate that the ballot was sampled twice in Round 1; "Round 1: 0.0123, Round 2: 0.0234" would indicate that the ballot was sampled once in Round 1, and once in Round 2)
-- Whether data was entered for that ballot yet, since ballots are sampled prior to the audit board entering data about them (Audited/Not audited)
-- If audited, what data was entered by the audit board. This could be: 1 or more candidate/choice selections OR no consensus/audit board can't agree OR ballot not found OR blank vote/no mark. If not yet audited, this field can be blank.
-- An indication of whether this is the initial entry or an edited (re-audited) entry for the ballot, since we now give audit boards the ability to go back and correct entries within a particular audit round. In either case, the most recent interpretation should show.
-- Any comments that were entered by the audit board for that ballot.

Note: I realize this is *a lot* of data, so if this requires us to move from .csv to .xls and start using tabs, that's fine with me. Whatever's cleanest/clearest.

jonahkagan:
As discussed with @mcchilders, we don't currently record when a ballot is re-audited in the database, so leaving that out of the audit report for this task. We can revisit later.

PR #330, opened 2020-03-11
Add ballot interpretations to audit report

jonahkagan:
**Description**

Task: #239 

Adds a section labeled "All Sampled Ballots" to the end of the audit report (only for online audits). In this section, there's a row for each ballot containing info about what happened during the audit:
- The ballot id
- The ticket numbers for each round it was sampled in
- Whether it was audited or not
- The result of that audit
- Any comments

**Testing**

Added a few new test cases:
- Reports for offline audits should not have the "All Sampled Ballots" section
- Ballot interpretations for a one-round audit
- Ballot interpretations for a two-round audit

**Progress**

Ready for review. One outstanding question that I'll note with a PR comment.

### DIFF
--- a/arlo_server/__init__.py
+++ b/arlo_server/__init__.py
@@ -1148,21 +1148,22 @@ def audit_report(election_id):
     report_writer.writerow(["Risk Limit", "{:d}%".format(election.risk_limit)])
     report_writer.writerow(["Random Seed", election.random_seed])
 
-    for audit_board in jurisdiction.audit_boards:
-        report_writer.writerow(
-            [
-                audit_board.name,
-                audit_board.member_1,
-                pretty_affiliation(audit_board.member_1_affiliation),
-            ]
-        )
-        report_writer.writerow(
-            [
-                audit_board.name,
-                audit_board.member_2,
-                pretty_affiliation(audit_board.member_2_affiliation),
-            ]
-        )
+    if election.online:
+        for audit_board in jurisdiction.audit_boards:
+            report_writer.writerow(
+                [
+                    audit_board.name,
+                    audit_board.member_1,
+                    pretty_affiliation(audit_board.member_1_affiliation),
+                ]
+            )
+            report_writer.writerow(
+                [
+                    audit_board.name,
+                    audit_board.member_2,
+                    pretty_affiliation(audit_board.member_2_affiliation),
+                ]
+            )
 
     for round in election.rounds:
         round_contest = round.round_contests[0]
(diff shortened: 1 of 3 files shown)


arlo_server/__init__.py
```diff
@@ -1218,36 +1221,82 @@ def audit_report(election_id):
...
+                )
+
+            report_writer.writerow(
+                [
+                    "Batch {:s}, #{:d}".format(b.batch.name, b.ballot_position),
+                    ", ".join(ticket_numbers),
+                    "Audited" if b.sampled_ballot.vote else "Not audited",
+                    # TODO need to make sure the vote/comment are guaranteed on the first ballot
+                    b.sampled_ballot.vote,
+                    b.sampled_ballot.comment,
```
jonahkagan: I'm not sure about this. When a ballot is sampled multiple times (whether in one round or across multiple rounds), is there one and only one `SampledBallot` object? Or can there be multiple? If multiple, what do we do here?

jonahkagan:
@benadida @eventualbuddha I'm very open to any feedback on style/architecture/whatever! Getting used to the codebase and norms still.

arlo_server/__init__.py
benadida: When a ballot is sampled multiple times, it appears *once* in `SampledBallot`, and once per draw in `SampledBallotDraw`.

arlo_server/__init__.py
```diff
@@ -1218,36 +1221,82 @@ def audit_report(election_id):
...
+            ["Ballot", "Ticket Numbers", "Audited?", "Audit Result", "Comments"]
+        )
+        # Write a row for each ballot that looks like this:
+        # "Batch 1, #13",Round 1: 0.123,Audited,some_candidate_id,A comment
+        # The Ticket Numbers column is a bit tricky:
+        # If a ballot was sampled multiple times in a round: Round 1: 0.123, 0.456
+        # If a ballot was sampled in multiple rounds: Round 1: 0.123, Round 2: 0.456
+
+        # First group all the ballot draws by the actual ballot
+        for _, ballot_draws in group_by(
```
benadida: I can see how using the `group_by` here in code vs the DB might make it easier to generate the Tickets fields the way you want, but then I would consider joining, in the database, with `SampledBallot`, otherwise you'll be doing an extra query every time you reference `b.sampled_ballot`.

jonahkagan: This is very helpful! I was wondering how all the foreign objects I needed were magically available haha. Guess I could have thought a little harder about that.

I also didn't think super hard about doing the grouping in the database, since I was just reusing the data that was already being queried. Do you think it's worth thinking through a better query that does some aggregation? Part of me thinks it's not worthwhile since we need the unaggregated ballot samples as well for the existing report behavior.

jonahkagan: Added join here: e334e08

Issue #309, opened 2020-03-05
backend support for targeted + opportunistic contests


benadida:
provides support for [votingworks/arlo#289 (votingworks/arlo#287)](https://zube.io/votingworks/arlo/c/289)  and [votingworks/arlo#222 (votingworks/arlo#221)](https://zube.io/votingworks/arlo/c/222)  and [votingworks/arlo#288 (votingworks/arlo#286)](https://zube.io/votingworks/arlo/c/288).

- calls into the sampler should take all contests, targeted and opportunistic, and return estimated sample sizes
- data model should have one table for all contests with flag for opportunistic/targeted
- JSON model and API calls from frontend/backend should include this flag.

PR #331, opened 2020-03-12
Change TargetedContest model to Contest, adding is_targeted flag

jonahkagan:
**Description**

Task: #309 

To support both targeted and opportunistic contests, the first step is
to change the TargetedContest data type to be generic to all contests.
We add `is_targeted` as a boolean flag to specify which contests are
targeted and which are opportunistic. There are no other types of
contests, so this should be sufficient.

**Testing**

Covered by existing tests

**Progress**

Ready for review. This will require some sort of db migration - need advice on that.

### DIFF
--- a/models.py
+++ b/models.py
@@ -40,7 +40,7 @@ class Election(db.Model):
     jurisdictions = relationship(
         "Jurisdiction", backref="election", passive_deletes=True
     )
-    contests = relationship("TargetedContest", backref="election", passive_deletes=True)
+    contests = relationship("Contest", backref="election", passive_deletes=True)
     rounds = relationship("Round", backref="election", passive_deletes=True)
 
     jurisdictions_file = db.Column(db.Text, nullable=True)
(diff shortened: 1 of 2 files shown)


models.py
```diff
@@ -74,7 +74,7 @@ class Jurisdiction(db.Model):
         "AuditBoard", backref="jurisdiction", passive_deletes=True
     )
     contests = relationship(
-        "TargetedContestJurisdiction", backref="jurisdiction", passive_deletes=True
+        "ContestJurisdiction", backref="jurisdiction", passive_deletes=True
```
umbernhard: Fwiw, I don’t think we’re ever gonna have a jurisdiction that is not targeted, for the record (I.e. one that only has opportunistic auditing, since opportunistic auditing is done only on ballots that have a targeted contest)

jonahkagan: I think that the `TargetedContestJurisdiction` table was just an association table to link `TargetedContest`s with `Jurisdiction`s, less a commentary on whether or not the jurisdiction is targeted. If that understanding is correct, then I think renaming both does makes sense.

PR #333, opened 2020-03-12
fix: restrict access to elections

### LINKED ISSUE #290: add authorization to sensitive API calls
Ensure all API calls check action against user role and access scope.

- Creating an election
- Uploading a jurisdiction file

eventualbuddha:
**Description**

This replaces the checks in the individual views with one inside `get_election`, which ensures that you can't get the election record without checking that the logged in user has access to that election.

**Testing**

Existing tests should cover this.

**Progress**

I believe this closes #290.

### DIFF
--- a/arlo_server/__init__.py
+++ b/arlo_server/__init__.py
@@ -65,7 +65,9 @@ def init_db():
 
 
 def get_election(election_id):
-    return Election.query.filter_by(id=election_id).one()
+    election = Election.query.filter_by(id=election_id).one()
+    require_audit_admin_for_organization(election.organization_id)
+    return election
 
 
 def contest_status(election):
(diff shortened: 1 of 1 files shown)


eventualbuddha:
> I'm kind of into having permissions specified on every route. Makes it easy to quickly figure out from reading the code which users have access to the endpoint. What do you think about starting that as a convention?

I worry that it'll be easy to forget to specify who has access. I think we'll have a clearer sense of how permissions should work once we have a few endpoints that are for jurisdiction admins only. In that case, I suppose this would have to change. Can you think of something that fits both requirements?

eventualbuddha:
I'm going to close this since I see it won't work well with jurisdiction admins. I'll create a new ticket where we can brainstorm an approach.

PR #334, opened 2020-03-12
fix: restrict access to elections

### LINKED ISSUE #290: add authorization to sensitive API calls

eventualbuddha:
**Description**

This puts the line to restrict access in each route that accesses an election by ID.

**Testing**

Covered by existing tests.

**Progress**

I believe this closes #290. However, it does so in a way that is fairly easy to forget to apply to a given route. We should open a new issue detailing a better way (edit: this is now https://github.com/votingworks/arlo/issues/335).

### DIFF
--- a/arlo_server/__init__.py
+++ b/arlo_server/__init__.py
@@ -438,6 +438,7 @@ def update_jurisdictions_file(election_id=None):
 @app.route("/election/<election_id>/audit/status", methods=["GET"])
 def audit_status(election_id=None):
     election = get_election(election_id)
+    require_audit_admin_for_organization(election.organization_id)
 
     return jsonify(
         organizationId=election.organization_id,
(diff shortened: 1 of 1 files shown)


eventualbuddha:
I could add tests for each endpoint. It might be a bit redundant, but perhaps worth it. The main question I have for @benadida and @MorganLove (and perhaps also @mcchilders) is this: do all of these endpoints require the logged-in user to be an audit admin?

jonahkagan:
I'm currently trying to take baby steps towards having a set of tests for each endpoint anyway. Whenever I touch an endpoint I add a barebones test suite for it. Also see some new test helpers I'm adding in #336 that support that project

mcchilders:
@eventualbuddha when you say "all of these endpoints" - can you define which ones? Or do you just mean literally any endpoint we've created? Not everything will require the user to be an AA, no, and we should prob be a little careful because some of these are likely getting reused between the single-page existing workflow (which we're keeping around for un-auth'd users) and the new multi-jurisdiction workflow ... don't want to inadvertently require it for something that's being used in both places?

eventualbuddha:
> …some of these are likely getting reused between the single-page existing workflow (which we're keeping around for un-auth'd users) and the new multi-jurisdiction workflow ... don't want to inadvertently require it for something that's being used in both places?

That's a good point. Alright, sounds like we need to have more discussion around access permissions before we dig into this any more. Abandoning for now.

PR #337, opened 2020-03-12
/audit/basic: Require contest.isTargeted

jonahkagan:
**Description**

Task: votingworks/arlo#309 

votingworks/arlo#336 added `contest.isTargeted` to `/audit/basic` as an optional field. Once the frontend starts sending the field, this PR will make it required (as well as all the other fields on the contest object).

**Testing**

Added new test for all required fields in the contest object.

**Progress**

Ready for review. Can't be merged til the frontend sends `isTargeted`.

### DIFF
--- a/arlo_server/__init__.py
+++ b/arlo_server/__init__.py
@@ -6,7 +6,7 @@ from flask import Flask, jsonify, request, Response, redirect, session
 from flask_httpauth import HTTPBasicAuth
 
 from audits import sampler, bravo, sampler_contest
-from werkzeug.exceptions import InternalServerError, Unauthorized, Forbidden
+from werkzeug.exceptions import InternalServerError, BadRequest, Unauthorized, Forbidden
 from xkcdpass import xkcd_password as xp
 
 from sqlalchemy import event, func
@@ -17,6 +17,7 @@ from authlib.flask.client import OAuth
 from arlo_server.base import app
 from arlo_server.db import db
 from models import *
+from .validation import check_required_fields_for_contest
 
 from config import HTTP_ORIGIN
 from config import (
@@ -538,6 +539,8 @@ def audit_basic_update(election_id):
     db.session.query(Contest).filter_by(election_id=election.id).delete()
 
     for contest in info["contests"]:
+        check_required_fields_for_contest(contest)
+
         total_allowed_votes_in_contest = (
             contest["totalBallotsCast"] * contest["votesAllowed"]
         )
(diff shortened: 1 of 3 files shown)


arlo_server/validation.py
```diff
@@ -0,0 +1,15 @@
+from werkzeug.exceptions import BadRequest
+
+
+def check_required_fields(json, fields, prefix=""):
```
eventualbuddha: I looked into something like this a while ago. Other than using something like GraphQL, something like this might be best: https://json-schema.org. I wouldn't say it's required for this, but that's the direction I think we should go.

jonahkagan: I've used json schema in the past and it worked well. Think it's worth setting up now?

jonahkagan: Ooh I've never used GraphQL before but it looks cool too...

arlo_server/__init__.py
```diff
@@ -538,6 +539,8 @@ def audit_basic_update(election_id):
     db.session.query(Contest).filter_by(election_id=election.id).delete()
 
     for contest in info["contests"]:
+        check_required_fields_for_contest(contest)
```
jonahkagan: I just noticed that further down in this function, there's a different error reporting scheme for another validation concern. It calls `db.session.rollback()` before reporting the error. Is that generally necessary? Do we need to think about that here?

eventualbuddha: I think we don’t want to leave the database in an inconsistent state, so if we mutate the db before we can do all validation then I think a rollback is appropriate.

arlo_server/validation.py
eventualbuddha: If you feel you can set up JSON schema without too much fuss, I’d love to see it. I did a spike with GraphQL and personally really liked it. That opinion was not universally shared.

benadida: that would be my fault for pushing back on it :) From everyhing I've seen, graphql adds can rapidly complicate an API surface, and I tend to lean hard towards keeping API contracts as simple as possible.

I'm all for schema validation using JSON schema.

arlo_server/__init__.py
benadida: I like the explicit rollback. I believe it's implied if you don't call `commit`, but yeah let's be explicit.

MorganLove:
@jonahkagan Do you want me to update the frontend in this PR to send back a default value?

jonahkagan:
@MorganLove sure, but it's not high priority. This is just meant to be a cleanup after we're done with all the changes we're making

jonahkagan:
Since we are not using /audit/basic in the multi-jurisdiction flow, I don't think this really matters anymore.

PR #339, opened 2020-03-14
Multi-jurisdiction backend API proposal

jonahkagan:
Make sure you click "view file" on proposed-routes.md so you can actually see the pictures!


docs/api-proposal/proposed-routes.md
```diff
@@ -0,0 +1,76 @@
...
+- Avoid reusing routes across roles. Instead of reusing one route and returning different scoped responses, make different roles query different endpoints with explicit scoping in the route. That way we just have to check if a user has access to the data they are asking for, as opposed to trying to filter data based on their access. (I learned this one the hard way.)
+
+New routes are in **bold**.
+
+# Audit Admin: Audit setup
+
+![](./create-audit.png)
+
+- POST /election/new - create new audit
+- **GET /election** - get all audits
```
benadida: nit: `GET /election/` for listing a set of audits / elecitons.

docs/api-proposal/proposed-routes.md
```diff
@@ -0,0 +1,76 @@
...
+
+![](upload-jurisdiction-file.png)
+
+- POST /election/<election_id>/jurisdiction/csv - upload CSV file of jurisdictions with JAs
+- GET /election/<election_id>/jurisdiction/csv - download CSV file of jurisdictions with JAs
+
+![](contests.png)
+
+- **PUT /election/<election_id>/contests** - create/update all contests, including contest universe
+- **GET /election/<election_id>/contests** - get all contests
```
benadida: :+1:

docs/api-proposal/proposed-routes.md
```diff
@@ -0,0 +1,76 @@
...
+
+![](audit-settings.png)
+
+- **GET /election/<election_id>** - get audit settings (risk limit, name, etc)
+- **PUT /election/<election_id>** - update audit settings
```
benadida: :+1: would be good to enumerate exactly which fields go here.

docs/api-proposal/proposed-routes.md
```diff
@@ -0,0 +1,76 @@
...
+
+![](review-and-launch.png)
+
+- **GET /election/<election_id>/round** - check if we have created the first round object yet
```
benadida: suggest:
`GET /election/<election_id>/round/` to list all the rounds. No special API for the first round, just get all of them.

docs/api-proposal/proposed-routes.md
```diff
@@ -0,0 +1,76 @@
...
+- **POST /election/<election_id>/round** - create the first round object, if we need to
```
benadida: same nit: add trailing slash to indicate a collection of rounds.

docs/api-proposal/proposed-routes.md
```diff
@@ -0,0 +1,76 @@
...
+- **GET /election/<election_id>/round/<round_id>** - poll the round, which will eventually have the estimated sample sizes on it
```
benadida: in the interest of fewer APIs to start, I'd suggest just polling `/round/` which lists all the rounds and which we need up there anyways. So this API could be removed.

docs/api-proposal/proposed-routes.md
```diff
@@ -0,0 +1,76 @@
...
+- **PUT /election/<election_id>/round/<round_id>** - set the sample size, which starts the round
```
benadida: Given that we only want to set the sample size and, as a side effect, start the round, I suggest:

`POST /election/<election_id>/round/<round_id>/start`

docs/api-proposal/proposed-routes.md
```diff
@@ -0,0 +1,76 @@
...
+
+# Audit Admin: Running the audit
+
+![](jurisdiction-status.png)
+
+- **GET /election/<election_id>/round** - get all rounds to figure out what round we're on
+- **GET /election/<election_id>/round/<round_id>/jurisdiction** - get round status by jurisdiction
```
benadida: nit trailing slash :)

docs/api-proposal/proposed-routes.md
```diff
@@ -0,0 +1,76 @@
...
+
+![](contest-status.png)
+
+- **GET /election/<election_id>/round** - get all rounds to figure out what round we're on
+- **GET /election/<election_id>/round/<round_id>/contest** - get round status by contest
```
benadida: nit trailing slash :)

docs/api-proposal/proposed-routes.md
```diff
@@ -0,0 +1,76 @@
...
+
+![](start-round2.png)
+
+- **POST /election/<election_id>/round** - start a new round
```
benadida: if you go with my earlier suggestion, then this would be

`POST /election/<election_id>/round/` to create the round
`POST /election/<election_id>/round/<round_id>/start` to start it with a sample size (which could be picked automatically)

docs/api-proposal/proposed-routes.md
```diff
@@ -0,0 +1,76 @@
...
+
+- **GET /election** - get all audits
+
+![](ja-upload-manifest.png)
+
+- PUT /election/<election_id>/jurisdiction/<jurisdiction_id>/manifest - upload the manifest (needs to be modified, since this currently initiates sampling of the ballots)
+
+![](ja-create-audit-boards.png)
+
+- **GET /election/<election_id>/jurisdiction/<jurisdiction_id>/round** - figure out what round we're on
```
benadida: list all the rounds I imagine? and nit trailing slash :)

docs/api-proposal/proposed-routes.md
```diff
@@ -0,0 +1,76 @@
...
+- **POST /election/<election_id>/jurisdiction/<jurisdiction_id>/round/<round_id>/audit-board** - create audit boards for this round (could potentially reuse existing audit-board endpoint, but it doesn't know anything about rounds currently)
```
benadida: no don't reuse the other endpoint, this is right.

docs/api-proposal/proposed-routes.md
```diff
@@ -0,0 +1,76 @@
+This doc proposes a new set of API endpoints for the multi-jurisdiction audit flow.
+
+Some principles I tried to follow:
+
+- Reuse existing routes when available
+- Use REST conventions
+- Avoid reusing routes across roles. Instead of reusing one route and returning different scoped responses, make different roles query different endpoints with explicit scoping in the route. That way we just have to check if a user has access to the data they are asking for, as opposed to trying to filter data based on their access. (I learned this one the hard way.)
```
benadida: 110%. Very RESTish and good security practice to make a given URL always return the same information if it is authorized. (With typically one exception, `/me`, which returns the user's core confirmation.)

docs/api-proposal/existing-api-routes.md
```diff
@@ -0,0 +1,49 @@
+# React app
+
+GET /
```
umbernhard: When these are displayed in markdown, they aren't split over lines (see below)
image Maybe drop them into a bulleted list or something?

docs/api-proposal/proposed-routes.md
```diff
@@ -0,0 +1,76 @@
...
+New routes are in **bold**.
+
+# Audit Admin: Audit setup
+
+![](./create-audit.png)
+
+- POST /election/new - create new audit
+- **GET /election** - get all audits
+
+![](upload-jurisdiction-file.png)
```
umbernhard: This image is missing.

benadida:
> One thing that just occurred to me: it kinda makes no sense that we're calling everything "election" when we mean "audit". Would this be a good time to change that? @benadida @mcchilders thoughts?

It makes sense except I think if we make this change, we then would have to change the data model too for consistency, and since we're mapping elections and audits 1:1, I could see us keeping things as elections and not worrying about it.

docs/api-proposal/proposed-routes.md
benadida: actually, thinking about this some more: this is one call that depends on authentication, and as per the principle outlined by @jonahkagan (which I agree with) that API calls should not return values that depend on authentication -- except one call --, we already have the list of audits that map to a user in `/auth/me`. How about we kill `GET /election` and just use `/auth/me` to get the elections that a user is mapped to. `/auth/me` is then the only auth-dependent API return value.

docs/api-proposal/proposed-routes.md
eventualbuddha: Relying on `/foo` being different from `/foo/` seems likely to bite us down the road. I don’t have a clear example as to why, but… that’s just not how paths work and it feels pretty unintuitive.

docs/api-proposal/existing-api-routes.md
eventualbuddha: Yes, please.

docs/api-proposal/proposed-routes.md
```diff
@@ -0,0 +1,76 @@
...
+
+- POST /election/<election_id>/jurisdiction/csv - upload CSV file of jurisdictions with JAs
+- GET /election/<election_id>/jurisdiction/csv - download CSV file of jurisdictions with JAs
```
eventualbuddha: Technically these endpoints exist already. I’d be wary of hard-coding the expected format in the URL.

docs/api-proposal/proposed-routes.md
eventualbuddha: What does this return if we don’t have a first round yet? 404 Not Found? Would it be better to just have a flag on the `election` response?

docs/api-proposal/proposed-routes.md
eventualbuddha: The way we'd figure out what round we're on is to get all the rounds and take the last one? We need to guarantee and document an order, then.

docs/api-proposal/proposed-routes.md
```diff
@@ -0,0 +1,76 @@
...
+![](ja-create-audit-boards.png)
+
+- **GET /election/<election_id>/jurisdiction/<jurisdiction_id>/round** - figure out what round we're on
+- **POST /election/<election_id>/jurisdiction/<jurisdiction_id>/round/<round_id>/audit-board** - create audit boards for this round (could potentially reuse existing audit-board endpoint, but it doesn't know anything about rounds currently)
+
+![](ja-run-audit.png)
+
+- **GET /election/<election_id>/jurisdiction/<jurisdiction_id>/round** - figure out what round we're on
+- **GET /election/<election_id>/jurisdiction/<jurisdiction_id>/round/<round_id>** - get round status
+- GET /election/<election_id>/jurisdiction/<jurisdiction_id>/round/<round_id>/retrieval-list - get the retrieval list (technically the existing endpoint is /election/<election_id>/jurisdiction/<jurisdiction_id>/<round_num>/retrieval-list, which is painfully inconsistent, but I guess we can reuse it)
```
eventualbuddha: There's no real need to keep it, is there? Or at least we can make a new one and deprecate the old one.

docs/api-proposal/proposed-routes.md
jonahkagan: This is intended to list all rounds. I was just using the singular `round` to match how the rest of the API is named. I would like to eventually move everything to the plural naming convention with no trailing slash.

docs/api-proposal/proposed-routes.md
jonahkagan: This was meant to be singular `/contest` to match the naming convention elsewhere

docs/api-proposal/proposed-routes.md
jonahkagan: Good call

docs/api-proposal/proposed-routes.md
jonahkagan: Yeah I like that it clarifies the side effect

docs/api-proposal/proposed-routes.md
jonahkagan: The rounds have a `round_num` property that could be used, right?

jonahkagan:
@umbernhard I plan to come back later and rename everything to make more sense :) But since that will touch frontend and backend I figured it could happen in a separate PR after this sprint.

docs/api-proposal/proposed-routes.md
jonahkagan: A few thoughts:
- I was thinking we could rename them now since they aren't used yet. Will be harder later.
- I was thinking of the file upload as part of the jurisdictions API, which is why I was thinking of nesting it under `/jurisdiction/`
- Fair point about hardcoding the format. Though I generally think once we make an API with a format, we're kind of stuck with it, so that's why I was thinking about being explicit. If we wanted to change it, we'd probably have to make a new endpoint anyway.

But I don't have a strong opinion here

docs/api-proposal/proposed-routes.md
jonahkagan: Empty list. Sorry, should have been clearer about this being the "list" endpoint for rounds

docs/api-proposal/proposed-routes.md
jonahkagan: I would be happy to make a new one. Just was trying to minimize the number of new endpoints created

docs/api-proposal/proposed-routes.md
eventualbuddha: With all of this, it is harder to change later but since we're the only client we can change things as long as @MorganLove knows we're doing it.

benadida:
@jonahkagan should we close this PR?

jonahkagan:
sure!

PR #340, opened 2020-03-15
run black formatter out of pipenv

benadida:


package.json
```diff
@@ -15,7 +15,7 @@
       "prettier --write"
     ],
     "*.py": [
-      "black"
+      "pipenv run black"
```
eventualbuddha:
```suggestion
      "python3.7 -m pipenv run black"
```

benadida: I think `python3` vs `python3.7` might be more forward-compatible once we move to 3.8.

eventualbuddha: That’s true, but we just hardcoded 3.7 in the makefile and I think it makes sense to do here for the same reasons.

PR #341, opened 2020-03-15
add optional round_id to AuditBoard

benadida:
Following the direction of a new, separate API for multi-jurisdiction, this PR makes it possible to map an audit board to a round, without requiring it. Much simpler change that lets the current flow work with no front-end changes and only minor backend changes.

Then, the new API calls that @jonahkagan is planning can map audit boards to an explicit round. This might require some slightly more complicated JOINs on API calls that are shared between both flows, but that should be trivial compared to trying to remap the single workflow onto the new data model.

Addresses votingworks/arlo#299

### DIFF
--- a/arlo_server/models.py
+++ b/arlo_server/models.py
@@ -203,6 +203,10 @@ class AuditBoard(db.Model):
         db.ForeignKey("jurisdiction.id", ondelete="cascade"),
         nullable=False,
     )
+    round_id = db.Column(
+        db.String(200), db.ForeignKey("round.id", ondelete="cascade"), nullable=True
+    )
+
     name = db.Column(db.String(200))
     member_1 = db.Column(db.String(200), nullable=True)
     member_1_affiliation = db.Column(db.String(200), nullable=True)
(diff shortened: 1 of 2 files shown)


arlo_server/__init__.py
```diff
@@ -867,8 +867,8 @@ def ballot_list(election_id, jurisdiction_id, round_id):
     query = (
         SampledBallotDraw.query.join(SampledBallot)
         .join(SampledBallotDraw.batch)
-        .join(AuditBoard)
         .join(Round)
+        .join(AuditBoard, AuditBoard.id == SampledBallot.audit_board_id)
```
benadida: the explicit join clause here is so SQLAlchemy doesn't try to be too smart and automatically join on `round_id`, which then makes the existing flow break.

jonahkagan:
Does `Round` get an `audit_boards` field as well? (This is half a question about SQLAlchemy and half a suggestion)

benadida:
> Add a test?

couldn't think of a test right now that wouldn't be just testing the SQLAlchemy object. We have no code path that sets a `round_id` yet, so when we add one, we should def add a test.

benadida:
> Does `Round` get an `audit_boards` field as well? (This is half a question about SQLAlchemy and half a suggestion)

It does now, I've added the relationship :)

PR #346, opened 2020-03-16
Reorganize server modules

jonahkagan:
**Description**

This should make it easier to add new routes in separate files.

**Testing**

Existing tests run and pass.

**Progress**

Ready for review.


arlo_server/models.py
```diff
@@ -1,8 +1,10 @@
 # type: ignore
-from arlo_server import db
 from sqlalchemy.orm import relationship, backref
 from typing import Union, List
 from enum import Enum
+from flask_sqlalchemy import SQLAlchemy
+
+db = SQLAlchemy()
```
jonahkagan: This was the key change - not passing `app` into SQLAlchemy, and closing that loop later (see __init__.py)

PR #347, opened 2020-03-16
Audit setup structure [wip]

MorganLove:
**Description**

- Renamed `/AuditFlow` to `/DataEntry` and `/AuditForms` to `/Audit` and added a substructure under `/Audit` for the multi-jurisdiction flow separate from the single page flow
- Created a sidebar for navigating the stages of the setup
- Made the flows separated based on authentication status
- Created structure for the forms in the setup stages (right now just the participants and contests stages have forms, and those just go to console.log, and there's no file upload field just yet).

**Testing**

- none yet

**Progress**

- Need to link into the backend api changes
- Need to add the file upload
- Need to create the other stages
- Need to add the navigation in the header
- Need to add the audit name field on the create audit screen
- Need to add a confirmation dialog for when you hit the 'back' button


benadida:
looks like the single page workflow got messed up a bit in layout:

image

MorganLove:
> If the next PRs can be a lot smaller thanks to this one, given that a lot of this is just refactoring/renaming, I can live with it.
> 
> Single change request: fix single-page workflow to still be one column. Otherwise, good to go.

@benadida Yeah, the rest should be able to be much more contained. :slightly_smiling_face: 

Good catch! Fixed now.

PR #349, opened 2020-03-16
Audit name

MorganLove:
**Description**

- Has the front end post an `auditName` field to the `/election/new` endpoint

**Testing**

- Tests that it passes as expected for both authenticated and unauthenticated users, and is required
- Updates snapshots

**Progress**

- Do we need to verify the uniqueness of the audit name at this time?
- The back end needs to use the value as passed back when creating an audit, still

image

### DIFF
--- a/arlo-client/src/components/CreateAudit.test.tsx
+++ b/arlo-client/src/components/CreateAudit.test.tsx
@@ -45,7 +45,13 @@ describe('CreateAudit', () => {
 
   it('calls the /election/new endpoint for nonauthenticated user', async () => {
     apiMock.mockImplementation(async () => ({ electionId: '1' }))
-    const { getByText } = render(<CreateAudit {...routeProps} />)
+    const { getByText, getByLabelText } = render(
+      <CreateAudit {...routeProps} />
+    )
+
+    fireEvent.change(getByLabelText('Give your new audit a unique name.'), {
+      target: { value: 'Audit Name' },
+    })
 
     fireEvent.click(getByText('Create a New Audit'), { bubbles: true })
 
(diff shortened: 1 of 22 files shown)


MorganLove:
I think this PR should cover everything that closes #349 once it's done.

arlo-client/src/components/CreateAudit.test.tsx
```diff
@@ -86,15 +104,20 @@ describe('CreateAudit', () => {
...
 
     await wait(() => {
       expect(apiMock).toBeCalledTimes(2)
       expect(apiMock).toHaveBeenNthCalledWith(2, '/election/new', {
         method: 'POST',
-        // eslint-disable-next-line @typescript-eslint/camelcase
-        body: JSON.stringify({ organization_id: 'org-id' }),
+        body: JSON.stringify({
+          // eslint-disable-next-line @typescript-eslint/camelcase
+          organization_id: 'org-id',
```
jonahkagan: Oops, should we make a follow up task to camelCase this on the backend?

MorganLove: @jonahkagan I think @eventualbuddha already created a PR to do that, just need to update this to match when it gets rebased.

eventualbuddha: Yep, this was merged in #353.

jonahkagan: giphy

arlo-client/src/components/CreateAudit.tsx
```diff
@@ -60,18 +71,37 @@ const CreateAudit = ({ history }: RouteComponentProps<ICreateAuditParams>) => {
   return (
     <Wrapper>
       <img height="50px" src="/arlo.png" alt="Arlo, by VotingWorks" />
-      <Button
-        verticalSpaced
```
eventualbuddha: This line seems to have been lost when I rebased. Do we want it, or is it right that it's gone? @MorganLove

MorganLove: @jonahkagan That works with it being gone. I just compared with and without and it looks better without. Thanks for checking! 👍

MorganLove:
@eventualbuddha Any idea why this is failing backend tests now?

Issue #326, opened 2020-03-11
Separate endpoint for creating contests


MorganLove:
There should be a separate endpoint for creating contests, instead of using the `/basic` endpoint.

mcchilders:
This is a placeholder card requested by @benadida to hold space for future work (see https://github.com/votingworks/arlo/pull/318#pullrequestreview-370800440 for the initial convo that prompted this.) Don't start this until we have a conversation!

nikhilb4a:
Completed back in 2020

PR #350, opened 2020-03-17
Add /election/<election_id>/contest API

jonahkagan:
**Description**

Task: #309, #326 

Adds new endpoints for getting and creating/updating contests when setting up an audit:
- `GET /election/<election_id>/contest` - gets a list of all contests for the audit
- `PUT /election/<election_id>/contest` - sets the contests for the audit, overwriting any existing contests

Notably, this new API includes a `jurisdictionIds` field to set the contest universe.

The new endpoints use JSON schema to validate the contests.

**Testing**

Added a new test file with quite a few cases.

**Progress**

Ready for review. Left some open questions as PR comments.

### DIFF
--- a/arlo_server/contests.py
+++ b/arlo_server/contests.py
@@ -1,0 +1,139 @@
+from typing import List
+from flask import request, jsonify
+from jsonschema import validate
+from werkzeug.exceptions import BadRequest
+
+from arlo_server import app, db
+from arlo_server.routes import get_election, require_audit_admin_for_organization
+from arlo_server.models import Contest, ContestChoice, Jurisdiction
+
+contest_choice_schema = {
+    "type": "object",
+    "properties": {
+        "id": {"type": "string"},
+        "name": {"type": "string"},
+        "numVotes": {"type": "integer", "minimum": 0},
+    },
+    "required": ["id", "name", "numVotes"],
+}
+
+contest_schema = {
+    "type": "object",
+    "properties": {
+        "id": {"type": "string"},
+        "name": {"type": "string"},
+        "isTargeted": {"type": "boolean"},
+        "choices": {"type": "array", "items": contest_choice_schema},
+        "totalBallotsCast": {"type": "integer", "minimum": 0},
+        "numWinners": {"type": "integer", "minimum": 1},
+        "votesAllowed": {"type": "integer", "minimum": 1},
+        "jurisdictionIds": {"type": "array", "items": {"type": "string"}},
+    },
+    "required": [
+        "id",
+        "name",
+        "isTargeted",
+        "choices",
+        "totalBallotsCast",
+        "numWinners",
+        "votesAllowed",
+        "jurisdictionIds",
+    ],
+}
+
+
+def serialize_contest_choice(db_contest_choice: ContestChoice) -> dict:
+    return {
+        "id": db_contest_choice.id,
+        "name": db_contest_choice.name,
+        "numVotes": db_contest_choice.num_votes,
+    }
+
+
+def serialize_contest(db_contest: Contest) -> dict:
+    return {
+        "id": db_contest.id,
+        "name": db_contest.name,
+        "isTargeted": db_contest.is_targeted,
+        "choices": [serialize_contest_choice(c) for c in db_contest.choices],
+        "totalBallotsCast": db_contest.total_ballots_cast,
+        "numWinners": db_contest.num_winners,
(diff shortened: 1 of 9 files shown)


arlo_server/models.py
```diff
@@ -183,17 +186,32 @@ class ContestChoice(db.Model):
...
         nullable=False,
-    )
+    ),
+    db.Column(
+        "jurisdiction_id",
+        db.String(200),
+        db.ForeignKey("jurisdiction.id", primary_key=True, ondelete="cascade"),
+        nullable=False,
+    ),
+)
```
jonahkagan: This is the recommended way to set up the relationship according to the flask-sqlalchemy docs. I'm a bit concerned about performance, because we have to look up all the jurisdictions when we set the `jurisdiction_ids` field on the contests.

https://flask-sqlalchemy.palletsprojects.com/en/2.x/models/#many-to-many-relationships

arlo_server/contests.py
```diff
@@ -0,0 +1,137 @@
...
+        contest_id=contest_id,
+        name=json_contest_choice["name"],
+        num_votes=json_contest_choice["numVotes"],
+    )
+
+
+def deserialize_contest(json_contest, election_id):
+    jurisdictions = (
+        db.session.query(Jurisdiction)
+        .filter(Jurisdiction.id.in_(json_contest["jurisdictionIds"]))
```
benadida: add a filter here that the jurisdiction belong to the `election_id`.

arlo_server/contests.py
```diff
@@ -0,0 +1,137 @@
...
+    return Contest(
+        election_id=election_id,
+        id=json_contest["id"],
+        name=json_contest["name"],
+        is_targeted=json_contest["isTargeted"],
+        choices=choices,
+        total_ballots_cast=json_contest["totalBallotsCast"],
+        num_winners=json_contest["numWinners"],
+        votes_allowed=json_contest["votesAllowed"],
+        jurisdictions=jurisdictions,
```
benadida: this works for adding new relationships? Cool!

arlo_server/models.py
benadida: not at all worried, that field will be auto-indexed and the lookups will be fast, plus you're already doing it in one query for all jurisdictions.  We'll want to review indexes at some point, but architecturally, this is sound.

benadida: actually, the thing to be potentially worried about is not the query, but the *inserts* into the mapping table. We'll almost certainly want to background this at some point, cause it could be ~1600 inserts in a single contest. But that doesn't have to be in this PR.

bgcompute.py
```diff
@@ -4,13 +4,15 @@
...
-from arlo_server.models import RoundContest
+from arlo_server.models import RoundContest, Contest
 from arlo_server.routes import compute_sample_sizes
 
 
 def bgcompute():
-    # round contests that don't have sample_size_options
-    round_contests = RoundContest.query.filter_by(sample_size_options=None)
+    # Round contests for targeted contests that don't have sample_size_options
+    round_contests = RoundContest.query.filter_by(Contest.is_targeted).filter_by(
```
benadida: sneaking in an unrelated change? ;)

arlo_server/contests.py
jonahkagan: This is straight outta the docs

bgcompute.py
jonahkagan: whoops, that was from my earlier poking around. thanks for catching, wasn't supposed to be in here

arlo_server/contests.py
jonahkagan: added

jonahkagan:
Ok, a few things happened since last review:
- There were type errors, and I ended up figuring out how to type the db models to get around it, which is pretty cool!
- I added type annotations to all my new code
- The tests were failing due to the db model changes, so I made some fixes to existing code to fix those tests

arlo_server/routes.py
```diff
@@ -464,7 +465,7 @@ def audit_status(election_id=None):
             {
                 "id": j.id,
                 "name": j.name,
-                "contests": [c.contest_id for c in j.contests],
+                "contests": [c.id for c in j.contests],
```
benadida: wow, was this just wrong code prior to your fix?

jonahkagan: No, it worked before. My database model change broke it

Issue #343, opened 2020-03-16
new API call for jurisdiction upload status


benadida:
so we can implement the screen where AA sees the state of jurisdiction ballot manifest upload

PR #351, opened 2020-03-17
Add GET /election/<election_id>/jurisdiction

jonahkagan:
**Description**

Task: #343 

An endpoint to list all of the jurisdictions in an election, including details about their ballot manifest upload.

Also tweaks the existing /election/<election_id>/jurisdiction/<jurisdiction_id>/manifest endpoint to only initiate ballot sampling when used by the single-jurisdiction flow.

**Testing**

Added new test file testing the endpoint with and without uploading the manifest.

**Progress**

Ready to review.

### DIFF
--- a/arlo_server/jurisdictions.py
+++ b/arlo_server/jurisdictions.py
@@ -1,0 +1,32 @@
+from flask import jsonify
+
+from arlo_server import app, db
+from arlo_server.routes import get_election, require_audit_admin_for_organization
+from arlo_server.models import Jurisdiction
+
+
+def serialize_jurisdiction(db_jurisdiction: Jurisdiction) -> dict:
+    return {
+        "id": db_jurisdiction.id,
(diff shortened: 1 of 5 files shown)


arlo_server/jurisdictions.py
```diff
@@ -0,0 +1,28 @@
...
+        },
+    }
+
+
+@app.route("/election/<election_id>/jurisdiction", methods=["GET"])
+def list_jurisdictions(election_id: str = None):
+    election = get_election(election_id)
+    require_audit_admin_for_organization(election.organization_id)
+
+    jurisdictions = Jurisdiction.query.filter_by(election_id=election.id).all()
```
benadida: Maybe worth ordering by name?

jonahkagan: Good idea, updated here: 17a252a

PR #352, opened 2020-03-17
Remove unused fields from Jurisdictions model

jonahkagan:
**Description**

Couldn't find anywhere that references these fields, so removing them. Maybe they are an artifact of an older version of the app. Let me know if I'm wrong.

**Testing**

Existing tests should cover.

**Progress**

Ready to review.


benadida:
We're not using them yet, but pretty sure we'll want them when we go async on manifest handling. I wouldn't remove these now

jonahkagan:
Ok, my bet is that we won't want exactly these fields. But we'll see! Closing for now

PR #354, opened 2020-03-17
refactor: extract `File` from `Election` jurisdictions file

eventualbuddha:
**Description**

Since it seems likely we're going to have more "file" objects we're going to have to keep track of, this codifies that into a model.

**Testing**

Covered by existing tests.

**Progress**

This was an opportunistic refactor on the road to making the bulk jurisdictions file upload async.

### DIFF
--- a/arlo_server/models.py
+++ b/arlo_server/models.py
@@ -45,11 +45,10 @@ class Election(db.Model):
     contests = relationship("Contest", backref="election", passive_deletes=True)
     rounds = relationship("Round", backref="election", passive_deletes=True)
 
-    jurisdictions_file = db.Column(db.Text, nullable=True)
-    jurisdictions_filename = db.Column(db.String(250), nullable=True)
-    jurisdictions_file_uploaded_at = db.Column(
-        db.DateTime(timezone=False), nullable=True
+    jurisdictions_file_id = db.Column(
+        db.String(200), db.ForeignKey("file.id", ondelete="set null"), nullable=True
     )
+    jurisdictions_file = relationship("File")
 
 
 # these are typically counties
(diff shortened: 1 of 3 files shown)


arlo_server/models.py
```diff
@@ -326,3 +325,10 @@ class RoundContestResult(db.Model):
         nullable=False,
     )
     result = db.Column(db.Integer)
+
+
+class File(db.Model):
+    id = db.Column(db.String(200), primary_key=True)
+    name = db.Column(db.String(250), nullable=False)
+    contents = db.Column(db.Text, nullable=False)
+    uploaded_at = db.Column(db.DateTime(timezone=False), nullable=False)
```
benadida: We'll probably want things like processing_started_at and processing_succeeded_at and possibly processing_errors, but that can wait till first async implementation.

eventualbuddha: My thoughts exactly.

arlo_server/routes.py
```diff
@@ -394,9 +401,15 @@ def update_jurisdictions_file(election_id=None):
     jurisdictions_file = request.files["jurisdictions"]
     jurisdictions_file_string = jurisdictions_file.read().decode("utf-8-sig")
 
-    election.jurisdictions_file = jurisdictions_file_string
-    election.jurisdictions_filename = jurisdictions_file.filename
-    election.jurisdictions_file_uploaded_at = datetime.datetime.utcnow()
+    if election.jurisdictions_file:
+        db.session.delete(election.jurisdictions_file)
```
jonahkagan: do you have a test for overwriting?

eventualbuddha: I do here: https://github.com/votingworks/arlo/pull/356/files#diff-0e2be5baf85cbad5135cb3e8ad217724R151

PR #356, opened 2020-03-17
feat: process jurisdiction file uploads in the background

eventualbuddha:
**Description**

In an effort to not block requests on things that could potentially take a long time, this puts the processing of jurisdiction files in the background.

**Testing**

Adds some new tests for the helper function `process_file` and updates tests to use the new backgrounding flow.

**Progress**

This doesn't require any more changes to make the feature work, unless I missed something. Some concerns I have:
- `bgcompute.py` doesn't seem to catch errors. I'm assuming Heroku will run the `worker` process in the `Procfile` again if it dies, but I don't know for sure.
- What happens if a file is being processed and a new file is uploaded while that is happening? In theory that could leave things in an inconsistent state.
- Did I get the `File` and `Election` relationship right?
- We should probably use ISO-8601 for all our JSON dates. This PR doesn't change any others that might exist, but we should do that.
- There's a little bit of duplication of the CSV logic, but it's fairly minimal. Wasn't sure what the right abstraction would be so I just left it.

### DIFF
--- a/arlo_server/models.py
+++ b/arlo_server/models.py
@@ -343,3 +343,15 @@ class File(BaseModel):
     name = db.Column(db.String(250), nullable=False)
     contents = db.Column(db.Text, nullable=False)
     uploaded_at = db.Column(db.DateTime(timezone=False), nullable=False)
+
+    # Metadata for processing files in the background.
+    processing_started_at = db.Column(db.DateTime(timezone=False), nullable=True)
+    processing_completed_at = db.Column(db.DateTime(timezone=False), nullable=True)
+    processing_error = db.Column(db.Text, nullable=True)
+
+
+class ProcessingStatus(str, Enum):
+    READY_TO_PROCESS = "READY_TO_PROCESS"
+    PROCESSING = "PROCESSING"
+    PROCESSED = "PROCESSED"
+    ERRORED = "ERRORED"
--- a/arlo_server/routes.py
+++ b/arlo_server/routes.py
@@ -32,6 +32,7 @@ from config import (
 
 from util.binpacking import Bucket, BalancedBucketList
 from util.jurisdiction_bulk_update import bulk_update_jurisdictions
+from util.process_file import process_file
 
 AUDIT_BOARD_MEMBER_COUNT = 2
 WORDS = xp.generate_wordlist(wordfile=xp.locate_wordfile())
@@ -305,6 +306,13 @@ def serialize_members(audit_board):
     return members
 
 
+def isoformat(datetime: Optional[datetime.datetime]) -> Optional[str]:
+    if datetime is not None:
+        return datetime.isoformat()
+    else:
+        return None
+
+
 ADMIN_PASSWORD = os.environ.get("ARLO_ADMIN_PASSWORD", None)
 
 # this is a temporary approach to getting all running audits
@@ -369,16 +377,31 @@ def get_jurisdictions_file(election_id=None):
     require_audit_admin_for_organization(election.organization_id)
     jurisdictions_file = election.jurisdictions_file
 
+    if jurisdictions_file.processing_error:
+        status = ProcessingStatus.ERRORED
+    elif jurisdictions_file.processing_completed_at:
+        status = ProcessingStatus.PROCESSED
+    elif jurisdictions_file.processing_started_at:
+        status = ProcessingStatus.PROCESSING
+    else:
+        status = ProcessingStatus.READY_TO_PROCESS
+
     if jurisdictions_file:
         return jsonify(
             file={
                 "contents": jurisdictions_file.contents,
                 "name": jurisdictions_file.name,
-                "uploadedAt": jurisdictions_file.uploaded_at,
-            }
+                "uploadedAt": isoformat(jurisdictions_file.uploaded_at),
+            },
+            processing={
+                "status": status,
+                "startedAt": isoformat(jurisdictions_file.processing_started_at),
+                "completedAt": isoformat(jurisdictions_file.processing_completed_at),
+                "error": jurisdictions_file.processing_error,
+            },
         )
     else:
-        return jsonify(file=None)
+        return jsonify(file=None, processing=None)
 
 
 @app.route("/election/<election_id>/jurisdictions/file", methods=["PUT"])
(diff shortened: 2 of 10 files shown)


arlo_server/routes.py
```diff
@@ -304,6 +305,13 @@ def serialize_members(audit_board):
     return members
 
 
+def isoformat(datetime: Optional[datetime.datetime]) -> Optional[str]:
```
jonahkagan: YES I was thinking the same thing

arlo_server/models.py
```diff
@@ -332,3 +332,16 @@ class File(db.Model):
     name = db.Column(db.String(250), nullable=False)
     contents = db.Column(db.Text, nullable=False)
     uploaded_at = db.Column(db.DateTime(timezone=False), nullable=False)
+
+    # Metadata for processing files in the background.
+    processing_started_at = db.Column(db.DateTime(timezone=False), nullable=True)
+    processing_succeeded_at = db.Column(db.DateTime(timezone=False), nullable=True)
+    processing_errored_at = db.Column(db.DateTime(timezone=False), nullable=True)
```
jonahkagan: What happens with these fields if it errors on the first attempt, then succeeds on the second attempt? (I'm assuming you can retry, which maybe is not the case.) Would it be simpler to have one field `processing_ended_at` and rely on `processing_error` as the flag for whether or not it ended in success or error?

arlo_server/models.py
```diff
@@ -332,3 +332,16 @@ class File(db.Model):
     name = db.Column(db.String(250), nullable=False)
     contents = db.Column(db.Text, nullable=False)
     uploaded_at = db.Column(db.DateTime(timezone=False), nullable=False)
+
+    # Metadata for processing files in the background.
+    processing_started_at = db.Column(db.DateTime(timezone=False), nullable=True)
+    processing_succeeded_at = db.Column(db.DateTime(timezone=False), nullable=True)
```
benadida: suggest `processing_completed_at` with `processing_error` null if successful, not-null if not. Otherwise data model could be inconsistent, with both `succeeded_at` and `errored_at` being non-null.

bgcompute.py
```diff
@@ -30,6 +37,24 @@ def bgcompute():
         )
 
 
+def bgcompute_update_election_jurisdictions_file() -> int:
+    elections = (
+        Election.query.join(File, File.id == Election.jurisdictions_file_id)
+        .filter(
+            Election.jurisdictions_file_id != None, File.processing_started_at == None
+        )
+        .options(joinedload(Election.jurisdictions_file))
```
benadida: nit: simpler code might be to query the `File` table joining against `Election` and adding the `Election` entity.

arlo_server/routes.py
```diff
@@ -401,16 +425,13 @@ def update_jurisdictions_file(election_id=None):
     jurisdictions_file = request.files["jurisdictions"]
     jurisdictions_file_string = jurisdictions_file.read().decode("utf-8-sig")
 
-    if election.jurisdictions_file:
-        db.session.delete(election.jurisdictions_file)
-
+    old_jurisdictions_file = election.jurisdictions_file
```
jonahkagan: any particular reason this gets defined up here? as far as i can read, it only gets used once below

arlo_server/routes.py
```diff
@@ -437,11 +458,9 @@ def update_jurisdictions_file(election_id=None):
...
         )
 
-    bulk_update_jurisdictions(
-        db.session,
-        election,
-        [(row[JURISDICTION_NAME], row[ADMIN_EMAIL]) for row in jurisdictions_csv],
-    )
+    if old_jurisdictions_file:
+        db.session.delete(old_jurisdictions_file)
+    db.session.add(election)
```
jonahkagan: Is there a reason the `election` needs to be added to the session? I don't quite understand SQLAlchemy yet so I'm curious

util/jurisdiction_bulk_update.py
```diff
@@ -1,7 +1,32 @@
+import csv
+import io
 import uuid
-from arlo_server.models import Election, JurisdictionAdministration, User, Jurisdiction  # type: ignore
-from typing import Tuple, List
 from sqlalchemy import func
+from typing import Tuple, List
+
+from arlo_server.models import Election, JurisdictionAdministration, User, Jurisdiction  # type: ignore
```
jonahkagan: Shouldn't need this after I merge https://github.com/votingworks/arlo/pull/351

util/process_file.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def process_file(session: Session, file: File, callback: Callable[[], None]) -> bool:
+    if file.processing_started_at:
+        return False
+
+    # Claim this file by updating the `processing_started_at` timestamp in such
+    # a way that it must not have been set before.
+    file.processing_started_at = datetime.datetime.utcnow()
+    result = session.execute(
```
jonahkagan: Does `session.execute` work before you call `session.commit`? Is that why this is different from how you would normally update the db through a model?

arlo_server/models.py
eventualbuddha: This crossed my mind but didn't think through it all the way. I like it, will do.

util/process_file.py
benadida: the `commit` just commits the transaction, the SQL statements are still issued in the interim, and the update locking of the DB ensures that only one of these updates comes through. Great way to do this and be somewhat prepared for multi-threaded job processing!

bgcompute.py
eventualbuddha: Good call. That is better.

arlo_server/models.py
eventualbuddha: That's more or less what @benadida suggested, and I just implemented that. Right now there's no way to retry processing, you just upload a new file.

arlo_server/routes.py
eventualbuddha: Because `election.jurisdictions_file` gets clobbered on the next line.

arlo_server/routes.py
eventualbuddha: I don't quite know for sure. I've been conservative about my assumptions on how smart it is and just add anything that changes. In this case it's changing `Election#jurisdictions_file_id` so I'm adding it.

util/jurisdiction_bulk_update.py
eventualbuddha: Cool, I removed this and others.

util/process_file.py
eventualbuddha: So if there were another thread/process/whatever that tried to grab the same one at the same time, they'd have different sessions/transactions and could both grab the same one, right? So maybe I do want an immediate commit?

benadida: No, what would happen is that only one of the updates would go through, as any reasonable database (postgres included), would serialize the updates to the same row and only one thread would successfully get a rowcount return of 1.

There will be a problem if the thread processing the job fails and we need to retry, but that's a separate issue.

arlo_server/routes.py
eventualbuddha: I miss TypeScript: `file.processing_started_at?.isoformat()` FTW.

Issue #345, opened 2020-03-16
make sure we have audit_name and election_name as separate fields in DB and API


benadida:

MorganLove:
We also need to have an endpoint for returning whether or not a given string is valid as a unique audit_name or not.

jonahkagan:
Could that be an error from the endpoint that creates audits when you try to create an audit? Or do you want it to be something you can validate before trying to create the audit?

benadida:
> Could that be an error from the endpoint that creates audits when you try to create an audit? Or do you want it to be something you can validate before trying to create the audit?

+1 to this, @mcchilders had mentioned in a previous ticket how an error on creation would be just fine for now.

PR #359, opened 2020-03-18
Add Election.audit_name required field

jonahkagan:
**Description**

Task: #345 
Unblocks: #349 

Adds `auditName` as a required field to `POST /election/new`. Also moves `Election.name` to `Election.election_name`.

Note: does not break `/audit/basic`, which still expects the `name` field (which corresponds to `election_name`.

**Testing**

Tests have to be updated everywhere because they all rely on `/election/new` to initialize the election. We pull this out into a universal fixture. Yay dependency injection!

Also adds one test to ensure `auditName` is required.

**Progress**

Ready for review. Sorry for the tedious test changes (just know they were even more tedious to make than they are to review!)

### DIFF
--- a/arlo_server/models.py
+++ b/arlo_server/models.py
@@ -21,9 +21,13 @@ class Organization(BaseModel):
     elections = relationship("Election", backref="organization", passive_deletes=True)
 
 
+# Election is a slight misnomer - this model represents an audit.
 class Election(BaseModel):
     id = db.Column(db.String(200), primary_key=True)
-    name = db.Column(db.String(200), nullable=True)
+    # audit_name must be unique within each Organization
+    audit_name = db.Column(db.String(200), nullable=False)
+    # election_name can be the same across audits
+    election_name = db.Column(db.String(200), nullable=True)
     state = db.Column(db.String(100), nullable=True)
     election_date = db.Column(db.Date, nullable=True)
     election_type = db.Column(db.String(200), nullable=True)
@@ -53,6 +57,8 @@ class Election(BaseModel):
         db.String(200), db.ForeignKey("file.id", ondelete="set null"), nullable=True
     )
     jurisdictions_file = relationship("File")
+
+    __table_args__ = (db.UniqueConstraint("organization_id", "audit_name"),)
 
 
 # these are typically counties
--- a/arlo_server/routes.py
+++ b/arlo_server/routes.py
@@ -1,17 +1,25 @@
 import os, datetime, csv, io, math, json, uuid, locale, re, hmac, urllib.parse, itertools
 from enum import Enum, auto
 from typing import Optional, Tuple, Union
+from jsonschema import validate
 from jsonschema.exceptions import ValidationError
 
 from flask import Flask, jsonify, request, Response, redirect, session
 from flask_httpauth import HTTPBasicAuth
 
 from audits import sampler, bravo, sampler_contest
-from werkzeug.exceptions import InternalServerError, Unauthorized, Forbidden, BadRequest
+from werkzeug.exceptions import (
+    InternalServerError,
+    Unauthorized,
+    Forbidden,
+    BadRequest,
+    Conflict,
+)
 from xkcdpass import xkcd_password as xp
 
 from sqlalchemy import event, func
 from sqlalchemy.dialects.postgresql import aggregate_order_by
+from sqlalchemy.exc import IntegrityError
 
 from authlib.flask.client import OAuth
 
@@ -41,15 +49,6 @@ WORDS = xp.generate_wordlist(wordfile=xp.locate_wordfile())
 class UserType(str, Enum):
     AUDIT_ADMIN = "audit_admin"
     JURISDICTION_ADMIN = "jurisdiction_admin"
-
-
-def create_election(election_id=None, organization_id=None):
-    if not election_id:
-        election_id = str(uuid.uuid4())
-    e = Election(id=election_id, organization_id=organization_id, name="")
-    db.session.add(e)
-    db.session.commit()
-    return election_id
 
 
 def create_organization(name=""):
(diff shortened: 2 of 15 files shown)


arlo_server/routes.py
```diff
@@ -1416,9 +1434,10 @@ def clear_loggedin_user():
 def serialize_election(election):
     return {
         "id": election.id,
-        "name": election.name,
+        "auditName": election.audit_name,
+        "electionName": election.election_name,
         "state": election.state,
-        "election_date": election.election_date,
+        "electionDate": election.election_date,
```
eventualbuddha:
```suggestion
        "electionDate": isoformat(election.election_date),
```

If we do this, we should change the signature of `isoformat` to this:

```py
def isoformat(datetime: Optional[Union[datetime.datetime, datetime.date]]) -> Optional[str]:
```

arlo_server/routes.py
```diff
@@ -360,15 +352,33 @@ def require_audit_admin_for_organization(organization_id: Optional[str]):
...
+}
+
+
 @app.route("/election/new", methods=["POST"])
 def election_new():
-    info = request.get_json()
+    election = request.get_json()
+    validate(election, election_new_schema)
 
-    organization_id = info.get("organizationId", None) if info else None
```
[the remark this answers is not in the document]
jonahkagan: I don't believe it's necessary anymore for two reasons:
- We validate that `info` (now `election`) against a schema that expects an object
- `info` should never be `None` since `auditName` is required

arlo_server/routes.py
```diff
@@ -1333,10 +1343,18 @@ def pretty_affiliation(affiliation):
...
     # deleting the election cascades to all the data structures
-    Election.query.filter_by(id=election_id).delete()
-    db.session.commit()
+    election = Election.query.filter_by(id=election_id).one_or_none()
+    if election:
+        db.session.delete(election)
+        db.session.commit()
 
-    create_election(election_id)
+    # TODO election.audit_name doesn't exist if we didn't find an election for election_id
```
jonahkagan: I could use help understanding what to do in this case @benadida

arlo_server/routes.py
```diff
@@ -360,15 +352,33 @@ def require_audit_admin_for_organization(organization_id: Optional[str]):
     )
 
 
+election_new_schema = {
```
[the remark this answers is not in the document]
jonahkagan: will do

arlo_server/routes.py
eventualbuddha: My proposal would be to just return a 404 right off the bat. The behavior of creating a new election with the ID of the old one even if there never was an old one just seemed like a byproduct of a SQL `DELETE` call operating on zero records not being an error, rather than an intentional choice.

arlo_server/models.py
```diff
@@ -56,6 +58,8 @@ class Election(BaseModel):
     )
     jurisdictions_file = relationship("File")
 
+    __table_args__ = (db.UniqueConstraint("audit_name", "organization_id"),)
```
eventualbuddha: 💯

tests/test_new_election.py
```diff
@@ -100,3 +87,52 @@ def test_in_org_with_logged_in_jurisdiction_admin(client: FlaskClient):
...
+    rv = post_json(client, "/election/new", {"auditName": "Test Audit"})
+    assert rv.status_code == 200
+    assert json.loads(rv.data)["electionId"]
+
+    rv = post_json(client, "/election/new", {"auditName": "Test Audit"})
+    assert rv.status_code == 200
+    assert json.loads(rv.data)["electionId"]
+
+
+def test_in_org_duplicate_audit_name(client: FlaskClient):
```
eventualbuddha: Also test that two orgs can have audits with the same name?

arlo_server/routes.py
jonahkagan: Nevermind, I don't know where I got the idea to do `one_or_none`... I thought the old code also checked for this but I must have misread

eventualbuddha: It had the same effect since the `delete` would succeed even if it deleted 0 rows.

arlo_server/routes.py
jonahkagan: Will do

arlo_server/routes.py
```diff
@@ -1540,6 +1577,14 @@ def handle_401(e):
     )
 
 
+@app.errorhandler(Conflict)
+def handle_409(e):
```
eventualbuddha: Love the appropriate use of lesser-known HTTP error codes 😄

arlo_server/routes.py
jonahkagan: added 404 here: 3c144d9

arlo_server/models.py
benadida: FYI it's better practice to do `(organization_id, audit_name)`, as that index can be used to select all audits by organization, whereas the other order doesn't let you do that.

arlo_server/routes.py
```diff
@@ -360,15 +361,42 @@ def require_audit_admin_for_organization(organization_id: Optional[str]):
...
+        id=str(uuid.uuid4()),
+        audit_name=election["auditName"],
+        organization_id=organization_id,
+    )
+    db.session.add(election)
+
+    try:
+        db.session.commit()
+    except IntegrityError as e:
+        if e.orig.diag.constraint_name == "election_audit_name_organization_id_key":
```
benadida: depending on the constraint name could be a tad dangerous, but I guess with tests maybe it's okay?

We should at some point figure out how to use Postgres's `on conflict`.

arlo_server/routes.py
benadida: ❤️ ❤️ ❤️

arlo_server/models.py
jonahkagan: Oh right duh! Can you tell I've been out of the game for a year?

arlo_server/routes.py
jonahkagan: I've used `on conflict` before, but I pretty much thought it was just for `upsert` logic. Not sure how it would help here.

Open to ideas about a better thing to key off of then constraint name, but I couldn't think of anything!

PR #360, opened 2020-03-18
fix: require audit admin role to reset elections

eventualbuddha:
**Description**

Refs #290 
Piggybacks on ~#359~ #362

**Testing**

Adds a test case ensuring that only audit admins for an organization can reset elections in that organization. Also adds a test helper to clear the logged-in user.

**Progress**

I know authorization isn't the highest priority thing right now, but this is a particularly destructive action so I wanted it to be restricted sooner rather than later.

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -6,7 +6,14 @@ from flask.testing import FlaskClient
 import pytest
 
 from arlo_server import app, db
-from tests.helpers import post_json
+from arlo_server.routes import UserType
+from tests.helpers import (
+    clear_logged_in_user,
+    create_org_and_admin,
+    create_organization,
+    post_json,
+    set_logged_in_user,
+)
 import bgcompute
 
 manifest_file_path = os.path.join(os.path.dirname(__file__), "manifest.csv")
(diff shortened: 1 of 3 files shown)


eventualbuddha:
I changed this to be based on #360 instead of #359, and changed the code to enforce that there is an election before resetting it.

eventualbuddha:
**REMINDER TO MYSELF:** Merge #360 first, then re-target this and merge into master.

tests/test_app.py
```diff
@@ -78,6 +85,22 @@ def test_whole_audit_flow(client):
     assert result2["riskLimit"] == 5
 
 
+def test_reset_election(client):
```
jonahkagan: I was under the impression that the reset button was only part of the single-jurisdiction flow, so should only work with anonymous users.

jonahkagan:
I'm confused, isn't this PR #360?

eventualbuddha:
> I'm confused, isn't this PR #360?

Indeed. I meant #362.

tests/test_app.py
eventualbuddha: Ah. I suppose you're probably right. In that case… should I just decline this or require that the user be anonymous?

eventualbuddha: (Aside: if we do want to keep single vs multi as separate flows, I think we should heavily segregate them in the backend code.)

jonahkagan: What do you mean by "decline" vs "require user be anonymous"? I think this could just return a Forbidden or Unauthorized, no strong preference

I agree on segregating the flows. That's why I pushed for a separate API. I don't really want to keep both but I don't understand the product context well enough to know what makes sense

eventualbuddha: I meant "decline this PR" 😉  I think I'll do that for now.

PR #363, opened 2020-03-18
Hiding reset button when user is authenticated

MorganLove:
**Description**

- Made the reset button not show if the user is authenticated, since it's only useful on the single page flow, and hazardous on the multi jurisdiction flow.

**Testing**

- Testing that it vanishes when it isn't wanted

**Progress**

- Should be good to go!


MorganLove:
> Seems sane, assuming this is what we want product-wise.

That's what it sounds like @mcchilders wants for now anyway. :slightly_smiling_face:

PR #364, opened 2020-03-18
feat: initial version of {GET,PUT} /election/<id>/settings

eventualbuddha:
**Description**

Implements `{GET,PUT} /election/<id>/settings`.

**Testing**

Tests are in a new file, which is a pattern I'd like to do more of.

**Progress**

~I put the routes in a new file. I followed a convention where each path component in the URL is a path component in the file system. I wasn't sure what to name the `<id>` part, so I called it `__element__`. Suggestions welcome, or maybe folks don't like that direction?~

~This is still missing some tests and `test_get_empty` isn't working because response validation is failing. I can pick this up tomorrow, or someone else can take it.~

This should be ready to review & merge.


arlo_server/all_routes/election/__element__.py
```diff
@@ -0,0 +1,91 @@
...
+
+from arlo_server import app, db
+from arlo_server.auth import require_audit_admin_for_organization
+from arlo_server.models import Election
+
+ELECTION_NAME_KEY = "electionName"
+ELECTION_ONLINE_KEY = "online"
+ELECTION_RANDOM_SEED_KEY = "randomSeed"
+ELECTION_RISK_LIMIT_KEY = "riskLimit"
+ELECTION_STATE_KEY = "state"
```
jonahkagan: I know that this helps us from making mistakes, but I think it makes the code more confusing to read for me. It's helpful for my mental model to see the JSON in the code.

Do you have a strong reason for pulling out these constants?

arlo_server/all_routes/election/__element__.py
```diff
@@ -0,0 +1,91 @@
...
+        "election": {
+            ELECTION_NAME_KEY: election.name,
+            ELECTION_ONLINE_KEY: election.online,
+            ELECTION_RANDOM_SEED_KEY: election.random_seed,
+            ELECTION_RISK_LIMIT_KEY: election.risk_limit,
+            ELECTION_STATE_KEY: election.state,
+        }
+    }
+
+    validate(schema=GET_ELECTION_RESPONSE_SCHEMA, instance=response_data)
```
jonahkagan: What's the thinking behind validating this? Should already be valid in the db, right?

arlo_server/all_routes/election/__element__.py
```diff
@@ -0,0 +1,91 @@
...
+PUT_ELECTION_REQUEST_SCHEMA = {
+    "type": "object",
+    "properties": {
+        ELECTION_NAME_KEY: {"type": "string"},
+        ELECTION_ONLINE_KEY: {"type": "boolean"},
+        ELECTION_RANDOM_SEED_KEY: {"type": "string"},
+        ELECTION_RISK_LIMIT_KEY: {"type": "integer"},
+        ELECTION_STATE_KEY: {"type": "string"},
+    },
+    "additionalProperties": False,
```
jonahkagan: Nice, we should add this to more schemas!

arlo_server/all_routes/election/__element__.py
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    validate(schema=GET_ELECTION_RESPONSE_SCHEMA, instance=response_data)
+
+    return jsonify(response_data)
+
+
+@app.route("/election/<election_id>", methods=["PUT"])
+def put_election(election_id: str) -> Response:
+    election = Election.query.filter_by(id=election_id).one()
```
jonahkagan: I just found `Election.query.get_or_404(election_id)`. What do you think of starting to use that? https://flask-sqlalchemy.palletsprojects.com/en/2.x/queries/#queries-in-views

arlo_server/all_routes/election/__element__.py
```diff
@@ -0,0 +1,91 @@
...
+    require_audit_admin_for_organization(election.organization_id)
+
+    settings = request.get_json()
+    validate(schema=PUT_ELECTION_REQUEST_SCHEMA, instance=settings)
+
+    if ELECTION_NAME_KEY in settings:
+        election.name = settings[ELECTION_NAME_KEY]
```
jonahkagan: I'm realizing that these semantics are for `PATCH`, not `PUT`, which is my mistake. I'm thinking we should either:
- Change to `PATCH`
- Require every field be sent like a regular `PUT`, but accept null values

My leaning is towards the latter for consistency with other APIs. A world in which we always PUT the whole object seems simpler than a world in which some APIs PUT and some PATCH.

What do you think?

arlo_server/routes.py
```diff
@@ -42,6 +42,8 @@
 from util.jurisdiction_bulk_update import bulk_update_jurisdictions
 from util.process_file import process_file
 
+import arlo_server.all_routes.election.__element__
```
jonahkagan: I'm down with putting routes into a subdirectory, but this nested structure feels a bit more confusing to me given the size/structure of the API. The pattern I already started was to put each logical set of routes (loosely defined) in its own file within `arlo_server`, and then import them in `__init__.py`.

I think this kind of thing can get confusing when there are multiple organization systems in place, so I prefer to either do small changes that don't make a radically new scheme or migrate everything to the new scheme at once.

Could you follow the pattern I established for now and we can come back and reorg and break up routes.py in a separate PR?

For this route specifically, I think it would group well with `/election/new` and `/election/audit/reset` since those are the C and the D that accompany these endpoints (the R and U) of the election CRUD methods.

tests/test_election_show_route.py
```diff
@@ -0,0 +1,71 @@
...
+    app.config["TESTING"] = True
+    client = app.test_client()
+
+    with app.app_context():
+        db.drop_all()
+        db.create_all()
+
+    yield client
+
+    db.session.commit()
```
jonahkagan: Once I land #359 this will be unnecessary

tests/test_election_show_route.py
```diff
@@ -0,0 +1,71 @@
...
+
+@pytest.fixture
+def election_id(client: FlaskClient, request) -> str:
+    rv = post_json(
+        client,
+        "/election/new",
+        {"auditName": f"Test Audit {request.function.__name__}"},
+    )
+    election_id = json.loads(rv.data)["electionId"]
+    yield election_id
```
jonahkagan: And this

tests/test_election_show_route.py
```diff
@@ -0,0 +1,71 @@
...
+    )
+    assert rv.status_code == 400
+    assert json.loads(rv.data) == {
+        "errors": [
+            {
+                "message": "Additional properties are not allowed ('electionNameTypo' was unexpected)",
+                "errorType": "Bad Request",
+            }
+        ]
+    }
```
jonahkagan: Nice! Would love to start adding this test elsewhere

arlo_server/all_routes/election/__element__.py
```diff
@@ -0,0 +1,91 @@
...
+}
+
+
+@app.route("/election/<election_id>", methods=["GET"])
+def get_election(election_id: str) -> Response:
+    election = Election.query.filter_by(id=election_id).one()
+    require_audit_admin_for_organization(election.organization_id)
+
+    response_data = {
+        "election": {
```
jonahkagan: Why nest like this? So far, the endpoints I've seen don't use this pattern.

arlo_server/all_routes/election/__element__.py
eventualbuddha: > Do you have a strong reason for pulling out these constants?

Same as usual, preventing typos. I could buy that the schema validation is enough and that they could be duplicated there.

arlo_server/all_routes/election/__element__.py
eventualbuddha: It's not so much validating the types of the fields as forcing us to have some sort of defined structure for our responses. At Square, we used [Protocol Buffers](https://en.wikipedia.org/wiki/Protocol_Buffers), which I found to be a nice way to define our APIs. It was awesome to have files that clearly documented our request and response data structures in a way that could not get out of sync with the actual implementation. That's my main problem with the `.md` files we have for API docs. They're out of date practically instantly.

This at least ensures that we adhere to the schema we say we do.

arlo_server/all_routes/election/__element__.py
eventualbuddha: For sure. 💯

arlo_server/all_routes/election/__element__.py
eventualbuddha: Another approach would be at add this to `handle_500`:

```py
    if isinstance(original, NoResultFound):
        return (
            jsonify(
                errors=[{"message": str(original), "errorType": NotFound.__name__}]
            ),
            NotFound.code,
        )
```

The advantage is we can just keep using `.one()` and the like and it'll just work. The disadvantage is we can't easily change the error message which would look like this:

```
{"errors":[{"errorType":"NoResultFound","message":"No row was found for one()"}]}
```

arlo_server/all_routes/election/__element__.py
eventualbuddha: I'm fine with making it `PUT`.

arlo_server/routes.py
eventualbuddha: I didn't see that you'd already established a pattern here. I'll take a look.

arlo_server/all_routes/election/__element__.py
eventualbuddha: The main reason to do this is if we ever want to include metadata in the response. We could add a key that's something like `_meta` or whatever, but doing it like this makes it much easier and less messy. It's the same reason I don't consider it a good idea to return JSON with an array as the root object: it makes it harder to extend later in a non-breaking way. I'll remove the wrapper object for now, but I think we still have a lot of work to do to come up with a good extensible API.

arlo_server/all_routes/election/__element__.py
jonahkagan: I definitely like the idea of having schemas to document our request/response data structures. I agree that it's way better than written docs, especially if they are tied to the code.

I did a quick google and found this: https://pypi.org/project/jsonschema-typed/. Could that help?

arlo_server/all_routes/election/__element__.py
jonahkagan: Why couldn't we change the error message? Couldn't we just change the "message" property in the snippet you wrote?

I am torn. That is very clever and I like it a lot. I also value being explicit and minimizing magic.

arlo_server/all_routes/election/__element__.py
jonahkagan: @MorganLove do you have any thoughts on this? Are you storing the election settings in a way that makes it easy for you to send them all to the backend at once with a `PUT` request?

arlo_server/all_routes/election/__element__.py
jonahkagan: Cool, that's what I figured. I agree with you 💯 %. Want to make a follow up task?

jonahkagan: Or I could be easily convinced to just start this pattern now and go back and fix existing endpoints later.

arlo_server/all_routes/election/__element__.py
eventualbuddha: Yeah, something like that might be a good way to go. Let's explore in a follow-up.

arlo_server/all_routes/election/__element__.py
eventualbuddha: I added `util.jsonschema` to make this sort of thing easier.

arlo_server/all_routes/election/__element__.py
eventualbuddha: > Why couldn't we change the error message? Couldn't we just change the "message" property in the snippet you wrote?

We could, but that message is generic for every `NoResultFound` so we'd have to use the same message for everything unless the exception gives us enough information to figure out what happened.

Anyway, I updated the PR to use `get_or_404`.

arlo_server/routes.py
eventualbuddha: I moved to `arlo_server/election_settings.py`.

arlo_server/all_routes/election/__element__.py
eventualbuddha: We don't have to fix it now, and you and @MorganLove are likely to be the ones to own this going forward. So why don't you two think about it.

jonahkagan:
Oh man, I guess we should really put all these endpoints this under `/api` or something

eventualbuddha:
That is probably a good idea.

arlo_server/__init__.py
```diff
@@ -11,6 +11,9 @@
 db.app = app
 db.init_app(app)
 
+# The order of these imports is important as it defines route precedence.
+# Be careful when re-ordering them.
```
jonahkagan: 💯

arlo_server/routes.py
```diff
@@ -593,6 +593,12 @@ def audit_basic_update(election_id):
     return jsonify(status="ok")
 
 
+@app.route("/election/foo")
+def foooo():
+    Election.query.filter_by(id="abcdefg").one()
+    return jsonify()
```
jonahkagan: what this?

arlo_server/election_settings.py
```diff
@@ -0,0 +1,63 @@
...
+from arlo_server import app, db
+from arlo_server.auth import require_audit_admin_for_organization
+from arlo_server.models import Election
+
+from util.jsonschema import nullable, Obj, Str, Bool, Int
+
+
+PUT_ELECTION_SETTINGS_REQUEST_SCHEMA = Obj(
+    electionName=Str,
+    online=Bool,
```
jonahkagan: Any reason these two aren't nullable as well?

arlo_server/election_settings.py
```diff
@@ -0,0 +1,63 @@
...
+    randomSeed=nullable(Str),
+    riskLimit=nullable(Int),
+    state=nullable(Str),
```
jonahkagan: If you wanted to, could add an enum to validate the `state`

arlo_server/election_settings.py
```diff
@@ -0,0 +1,63 @@
...
+from arlo_server.models import Election
+
+from util.jsonschema import nullable, Obj, Str, Bool, Int
+
+
+PUT_ELECTION_SETTINGS_REQUEST_SCHEMA = Obj(
+    electionName=Str,
+    online=Bool,
+    randomSeed=nullable(Str),
+    riskLimit=nullable(Int),
```
jonahkagan: Could add `minimum` and `maximum` validators for this

util/jsonschema.py
```diff
@@ -0,0 +1,19 @@
...
+    return {"anyOf": [schema, {"type": "null"}]}
+
+
+def Obj(**properties: Dict[str, JSONSchema]) -> JSONSchema:
+    return {
+        "type": "object",
+        "properties": properties,
+        "additionalProperties": False,
+        "required": list(properties.keys()),
+    }
```
jonahkagan: 💯 💯 💯 

This is a cool module

benadida:
> Oh man, I guess we should really put all these endpoints this under `/api` or something

tests/test_election_settings.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+    assert rv.status_code == 400, f"unexpected response: {rv.data}"
+    assert json.loads(rv.data) == {
+        "errors": [
+            {
+                "message": "Additional properties are not allowed ('electionNameTypo' was unexpected)",
+                "errorType": "Bad Request",
+            }
+        ]
+    }
```
jonahkagan: Do you need a test for invalid values as well?

tests/test_election_settings.py
```diff
@@ -0,0 +1,79 @@
...
+            "electionName": "An Updated Name",
+            "online": False,
+            "riskLimit": None,
+            "randomSeed": None,
+            "state": None,
+        },
+    )
+    assert rv.status_code == 200, f"unexpected response: {rv.data}"
+    assert json.loads(rv.data) == {"status": "ok"}
+    assert Election.query.filter_by(id=election_id).one().name == "An Updated Name"
```
jonahkagan: Do you think it's worth trying out updating all the fields?

jonahkagan: Also the pattern I noticed in other tests was to use the GET API to test the results of the PUT. I kind of like that because it exercises the interface that a consumer uses.

arlo_server/routes.py
eventualbuddha: Whoops. I was testing `get_or_404`.

arlo_server/election_settings.py
eventualbuddha: They aren’t nullable in the database.

arlo_server/election_settings.py
eventualbuddha: I like that. Have the values handy?

arlo_server/election_settings.py
eventualbuddha: Will do. 0-100?

tests/test_election_settings.py
eventualbuddha: Sure, will do.

tests/test_election_settings.py
eventualbuddha: Sounds good, will do.

arlo_server/election_settings.py
jonahkagan: Good question. I found this: https://www.50states.com/abbreviations.htm

arlo_server/election_settings.py
jonahkagan: Seems reasonable

arlo_server/election_settings.py
eventualbuddha: 🤦‍♂ I took this to mean "audit state" like "ready" or "frozen" or "completed" 🤣

jonahkagan: Lol, I was wondering. I mean it's not necessarily as simple as the 50 states...

arlo_server/all_routes/election/__element__.py
MorganLove: @jonahkagan If the information is all on one screen in the setup stage, then it's all able to be sent at once. It's when it's split across multiple screens that things get sketchy. I think all the audit settings data is in that one screen, though, so we should be good.

arlo_server/election_settings.py
MorganLove: @jonahkagan @eventualbuddha :laughing: That is an unfortunate ambiguity. Should we perhaps change the name of `state` to something more explicit?

Also, I'm sending them as the full name of the state like `'Alabama'`, not abbreviations. I could implement it so that it displays the option as the full name but sends back the abbreviation I suppose... depends on where we're displaying it on client-side and which one we want to see I suppose. Thoughts, @mcchilders?

eventualbuddha: I would be inclined to store abbreviations and to have a mapping for display. Otherwise localization becomes very annoying.

arlo_server/election_settings.py
MorganLove: Right now the front end limits it to 1-20. We should probably have them match?

arlo_server/election_settings.py
jonahkagan: An amazing and rare case where the government defines the enum for you!

arlo_server/all_routes/election/__element__.py
jonahkagan: They are all on one screen except `state`, I believe.

arlo_server/election_settings.py
eventualbuddha: Updated.

arlo_server/election_settings.py
eventualbuddha: Updated.

eventualbuddha:
Alright, thanks for all the feedback @jonahkagan and @MorganLove. I believe this is ready to go again. PTAL?

arlo_server/models.py
```diff
@@ -361,3 +361,64 @@ class ProcessingStatus(str, Enum):
     PROCESSING = "PROCESSING"
     PROCESSED = "PROCESSED"
     ERRORED = "ERRORED"
+
+
+class USState(str, Enum):
```
jonahkagan: This looks great.

Is there no way to actually use this in the db model? Or did you just choose to not validate at the db level?

util/jsonschema.py
```diff
@@ -0,0 +1,42 @@
+from typing import Any, Dict, List
```
jonahkagan: I have a small fear that over time it is going to become difficult to design an interface that allows us to use all the features of json schema that we need. A small fear though...

jonahkagan:
Except failing tests

MorganLove:
@eventualbuddha Just want to clarify that you're expecting the abbreviation for the state value from the frontend, right?

Also, does this support sending the `state` before all the others? That's the only split, as @jonahkagan mentioned. I can send `state` with the other values, but I can't send the other values when I first set `state`, if that makes sense.

arlo_server/models.py
eventualbuddha: I chose not to validate at the db level.

util/jsonschema.py
eventualbuddha: We can always mix and match.

eventualbuddha:
@MorganLove as-is you must pass all properties. You can pass `null` for ones that you don't have. I recommend `GET`ing first, updating the object, and `PUT`ing it back.

PR #365, opened 2020-03-18
Initial jurisdictions file upload setup

MorganLove:
**Description**

- Added a field for file upload for the jurisdictions csv
- Added the sample csv file
- Attempted the PUT and POST for the file and the state name

**Testing**

- Added endpoint tests and updated snapshots

**Progress**

- File upload is complete, though the polling for processing completion of the file might be better done in a separate PR, since that will incorporate some state management for changing from stage to stage (we don't want to go forward in a stage until it's ready, so we need a system for unlocking them as you go), and the state selection isn't submitted to the endpoint yet.
- Given the above, this PR might be ready to go after tests


arlo-client/public/sample_jurisdiction_filesheet.csv
```diff
@@ -0,0 +1,4 @@
+"Jurisdiction","Admin Email"
+"LaSalle Twp","admin@lasalletwp.com"
+"Detroit","lbeech@detroitelections.gov"
+"Ann Arbor","elections@annarbor.mi.us"
```
eventualbuddha: Are these real? If so, change them to not be real?

arlo-client/src/components/Audit/Setup/Participants/index.tsx
```diff
@@ -26,18 +31,60 @@ interface IProps {
...
+      //     body: JSON.stringify(data),
+      //     headers: {
+      //       'Content-Type': 'application/json',
+      //     },
+      //   }
+      // )
+      // if (checkAndToast(response)) {
+      //   setIsLoading(false)
+      //   return
+      // }
```
eventualbuddha:
```suggestion
```

arlo-client/src/components/Audit/Setup/Participants/index.tsx
```diff
@@ -56,6 +103,46 @@ const Participants: React.FC<IProps> = ({ audit, nextStage }: IProps) => {
...
+                }}
+                onInputChange={e => {
+                  setFieldValue(
+                    'csv',
+                    (e.currentTarget.files && e.currentTarget.files[0]) ||
+                      undefined
+                  )
+                }}
+                hasSelection={!!values.csv}
+                text={values.csv ? values.csv.name : 'Select csv...'}
```
eventualbuddha:
```suggestion
                text={values.csv ? values.csv.name : 'Select CSV...'}
```

arlo-client/src/components/Audit/Setup/Participants/index.tsx
```diff
@@ -56,6 +103,46 @@ const Participants: React.FC<IProps> = ({ audit, nextStage }: IProps) => {
               />
             </label>
             <ErrorMessage name="state" component={ErrorLabel} />
+            {/* When one is already uploaded, this will be toggled to show its details, with a button to reveal the form to replace it */}
+            <FormSection>
+              <FormSectionDescription>
+                Click &quot;Browse&quot; to choose the appropriate field from
+                your computer. This file should be a comma-ceparated list of all
```
eventualbuddha:
```suggestion
                your computer. This file should be a comma-separated list of all
```

arlo-client/src/components/Audit/Setup/Participants/index.tsx
```diff
@@ -56,6 +103,46 @@ const Participants: React.FC<IProps> = ({ audit, nextStage }: IProps) => {
               />
             </label>
             <ErrorMessage name="state" component={ErrorLabel} />
+            {/* When one is already uploaded, this will be toggled to show its details, with a button to reveal the form to replace it */}
+            <FormSection>
+              <FormSectionDescription>
+                Click &quot;Browse&quot; to choose the appropriate field from
```
eventualbuddha:
```suggestion
                Click &quot;Browse&quot; to choose the appropriate file from
```

arlo-client/src/setupProxy.js
```diff
@@ -16,6 +16,7 @@ module.exports = function(app) {
   app.use(proxy('/election/new', { target }))
   app.use(proxy('/election/*/audit/**', { target }))
   app.use(proxy('/election/*/jurisdiction/**', { target }))
+  app.use(proxy('/election/*/jurisdictions/**', { target }))
```
eventualbuddha: Should we use `…/jurisdiction/…` for this? cc @benadida @jonahkagan

eventualbuddha:
Approved as a WIP, meaning "so far so good". I'm assuming there's a lot more work to do including tests.

arlo-client/src/setupProxy.js
benadida: Do we have both `jurisdiction` and `jurisdictions`? If so would be good to standardize, yes.

arlo-client/public/sample_jurisdiction_filesheet.csv
MorganLove: Good question. That's just the one @mcchilders uploaded. I'll edit it to be obviously fake.

arlo-client/src/setupProxy.js
MorganLove: @eventualbuddha @benadida Interestingly, now I take a closer look, the actual endpoint we were using for the single page flow is `.../jurisdictions` and the new one just has `/file` appended to it. It doesn't look like we *have* a singular jurisdiction endpoint proper. :thinking: I wonder why it worked with this file as it was. Maybe I'm missing something?

eventualbuddha: In a separate PR, we should change `jurisdictions` to `jurisdiction` and then leave this line out.

MorganLove:
I am running into issues resolving bugs in the last few tests. For some reason in the test environment the form is not getting submitted.

arlo-client/src/setupProxy.js
eventualbuddha: https://github.com/votingworks/arlo/pull/368

arlo-client/src/components/Audit/Setup/Participants/index.tsx
```diff
@@ -26,18 +31,49 @@ interface IProps {
 }
 
 const Participants: React.FC<IProps> = ({ audit, nextStage }: IProps) => {
+  const { electionId } = useParams()
   const [isLoading, setIsLoading] = useState(false)
+  const submit = async (values: IValues) => {
+    console.log('SUBMIT')
```
eventualbuddha:
```suggestion
```

arlo-client/src/components/Audit/Setup/Participants/index.tsx
```diff
@@ -26,18 +31,49 @@ interface IProps {
...
+    try {
+      setIsLoading(true)
+      console.log('Is this posting?')
```
eventualbuddha:
```suggestion
```

arlo-client/src/components/Audit/Setup/Participants/index.tsx
```diff
@@ -26,18 +31,49 @@ interface IProps {
...
+  }
   return (
     <Formik
       initialValues={initialValues}
       validationSchema={schema}
       onSubmit={v => {
-        setIsLoading(true)
-        console.log(v)
-        nextStage()
+        console.log('ONSUBMIT')
```
eventualbuddha:
```suggestion
```

arlo-client/src/components/Audit/Setup/Participants/index.tsx
```diff
@@ -64,7 +140,10 @@ const Participants: React.FC<IProps> = ({ audit, nextStage }: IProps) => {
                 type="submit"
                 intent="primary"
                 disabled={!!audit.frozenAt}
-                onClick={handleSubmit}
+                onClick={e => {
+                  console.log('CLICK')
```
eventualbuddha:
```suggestion
```

arlo-client/src/components/Audit/Setup/Participants/states.ts
```diff
@@ -1,53 +1,68 @@
...
+  WI: 'Wisconsin',
+  WY: 'Wyoming',
+}
+
+export const labelValueStates = Object.keys(states).map(value => ({
+  value,
+  label: states[value],
+}))
+
+export default states
```
eventualbuddha: Why not use an `enum`? It'd be something like this:

```ts
enum USState {
  AL = 'Alabama'
  …
}
```

You end up with pretty much the same object, but more type-safe.

Also, we seem to have used different values for the states: [`USState` in models.py](https://github.com/votingworks/arlo/blob/f75ce2a5e388ec95bde93a3a2ec75f1626bc6990/arlo_server/models.py#L370-L428). Perhaps we should stick this in a text file and generate backend and frontend models? Or the frontend could just ask the backend for the list. This seems to be a pretty complete list: https://en.wikipedia.org/wiki/List_of_U.S._state_abbreviations. I'd probably include everything but the ones listed "obsolete". @mcchilders, does it matter what we include or exclude here?

mcchilders: It needs all 50 states, none of the territories for now.

MorganLove: @eventualbuddha I just tried an enum, but it broke the generation of the array of labels and values needed for the dropdown. It doesn't look like enums play nicely with mapping, possibly? Do you have an example of how this would work?

On the other hand, we are never going to be actually dealing with these values directly. They're always going to be behind the scenes, and interacted with as plain strings. So I'm not sure they need to be more type-safe than they already are. Are you thinking of a particular situation where it would be needed?

MorganLove:
This could be merged as-is, with the state selection submission to its endpoint in a separate PR.

arlo-client/src/components/Audit/Setup/Participants/_mocks.ts
```diff
@@ -0,0 +1,12 @@
+const jurisdictionFile = new File(
```
benadida: any way to load that from the `.csv` file that has the same data, so as to minimize duplication?

eventualbuddha: I fixed this in 8a8cecd.

arlo-client/src/components/Audit/Setup/Participants/index.tsx
```diff
@@ -26,19 +31,45 @@ interface IProps {
...
+        const errorResponse: IErrorResponse = await api(
+          `/election/${electionId}/jurisdictions/file`,
+          {
+            method: 'PUT',
+            body: formData,
+          }
+        )
+        if (checkAndToast(errorResponse)) return
+      }
+      nextStage()
```
eventualbuddha: I don't see where this is polling `…/jurisdictions/file` to ensure the upload is processed before moving on. Does it do that anywhere? Should it?

MorganLove: @eventualbuddha It is not yet. When I got to that point, I realized that there is no mechanism in place to manage going from stage to stage in a controlled way. What if they click on the sidebar to jump ahead when the data isn't ready yet? What if they jump backwards and edit something and then things have to re-process before going forward? Polling whether a particular stage is ready to be accessed is part of that control flow, and I'm currently developing that in a separate branch, in order to keep PR size down.

PR #369, opened 2020-03-20
test: improve error messages for `compare_json`

eventualbuddha:
**Description**

Changes `compare_json` to keep track of the key path that lead to the values being compared and ensures that the path is included in the error message. The original assertion message is still included, but we now include a key path like so: `root["rounds"][0]["startedAt"]`.

**Testing**

`compare_json` doesn't have any tests of its own behavior, but is used in lots of tests.

**Progress**

I made this change while working on #344 since I was changing some deeply-nested fields and wasn't sure exactly which value `compare_json` was complaining about.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -1,6 +1,7 @@
 import uuid, json, re
+import pytest
 from email.utils import parsedate_to_datetime
-from typing import Any, Optional
+from typing import Any, List, Optional, Union
 from flask.testing import FlaskClient
 
 from arlo_server.auth import UserType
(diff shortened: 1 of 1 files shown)


tests/helpers.py
```diff
@@ -90,20 +90,41 @@ def assert_startswith(x: str):
 
 # Checks that a json blob (represented as a Python dict) is equal-ish to an expected
 # dict. The expected dict can contain assertion functions in place of any non-deterministic values.
-def compare_json(actual_json, expected_json):
+def compare_json(
+    actual_json, expected_json, keypath: Optional[List[Union[str, int]]] = None
```
benadida: suggest calling the argument something like `current_keypath` or `keypath_breadcrumbs`, or something that indicates that you're just using it to keep track of the keypath so far.

eventualbuddha: Renamed it `current_keypath`, moved the guts to an inner function, and added tests for `compare_json` specifically.

jonahkagan:
Wow cool!! We could almost publish this as its own library now!

PR #370, opened 2020-03-20
fix: use ISO-8601 date format

eventualbuddha:
**Description**

As discussed elsewhere, this changes our API to consistently use ISO-8601 dates as understood by `datetime.datetime.isoformat` and `datetime.datetime.fromisoformat`.

cc @jonahkagan @MorganLove 

**Testing**

Existing tests cover this change.

**Progress**

There's nothing here that actively prevents non-ISO-8601 dates from being used. Perhaps there's a way to teach `jsonify` to use this format specifically?


jonahkagan:
Hurrah! I do think there is a way to give `jsonify` a custom encoder: https://stackoverflow.com/a/44158611/1472662

PR #371, opened 2020-03-20
refactor: extract ballot manifest file fields

eventualbuddha:
**Description**

This will make it easier to process the ballot manifest file in the background (#344), re-using existing background file processing code.

**Testing**

Covered by existing tests as this should not change behavior.

**Progress**

This is progress toward #344.

### DIFF
--- a/arlo_server/models.py
+++ b/arlo_server/models.py
@@ -72,17 +72,16 @@ class Jurisdiction(BaseModel):
         db.String(200), db.ForeignKey("election.id", ondelete="cascade"), nullable=False
     )
     name = db.Column(db.String(200), nullable=False)
-    manifest = db.Column(db.Text, nullable=True)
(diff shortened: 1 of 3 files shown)


arlo_server/models.py
```diff
@@ -72,18 +72,17 @@ class Jurisdiction(BaseModel):
...
+    manifest_file_id = db.Column(
+        db.String(200), db.ForeignKey("file.id", ondelete="set null"), nullable=True
+    )
+    manifest_file = relationship("File")
+
     # any error in the upload? null == none
     manifest_errors = db.Column(db.Text, nullable=True)
 
-    # a JSON array of field names that are included in the CSV
-    manifest_fields = db.Column(db.Text, nullable=True)
```
eventualbuddha: This was unused, so I figured I remove it while I was already in here changing the model.

PR #375, opened 2020-03-23
feat: process ballot manifest uploads in the background

eventualbuddha:
**Description**

This tries to maintain as much backward-compatibility as possible, but one breaking change is that the endpoint changed from POST to PUT. I updated the API docs and the frontend code to use PUT.

**Testing**

Adjusts current tests, but doesn't really add any new ones.

**Progress**

The front-end still needs to be updated to check the `ballotManifest.processing.status` value.

### DIFF
--- a/arlo_server/jurisdictions.py
+++ b/arlo_server/jurisdictions.py
@@ -7,6 +7,7 @@ from arlo_server.routes import (
     isoformat,
 )
 from arlo_server.models import Jurisdiction
+from util.process_file import serialize_file, serialize_file_processing
 
 
 def serialize_jurisdiction(db_jurisdiction: Jurisdiction) -> dict:
@@ -14,14 +15,14 @@ def serialize_jurisdiction(db_jurisdiction: Jurisdiction) -> dict:
         "id": db_jurisdiction.id,
         "name": db_jurisdiction.name,
         "ballotManifest": {
-            "filename": db_jurisdiction.manifest_file.name
+            "file": serialize_file(db_jurisdiction.manifest_file)
+            if db_jurisdiction.manifest_file
+            else None,
+            "processing": serialize_file_processing(db_jurisdiction.manifest_file)
             if db_jurisdiction.manifest_file
             else None,
             "numBallots": db_jurisdiction.manifest_num_ballots,
             "numBatches": db_jurisdiction.manifest_num_batches,
-            "uploadedAt": isoformat(db_jurisdiction.manifest_file.uploaded_at)
-            if db_jurisdiction.manifest_file
-            else None,
         },
     }
 
--- a/arlo_server/models.py
+++ b/arlo_server/models.py
@@ -80,9 +80,6 @@ class Jurisdiction(BaseModel):
     )
     manifest_file = relationship("File")
 
-    # any error in the upload? null == none
-    manifest_errors = db.Column(db.Text, nullable=True)
-
     batches = relationship("Batch", backref="jurisdiction", passive_deletes=True)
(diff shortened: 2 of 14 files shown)


arlo_server/jurisdictions.py
```diff
@@ -7,18 +7,26 @@
...
+            else None,
+            "processing": serialize_file_processing(db_jurisdiction.manifest_file)
             if db_jurisdiction.manifest_file
             else None,
             "numBallots": db_jurisdiction.manifest_num_ballots,
             "numBatches": db_jurisdiction.manifest_num_batches,
+            # Deprecated fields.
+            "filename": db_jurisdiction.manifest_file.name
+            if db_jurisdiction.manifest_file
+            else None,
```
jonahkagan: I don't think this needs to be backwards compatible. I just added this endpoint and I don't think it's being used yet.

arlo_server/models.py
```diff
@@ -80,9 +80,6 @@ class Jurisdiction(BaseModel):
     )
     manifest_file = relationship("File")
 
-    # any error in the upload? null == none
-    manifest_errors = db.Column(db.Text, nullable=True)
-
```
jonahkagan: 💯

util/process_file.py
```diff
@@ -34,3 +35,33 @@ def process_file(session: Session, file: File, callback: Callable[[], None]) ->
...
+        "uploadedAt": isoformat(file.uploaded_at),
+    }
+
+    if contents:
+        result["contents"] = file.contents
+
+    return result
+
+
+def serialize_file_processing(file: File) -> Dict[str, Optional[str]]:
```
jonahkagan: Think it's worth making a `JSON` type alias?

eventualbuddha: Maybe. I'll tinker with it.

eventualbuddha: The simplest version of type declarations for JSON that are better than `Any` look something like this:

```py
JSONDict = Dict[str, "JSON"]
JSONArray = List["JSON"]
JSONValue = Union[str, int, float, None]
JSON = Union[JSONDict, JSONArray, JSONValue]
```

Unfortunately, [`mypy` does not support recursive types](https://github.com/python/mypy/issues/731). That leaves us with this:

```py
JSONDict = Dict[str, Any]
JSONArray = List[Any]
JSONValue = Union[str, int, float, None]
JSON = Union[JSONDict, JSONArray, JSONValue]
```

I don't think that's really better, so I'm going to just make these `Dict[str, Any]`.

audits/sampler.py
```diff
@@ -42,15 +45,21 @@ def draw_sample(seed, manifest, sample_size, num_sampled=0):
...
+        list(
+            consistent_sampler.sampler(
+                ballots,
+                seed=seed,
+                take=sample_size + num_sampled,
+                with_replacement=True,
+                output="tuple",
+            )
+        )[num_sampled:],
+    )
```
eventualbuddha: cc @umbernhard

arlo_server/jurisdictions.py
eventualbuddha: Removed.

util/process_file.py
jonahkagan: What if we just use `JSONDict` as you defined it, since that's mostly what we'll be using it for? I think it will help the readability of the code to know what functions are spitting out data that is intended to be a JSON object.

audits/sampler.py
umbernhard: Looks right to me!

PR #376, opened 2020-03-23
Add GET /election/<election_id>/sample-sizes

jonahkagan:
**Description**

This endpoint returns the estimated sample size options for the current round (the frontend will only use this for the first round for now, but we'll need the logic for later rounds as well, so it's implemented). This is based on the `compute_sample_sizes` function from the single-jurisdiction flow.

**Testing**

Added some basic tests for the new endpoint - two error cases where we don't have enough of the audit set up done to compute the sample sizes, one happy path.

**Progress**

Ready for review.

### DIFF
--- a/arlo_server/sample_sizes.py
+++ b/arlo_server/sample_sizes.py
@@ -1,0 +1,55 @@
+from flask import jsonify, request
+from collections import defaultdict
+import math
+from werkzeug.exceptions import BadRequest
+from typing import Dict
+
+from arlo_server import app, db
+from arlo_server.models import Election, Contest
+from arlo_server.routes import require_audit_admin_for_organization
+from audits import bravo, sampler_contest
+
+
+@app.route("/election/<election_id>/sample-sizes", methods=["GET"])
+def get_sample_sizes(election_id: str):
+    election = Election.query.get_or_404(election_id)
+    require_audit_admin_for_organization(election.organization_id)
+
+    if not election.contests:
+        raise BadRequest("Cannot compute sample sizes until contests are set")
+    if not election.risk_limit:
+        raise BadRequest("Cannot compute sample sizes until risk limit is set")
+
+    # For now, we only support one targeted contest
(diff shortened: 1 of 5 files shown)


arlo_server/sample_sizes.py
```diff
@@ -0,0 +1,54 @@
...
+    for result in contest.results:
+        results_by_choice[result.contest_choice_id] += result.result
+    results_so_far = {contest.id: results_by_choice}
+
+    # Do the math!
+    sample_sizes = bravo.get_sample_size(
+        election.risk_limit / 100,
+        sampler_contest.from_db_contest(contest),
+        results_so_far,
+    )
```
jonahkagan: @umbernhard I noticed that `bravo.get_sample_size` takes in the `sample_results` object (which I named `results_so_far` here), and that this object in the old single-jurisdiction code tried to aggregate results for every contest (even though there was only ever one contest). I figured that `bravo.get_sample_size` only considers the one contest that its passed anyway, so there was no point in passing in results for other contests. Does that sound right?

If so, might be worth changing the interface to `get_sample_size` to just take in the results for that one contest instead of a mapping of contest ids to results.

arlo_server/sample_sizes.py
```diff
@@ -0,0 +1,55 @@
...
+from arlo_server import app, db
+from arlo_server.models import Election, Contest
+from arlo_server.routes import require_audit_admin_for_organization
+from audits import bravo, sampler_contest
+
+
+@app.route("/election/<election_id>/sample-sizes", methods=["GET"])
+def get_sample_sizes(election_id: str):
+    election = Election.query.get_or_404(election_id)
+    require_audit_admin_for_organization(election.organization_id)
```
benadida: suggestion for later improvement: a decorator that transforms `election_id` param into `election` param, returns 404 if not found, returns 403 is not right permission.

eventualbuddha: #380

Issue #338, opened 2020-03-13
Back-end work to correctly represent multi-jurisdiction contest universes


mcchilders:
*User Story:* 
As an AA I want to be able to define the universe for a particular contest across multiple jurisdictions so that I can audit contests that appear in one, many, or all of the jurisdictions involved in an audit.

*Background:* 
In plain language, a "contest universe" is the geography where a particular contest appeared on the ballot. For this milestone in Arlo, this translates to a list of jurisdictions where that contest appeared. (Eventually, we will want to add another level of granularity, so that a contest universe can be limited to particular ballot batches from the ballot manifest for a particular jurisdiction -- that will allow us to correctly audit contests that don't span an entire jurisdiction, but are only in certain precincts in a particular jurisdiction etc. For now though, that level isn't necessary, and the smallest notion of a contest universe will be 1 jurisdiction.)

Right now, in the single page, single-jurisdiction flow, there is assumed to be a one-to-one relationship between contest, contest universe, & jurisdiction - we support only one of each, so the contest universe for the (only) contest can be assumed to be the (only) jurisdiction. In the new multi-jurisdiction flow, the universe for any of our contest(s) can be: 1 jurisdiction, many jurisdictions, or all of the jurisdictions. Rather than being an assumed 1-to-1, the contest universe is defined by the AA during the audit setup for each targeted and opportunistic contest, and stored on the back end.

The contest universe is used for two things: 
1) on the front end, if the AB's are using online data entry, the contest(s) that appear on the on the data entry screens for the audit boards in a particular jurisdiction include *all the contests that have that jurisdiction in their contest universe.*
2) on the back end, the universe of ballots from which you select random ballots for audit for a given contest is the combined ballot manifests for all the jurisdictions in the contest universe for that contest

*We need to be able to:*
- Associate each ballot manifest with the jurisdiction that uploaded it  **DONE**
- Associate a set of jurisdictions (one, many, or all available) with each contest (targeted or opportunistic) as the contest universe **DONE**
- Associate each sampled ballot with the contest(s) that it is sampled for in each round **DONE** *(implicitly - we know the jurisdiction for each sampled ballot, and the previous bullet associates jurisdictions with contests)*
-- This universe is defined by the AA user in the AA audit setup workflow, when the rest of the contest data is entered
- When creating the pool of ballots for a particular contest, in order to randomly select ballots for audit, the pool of available ballots should include all the ballots from all the manifests in the jurisdictions in that contest's universe **DONE**
[...]

PR #377, opened 2020-03-23
GET/POST rounds endpoints

jonahkagan:
**Description**

Task: #338 
Adds two new endpoints:
- GET /election/<election_id>/round - list the rounds in the audit so far
- POST /election/<election_id>/round - create a new round, which has a side effect of sampling the ballots

**Testing**

Added lots of new tests for validation as well as functionality.

**Progress**

Ready for review.

### DIFF
--- a/arlo_server/rounds.py
+++ b/arlo_server/rounds.py
@@ -1,0 +1,156 @@
+from flask import jsonify, request
+from jsonschema import validate
+from werkzeug.exceptions import BadRequest, Conflict
+import uuid, datetime
+
+from arlo_server import app, db
+from arlo_server.models import (
+    Round,
+    RoundContest,
+    Election,
+    SampledBallot,
+    SampledBallotDraw,
+)
+from arlo_server.auth import with_election_access, UserType
+from arlo_server.sample_sizes import sample_size_options
+from util.isoformat import isoformat
+from audits import sampler
+
+
+CREATE_ROUND_REQUEST_SCHEMA = {
+    "type": "object",
+    "properties": {
+        "roundNum": {"type": "integer", "minimum": 1,},
+        "sampleSize": {"type": "integer", "minimum": 1,},
+    },
+    "additionalProperties": False,
+    "required": ["roundNum"],
+}
+
+
+def serialize_round(r: Round) -> dict:
+    return {
+        "id": r.id,
+        "roundNum": r.round_num,
+        "startedAt": isoformat(r.started_at),
+        "endedAt": isoformat(r.ended_at),
+    }
+
+
+# Raises if invalid
+def validate_round(r: dict, election: Election):
+    validate(r, CREATE_ROUND_REQUEST_SCHEMA)
+
+    rounds = sorted(election.rounds, key=lambda r: r.round_num, reverse=True)
+    current_round = next(iter(rounds), None)
+    next_round_num = current_round.round_num + 1 if current_round else 1
+    if r["roundNum"] != next_round_num:
+        raise BadRequest(f"The next round should be round number {next_round_num}")
+
+    if current_round and not current_round.ended_at:
+        raise Conflict(f"The current round is not complete")
+
+    if r["roundNum"] == 1 and "sampleSize" not in r:
+        raise BadRequest(f"Sample size is required for round 1")
+
+
+def sample_ballots(election: Election, round: Round, sample_size: int):
+    # For now, we only support one targeted contest
+    targeted_contest = next(c for c in election.contests if c.is_targeted)
+
+    # Compute the total number of ballot samples in all rounds leading up to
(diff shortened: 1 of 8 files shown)


arlo_server/rounds.py
```diff
@@ -0,0 +1,155 @@
...
+@app.route("/election/<election_id>/round", methods=["GET"])
+def list_rounds(election_id: str):
+    election = Election.query.get_or_404(election_id)
+    require_audit_admin_for_organization(election.organization_id)
+
+    rounds = (
+        Round.query.filter_by(election_id=election_id).order_by(Round.round_num).all()
+    )
+
+    return jsonify({"rounds": [serialize_round(r) for r in rounds]})
```
jonahkagan: @eventualbuddha you convinced me to have a top-level wrapper object. I know I convinced you not to, lol, but I think you're right that we should move in this direction

arlo_server/rounds.py
```diff
@@ -0,0 +1,154 @@
...
+    # Compute the total number of ballot samples in all rounds leading up to
+    # this one. Note that this corresponds to the number of SampledBallotDraws,
+    # not SampledBallots.
+    previous_rounds = (
+        db.session.query(RoundContest)
+        .join(Round)
+        .filter(Round.election_id == election.id, Round.id != round.id)
+        .all()
+    )
+    num_sampled = sum(r.sample_size for r in previous_rounds)
```
jonahkagan: @benadida @umbernhard please check this. The single-jurisdiction flow computed this by actually counting `SampledBallotDraw`s in the db, but I thought this would be a bit simpler of a db query. I want to be sure they are equivalent.

benadida: They should be the same, but given that the instructions we give audit boards are taken from `SampledBallotDraw`, I have a small preference to use that as the reference for how many were sampled. Small preference, your call.

jonahkagan: It ended up being simpler that way haha! ba5fbe7

arlo_server/rounds.py
eventualbuddha: For an array I really think it is necessary, and then it kind of trickles out…

jonahkagan: Do you also think it's a good practice for the body of POST/PUT requests? Or can those take the data without any wrapper?

benadida: +1 on wrapping arrays in an object. It stems from an old security bug with cross-site JavaScript, which I believe is mostly fixed, but I think is worthwhile sticking to :)

arlo_server/rounds.py
```diff
@@ -0,0 +1,150 @@
...
+
+    # Create the pool of ballots to sample (aka manifest) by combining the
+    # manifests from every jurisdiction in the contest's universe.
+    # Audits must be deterministic and repeatable for the same real world
+    # inputs. So the sampler expects the same input for the same real world
+    # data. Thus, we use the jurisdiction and batch names (deterministic real
+    # world ids) instead of the jurisdiction and batch ids (non-deterministic
+    # uuids that we generate for each audit).
+    manifest = {
+        (jurisdiction.name, batch.name): batch.num_ballots
```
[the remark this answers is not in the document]
jonahkagan: It seems sufficient for now to me, since each batch in the contest universe should be uniquely identified by jurisdiction+batch.

I guess the only caveat I could think of (which is maybe what you're getting at) is that if we need to change this later to add more fields, then audits run before that change might not sample the same ballots as identical audits run after that change. Is that an issue? I can't really think of any way to work around it if that's the case.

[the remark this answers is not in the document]
umbernhard: I think future changes to this will be for batch manifests, where we have (jurisdiction, batch name, number of ballots). Or CVRs too. I'm not super sure this is very important right now, I just wanted to flag it so we could think about it a little more. 

Another thing to worry about is when there's a collision between jurisdiction and batch names, e.g. if Ann Arbor City and Ann Arbor Township both omit their full designations and they share a precinct name. So you'd get (Ann Arbor, PS1) twice, with no way to disambiguate. That can be handled by assertions or something like that during data entry, I suppose.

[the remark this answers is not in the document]
jonahkagan: Your point about collisions seems very important. I think we should definitely enforce the uniqueness of jurisdiction names within an election and batch names within a jurisdiction. We can do this at the db level. cc @benadida to confirm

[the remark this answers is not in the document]
benadida: @jonahkagan I agree on requiring unique jurisdiction names within an audit, and unique batch names within a manifest. Worth asking @mcchilders if she foresees any issue, but my sense is the complexity goes up a LOT if we can't make this assumption.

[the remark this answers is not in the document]
jonahkagan: Confirmed with @mcchilders. Added task: #389

PR #378, opened 2020-03-24
Audit setup stages ui management

MorganLove:
**Description**

- Lifting up and extracting out state management for the setup stages so we can add in logic for controlling navigation between them based on polling the data.
- Corrected `/jurisdictions/file` to `/jurisdiction/file` to comply with the recent endpoint fix

**Testing**

- Updating snapshots and test implementation

**Progress**

- I'm still working on doing the actual polling logic, but I wanted to get this one pushed up before it got too big and scope creep set in.

### DIFF
--- a/arlo-client/src/components/Audit/index.tsx
+++ b/arlo-client/src/components/Audit/index.tsx
@@ -74,7 +74,11 @@ const Audit: React.FC<{}> = () => {
     'Participants'
   )
 
-  const [menuItems] = useSetupMenuItems(stage, setStage)
+  const [menuItems, refresh] = useSetupMenuItems(stage, setStage)
+
+  useEffect(() => {
+    refresh()
+  }, [])
 
   const activeStage = menuItems.find(m => m.title === stage)
   const nextStage = menuItems[menuItems.indexOf(activeStage!) + 1]
--- a/arlo-client/src/components/Audit/index.tsx
+++ b/arlo-client/src/components/Audit/index.tsx
@@ -78,7 +78,7 @@ const Audit: React.FC<{}> = () => {
 
   useEffect(() => {
     refresh()
-  }, [])
+  }, [refresh])
 
   const activeStage = menuItems.find(m => m.title === stage)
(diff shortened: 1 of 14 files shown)


arlo-client/src/components/Audit/index.tsx
```diff
@@ -79,17 +74,15 @@ const Audit: React.FC<{}> = () => {
...
-  )
+  const [menuItems, refresh] = useSetupMenuItems(stage, setStage)
+
+  useEffect(() => {
+    refresh()
+  }, [])
+
+  const activeStage = menuItems.find(m => m.title === stage)
+  const nextStage = menuItems[menuItems.indexOf(activeStage!) + 1]
+  const prevStage = menuItems[menuItems.indexOf(activeStage!) - 1]
```
benadida: Any edge cases to consider here when `indexOf` is first or last element?

MorganLove: > Any edge cases to consider here when `indexOf` is first or last element?

@benadida Right now, those functions are only accessible when they wouldn't be undefined, but that's not checked programmatically. I'll take a look and see if there's a place where it would be efficacious to check for that to prevent human error.

benadida: @MorganLove are we talking about the same thing? I'm wondering what happens if `menuItems.indexOf(activeStage!)` returns `0`, and then `prevStage` is set to `menuItems[-1]`. Is that simply a null and it's ok?

MorganLove: @benadida I believe we are. If it returns `0` then `prevStage` is `undefined`, but if the index of the active stage is 0, then that's the first item in the menu and the first item never gets passed `prevStage`, since it's meaningless. Therefore, nothing ever tries to access values on the object in a situation in which it would be undefined.

I just pushed a commit that makes it a bit more explicit, by marking the types in these lines as `| undefined` and then in the `Setup` component marking where we know it isn't undefined and commenting where we shouldn't be sending in the undefined values. Hope that helps. :slightly_smiling_face:

Issue #258, opened 2020-02-09
Define docstring approach for Arlo, support automatic documentation generation


nealmcb:
Projects benefit when documentation can be automatically generated from the docstrings in the code. Popular tools for doing so in the Python world include Sphinx and pdoc, as discussed at [Comparison of Python documentation generators - Peter Kong](https://medium.com/@peterkong/comparison-of-python-documentation-generators-660203ca3804)

The current Arlo docstring style is most similar to the [Google docstring format](http://google.github.io/styleguide/pyguide.html), but uses a dash ` - ` instead of a colon `: ` to set off names from definitions. That breaks the documentation generators.

I suggest that Arlo should define its docstring format, complying with something that can be leveraged automatically, and identify a tool to use for generating documentation. Over time the
code can be brought into compliance.

nealmcb:
It still seems to me like a good idea to format comments to help tools generate good documentation, but the current practice doesn't support that. Did I miss something? Can we re-open this?

PR #379, opened 2020-03-24
Improves documentation and cleans up.

umbernhard:
**Description**

* Fixed documentation in audit classes to reflect their inputs and
outputs.
* Removes audit.py, since we're no longer using objects.
* Generally cleans code.

**Testing**

None

**Progress**

Ready for review and merge.


nealmcb:
I was going to comment on the move away from objects/classes for audits. I still think they would be very valuable. E.g. in Poorvi's R2B2 teams' Athena statistical module implementation, we want much of the code in bravo.py, including margin calculations, checks for ties and single-winners, etc. Did I miss a conversation about the shift away from classes?

nealmcb:
As we work on documentation, I want to highlight #258 on refining the docstring approach and suggest adopting it here mostly by changing dashes to colons. Automatically generating API documentation is super helpful in my mind.

Issue #335, opened 2020-03-12
Improve permissions in routes


eventualbuddha:
We should improve the permissions access in routes such that:
* it's hard/impossible to forget to apply _and_
* it's easy to tell by looking at a route what permissions are required _and_
* it applies all types of user access, i.e. audit admins, jurisdiction admins, and audit board members.

@benadida @jonahkagan and I discussed it and believe that a decorator could work as long as requests to routes that do not have the decorator simply fail.

jonahkagan:
More thoughts from @benadida: https://github.com/votingworks/arlo/pull/376#discussion_r396809885

nikhilb4a:
Closing as completed - Claude review: "PR #742 (unified restrict_access decorator for all user types) and PR #1670 (startup enforcement that every route has a decorator)

The issue (March 2020, by Brian Donovan) asked for route permissions where (1) it's hard/impossible to forget to apply them, (2) you can tell at a glance what permissions a route requires, (3) all user types are covered, and specifically proposed a decorator approach where undecorated routes simply fail. All four points are now true in the codebase:

Explicit, readable decorators: every route is decorated with restrict_access([UserType.AUDIT_ADMIN]), restrict_access_support, allow_public_access, or allow_any_logged_in_user_access (defined in [auth_helpers.py:164-278](vscode-webview://0k4k6l6jq1j000v6br37se94obt1aph4jrn6rfigbb226lgpt5dc/server/auth/auth_helpers.py#L164-L278)). This landed incrementally starting with @with_jurisdiction_access (March 31, 2020 — just weeks after the issue was filed) and was unified into the current restrict_access in PR #742 (Sep 2020).
All user types: the UserType enum covers audit admins, jurisdiction admins, audit boards, and even tally entry users (added later), plus support users via a separate decorator.
Impossible to forget: PR #1670, "Ensure access control on all routes" (Nov 2022, Jonah Kagan), added a startup check in [app.py:64-69](vscode-webview://0k4k6l6jq1j000v6br37se94obt1aph4jrn6rfigbb226lgpt5dc/server/app.py#L64-L69) that iterates every registered route and raises an exception if any lacks an access-control decorator. That's actually stronger than the issue's ask — the app refuses to boot rather than failing requests at runtime. Public routes must opt in explicitly with allow_public_access, so there's no implicit default."

PR #381, opened 2020-03-24
refactor: introduce election access route decorator

### LINKED ISSUE #380: Add decorator that transforms `election_id` param into `election` param, returns 404 if not found, returns 403 is not right permission
suggestion for later improvement: a decorator that transforms `election_id` param into `election` param, returns 404 if not found, returns 403 is not right permission.

_Originally posted by @benadida in https://github.com/votingworks/arlo/pull/376_

eventualbuddha:
**Description**

Closes #380
Refs #290

Decorating a route using `with_election_access` allows the route to be defined with an `election` parameter that will be be provided only if the election exists and the current user has access to it.

**Testing**

Existing tests should cover this okay, but to really test it I'd have to figure out how to test this more directly. I'm not sure what the right approach is there. Maybe create another Flask app per test that I then add test routes to?

**Progress**

This is a building block toward #290.

### DIFF
--- a/arlo_server/auth.py
+++ b/arlo_server/auth.py
@@ -1,9 +1,10 @@
+import functools
 from enum import Enum
 from flask import session
-from typing import Optional, Tuple, Union
+from typing import Callable, Optional, Tuple, Union
 from werkzeug.exceptions import Unauthorized, Forbidden
 
-from arlo_server.models import User
+from arlo_server.models import Election, User
 
 
 class UserType(str, Enum):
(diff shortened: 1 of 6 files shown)


arlo_server/routes.py
```diff
@@ -440,11 +438,15 @@ def update_jurisdictions_file(election_id=None):
...
-    ]
+    missing_fields = (
+        [
+            field
+            for field in [JURISDICTION_NAME, ADMIN_EMAIL]
+            if field not in jurisdictions_csv.fieldnames
+        ]
+        if jurisdictions_csv.fieldnames
+        else [JURISDICTION_NAME, ADMIN_EMAIL]
+    )
```
eventualbuddha: I'm not sure why this change was required, but `mypy` all of a sudden started complaining:

```
arlo_server/routes.py:444: error: Unsupported right operand type for in ("Optional[Sequence[str]]")
```

I infer that `jurisdictions_csv.fieldnames` is `Optional[Sequence[str]]` and that `in` does not apply to `None`. But I don't understand why this is only now a problem since the type of this shouldn't have changed.

jonahkagan:
Also refs #335

arlo_server/routes.py
```diff
@@ -25,12 +25,13 @@
...
 from arlo_server import app
 from arlo_server.auth import (
+    UserType,
     clear_loggedin_user,
     get_loggedin_user,
     get_loggedin_user_record,
     require_audit_admin_for_organization,
     set_loggedin_user,
-    UserType,
+    with_election_access,
```
jonahkagan: Could we add this to all the routes that have an `election_id` param and call `require_audit_admin_for_organization`?

arlo_server/auth.py
```diff
@@ -55,3 +56,22 @@ def require_audit_admin_for_organization(organization_id: Optional[str]):
     raise Forbidden(
         description=f"{user.email} does not have access to organization {organization_id}"
     )
+
+
+def with_election_access(user_type: UserType):
```
jonahkagan: Maybe a comment explaining what this does?

arlo_server/routes.py
jonahkagan: So weird! Here's an idea for something less verbose: 
```python
     missing_fields = [
        field
        for field in [JURISDICTION_NAME, ADMIN_EMAIL]
        if field not in (jurisdictions_csv.fieldnames or [])
    ]
```

jonahkagan:
For testing, maybe just pick a really simple/basic endpoint and write all the tests for the logic of this decorator against that endpoint?

arlo_server/routes.py
eventualbuddha: Sure, forgot about other files!

PR #382, opened 2020-03-24
refactor: validate JSON schemas thoroughly

eventualbuddha:
**Description**

This ensures that `additionalProperties` and `required` are set on objects, and recursively checks schemas. This will not help with situations such as adding a property but forgetting to add it to `required`, but is better than the status quo.

**Testing**

Tests are covered by validation.

**Progress**

This was just something I noticed and wanted to fix.

### DIFF
--- a/util/jsonschema.py
+++ b/util/jsonschema.py
@@ -1,42 +1,63 @@
-from typing import Any, Dict, List
+import jsonschema
+import jsonschema.validators
+from typing import Any, Dict, List, Union
 
-JSONSchema = Any
-Str = {"type": "string"}
-Int = {"type": "integer"}
-Bool = {"type": "boolean"}
-
-
-def nullable(schema: JSONSchema) -> JSONSchema:
-    return {"anyOf": [schema, {"type": "null"}]}
+JSONSchema = Dict[str, Any]
 
 
-def Obj(**properties: Dict[str, JSONSchema]) -> JSONSchema:
-    return {
-        "type": "object",
-        "properties": properties,
-        "additionalProperties": False,
-        "required": list(properties.keys()),
-    }
+def validate(instance: Any, schema: JSONSchema):
+    jsonschema.validators.validator_for(schema).check_schema(schema)
+    validate_schema(schema)
+    jsonschema.validate(instance=instance, schema=schema)
 
 
-def Enum(values: List[str]) -> JSONSchema:
-    return {
-        "type": "string",
-        "enum": values,
-    }
+def validate_schema(schema: JSONSchema):
+    def validate_schema_node(node: JSONSchema, current_keypath: List[Union[str, int]]):
+        assert isinstance(node, dict)
+        if node.get("type", None) == "object":
+            properties = node["properties"]
+            assert isinstance(properties, dict)
+
+            if "additionalProperties" not in node:
+                raise jsonschema.exceptions.ValidationError(
+                    f"'additionalProperties' must be present on objects, and should probably be False (at {_serialize_keypath(current_keypath)})"
+                )
+
+            if "required" not in node:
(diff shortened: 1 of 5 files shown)


jonahkagan:
I am starting to feel like just writing the raw JSON schema is preferable for me. I also ran into the optional object properties problem recently.

Here's my analysis of this approach:
Pros:
- More concise
- Avoids typos

Cons:
- Adds a level of complexity (now you have to learn both JSON schema and our DSL)
- Hard to design an interface that maps well to all the JSON schema functionality

Thanks to the meta-schema validation built into the JSON schema library (it validates your schema against a meta-schema for schemas), I'm not so worried about typos.

For me, just being more concise isn't worth it. Are there other advantages you see?

eventualbuddha:
> Are there other advantages you see?

Can’t forget to disallow additional properties, can’t forget to mark properties as required. Does jsonschema validate that required property names match actual properties on the object?

jonahkagan:
> Can’t forget to disallow additional properties, can’t forget to mark properties as required.

Ok, those are both reasonable concerns. I wonder if we could write a wrapper around `jsonschema.validate` that checks some of those constraints for us.

> Does jsonschema validate that required property names match actual properties on the object?

I don't think entries in `required` need to also be in `properties`, since they describe two diff things (which properties are required vs the types for specific properties). Not sure if that's what you were asking though.

eventualbuddha:
> I don't think entries in `required` need to also be in `properties`, since they describe two diff things (which properties are required vs the types for specific properties). Not sure if that's what you were asking though.

How can you have a required property that doesn’t have a type? Would it implicitly be any type?

eventualbuddha:
Here's an example of a schema that looks right at first glance, but actually doesn't constrain the type of `name` at all:

```py
jsonschema.validate(
  schema={"type": "object", "required": ["name"], "properties": {"naem": {"type": "string"}}},
  instance={"name": 3}
)
```

If you fix the typo, then the instance in that example is rejected properly:

```
jsonschema.exceptions.ValidationError: 3 is not of type 'string'

Failed validating 'type' in schema['properties']['name']:
    {'type': 'string'}

On instance['name']:
    3
```

jonahkagan:
> How can you have a required property that doesn’t have a type? Would it implicitly be any type?

Definitely seems weird for a type system. I guess JSON schema is not really a type system, but rather a system for expressing constraints on data. You can say "this field is required" and that's a constraint by itself. You can also say "this field is a string." But they don't seem to depend on each other.

jonahkagan:
That example def looks bad. What do you think about the idea of writing some code to validate that our schemas don't have those mistakes? (As opposed to writing code to construct schemas that don't have mistakes)

eventualbuddha:
@jonahkagan I changed this to do what you suggested. WDTY?

util/jsonschema.py
```diff
@@ -1,42 +1,61 @@
...
-    }
+import jsonschema
+from typing import Any, Dict, List, Union
+
+JSONSchema = Dict[str, Any]
+
+
+def validate(instance: Any, schema: JSONSchema):
+    validate_schema(schema)
+    jsonschema.validate(instance=instance, schema=schema)
```
jonahkagan: Do you think it's worth calling the `jsonschema` library's validator on the schema before calling our validator? I was thinking it might be able to catch some more basic structural errors and provide good error messages. It looks like this method would do it: https://python-jsonschema.readthedocs.io/en/stable/validate/#jsonschema.IValidator.check_schema

eventualbuddha: I think I got it working, though it's clearly not an API they really intend for outside folks to use for this purpose.

jonahkagan: 😳

PR #384, opened 2020-03-25
Audit settings

MorganLove:
**Description**

- Add form to the Audit Settings screen
- Add submission to the `/election/:electionId/settings` endpoint from the Participants and Audit Settings screens

**Testing**

- none yet

**Progress**

- ~~TODO: tests~~ Done!

### DIFF
--- a/arlo-client/src/components/Audit/Setup/Participants/index.tsx
+++ b/arlo-client/src/components/Audit/Setup/Participants/index.tsx
@@ -82,7 +82,7 @@ const Participants: React.FC<IProps> = ({ audit, nextStage }: IProps) => {
                   setFieldValue('state', e.currentTarget.value)
                 }
                 disabled={!!audit.frozenAt}
-                value={values.state}
+                value={values.state || ''}
                 options={[{ value: '' }, ...labelValueStates]}
               />
             </label>
--- a/arlo-client/src/components/Audit/Setup/Participants/index.tsx
+++ b/arlo-client/src/components/Audit/Setup/Participants/index.tsx
@@ -16,15 +16,11 @@ import { ErrorLabel } from '../../../Form/_helpers'
 import FormSection, { FormSectionDescription } from '../../../Form/FormSection'
 import { api, checkAndToast } from '../../../utilities'
 import { ISidebarMenuItem } from '../../../Atoms/Sidebar'
+import useAuditSettings from '../useAuditSettings'
 
 export const Select = styled(HTMLSelect)`
   margin-top: 5px;
 `
-
-const initialValues = {
-  state: '',
-  csv: null,
-}
 
 interface IProps {
   audit: IAudit
@@ -34,9 +30,11 @@ interface IProps {
 const Participants: React.FC<IProps> = ({ audit, nextStage }: IProps) => {
   const { electionId } = useParams()
   const [isLoading, setIsLoading] = useState(false)
+  const [{ state = '' }, updateState] = useAuditSettings(electionId!)
   const submit = async (values: IValues) => {
     try {
       setIsLoading(true)
+      updateState({ state: values.state })
       /* istanbul ignore else */
       if (values.csv) {
         const formData: FormData = new FormData()
@@ -57,9 +55,10 @@ const Participants: React.FC<IProps> = ({ audit, nextStage }: IProps) => {
   }
   return (
     <Formik
-      initialValues={initialValues}
+      initialValues={{ state, csv: null }}
       validationSchema={schema}
       onSubmit={submit}
+      enableReinitialize
     >
       {({
         handleSubmit,
@@ -83,6 +82,7 @@ const Participants: React.FC<IProps> = ({ audit, nextStage }: IProps) => {
                   setFieldValue('state', e.currentTarget.value)
                 }
                 disabled={!!audit.frozenAt}
+                value={values.state}
                 options={[{ value: '' }, ...labelValueStates]}
               />
             </label>
@@ -129,7 +129,7 @@ const Participants: React.FC<IProps> = ({ audit, nextStage }: IProps) => {
             </FormSection>
           </FormWrapper>
           {isLoading && <Spinner />}
-          {!audit.contests.length && !isLoading && (
+          {!isLoading && (
             <FormButtonBar>
               <FormButton
                 type="submit"
--- a/arlo-client/src/components/Audit/Setup/Participants/index.tsx
+++ b/arlo-client/src/components/Audit/Setup/Participants/index.tsx
@@ -30,7 +30,7 @@ interface IProps {
 const Participants: React.FC<IProps> = ({ audit, nextStage }: IProps) => {
   const { electionId } = useParams()
   const [isLoading, setIsLoading] = useState(false)
-  const [{ state = '' }, updateSettings] = useAuditSettings(electionId!)
+  const [{ state }, updateSettings] = useAuditSettings(electionId!)
   const submit = async (values: IValues) => {
     try {
       setIsLoading(true)
@@ -56,7 +56,7 @@ const Participants: React.FC<IProps> = ({ audit, nextStage }: IProps) => {
   }
   return (
     <Formik
-      initialValues={{ state, csv: null }}
+      initialValues={{ state: state || '', csv: null }}
       validationSchema={schema}
       onSubmit={submit}
       enableReinitialize
--- a/arlo-client/src/components/Audit/Setup/Participants/index.tsx
+++ b/arlo-client/src/components/Audit/Setup/Participants/index.tsx
@@ -30,11 +30,11 @@ interface IProps {
 const Participants: React.FC<IProps> = ({ audit, nextStage }: IProps) => {
   const { electionId } = useParams()
   const [isLoading, setIsLoading] = useState(false)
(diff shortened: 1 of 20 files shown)


arlo-client/src/setupProxy.js
```diff
@@ -15,6 +15,7 @@ module.exports = function(app) {
   app.use(proxy('/admin', { target }))
   app.use(proxy('/election/new', { target }))
   app.use(proxy('/election/*/audit/**', { target }))
+  app.use(proxy('/election/*/settings', { target }))
   app.use(proxy('/election/*/jurisdiction/**', { target }))
   app.use(proxy('/election/*/jurisdictions/**', { target }))
   app.use(proxy('/election/*/admin/**', { target }))
```
benadida: is there any reason all of these aren't just `/election/*`?

arlo-client/src/components/Audit/Setup/useAuditSettings.tsx
```diff
@@ -0,0 +1,73 @@
...
+    return true
+  }
+
+  useEffect(() => {
+    ;(async () => {
+      const newSettings = await getSettings()
+      setSettings(newSettings)
+    })()
+  }, [getSettings])
+  return [settings, updateSettings]
```
benadida: if I'm reading this right, I think the settings are fetched once every time a `useAuditSettings` is called, then cached in state until updated. Is that right? If so, I'm good with it, I have a small bit of concern about there being a "hidden" cache here that could bite us later, but I'm not seeing an explicit way in which this would fail, so I can live with it.

arlo-client/src/setupProxy.js
eventualbuddha: It breaks development, since it sends e.g. `/election/<election_id>` to the python app which then serves the production-built React app rather than the development one.

arlo-client/src/components/Audit/Setup/Participants/index.tsx
```diff
@@ -34,9 +30,12 @@ interface IProps {
 const Participants: React.FC<IProps> = ({ audit, nextStage }: IProps) => {
   const { electionId } = useParams()
   const [isLoading, setIsLoading] = useState(false)
+  const [{ state = '' }, updateState] = useAuditSettings(electionId!)
```
eventualbuddha: Naming this `updateState` implies that you can call `updateState(state)`, but you can't. Just call it `updateSettings`?

arlo-client/src/components/Audit/Setup/useAuditSettings.tsx
eventualbuddha: It looks like, for any given `useAuditSettings` instance, it'll cache the values as long as the election ID does not change. It will only use new values if they are set via `updateSettings` or the whole component that uses `useAuditSettings` is torn down and built again. Values changing on the server would therefore not be reflected here, but maybe that's okay?

arlo-client/src/components/Audit/Setup/Settings/types.ts
```diff
@@ -0,0 +1,8 @@
+import { IAuditSettings } from '../../../../types'
+
+export interface IValues {
+  electionName: IAuditSettings['electionName']
+  online: IAuditSettings['online']
+  randomSeed: IAuditSettings['randomSeed']
+  riskLimit: IAuditSettings['riskLimit']
+}
```
eventualbuddha: One of these might be better:

```ts
export type IValues = Pick<IAuditSettings, 'electionName' | 'online' | 'randomSeed' | 'riskLimit'>
export type IValues = Omit<IAuditSettings, 'state'>
```

arlo-client/src/components/Audit/Setup/Settings/index.tsx
```diff
@@ -1,24 +1,149 @@
...
+          <FormWrapper title="Audit Settings">
+            <FormSection>
+              {/* eslint-disable jsx-a11y/label-has-associated-control */}
+              <label htmlFor="audit-name" id="audit-name-label">
+                Election Name
+                <Field
+                  id="election-name"
+                  aria-labelledby="election-name-label"
+                  name="electionName"
+                  disabled={!!audit.frozenAt}
```
jonahkagan: Where is `frozenAt` being set? I know that @benadida had originally built an API to "freeze" the audit, but I took that out of the API proposal I wrote. In the world I proposed, an audit's settings should no longer be updated as soon as the first round starts (which you can tell when you hit the /election/<election_id>/round API), but there's no indicator in the election settings API.

arlo-client/src/components/Audit/Setup/useAuditSettings.tsx
```diff
@@ -0,0 +1,73 @@
...
+      ...newSettings,
+    }
+    const response: IErrorResponse = await api(
+      `/election/${electionId}/settings`,
+      {
+        method: 'PUT',
+        body: JSON.stringify(mergedSettings),
+        headers: {
+          'Content-Type': 'application/json',
+        },
```
jonahkagan: Would it be useful to make a helper function to send json to the API? The server doesn't have good error messages if you forget this header (as we learned a few days ago).

arlo-client/src/components/Audit/Setup/useAuditSettings.tsx
MorganLove: @benadida @eventualbuddha That's how I've set it up, yes. This is mainly for code reuse, since we're accessing the settings data from two separate forms.

If values change on the server in between the initial loading of the component and submitting values from the form, it will only overwrite the changes in the values represented in the form you're on. That way there are no surprises. Does that make sense?

arlo-client/src/components/Audit/Setup/Settings/index.tsx
MorganLove: @jonahkagan Oh, I missed that distinction. Hm. So, to clarify, what you're saying is that when the 'Launch Audit' button is hit and launching is confirmed on the last step of the setup flow, that starts the first round? And having rounds is exactly identical to setup being completed and locked?

If so, I'll need to add another state management hook to keep track of the rounds api responses. Right now, I'm just using the regular audit /status endpoint.

arlo-client/src/components/Audit/Setup/useAuditSettings.tsx
MorganLove: We could possibly have a couple variations of the `api` helper for sending different kinds of requests.

MorganLove: We could possibly have a couple variations of the `api` helper for sending different kinds of requests.

arlo-client/src/components/Audit/Setup/Settings/index.tsx
jonahkagan: Yes and yes to your questions. You understand me correctly.

I mentally bucketed `/audit/status` as a single-juris flow endpoint. It seems a bit tricky to try to make it return correct info for both single-juris and multi-juris flows. So better not to rely on it here.

Is hitting the rounds API a good way to manage the overall control flow? I figured the rounds are the real world "control flow" of the audit, so it makes sense. But if there's another API that would work better, let me know and let's talk through it.

arlo-client/src/components/Audit/Setup/Participants/index.tsx
MorganLove: I waffled over the naming here a bit, yeah. Good point.

arlo-client/src/components/Audit/Setup/Settings/index.tsx
MorganLove: @jonahkagan So would I be pinging `GET /election/<election_id>/round` then? What response will I be expecting pre-launch vs post-launch?

It's a shift from how the front end architecture was handling things before. On the single-page flow the /status endpoint was the single source of truth for all data. It's pinged at every interaction with the server (plus some more times) to update everything. Now, everything is broken up into a bunch of endpoints that appear to work more or less independently of each other, and there is no longer a single way to access the single source of truth. So I'm just trying to figure out which ones need to be pinged when and for which part of the app.

What I'm seeing for this particular piece, since whether or not all of the setup forms are locked or not is a pretty critical piece of information, I'll end up having to ping an endpoint (whether it's the rounds endpoint or something else) on every rerender to make sure the data is up-to-date. So as long as it's light-weight it doesn't particularly matter which endpoint it is, I don't think.

MorganLove: Regardless, for this PR, I'll need to leave the frozen bit as is for now, since that's a more involved change that needs its own PR.

jonahkagan: Yes, you'd hit `GET /election/<election_id>/round`. Pre-launch, you'd get this `{"rounds": []}`. Post-launch, you'd get `{"rounds": [ ... array of round objects, ordered by roundNum ... ]}`.

Currently, this endpoint simply reads the rounds from the db. It's a two line function, so I'd say pretty lightweight! 😄 

Makes me think we should also put in some server-side logic to block requests to all the set up endpoints after the first round is created.

MorganLove:
3/26 - complete except for tests

MorganLove:
Now complete, with tests passing!

PR #385, opened 2020-03-25
test(pylint): enable pylint in CI

eventualbuddha:
**Description**

This enables a single rule, `singleton-comparison`, which was pretty easy to fix all the violations for. The idea is to add rules over time to the `.pylintrc` and enforce them in CI to ensure we don't regress.

**Testing**

n/a

**Progress**

n/a

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,503 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code.
+extension-pkg-whitelist=
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=.git
+
+# Add files or directories matching the regex patterns to the blacklist. The
+# regex matches against base names, not paths.
+ignore-patterns=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
+# number of processors available to use.
+jobs=1
+
+# Control the amount of potential inferred values when inferring a single
+# object. This can help the performance when dealing with large functions or
+# complex, nested conditions.
+limit-inference-results=100
+
+# List of plugins (as comma separated values of python module names) to load,
+# usually to register additional checkers.
+load-plugins=
+
+# Pickle collected data for later comparisons.
+persistent=yes
+
+# Specify a configuration file.
+#rcfile=
+
+# When enabled, pylint would attempt to guess common misconfiguration and emit
+# user-friendly hints instead of false-positive error messages.
+suggestion-mode=yes
+
+# Allow loading of arbitrary C extensions. Extensions are imported into the
+# active Python interpreter and may run arbitrary code.
+unsafe-load-any-extension=no
+
+
+[MESSAGES CONTROL]
+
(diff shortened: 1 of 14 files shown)


tests/test_audit_basic_update.py
```diff
@@ -69,4 +69,4 @@ def test_audit_basic_update_sets_default_for_contest_is_targeted(client, electio
     assert json.loads(rv.data)["status"] == "ok"
 
     rv = client.get(f"/election/{election_id}/audit/status")
-    assert json.loads(rv.data)["contests"][0]["isTargeted"] == True
+    assert json.loads(rv.data)["contests"][0]["isTargeted"]
```
jonahkagan: The original assertion was attempting to check if the value was exactly `True`, not truthy. (Though based on the semantics of `==` it might not have been doing that)

tests/test_election_settings.py
```diff
@@ -37,7 +37,7 @@ def test_update_election(client: FlaskClient, election_id: str):
 
     election_record = Election.query.filter_by(id=election_id).one()
     assert election_record.election_name == "An Updated Name"
-    assert election_record.online == True
+    assert election_record.online
```
jonahkagan: same here

bgcompute.py
```diff
@@ -57,6 +58,7 @@ def bgcompute_update_election_jurisdictions_file() -> int:
 
 
 def bgcompute_update_ballot_manifest_file() -> int:
+    # pylint: disable=singleton-comparison
```
eventualbuddha: You might be wondering why there are some places where I disable this rule. It's because sqlalchemy filters do not work with `… is None`, but require `== None`. Alternatively, [we could do `.is_(None)`](https://docs.sqlalchemy.org/en/13/orm/tutorial.html#common-filter-operators). Preference?

jonahkagan: I think either `.is_(None)` or put the comment that disables the rule right on the relevant line. It was confusing to see the comment at the top of the function.

tests/test_audit_basic_update.py
eventualbuddha: Alright, that seems worth preserving. I added comments to disable the rules in these tests.

.pylintrc
```diff
@@ -0,0 +1,503 @@
+[MASTER]
```
eventualbuddha: I generated this file with this: `pylint --generate-rcfile > .pylintrc`. Then I edited the `disabled=` and `enabled=` to be what you see in this PR.

bgcompute.py
eventualbuddha: I'll go with `.is_(None)`. Trying to put it on the same line means you end up fighting with `black` since it wants to wrap the newly-long line and then the comment no longer applies to the right line.

tests/test_audit_basic_update.py
jonahkagan: Could you change the test to use `is True`?

eventualbuddha:
This LGTY, @jonahkagan?

tests/helpers.py
```diff
@@ -96,6 +96,23 @@ def assert_startswith(x: str):
...
+
+    return assert_contains_elements
+
+
+def dict_containing(**kwargs):
+    def assert_contains_elements(x):
+        assert isinstance(x, dict)
+        for k, v in kwargs.items():
+            assert k in x
+            compare_json(x[k], v)
```
jonahkagan: what are these for?

PR #391, opened 2020-03-25
Modular tests

umbernhard:
**Description**

Moves tests into subdirectories so they can be run modularly. Also adds make directives for each testing module. 

**Testing**

N/A

**Progress**

Ready for review.


Makefile
```diff
@@ -45,3 +45,17 @@ test-client:
 # To run a specific test: TEST=<test name> make test-server
 test-server:
 	FLASK_ENV=test python3.7 -m pipenv run python -m pytest -k '${TEST}' --ignore=arlo-client
+
+# Only tests audit_math
+test-math:
+	FLASK_ENV=test python3.7 -m pipenv run python -m pytest tests/audit_math_tests -k '${TEST}' --ignore=arlo-client
```
eventualbuddha: I think the `--ignore=arlo-client` is unnecessary in these since you're specifying a test file path.

PR #392, opened 2020-03-25
Add currentRoundStatus to list contests endpoint

jonahkagan:
**Description**

This adds a field `currentRoundStatus` to each contest in the output of
/election/<election_id>/contest. `currentRoundStatus` is null until the
audit starts. Once the audit starts, it is an object with two fields:

- `isRiskLimitMet` - whether the calculated risk measurement for that
contest meets the audit risk limit (or null if we're not done auditing
ballots for that contest yet)

- `numSampledBallots` - how many ballot samples were drawn for this contest

**Testing**

Updating existing tests for the null case. Added a new test to test the status once the round has started.

**Progress**

Ready for review.

### DIFF
--- a/arlo_server/contests.py
+++ b/arlo_server/contests.py
@@ -1,11 +1,25 @@
-from typing import List
+from typing import List, Dict, Any, Optional
 from flask import request, jsonify
 from werkzeug.exceptions import BadRequest
+from sqlalchemy import func
 
 from arlo_server import app, db
-from arlo_server.routes import with_election_access, UserType
-from arlo_server.models import Contest, ContestChoice, Election, Jurisdiction
+from arlo_server.auth import with_election_access, UserType
+from arlo_server.models import (
+    Contest,
+    ContestChoice,
+    Election,
+    Jurisdiction,
+    Round,
+    RoundContest,
+    SampledBallotDraw,
+    Batch,
+)
+from arlo_server.rounds import get_current_round
 from util.jsonschema import validate
+
+# An approximation of a JSON object type, since we can't do recursive types
+JSONDict = Dict[str, Any]
 
 CONTEST_CHOICE_SCHEMA = {
     "type": "object",
@@ -14,6 +28,7 @@ CONTEST_CHOICE_SCHEMA = {
         "name": {"type": "string"},
         "numVotes": {"type": "integer", "minimum": 0},
     },
+    "additionalProperties": False,
     "required": ["id", "name", "numVotes"],
     "additionalProperties": False,
 }
@@ -30,6 +45,7 @@ CONTEST_SCHEMA = {
         "votesAllowed": {"type": "integer", "minimum": 1},
         "jurisdictionIds": {"type": "array", "items": {"type": "string"}},
     },
+    "additionalProperties": False,
     "required": [
         "id",
         "name",
@@ -44,7 +60,7 @@ CONTEST_SCHEMA = {
 }
 
 
-def serialize_contest_choice(contest_choice: ContestChoice) -> dict:
+def serialize_contest_choice(contest_choice: ContestChoice) -> JSONDict:
     return {
         "id": contest_choice.id,
         "name": contest_choice.name,
@@ -52,7 +68,8 @@ def serialize_contest_choice(contest_choice: ContestChoice) -> dict:
     }
 
 
-def serialize_contest(contest: Contest) -> dict:
+def serialize_contest(contest: Contest, round_status: Optional[JSONDict]) -> JSONDict:
+    jurisdictions = sorted(contest.jurisdictions, key=lambda j: j.name)
     return {
         "id": contest.id,
         "name": contest.name,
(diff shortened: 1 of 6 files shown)


arlo_server/contests.py
```diff
@@ -107,13 +125,46 @@ def validate_contests(contests: List[dict]) -> None:
...
+        return {c.id: None for c in contests}
+
+    sampled_ballot_count_by_contest = dict(
+        SampledBallotDraw.query.filter_by(round_id=round.id)
+        .join(Batch)
+        .join(Jurisdiction)
+        .join(Jurisdiction.contests)
+        .group_by(Contest.id)
+        .values(Contest.id, func.count(SampledBallotDraw.batch_id))
+    )
```
jonahkagan: Would appreciate a sanity check on this. I'm still new to SQLAlchemy join logic.

arlo_server/contests.py
```diff
@@ -1,11 +1,25 @@
...
+    Round,
+    RoundContest,
+    SampledBallotDraw,
+    Batch,
+)
+from arlo_server.rounds import get_current_round
+from sqlalchemy import func
+
+# An approximation of a JSON object type, since we can't do recursive types
+JSONDict = Dict[str, Any]
```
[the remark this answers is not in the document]
jonahkagan: mypy doesn't support them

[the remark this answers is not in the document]
umbernhard: Oh I see, I was imagining just nesting the types like Dict[str, Dict[...]].

arlo_server/contests.py
eventualbuddha: I find it helpful to plug queries into the python REPL. Often I get back the SQL if I omit the `.all()` or `.one()` part. With this one I have to omit `.values(…)`, which doesn't give you the exact query but it's close. Here's the actual query:

```sql
SELECT contest.id AS contest_id, count(sampled_ballot_draw.batch_id) AS count_1 
FROM sampled_ballot_draw JOIN batch ON batch.id = sampled_ballot_draw.batch_id JOIN jurisdiction ON jurisdiction.id = batch.jurisdiction_id JOIN contest_jurisdiction AS contest_jurisdiction_1 ON jurisdiction.id = contest_jurisdiction_1.jurisdiction_id JOIN contest ON contest.id = contest_jurisdiction_1.contest_id 
WHERE sampled_ballot_draw.round_id = %(round_id_1)s GROUP BY contest.id
```

I only managed to get that by substituting in an invalid value (the int `1`) for `round_id`, which then showed me the query in an exception.

tests/conftest.py
```diff
@@ -85,3 +88,40 @@ def election_settings(client: FlaskClient, election_id: str) -> None:
...
+                    b"1,23\n"
+                    b"2,101\n"
+                    b"3,122\n"
+                    b"4,400"
+                ),
+                "manifest.csv",
+            )
+        },
+    )
+    assert rv.status_code == 200
```
eventualbuddha: One thing I've started doing with these is to [add a message that includes the response](https://github.com/votingworks/arlo/blob/6fdab2ff06d8cc08739e2eaf4a3f1016a3c81566/tests/routes_tests/test_election_settings.py#L34). Otherwise, it can become difficult to determine why this is happening with a `200 != 400` message.

arlo_server/contests.py
```diff
@@ -110,13 +125,46 @@ def validate_contests(json_contests: List[dict]) -> None:
...
+    # isRiskLimitMet will be None until we have computed the risk measurement
+    # for that contest, which happens once we're done auditing its sampled
+    # ballots. Once the risk measurement is calculated, isRiskLimitMet will be
+    # a boolean.
+    return {
+        c.id: {
+            "isRiskLimitMet": round_is_complete_by_contest[c.id],
+            "numBallotsSampled": sampled_ballot_count_by_contest.get(c.id, 0),
+        }
+        for c in contests
```
benadida: wow I've never used dictionary comprehension, that's so cool :)

arlo_server/contests.py
```diff
@@ -110,13 +125,46 @@ def validate_contests(json_contests: List[dict]) -> None:
...
+
+
 @app.route("/election/<election_id>/contest", methods=["PUT"])
 @with_election_access(UserType.AUDIT_ADMIN)
 def create_or_update_all_contests(election: Election):
     json_contests = request.get_json()
     validate_contests(json_contests)
 
-    db.session.query(Contest).filter_by(election_id=election.id).delete()
+    Contest.query.filter_by(election_id=election.id).delete()
```
benadida: I'm annoyed that:

`election.contests.delete()`

doesn't work.

arlo_server/contests.py
benadida: what is your goal with `func.count(SampledBallotDraw.batch_id)` as opposed to `func.count()`? Are you trying to count distinct batches? If so, that's not going to work, you'd need to `count(distinct(..))` or also group by the `batch_id`.

But I don't think that's what you're trying to do, so I imagine `func.count()` is all you want. I think that's the only tweak I've made, and as best as I can tell the query is right.

jonahkagan: Oooh cool trick!

arlo_server/contests.py
jonahkagan: I tried `election.contests = []` and that didn't seem to work either.

tests/conftest.py
jonahkagan: I guess that's a good argument for the pattern that existing tests were using: `assert json.loads(rv.data) == {"status": "ok"}`. I started checking status code instead because it seemed simpler, but maybe I'll just go back. Or make an `assert_ok(rv)` helper that does all the things...

arlo_server/contests.py
jonahkagan: Yeahhhh I'm so into the comprehensions!

arlo_server/contests.py
jonahkagan: @benadida I just didn't know you could do that. I assumed you had to count a column in particular.

PR #395, opened 2020-03-26
test(pylint): enable `unused-argument` rule

eventualbuddha:
**Description**

This fixes the violations in a variety of ways:
- underscore-prefix arguments we want to keep and can change the name
- disable specific instances where we can't fix yet
- remove unused arguments when possible

**Testing**

n/a

**Progress**

n/a

### DIFF
--- a/arlo_server/routes.py
+++ b/arlo_server/routes.py
@@ -199,7 +199,7 @@ if ADMIN_PASSWORD:
     auth = HTTPBasicAuth()
 
     @auth.verify_password
-    def verify_password(username, password):
+    def verify_password(_username, password):
         # use a comparison method that prevents timing attacks:
         # https://securitypitfalls.wordpress.com/2018/08/03/constant-time-compare-in-python/
         return password is not None and hmac.compare_digest(password, ADMIN_PASSWORD)
@@ -713,6 +713,7 @@ def ballot_list(election_id, jurisdiction_id, round_id):
         .add_entity(Batch)
         .add_entity(AuditBoard)
         .filter(Round.id == round_id)
+        .filter(Round.election_id == election_id)
         .filter(Batch.jurisdiction_id == jurisdiction_id)
         .order_by(
             AuditBoard.name,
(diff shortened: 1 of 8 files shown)


audit_math/sampler.py
```diff
@@ -62,7 +62,6 @@ def draw_sample(
 def draw_ppeb_sample(
     seed: str,
     contest: Contest,
-    manifest: Dict[str, int],
```
eventualbuddha: @umbernhard this argument was unused, and there wasn't any obvious plan to use it, so I removed it. Seem okay?

arlo_server/routes.py
```diff
@@ -1302,7 +1304,7 @@ def jurisdictionadmin_login_callback():
 @app.route("/")
 @app.route("/election/<election_id>")
 @app.route("/election/<election_id>/board/<board_id>")
-def serve(election_id=None, board_id=None):
+def serve(election_id=None, board_id=None):  # pylint: disable=unused-argument
```
jonahkagan: any reason not to just remove these args?

audit_math/sampler.py
umbernhard: No, that may be a bug, lemme look.

umbernhard: oh, I see, I'm using the batch results instead of the manifest. Gonna drop a convo in slack about this.

arlo_server/routes.py
eventualbuddha: I'm not positive, but I think that it'll crash if we do that:

```py
>>> def fn():
...   pass
... 
>>> fn(election_id='abc')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: fn() got an unexpected keyword argument 'election_id'
```

arlo_server/routes.py
```diff
@@ -713,6 +713,7 @@ def ballot_list(election_id, jurisdiction_id, round_id):
         .add_entity(Batch)
         .add_entity(AuditBoard)
         .filter(Round.id == round_id)
+        .filter(Round.election_id == election_id)
```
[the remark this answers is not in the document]
eventualbuddha: It doesn't support the audit board user type.

PR #396, opened 2020-03-26
Makefile improvements

jonahkagan:
**Description**

Makefile improvements:
- Pull out the command to run pipenv into a variable
- Add `FILE` and `FLAGS` variables to make test-server
- Use those variables to implement the modularized server test commands
- Name server targets consistently (always use "server", not "python")
- Add a command `make python-shell` to run a python shell

Also adds Python lint and typecheck to the pre-commit hooks

**Testing**

Ran a smattering of the commands

**Progress**

Ready for review

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+PIPENV=python3.7 -m pipenv
 
 deps:
 	sudo apt install python3.7 python3-pip nodejs libpython3.7-dev libpq-dev
@@ -11,13 +12,13 @@ initdevdb:
 	sudo -u postgres psql -c "create database arlo with owner arlo;"
 
 install:
-	python3.7 -m pipenv install
+	${PIPENV} install
 	yarn install
 	yarn --cwd arlo-client install
 	yarn --cwd arlo-client build
 
 install-development:
-	python3.7 -m pipenv install --dev
+	${PIPENV} install --dev
 	yarn install
 	yarn --cwd arlo-client install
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -11,51 +12,52 @@ initdevdb:
...
-
+	FILE=tests/util_tests make test-server
 
-# Only tests routes
 test-routes:
-	FLASK_ENV=test python3.7 -m pipenv run python -m pytest tests/routes_tests -k '${TEST}' 
+	FILE=tests/routes_tests make test-server
+
+python-shell:
+	${PIPENV} run python
```
eventualbuddha: It would be great that, when you ran this, you could continue to run `make` commands inside that shell. But alas, you can't.

package.json
```diff
@@ -15,7 +15,9 @@
       "prettier --write"
     ],
     "*.py": [
-      "python3.7 -m pipenv run black"
+      "python3.7 -m pipenv run black",
+      "python3.7 -m pipenv run pylint",
+      "make typecheck-server"
```
eventualbuddha: Seems like this will be called like e.g. `make typecheck-server app.py util/binpacking.py`, which is probably not what we want. Are you just trying to add a `pre-commit` step?

jonahkagan: Yes, I'm just trying to add a pre-commit step. I tried `python3.7 -m pipenv run mypy` but mypy didn't like receiving the files independently, it wanted to run on the whole app at once. I figured `make typecheck-server` would ignore the rest of the args so it would work, even though it's not ideal. Any other ideas?

eventualbuddha: You could just add it to the `husky.hooks.pre-commit` command in this file.

PR #397, opened 2020-03-26
fix: require uniqueness of jurisdiction & batch names

### LINKED ISSUE #389: Require unique jurisdiction names within an audit, and unique batch names within a jurisdiction
These constraints should be implemented with db indexes.

These endpoints will need to be updated with error messages:
- PUT /election/<election_id>/jurisdiction/file
- PUT /election/<election_id>/jurisdiction/<jurisdiction_id>/manifest

The frontend will need to receive and display these error messages.

eventualbuddha:
**Description**

`Jurisdiction` names must now be unique per `Election`, and `Batch` names must now be unique per `Jurisdiction`. Since these are handled via database constraints, we'll get an exception whenever we try to commit a session that would violate those constraints. It turns out that capturing and handling such an error correctly in `process_file` required some changes. We now roll back the session on error, then update the file processing metadata.

Closes #389

**Testing**

Adds a new test for `process_file` to ensure database constraint errors are recorded properly.

**Progress**

This puts the raw error database message into `processing_error`, which is probably not suitable for display to an end user. We should consider catching the error at a higher level than `process_file` and updating it.

### DIFF
--- a/tests/routes_tests/test_jurisdictions.py
+++ b/tests/routes_tests/test_jurisdictions.py
@@ -1,10 +1,12 @@
 import pytest
+from sqlalchemy.exc import SQLAlchemyError
 from flask.testing import FlaskClient
 
 import json, io
 from typing import List
 
-from helpers import compare_json, assert_is_date
+from helpers import compare_json, assert_is_date, asserts_startswith
+from arlo_server import db
 from arlo_server.models import Jurisdiction
 from bgcompute import (
     bgcompute_update_election_jurisdictions_file,
@@ -32,7 +34,11 @@ def jurisdiction_ids(client: FlaskClient, election_id: str) -> List[str]:
     )
     assert json.loads(rv.data) == {"status": "ok"}
     bgcompute_update_election_jurisdictions_file()
-    jurisdictions = Jurisdiction.query.filter_by(election_id=election_id).all()
+    jurisdictions = (
+        Jurisdiction.query.filter_by(election_id=election_id)
+        .order_by(Jurisdiction.name)
+        .all()
+    )
     assert len(jurisdictions) == 3
     yield [j.id for j in jurisdictions]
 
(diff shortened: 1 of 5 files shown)


tests/util_tests/test_process_file.py
```diff
@@ -16,7 +17,7 @@ def db():
 
     yield arlo_server.db
 
-    arlo_server.db.session.commit()
+    arlo_server.db.session.rollback()
```
eventualbuddha: This just mapped better to the behavior we actually want for a test. Just because a test got its db session wedged doesn't necessarily mean we should also throw _again_ on teardown.

tests/util_tests/test_process_file.py
```diff
@@ -29,29 +30,15 @@ def test_success(db):
     db.session.add(file)
     db.session.commit()
 
-    while_processing_started_at: Optional[datetime.datetime] = None
```
eventualbuddha: All this stuff was removed because we no longer update the `file` object before calling `process`.

tests/routes_tests/test_jurisdictions.py
```diff
@@ -32,7 +34,11 @@ def jurisdiction_ids(client: FlaskClient, election_id: str) -> List[str]:
     )
     assert json.loads(rv.data) == {"status": "ok"}
     bgcompute_update_election_jurisdictions_file()
-    jurisdictions = Jurisdiction.query.filter_by(election_id=election_id).all()
+    jurisdictions = (
+        Jurisdiction.query.filter_by(election_id=election_id)
+        .order_by(Jurisdiction.name)
```
eventualbuddha: Adding the constraint seemed to change the order of records returned by this query, so I added an `order by` to make it deterministic.

jonahkagan: Seems like an improvement

eventualbuddha:
> Maybe `process_file` takes a second optional callback to parse errors and return a better error message?

Something like that might be good. For now I'm going to merge this and you & @MorganLove can figure out a follow-up.

PR #398, opened 2020-03-27
Add currentRoundStatus to list jurisdictions endpoint

jonahkagan:
**Description**

Task: #338 
This adds a field `currentRoundStatus` to each jurisdiction in the output of /election/<election_id>/jurisdiction. `currentRoundStatus` is null until the audit starts. Once the audit starts, it is an object with three fields:

- `status`:
    - `NOT_STARTED` = the jurisdiction hasn’t set up any audit boards
    - `IN_PROGRESS` = the audit boards are set up
    - `COMPLETE` = all of the audit boards have signed off on their ballots

- `numBallotsSampled` - how many ballot samples were drawn for this jurisdiction
- `numBallotsAudited` - how many of the drawn ballot samples have been audited so far

Also adds a new field `signed_off_at` to the `AuditBoard` data model, signifying when the audit board members have finalized their audit results.

While I was at it, I wrapped the jurisdictions array returned by this endpoint in an object, similar to #388.

**Testing**

Updating existing tests (all testing before audit start). Added a new test case walking through the various statuses.

**Progress**

Ready for review.

### DIFF
--- a/arlo_server/jurisdictions.py
+++ b/arlo_server/jurisdictions.py
@@ -126,14 +126,12 @@ def round_status_by_jurisdiction(
 @app.route("/election/<election_id>/jurisdiction", methods=["GET"])
 @with_election_access(UserType.AUDIT_ADMIN)
 def list_jurisdictions(election: Election):
-    jurisdictions = sorted(election.jurisdictions, key=lambda j: j.name)
-
     current_round = get_current_round(election)
     round_status = round_status_by_jurisdiction(
-        current_round, jurisdictions, election.online
+        current_round, election.jurisdictions, election.online
     )
 
     json_jurisdictions = [
-        serialize_jurisdiction(j, round_status[j.id]) for j in jurisdictions
+        serialize_jurisdiction(j, round_status[j.id]) for j in election.jurisdictions
     ]
     return jsonify({"jurisdictions": json_jurisdictions})
--- a/arlo_server/jurisdictions.py
+++ b/arlo_server/jurisdictions.py
@@ -1,35 +1,139 @@
 from flask import jsonify
+from typing import List, Dict, Optional
+from enum import Enum
+from sqlalchemy import func
 
 from arlo_server import app
-from arlo_server.models import Election, Jurisdiction
+from arlo_server.models import (
+    Election,
+    Jurisdiction,
+    SampledBallotDraw,
+    SampledBallot,
+    Batch,
+    Round,
+    AuditBoard,
+)
 from arlo_server.auth import with_election_access, UserType
+from arlo_server.rounds import get_current_round
 from util.process_file import serialize_file, serialize_file_processing
+from util.jsonschema import JSONDict
 
 
-def serialize_jurisdiction(db_jurisdiction: Jurisdiction) -> dict:
+def serialize_jurisdiction(
+    jurisdiction: Jurisdiction, round_status: Optional[JSONDict]
+) -> JSONDict:
     return {
-        "id": db_jurisdiction.id,
-        "name": db_jurisdiction.name,
+        "id": jurisdiction.id,
+        "name": jurisdiction.name,
         "ballotManifest": {
-            "file": serialize_file(db_jurisdiction.manifest_file)
-            if db_jurisdiction.manifest_file
+            "file": serialize_file(jurisdiction.manifest_file)
+            if jurisdiction.manifest_file
             else None,
-            "processing": serialize_file_processing(db_jurisdiction.manifest_file)
-            if db_jurisdiction.manifest_file
+            "processing": serialize_file_processing(jurisdiction.manifest_file)
+            if jurisdiction.manifest_file
             else None,
-            "numBallots": db_jurisdiction.manifest_num_ballots,
-            "numBatches": db_jurisdiction.manifest_num_batches,
+            "numBallots": jurisdiction.manifest_num_ballots,
+            "numBatches": jurisdiction.manifest_num_batches,
         },
+        "currentRoundStatus": round_status,
+    }
+
+
+class JurisdictionStatus(str, Enum):
+    NOT_STARTED = "NOT_STARTED"
+    IN_PROGRESS = "IN_PROGRESS"
+    COMPLETE = "COMPLETE"
+
+
+def round_status_by_jurisdiction(
+    round: Optional[Round], jurisdictions: List[Jurisdiction], online: bool
+) -> Dict[str, Optional[JSONDict]]:
+    if not round:
+        return {j.id: None for j in jurisdictions}
+
+    audit_boards_set_up = dict(
+        AuditBoard.query.filter_by(round_id=round.id)
+        .group_by(AuditBoard.jurisdiction_id)
+        .values(AuditBoard.jurisdiction_id, func.count())
+    )
+    audit_boards_signed_off = dict(
+        AuditBoard.query.filter_by(round_id=round.id)
+        .filter(AuditBoard.signed_off_at.isnot(None))
+        .group_by(AuditBoard.jurisdiction_id)
+        .values(AuditBoard.jurisdiction_id, func.count())
+    )
+
+    sampled_ballot_count_by_jurisdiction = dict(
+        SampledBallotDraw.query.filter_by(round_id=round.id)
+        .join(Batch)
+        .group_by(Batch.jurisdiction_id)
+        .values(Batch.jurisdiction_id, func.count())
+    )
+    audited_ballot_count_by_jurisdiction = dict(
+        SampledBallotDraw.query.filter_by(round_id=round.id)
+        .join(SampledBallot)
+        .filter(SampledBallot.vote.isnot(None))
+        .join(Batch)
+        .group_by(Batch.jurisdiction_id)
+        .values(Batch.jurisdiction_id, func.count())
+    )
+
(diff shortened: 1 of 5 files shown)


tests/routes_tests/test_jurisdictions.py
```diff
@@ -32,57 +39,70 @@ def jurisdiction_ids(client: FlaskClient, election_id: str) -> List[str]:
     )
     assert json.loads(rv.data) == {"status": "ok"}
     bgcompute_update_election_jurisdictions_file()
-    jurisdictions = Jurisdiction.query.filter_by(election_id=election_id).all()
+    jurisdictions = (
+        Jurisdiction.query.filter_by(election_id=election_id)
+        .order_by(Jurisdiction.name)
+        .all()
+    )
```
jonahkagan: This copies some changes from #397

benadida:
@jonahkagan the pattern we've had that I typically find more useful is to have a startedAt and completedAt timestamps. Any reason for changjng the pattern here?

Same for is_signed_off, why not have signed_off_at?

Having these timestamps often comes in incredibly handy.

arlo_server/jurisdictions.py
```diff
@@ -1,35 +1,115 @@
...
+            "numBatches": jurisdiction.manifest_num_batches,
         },
+        "currentRoundStatus": round_status,
+    }
+
+
+class JurisdictionStatus(str, Enum):
+    NotStarted = "NOT_STARTED"
+    InProgress = "IN_PROGRESS"
+    Complete = "COMPLETE"
```
eventualbuddha: Should the keys be screaming snake case too? We do that for `UserType`, for example.

jonahkagan:
@benadida great point. No reason for changing the pattern - I didn't even think of that. I was just building to the product spec.

It's not quite as simple in this case, since `startedAt` and `completedAt` will have to be computed from other timestamps:
- `startedAt` would be the max `created_at` timestamp for the audit boards for the jurisdiction
- `completedAt` would be the max `signed_off_at` timestamp for the audit boards for the jurisdiction

Does that sound right?

benadida:
> @benadida great point. No reason for changing the pattern - I didn't even think of that. I was just building to the product spec.
> 
> It's not quite as simple in this case, since `startedAt` and `completedAt` will have to be computed from other timestamps:
> 
> * `startedAt` would be the max `created_at` timestamp for the audit boards for the jurisdiction
> * `completedAt` would be the max `signed_off_at` timestamp for the audit boards for the jurisdiction
> 
> Does that sound right?

I don't have good context on whether you'd want to have a separate field that explicitly starts and completes, so you should make that call. Assuming you want it derived, your description above makes sense to me.

jonahkagan:
Yeah, there's no db model that represents "the jurisdiction for the round," which is where that field would go. Could add one, but not sure that it's needed.

Thinking about this more, I don't quite see a point in deriving the timestamps yet since the product doesn't require them. But now we know that we can if we need to in the future. I do think changing to `signed_off_at` makes sense to me, so that we are set up if we need it in the future, so I'll add that change.

jonahkagan:
Also wondering if there's an easy way to add `created_at` and `updated_at` to every model...

eventualbuddha:
> Also wondering if there's an easy way to add `created_at` and `updated_at` to every model...

I know the [Rails version of this](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/TableDefinition.html#method-i-timestamps). 😉

arlo_server/jurisdictions.py
jonahkagan: Oh, I based this off `USState`, so I thought title case was the pattern. I've never heard the name screaming snake case before, but I love it so much that I will definitely switch to that

jonahkagan:
I have a few product notes to incorporate here, so hold off on reviews please.

jonahkagan:
Ok, this is ready for review again! @benadida

arlo_server/jurisdictions.py
```diff
@@ -1,35 +1,139 @@
...
+            "numBatches": jurisdiction.manifest_num_batches,
         },
+        "currentRoundStatus": round_status,
+    }
+
+
+class JurisdictionStatus(str, Enum):
+    NOT_STARTED = "NOT_STARTED"
+    IN_PROGRESS = "IN_PROGRESS"
+    COMPLETE = "COMPLETE"
```
benadida: ok, so I think I may not have been sufficiently direct with my feedback that I feel this would be better as `startedAt` and `completedAt` rather than this enum of states.

At the same time, I can live with this if you find it works as is, so all good :)

arlo_server/jurisdictions.py
```diff
@@ -1,35 +1,139 @@
...
 
 
 @app.route("/election/<election_id>/jurisdiction", methods=["GET"])
 @with_election_access(UserType.AUDIT_ADMIN)
 def list_jurisdictions(election: Election):
-    jurisdictions = (
-        Jurisdiction.query.filter_by(election_id=election.id)
-        .order_by(Jurisdiction.name)
-        .all()
+    jurisdictions = sorted(election.jurisdictions, key=lambda j: j.name)
```
benadida: could do this straight in the database query that has to run anyways:

https://docs.sqlalchemy.org/en/13/orm/extensions/declarative/relationships.html

arlo_server/jurisdictions.py
```diff
@@ -1,35 +1,139 @@
...
+
+    current_round = get_current_round(election)
+    round_status = round_status_by_jurisdiction(
+        current_round, jurisdictions, election.online
     )
 
-    return jsonify([serialize_jurisdiction(j) for j in jurisdictions])
+    json_jurisdictions = [
+        serialize_jurisdiction(j, round_status[j.id]) for j in jurisdictions
```
benadida: is there ever a case where we're not generating the `round_status`, and should that generation be within the `serialize_jurisdiction` so as not to have a potential mismatch between the jurisdiction and corresponding status?

arlo_server/jurisdictions.py
```diff
@@ -1,35 +1,139 @@
...
+        AuditBoard.query.filter_by(round_id=round.id)
+        .group_by(AuditBoard.jurisdiction_id)
+        .values(AuditBoard.jurisdiction_id, func.count())
+    )
+    audit_boards_signed_off = dict(
+        AuditBoard.query.filter_by(round_id=round.id)
+        .filter(AuditBoard.signed_off_at.isnot(None))
+        .group_by(AuditBoard.jurisdiction_id)
+        .values(AuditBoard.jurisdiction_id, func.count())
+    )
```
benadida: fyi, you could do this in one query with two `count` functions, one that is `count(*)` and one that is `count(signed_off_at)`, see: https://stackoverflow.com/questions/37328779/sqlalchemy-func-count-on-boolean-column

arlo_server/jurisdictions.py
```diff
@@ -1,35 +1,139 @@
...
+        .values(Batch.jurisdiction_id, func.count())
+    )
+    audited_ballot_count_by_jurisdiction = dict(
+        SampledBallotDraw.query.filter_by(round_id=round.id)
+        .join(SampledBallot)
+        .filter(SampledBallot.vote.isnot(None))
+        .join(Batch)
+        .group_by(Batch.jurisdiction_id)
+        .values(Batch.jurisdiction_id, func.count())
+    )
```
benadida: same, could do this as one query with two `count` functions.

arlo_server/jurisdictions.py
```diff
@@ -1,35 +1,139 @@
...
+
+    def num_ballots_sampled(jurisdiction_id: str) -> int:
+        return sampled_ballot_count_by_jurisdiction.get(jurisdiction_id, 0)
```
benadida: fancy nested functions! ;)

arlo_server/jurisdictions.py
jonahkagan: You were clear, I responded here: https://github.com/votingworks/arlo/pull/398#issuecomment-606176112

arlo_server/jurisdictions.py
jonahkagan: Yeah, I was playing around with doing it in one query. Two reason I ended up with this:
- I like the way the code reads when you return tuples and turn them right into a dict.
- I was having a hard time making it actually give me the right answer in a way that I felt confident about. I think I was using something like `.values(AuditBoard.jurisdiction_id, func.count(), func.count().filter(AuditBoard.signed_off_at.isnot(None))` and it seemed like the filter was applying to the whole query and not just the count. I'm sure there was a way to do it, but I decided to bail since it seemed like even if I figured it out, it would be brittle/easy to break if we ever tried to change it.

The only advantage was saving one db query which didn't seem worthwhile at this time.

arlo_server/jurisdictions.py
jonahkagan: `round_status_by_jurisdiction()` generates the round statuses for all the jurisdictions in bulk, while `serialize_jurisdiction` only looks at one jurisdiction at a time. The only way I can think of to make the round status generation within `serialize_jurisdiction` is to memoize `round_status_by_jurisdiction()`, but that seems more confusing/error-prone in this case. Do you have other ideas?

arlo_server/jurisdictions.py
jonahkagan: Love it! fa554f0

PR #400, opened 2020-03-27
chore: remove `console.log`

eventualbuddha:
**Description**

This seems intentional, but I wasn't sure why it was here so I'm asking with a PR.

**Testing**

n/a

**Progress**

n/a


MorganLove:
@eventualbuddha Yes, intentional. This is in-progress code, waiting to be connected to the api. It can be removed, but there will be a PR soon to replace it with the api call anyway.

eventualbuddha:
> It can be removed, but there will be a PR soon to replace it with the api call anyway.

👍

PR #403, opened 2020-03-30
fix: ensure `make resetdb` works

eventualbuddha:
**Description**

We now have to explicitly set `FLASK_ENV` to something other than `production` if we don't want it to be `production`. This makes `make resetdb` default to `development`, but allows overriding it since `make resettestdb` does exactly that.

**Testing**

n/a

**Progress**

This was broken by #399.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ resettestdb:
 	FLASK_ENV=test make resetdb
 
 resetdb:
-	${PIPENV} run python resetdb.py
+	FLASK_ENV=$${FLASK_ENV:-development} ${PIPENV} run python resetdb.py
 
 dev-environment: deps initdevdb install-development resetdb
 


Makefile
```diff
@@ -25,7 +25,7 @@ resettestdb:
 	FLASK_ENV=test make resetdb
 
 resetdb:
-	python3.7 -m pipenv run python resetdb.py
+	FLASK_ENV=$${FLASK_ENV:-development} python3.7 -m pipenv run python resetdb.py
```
jonahkagan: cool, didn't know about this syntax

eventualbuddha: It's a mashup of `make` and `bash`. `${FOO:-bar}` is a bash-ism which expands to the value of `FOO` if it is set, but defaults to `bar` if not. Prefixing that with `$` ensures `make` doesn't try to treat it as a `make` expansion.

PR #405, opened 2020-03-30
Remove Round.started_at and use Round.created_at

jonahkagan:
**Description**

After #404, the fields should have the same value, so no reason to have both around and confuse people. No change to any APIs.

**Testing**

Existing tests pass.

**Progress**

Ready for review.


jonahkagan:
Yeah, I figured you would say that 😜 

We don't need it now, and don't have plans to need it. I'm realizing that my bias is these situations is to not keep things around "in case," since product development is so unpredictable.

jonahkagan:
Accidentally merged this PR into the branch it originally based on instead of master. So I cherry-picked the commit into master instead. Just logging here for posterity

PR #406, opened 2020-03-31
[Sketch] Fix integration tests

jonahkagan:
**Description**

The frontend integration tests have been failing intermittently. This PR tests a hypothesis about why: after #344, which made ballot manifest uploads process in the background, we never changed the single-jurisdiction flow to wait for these uploads to finish processing. In other words, there was a breaking API change and we didn't update the consumer.

Evidence in support of this hypothesis:
- The failures started about the time when #344 was merged (you can see this in the commit log)
- The failures are unpredictable, which suggests a race condition
- This fix seems to work after a couple runs on CI

**Testing**

Existing tests.

**Progress**

This is a sketch. Work to be done:
- Actually implement this change fully according to the needs/standards of the frontend codebase
- Upgrade the frontend unit tests (I just skipped the ones I broke)

@MorganLove I think it will likely be faster for you to finish this, if we decide it's the way we wanna go.


jonahkagan:
cc @eventualbuddha @benadida

PR #408, opened 2020-03-31
Add @with_jurisdiction_access route decorator

jonahkagan:
**Description**

Adds a route decorator `@with_jurisdiction_access`. This decorator requires jurisdiction admin (JA) authorization, checks that it matches the given election/jurisdiction ids, and loads the election and jurisdiction objects.

Also modifies `@with_election_access` to not take a user type and only accept audit admin (AA) authorization.

To simplify permissions for the time being, AAs and JAs will not share any routes. JA routes will always have a `jurisdiction_id` in the path so we can easily check that they are trying to access the jurisdiction for which they are authorized.

**Testing**

Lots of new tests here, trying to cover every case of incorrect authorization - could use a good review to make sure I'm not missing something.

Here's a summary:
 - test_with_election_access_audit_admin
 - test_with_election_access_wrong_org
 - test_with_election_access_not_found
 - test_with_election_access_jurisdiction_admin
 - test_with_election_access_anonymous_user
 - test_with_jurisdiction_access_jurisdiction_admin
 - test_with_jurisdiction_access_wrong_org
 - test_with_jurisdiction_access_election_not_found
 - test_with_jurisdiction_access_jurisdiction_not_found
 - test_with_jurisdiction_access_audit_admin
 - test_with_jurisdiction_access_anonymous_user

**Progress**

Ready for review.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -10,7 +10,6 @@ from arlo_server.models import (
     db,
     AuditAdministration,
     User,
-    Election,
     Jurisdiction,
     JurisdictionAdministration,
 )
@@ -80,19 +79,21 @@ def create_jurisdiction_and_admin(
 
 
 def create_election(
+    client: FlaskClient,
     audit_name: str = "Test Audit",
     organization_id: str = None,
     is_multi_jurisdiction: bool = True,
 ):
-    election = Election(
-        id=str(uuid.uuid4()),
-        audit_name=audit_name,
-        organization_id=organization_id,
-        is_multi_jurisdiction=is_multi_jurisdiction,
+    rv = post_json(
+        client,
+        "/election/new",
+        {
+            "auditName": audit_name,
+            "organizationId": organization_id,
+            "isMultiJurisdiction": is_multi_jurisdiction,
+        },
     )
-    db.session.add(election)
-    db.session.commit()
-    return election.id
+    return json.loads(rv.data)["electionId"]
 
 
 def assert_is_id(x):
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -6,9 +6,17 @@ from flask.testing import FlaskClient
 
 from arlo_server.auth import UserType
 from arlo_server.routes import create_organization
-from arlo_server.models import db, AuditAdministration, User
+from arlo_server.models import (
+    db,
+    AuditAdministration,
+    User,
+    Election,
+    Jurisdiction,
+    JurisdictionAdministration,
+)
 
-DEFAULT_USER_EMAIL = "admin@example.com"
+DEFAULT_AA_EMAIL = "admin@example.com"
+DEFAULT_JA_EMAIL = "jurisdiction.admin@example.com"
 
 
 def post_json(client: FlaskClient, url: str, obj) -> Any:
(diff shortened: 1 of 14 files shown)


tests/helpers.py
```diff
@@ -24,46 +32,67 @@ def put_json(client: FlaskClient, url: str, obj) -> Any:
...
-        },
+    election = Election(
+        id=str(uuid.uuid4()),
+        audit_name=audit_name,
+        organization_id=organization_id,
+        is_multi_jurisdiction=is_multi_jurisdiction,
     )
-    return json.loads(rv.data)["electionId"]
+    db.session.add(election)
+    db.session.commit()
```
benadida, line 94: why not keep the more functional test of calling `/election/new`? I could see unit tests drifting a bit in coverage, and then we miss an edge case.

tests/routes_tests/test_auth.py
```diff
@@ -103,51 +126,224 @@ def test_jurisdictionadmin_start(client: FlaskClient):
...
-        id=str(uuid.uuid4()), email="bar@example.com", external_id="bar@example.com"
-    )
-    j_admin = JurisdictionAdministration(user_id=u.id, jurisdiction_id=jurisdiction.id)
-
-    db.session.add(election)
-    db.session.add(jurisdiction)
-    db.session.add(u)
-    db.session.add(j_admin)
-    db.session.commit()
-
```
benadida, line 127: did the behavior change here where you're no longer expecting a user in the table coming from back Auth0? I think that's worth a longer discussion if so.

tests/routes_tests/test_auth.py
```diff
@@ -103,51 +126,224 @@ def test_jurisdictionadmin_start(client: FlaskClient):
...
+
+
+def test_with_election_access_audit_admin(client: FlaskClient, election_id: str):
+    set_logged_in_user(client, UserType.AUDIT_ADMIN, AA_EMAIL)
+    rv = client.get(f"/election/{election_id}/test_auth")
+    assert rv.status_code == 200
+    assert json.loads(rv.data) == election_id
+
+
+def test_with_election_access_wrong_org(
```
benadida, line 171: 😍

tests/helpers.py
jonahkagan: I made this change for two reasons:
- With the way it was before, I was gonna have to first create an admin and then set them as the logged in user in the session in order to access the endpoint. (For some specific test I was working on.)
- All of the other helpers do it this way, so it seemed consistent.

But I do hear your point that in general it's good to test everything only using the external interface, so I could be convinced to change it back.

tests/routes_tests/test_auth.py
jonahkagan: I just removed this because the same code is now executed by the `jurisdiction_id` fixture (and the chain of fixtures it depends on).

Hopefully there's no behavior change. But maybe I missed something because I haven't actually worked with the auth0 integration yet.

tests/helpers.py
benadida: I would recommend keeping this as an API call, but it's not a hill I need to die on, so happy if you make the call you think is best.

tests/routes_tests/test_auth.py
benadida: got it, so no change in behavior, just a different place that creates the local admin jurisdiction user? Ok great.

tests/helpers.py
jonahkagan: Cool, changed back to API call

PR #410, opened 2020-04-03
refactoring Audit/_mocks

MorganLove:
**Description**

- I initially had the sequence of states of return values from the `/status` endpoint as a reference, and then repurposed it as mock data. The problem was that while initially it was fine to have it as an array, it quickly became bloated, unwieldy, and incredibly difficult to maintain and update. So I went back through and converted it to an object with named keys that (hopefully) make more sense.

**Testing**

- Passes existing tests.

**Progress**

- Should be good, though feedback on if the names I used are semantic enough would be nice.


eventualbuddha:
Yes please.

PR #411, opened 2020-04-03
make sure no double slash in redirect URIs for auth0

benadida:


MorganLove:
@benadida This is working for me! Looks like it's failing on formatting though?

arlo_server/routes.py
```diff
@@ -1257,7 +1257,7 @@ def logout():
 @app.route("/auth/auditadmin/start")
 def auditadmin_login():
     return auth0_aa.authorize_redirect(
-        redirect_uri=f"{request.host_url}{AUDITADMIN_OAUTH_CALLBACK_URL}"
+        redirect_uri=f"{request.host_url.rstrip('/')}{AUDITADMIN_OAUTH_CALLBACK_URL}"
```
jonahkagan, line 1260: would `urllib.parse.urljoin` work here?

tests/routes_tests/test_auth.py
```diff
@@ -100,6 +101,16 @@ def test_auth_me_jurisdiction_admin(
...
+    location = urlparse(rv.location)
+    query_vars = parse_qs(location.query)
+    assert query_vars["redirect_uri"]
+    redirect_uri = query_vars["redirect_uri"][0]
+
+    # common problem is a trailing slash on origin
+    # which makes a double slash like 'http://localhost//authorize'
+    # which won't work. So testing to make sure there is no '//'
+    # other than '://'
+    assert re.search("[^:]\/\/", redirect_uri) is None
```
jonahkagan, line 113: do we know what we expect the redirect_uri to be? could we just assert its that? or even just assert the whole query string?

benadida: can't assert the whole query string, as there is random state in there.

asserting the redirect_uri might work but might also break if some test config changes later to not be based on `localhost`, making the test flakier than we need it to be.

I *think* this is the right balance?

arlo_server/routes.py
benadida: yes, that's a good call.

tests/routes_tests/test_auth.py
jonahkagan: Do we have any test that the redirect_uri is the correct value (as opposed to just being well-formed)? That seems like something worth testing

benadida: yeah, I'm adding that now.

PR #416, opened 2020-04-06
Add assert_ok test helper and typecheck tests

jonahkagan:
**Description**

Adds an `assert_ok` test helper to check for successful API responses.
In order to get the assertion error messages to display well, we have to
do two things:
- Add `__tracebackhide__ = True` to the helper function, which treats it
as an inline assertion in the test error reporting
- Call `pytest.register_assert_rewrite("tests.helpers")`, which passes
through the assert error messages to the test error reporting

The latter change required making creating a `tests/__init__.py` file,
which had the bonus side-effect of causing `mypy` to actually run on the
tests for the first time (who knew?) So we also fix some type errors.

Last, we go and use `assert_ok` in all the appropriate places in existing tests (fairly straightforward find and replace).

**Testing**

Modifies existing tests.

**Progress**

Ready to review.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -1,8 +1,9 @@
 import uuid, json, re
 import pytest
 import datetime
-from typing import Any, List, Union
+from typing import Any, List, Union, Tuple
 from flask.testing import FlaskClient
+from werkzeug.wrappers import Response
 
 from arlo_server.auth import UserType
 from arlo_server.routes import create_organization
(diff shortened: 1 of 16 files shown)


tests/helpers.py
```diff
@@ -30,19 +32,27 @@ def put_json(client: FlaskClient, url: str, obj) -> Any:
     )
 
 
+def assert_ok(rv: Response):
+    __tracebackhide__ = True
```
benadida, line 36: I learned something new!

PR #417, opened 2020-04-07
Add endpoint for jurisdiction admins to create audit boards

jonahkagan:
**Description**

Adds an endpoint
`POST
/election/<election_id>/jurisdiction/<jurisdiction_id>/round/<round_id>/audit-board`
that jurisdiction admins can use to create audit boards.

When the audit boards are created, the sampled ballots for that round
are divvyed up as fairly as possible between the audit boards.

Also adds a db constraint that audit board names must be unique within a
jurisdiction for each round.

**Testing**

Added a variety of new test cases for the endpoint.

**Progress**

Ready for review.

### DIFF
--- a/arlo_server/audit_boards.py
+++ b/arlo_server/audit_boards.py
@@ -1,0 +1,121 @@
+from flask import jsonify, request
+import uuid
+from typing import List
+from xkcdpass import xkcd_password as xp
+from werkzeug.exceptions import Conflict
+from sqlalchemy.exc import IntegrityError
+
+from arlo_server import app, db
+from arlo_server.auth import with_jurisdiction_access
+from arlo_server.rounds import get_current_round
+from arlo_server.models import (
+    AuditBoard,
+    Round,
+    Election,
+    Jurisdiction,
+    SampledBallot,
+    Batch,
+)
+from arlo_server.errors import handle_unique_constraint_error
+from util.jsonschema import validate, JSONDict
+from util.binpacking import BalancedBucketList, Bucket
+from util.group_by import group_by
+
+WORDS = xp.generate_wordlist(wordfile=xp.locate_wordfile())
+
+CREATE_AUDIT_BOARD_REQUEST_SCHEMA = {
+    "type": "object",
+    "properties": {"name": {"type": "string"},},
+    "additionalProperties": False,
+    "required": ["name"],
+}
+
+# Raises if invalid
+def validate_audit_boards(
+    audit_boards: List[JSONDict],
+    election: Election,
+    jurisdiction: Jurisdiction,
+    round: Round,
+):
+    current_round = get_current_round(election)
+    if not current_round or round.id != current_round.id:
+        raise Conflict(f"Round {round.round_num} is not the current round")
+
+    if any(ab for ab in jurisdiction.audit_boards if ab.round_id == round.id):
+        raise Conflict(f"Audit boards already created for round {round.round_num}")
+
+    validate(
+        audit_boards, {"type": "array", "items": CREATE_AUDIT_BOARD_REQUEST_SCHEMA}
+    )
+
+
+def assign_sampled_ballots(
+    jurisdiction: Jurisdiction, round: Round, audit_boards: List[AuditBoard],
+):
+    # Collect the physical ballots for each batch that were sampled for this
+    # jurisdiction for this round
+    sampled_ballots = (
+        SampledBallot.query.join(Batch)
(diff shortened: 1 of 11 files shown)


tests/routes_tests/test_audit_status.py
```diff
@@ -77,6 +71,12 @@ def test_audit_status(client, election_id):
                             "name": "Audit Board #1",
                             "passphrase": assert_is_passphrase,
                         },
+                        {
+                            "id": audit_board_id_2,
+                            "members": [],
+                            "name": "audit board #2",
+                            "passphrase": assert_is_passphrase,
+                        },
```
jonahkagan, line 79: Order changed by adding the db unique constraint. I confirmed this by adding the db constraint on master and running the tests.

tests/routes_tests/test_report.py
```diff
@@ -286,16 +286,16 @@ def test_two_round_audit_report(client, election_id):
...
-        '"Batch 10, #151",Round 1: 0.012381762,Audited,Blank vote/no mark,Comment for ballot 3',
-        '"Batch 10, #175",Round 2: 0.021956866,Audited,Audit board can\'t agree,',
+        f'"Batch 10, #151",Round 1: 0.012381762,Audited,Blank vote/no mark,Comment for ballot 3',
+        f'"Batch 10, #175",Round 2: 0.021956866,Audited,{candidate_id_2},Comment for ballot 1017',
         f'"Batch 10, #200",Round 1: 0.000030407,Audited,{candidate_id_2},',
         f'"Batch 10, #59",Round 1: 0.002728647,Audited,{candidate_id_1},',
-        '"Batch 10, #72",Round 1: 0.009650515,Audited,Audit board can\'t agree,',
-        '"Batch 100, #106",Round 2: 0.015314474,Audited,Blank vote/no mark,Comment for ballot 3',
+        f'"Batch 10, #72",Round 1: 0.009650515,Audited,Audit board can\'t agree,',
+        f'"Batch 100, #106",Round 2: 0.015314474,Audited,{candidate_id_2},',
```
jonahkagan, line 298: Same here, changed by adding the db unique constraint. I confirmed this by adding the db constraint on master and running the tests.

arlo_server/audit_boards.py
```diff
@@ -0,0 +1,121 @@
...
+    db.session.commit()
+
+
+@app.route(
+    "/election/<election_id>/jurisdiction/<jurisdiction_id>/round/<round_id>/audit-board",
+    methods=["POST"],
+)
+@with_jurisdiction_access
+def create_audit_boards(election: Election, jurisdiction: Jurisdiction, round_id: str):
+    json_audit_boards = request.get_json()
```
benadida, line 94: so taking a raw array as input? I would suggest sticking to objects with a named property, kinda like named arguments in a function, especially for some measure of forward compatibility with potential optional arguments we add later.

tests/routes_tests/test_audit_status.py
benadida: should there be a defined order here that doesn't depend on the index? Meaning, are we not issuing an `order by`?

tests/routes_tests/test_report.py
benadida: same comment as above: I suggest having an implementation-independent ordering that is explicitly requested by the SQL query, so changes in indexing and other internal optimizations don't break these tests.

arlo_server/audit_boards.py
jonahkagan: I'm gonna leave it like this since this is the pattern established so far, and I'm not actually convinced that there's a need for forward compatibility in the way you describe right now.

tests/routes_tests/test_audit_status.py
jonahkagan: You wrote these endpoints hehe. I can go try to add an ordering if you'd like

tests/routes_tests/test_report.py
jonahkagan: Added here: 540f5f2dcb003ed68a068173fd16325f483a31ed

tests/routes_tests/test_audit_status.py
jonahkagan: Added: 540f5f2dcb003ed68a068173fd16325f483a31ed

tests/routes_tests/test_report.py
jonahkagan: Turns out this same call to `Jurisdiction.audit_boards` was at the root of both of these cases

PR #418, opened 2020-04-07
Add __init__ files to test packages

jonahkagan:
**Description**

This ensures that mypy will typecheck them. mypy doesn't recursively
check directories without an `__init__` file. See: https://github.com/python/mypy/issues/6385.

**Testing**

N/A

**Progress**

Ready for review.


jonahkagan:
cc @umbernhard, for future reference

PR #419, opened 2020-04-07
Add list rounds endpoint for jurisdiction admins

jonahkagan:
**Description**

Adds `GET /election/<election_id>/jurisdiction/<jurisdiction_id>/round` to list audit rounds when logged in as a jurisdiction admin.

This is pretty much the only piece of functionality that is identical for audit admins and jurisdiction admins, but instead of reusing the same endpoint, we make a new one. This has two benefits:
- All jurisdiction admin endpoints follow the same url scheme, which includes the jurisdiction id. So we can always check that the user is requesting info about district they are authorized for.
- Permissions are simpler to think about and implement. Every endpoint can only be accessed by one user type.

**Testing**

Updated the test cases to also try the same requests as a JA and expect the same results. It would be a bit nicer to have separate test cases, but since you need to do a lot of work to create the round before you can list it, the existing cases integrated creating/listing into the same test case. Factoring out the logic didn't seem worthwhile, since we don't plan to have other cases that need to test the same thing for different user types.

**Progress**

Ready for review.

### DIFF
--- a/arlo_server/rounds.py
+++ b/arlo_server/rounds.py
@@ -11,8 +11,9 @@ from arlo_server.models import (
     Election,
     SampledBallot,
     SampledBallotDraw,
+    Jurisdiction,
 )
-from arlo_server.auth import with_election_access
+from arlo_server.auth import with_election_access, with_jurisdiction_access
 from arlo_server.sample_sizes import sample_size_options
 from util.isoformat import isoformat
 from audit_math import sampler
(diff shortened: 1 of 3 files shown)


arlo_server/rounds.py
```diff
@@ -121,7 +122,22 @@ def sample_ballots(election: Election, round: Round, sample_size: int):
 
 @app.route("/election/<election_id>/round", methods=["GET"])
 @with_election_access
-def list_rounds(election: Election):
+def list_rounds_audit_admin(election: Election):
+    rounds = sorted(election.rounds, key=lambda r: r.round_num)
```
benadida, line 126: something we can do by default in the db relationship.

arlo_server/rounds.py
```diff
@@ -121,7 +122,22 @@ def sample_ballots(election: Election, round: Round, sample_size: int):
...
+# user type), even though the logic of this particular pair our routes is
+# identical.
+@app.route(
+    "/election/<election_id>/jurisdiction/<jurisdiction_id>/round", methods=["GET"]
+)
+@with_jurisdiction_access
+def list_rounds_jurisdiction_admin(
+    election: Election, jurisdiction: Jurisdiction  # pylint: disable=unused-argument
+):
     rounds = sorted(election.rounds, key=lambda r: r.round_num)
```
benadida, line 141: as per previous comment, then you wouldn't have to do this again over here :)

arlo_server/rounds.py
jonahkagan: yes, thanks!

PR #420, opened 2020-04-08
Add list audit boards endpoint for jurisdiction admins

jonahkagan:
**Description**

Adds `GET
/election/<election_id>/jurisdiction/<jurisdiction_id>/round/<round_id>/audit-board`, which a jurisdiction admin can use to get the list of audit boards they created for the round, as well as the status (ballots sampled, ballots audited) for each audit board.

**Testing**

Adds new test cases for no audit boards, one audit board, two audit boards, and audit boards in round 2 of the audit. In each test case, tries out various stages (before the audit boards start, once they've audited some ballots, once they've finished auditing ballots and signed off).

**Progress**

Ready to review.

### DIFF
--- a/arlo_server/audit_boards.py
+++ b/arlo_server/audit_boards.py
@@ -1,9 +1,10 @@
 from flask import jsonify, request
 import uuid
-from typing import List
+from typing import List, Dict
 from xkcdpass import xkcd_password as xp
 from werkzeug.exceptions import Conflict
 from sqlalchemy.exc import IntegrityError
+from sqlalchemy import func
 
 from arlo_server import app, db
 from arlo_server.auth import with_jurisdiction_access
@@ -14,12 +15,14 @@ from arlo_server.models import (
     Election,
     Jurisdiction,
     SampledBallot,
+    SampledBallotDraw,
     Batch,
 )
 from arlo_server.errors import handle_unique_constraint_error
 from util.jsonschema import validate, JSONDict
 from util.binpacking import BalancedBucketList, Bucket
 from util.group_by import group_by
+from util.isoformat import isoformat
 
 WORDS = xp.generate_wordlist(wordfile=xp.locate_wordfile())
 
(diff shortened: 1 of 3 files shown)


jonahkagan:
cc @MorganLove this endpoint wasn't in my original spec but I think you'll need it

arlo_server/audit_boards.py
```diff
@@ -119,3 +122,66 @@ def create_audit_boards(election: Election, jurisdiction: Jurisdiction, round_id
     assign_sampled_ballots(jurisdiction, round, audit_boards)
 
     return jsonify(status="ok")
+
+
+def round_status_by_audit_board(audit_boards: List[AuditBoard]) -> Dict[str, JSONDict]:
+    if len(audit_boards) == 0:
+        return {}
+
```
benadida, line 130: worth a check that all audit boards have the same `jurisdiction_id` to make sure we don't get into nonsensical results? I guess the `get()` will currently return `None`, so it's not a huge problem, but I think our future selves will enjoy an error message on feeding an improper list of audit boards.

benadida: put another way, I'm wondering if what you want is `round_status_by_jurisdiction`, since no matter how many audit boards you're feeding this function, it's going to do roughly the same amount of work overall.

jonahkagan: Can you share more about what that would be guarding against? This function gets called with the result of a db query that filters by jurisdiction. I originally had the signature `round_status_by_audit_board(jurisdiction_id, round_id, audit_boards)` but took out the first two args to make it more concise.

benadida: It *should* be called with the result of that query, but what if, at some point, it's called with a list of audit boards that come out of a poorly formed query, or I look at that signature and think somehow I can query all audit boards across jurisdictions?

I'm suggesting that a better sig might be `(jurisdiction_id, round_id)`

jonahkagan: How's this look? 3ab03f4

benadida: nice :)

PR #424, opened 2020-04-09
Copy retrieval list endpoint for jurisdiction admins

jonahkagan:
**Description**

We already have an endpoint for generating the ballot retrieval list for
a jurisdiction, but it had two shortcomings:
- It had a URL scheme inconsistent with the new API scheme, omitting the
`round` path segment and taking a round number instead of round id
- It didn't have any authorization

This change creates a new endpoint,
`/election/<election_id>/jurisdiction/<jurisdiction_id>/round/<round_id>/retrieval-list`,
which shares logic with the existing endpoint (I factored it out).

While I was at it, I factored out some common utilities for creating CSV
downloads used in the audit report as well.

**Testing**

Added new test cases for the new version of the endpoint that just compare the exact output (assuming the implementation is already correct).

**Progress**

Ready for review.


arlo_server/ballots.py
```diff
@@ -0,0 +1,118 @@
...
+        .add_entity(Batch)
+        .add_entity(AuditBoard)
+        .group_by(
+            Batch.name,
+            Batch.id,
+            Batch.storage_location,
+            Batch.tabulator,
+            AuditBoard.name,
+        )
+        .group_by(SampledBallotDraw.ballot_position)
```
benadida, line 64: Any reason not to combine the group_by's?

jonahkagan: I just copy-pasted your code without making any changes. So maybe I'll ask you: any reason not to combine the group_by's? 😝

jonahkagan: went ahead and did some cleanup: a5bbe58

PR #425, opened 2020-04-13
Authorization for audit boards

jonahkagan:
**Description**

Adds a user type for audit boards. This user type is a bit different for a couple of reasons:
- It doesn't correspond to one real world person, it corresponds to the entire audit board.
- It's not represented by a User record in the database, but rather by the AuditBoard record
- The audit board logs in by hitting `GET /auditboard/<passphrase>`, instead of using auth0 with email/password.

This PR:
- Adds the new user type
- Implements the login flow for audit boards
- Adds a `with_audit_board_access` decorator for routes that
should be restricted to audit board users
- Updates the session data format and auth helper functions to accommodate the differences outlined above

**Testing**

Adds a bunch of new test cases:
- test_audit_board_log_in
- test_auth_me_audit_board
- test_with_election_access_audit_board_user
- test_with_jurisdiction_access_audit_board_user
- test_with_audit_board_access_audit_board_user
- test_with_audit_board_access_audit_admin
- test_with_audit_board_access_jurisdiction_admin
- test_with_audit_board_access_anonymous_user
- test_with_audit_board_access_wrong_org
- test_with_audit_board_access_wrong_election
- test_with_audit_board_access_wrong_jurisdiction
- test_with_audit_board_access_wrong_round
- test_with_audit_board_access_wrong_audit_board
- test_with_audit_board_access_election_not_found
- test_with_audit_board_access_jurisdiction_not_found
- test_with_audit_board_access_round_not_found
- test_with_audit_board_access_audit_board_not_found

**Progress**

Ready for review

### DIFF
--- a/arlo_server/routes.py
+++ b/arlo_server/routes.py
@@ -15,7 +15,6 @@ from arlo_server.auth import (
     UserType,
     clear_loggedin_user,
     get_loggedin_user,
-    get_loggedin_user_record,
     require_audit_admin_for_organization,
     set_loggedin_user,
     with_election_access,
@@ -1054,8 +1053,9 @@ def audit_reset(election_id):
 @app.route("/auditboard/<passphrase>", methods=["GET"])
 def auditboard_passphrase(passphrase):
     auditboard = AuditBoard.query.filter_by(passphrase=passphrase).one()
+    set_loggedin_user(UserType.AUDIT_BOARD, auditboard.id)
     return redirect(
-        "/election/%s/board/%s" % (auditboard.jurisdiction.election.id, auditboard.id)
+        f"/election/{auditboard.jurisdiction.election.id}/board/{auditboard.id}"
     )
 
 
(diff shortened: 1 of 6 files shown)


arlo_server/routes.py
```diff
@@ -1053,8 +1052,9 @@ def audit_reset(election_id):
 @app.route("/auditboard/<passphrase>", methods=["GET"])
 def auditboard_passphrase(passphrase):
     auditboard = AuditBoard.query.filter_by(passphrase=passphrase).one()
+    set_loggedin_user(UserType.AUDIT_BOARD, auditboard.passphrase)
```
benadida, line 1055: Although the passphrase is unique right now, I would key the user on `auditboard.id` instead of `passphrase`. I don't see any advantage to keying it on passphrase.

In case it's useful, pointing out that the cookie is signed, so there's no way to forge the `auditboard.id`. So there is no authentication advantage to using `passphrase` in the cookie.

jonahkagan: Yep, that makes sense. Really not sure why I chose `passphrase` over `id` in the first place

PR #426, opened 2020-04-13
Add audit board sign-off endpoint

jonahkagan:
**Description**

Adds `POST /election/<election_id>/jurisdiction/<jurisdiction_id>/round/<round_id>/audit-board/<audit_board_id>/sign-off` (whew, that's a mouthful). This endpoint allows an audit board to finalize their audit results. When the last audit board in all the jurisdictions with sampled ballots signs off, the round ends and the risk measurements are calculated for each contest.

**Testing**

Added some new tests to check:
- The endpoint works
- Validation for the POST body (proper format, contains the right names)
- The audit board can't sign off til they finish auditing their ballots
- The round doesn't end until all jurisdictions set up audit boards and all audit boards sign off

**Progress**

Ready for review.

### DIFF
--- a/arlo_server/audit_boards.py
+++ b/arlo_server/audit_boards.py
@@ -1,13 +1,15 @@
 from flask import jsonify, request
 import uuid
+from datetime import datetime
 from typing import List, Dict
 from xkcdpass import xkcd_password as xp
-from werkzeug.exceptions import Conflict
+from werkzeug.exceptions import Conflict, BadRequest
 from sqlalchemy.exc import IntegrityError
-from sqlalchemy import func
+from sqlalchemy import func, and_
 
 from arlo_server import app, db
-from arlo_server.auth import with_jurisdiction_access
+from arlo_server.routes import sample_results
+from arlo_server.auth import with_jurisdiction_access, with_audit_board_access
 from arlo_server.rounds import get_current_round
 from arlo_server.models import (
     AuditBoard,
@@ -23,6 +25,7 @@ from util.jsonschema import validate, JSONDict
 from util.binpacking import BalancedBucketList, Bucket
 from util.group_by import group_by
 from util.isoformat import isoformat
+from audit_math import bravo, sampler_contest
 
 WORDS = xp.generate_wordlist(wordfile=xp.locate_wordfile())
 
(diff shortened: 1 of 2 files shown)


arlo_server/audit_boards.py
```diff
@@ -187,3 +190,101 @@ def list_audit_boards(
...
+    audit_board: AuditBoard,
+):
+    validate_sign_off(request.get_json(), audit_board)
+
+    audit_board.signed_off_at = datetime.utcnow()
+
+    if is_round_complete(election, round):
+        end_round(election, round)
+
+    db.session.commit()
```
jonahkagan, line 288: I wasn't totally sure whether the helper functions (`end_round`, `calculate_risk_measurement`) should each call `db.session.commit()` or whether it should only be called here once at the end. I opted for just calling it once because I figured if we wanted to rollback the changes, they should all be in one transaction. But not sure if that's actually what's going on under the hood or if that's the right approach.

arlo_server/audit_boards.py
```diff
@@ -187,3 +190,101 @@ def list_audit_boards(
...
+            election.risk_limit / 100,
+            sampler_contest.from_db_contest(contest),
+            results[contest.id],
+        )
+
+        round_contest = next(
+            rc for rc in round.round_contests if rc.contest_id == contest.id
+        )
+        round_contest.end_p_value = max(risk.values())
+        round_contest.is_complete = is_complete
```
benadida, line 209: @umbernhard can you double check this?

umbernhard: So the only thing about this is that we don't want to keep auditing contests that have met their risk limits. Is there logic somewhere else that keeps track of that?

jonahkagan: I copied this code from the single-page flow.

`is_complete` here signifies whether the contest met its risk limit this round, so yes, we do keep track of this.

Currently there's only one targeted contest that determines which ballots get sampled, so I don't think there's any logic to only audit contests that have not met their risk limits. We just keep auditing the ballots for the target contest until it meets its risk limit. Do we need to not show opportunistic contests that have met their risk limit to the audit boards when auditing ballots for the targeted contest?

mcchilders: I think you're talking about in the Audit Board workflow @jonahkagan - do we still include those contests on the page the audit boards fill out for each ballot? If I'm understanding your q correctly, then we don't need to exclude opportunistic contests - if we have to audit another ballot for the target contest anyway, we might as well keep gathering data for the opportunistic contests (if they appear on that particular ballot, otherwise they're left blank).

umbernhard: I dunno if that's a good idea. Because the test statistic for ballot polling is so ill-behaved, if we keep collecting data for opportunistic contests there's a chance that it could meet the risk-limit and then fall back above it once we have finished the targeted contest (e.g. if we only sample 10 more ballots for the targeted contest after the opportunistic one hits the risk limit, and the opportunistic votes are all for the loser), so our end result would be that the opportunistic contest didn't meet the risk-limit.

mcchilders: I think that's ok - for v.1 on the opportunistic contests I'm certainly comfortable with that. For future versions, since we're checking at the end of every round it's easy enough to check if a contest has *ever* met the risk limit (at the end of any round) for audits using offline data entry, and for online data entry we'll want to implement the logic for "if it hasn't reached the risk limit at the end of the current round, use the ballot-by-ballot accounting to figure out if it ever crossed that threshold in the interim" ... but none of that is necessary for v.1.

jonahkagan: Ok, so for now I'm going to leave things as they are.

PR #431, opened 2020-04-14
Fix Request.trusted_hosts to take host, not entire http origin url

jonahkagan:
**Description**

`Request.trusted_hosts` expects a hostname (e.g. foo.com), and was previously being given a full url (e.g. https://foo.com). This resulted in the app returning the error: `Host \"foo.com\" is not trusted"` when trying to access `request.host` in the audit admin login endpoint handler.

Also added a fix to the configuration of the `HTTP_ORIGIN` env var for the Heroku Review Apps flow.

**Testing**

Manually tested on this branch's staging deploy, which now works.

**Progress**

Ready for review.

### DIFF
--- a/arlo_server/__init__.py
+++ b/arlo_server/__init__.py
@@ -1,6 +1,7 @@
 from flask import Flask
 from flask_talisman import Talisman
 from werkzeug.wrappers import Request
+from urllib.parse import urlparse
 
 from config import (
     STATIC_FOLDER,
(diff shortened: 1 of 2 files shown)


jonahkagan:
cc @eventualbuddha

arlo_server/__init__.py
```diff
@@ -16,7 +17,7 @@
     # Restrict which hosts we trust when not in dev/test. This works by causing
     # anything accessing the request URL (i.e. `request.url` or similar) to
     # throw an exception if it doesn't match one of the values in this list.
-    Request.trusted_hosts = [HTTP_ORIGIN]
+    Request.trusted_hosts = [str(urlparse(HTTP_ORIGIN).hostname)]
```
jonahkagan, line 20: Have to wrap with `str()` because the type of `.hostname` is `Optional[str]`. Dunno if anyone knows a better way to deal with this.

PR #434, opened 2020-04-14
Adapt auth0 env vars for single tenant

jonahkagan:
**Description**

Shifting from having a separate auth0 tenant for audit admins and jurisdiction admins to having one single tenant for both user types. This makes our configuration/code simpler and also is a more recommended way to handle multiple user types by auth0's docs.

**Testing**

Tested locally by logging in with each user type.

**Progress**

Ready for review.


jonahkagan:
Decided to stick with two tenants. Closing

PR #435, opened 2020-04-15
Require jurisdiction admin authorization to upload ballot manifest

jonahkagan:
**Description**

Ok, this PR is a bit complicated. The goal is to require jurisdiction admin authorization for uploading/deleting ballot manifests for the multi-jurisdiction flow, while leaving the existing endpoint for the single-jurisdiction flow without authorization.

Here's how it's done:
- Factor out the logic from the existing endpoint `PUT/DELETE /election/<election_id>/jurisdiction/<jurisdiction_id>/manifest`
- Adding two corresponding endpoints, `PUT` and `DELETE /election/<election_id>/jurisdiction/<jurisdiction_id>/ballot-manifest` (yes, it's a very similar name - eventually I'm gonna come back and namespace the endpoints for the two flows)
- Add another new endpoint to `GET` the results of the upload
- Reorganize some code - everything for single-jurisdiction goes into `routes.py` (e.g. the `sample_ballots` helper) and everything for multi-jurisdiction goes into a new file `ballot_manifest.py`, following the established pattern.
- Update all of the tests for other multi-jurisdiction flow endpoints that were relying on being able to upload the manifest without authorization to use the new endpoint
- Add some new tests for the new endpoints

**Testing**

Added tests for the new endpoints.

**Progress**

Ready for review.


arlo_server/ballot_manifest.py
```diff
@@ -0,0 +1,157 @@
...
+    UserError,
+)
+
+BATCH_NAME = "Batch Name"
+NUMBER_OF_BALLOTS = "Number of Ballots"
+STORAGE_LOCATION = "Storage Location"
+TABULATOR = "Tabulator"
+
+
+def process_ballot_manifest_file(
```
jonahkagan, line 29: this method was basically copy-pasted from `util.ballot_manifest` - there's no new logic/changes to logic here

arlo_server/ballot_manifest.py
```diff
@@ -0,0 +1,157 @@
...
+
+        num_batches = 0
+        num_ballots = 0
+        for row in manifest_csv:
+            num_ballots_in_batch_csv = row[NUMBER_OF_BALLOTS]
+
+            try:
+                num_ballots_in_batch = locale.atoi(num_ballots_in_batch_csv)
+            except ValueError as error:
+                raise UserError(
```
jonahkagan, line 54: I changed this from `Exception` to a new `UserError` class so that we don't crash the server on validation errors. see `process_file.py` where we catch it

arlo_server/ballot_manifest.py
```diff
@@ -0,0 +1,157 @@
...
+                    f"Invalid value for '{NUMBER_OF_BALLOTS}' on line {manifest_csv.line_num}: {num_ballots_in_batch_csv}"
+                ) from error
+
+            batch = Batch(
+                id=str(uuid.uuid4()),
+                name=row[BATCH_NAME],
+                jurisdiction_id=jurisdiction.id,
+                num_ballots=num_ballots_in_batch,
+                storage_location=row.get(STORAGE_LOCATION) or None,
+                tabulator=row.get(TABULATOR) or None,
```
jonahkagan, line 64: I changed these from `row.get(field, None)` because they were returning empty strings in the case where the value was omitted, but `None` in the case where the entire column was omitted. I wanted it to be consistent

arlo_server/ballot_manifest.py
```diff
@@ -0,0 +1,157 @@
...
+
+    election = jurisdiction.election
+
+    # If we're in the single-jurisdiction flow, posting the ballot manifest
+    # starts the first round, so we need to sample the ballots.
+    # In the multi-jurisdiction flow, this happens after all jurisdictions
+    # upload manifests, and is triggered by a different endpoint.
+    if not election.is_multi_jurisdiction:
+        # Import this here to avoid circular dependencies
+        from arlo_server.routes import sample_ballots
```
jonahkagan, line 83: I moved this import here to accommodate the division of logic between the two files. `routes` has all the logic that is for the single-page flow generally, but we need to include `sample_ballots` for this one case here. This file has some logic that's shared for saving/deleting ballot manifests, so it needs to be imported in `routes`.

arlo_server/routes.py
```diff
@@ -157,6 +164,97 @@ def check_round(election, jurisdiction_id, round_id):
     return is_complete
 
 
+def sample_ballots(session: Session, election: Election, round: Round):
```
jonahkagan, line 167: this just got moved back here from util.ballot_manifest, no changes

arlo_server/routes.py
```diff
@@ -525,45 +623,19 @@ def jurisdiction_manifest(jurisdiction_id, election_id):
     election = get_election(election_id)
     jurisdiction = Jurisdiction.query.filter_by(
         election_id=election.id, id=jurisdiction_id
-    ).one()
```
jonahkagan, line 528: `.one()` throws on not found, so the next condition wasn't necessary

tests/conftest.py
```diff
@@ -270,12 +269,6 @@ def audit_board_round_2_ids(
     yield [ab["id"] for ab in audit_boards]
 
 
-@pytest.fixture
-def as_jurisdiction_admin(client: FlaskClient, jurisdiction_ids: List[str]):
-    create_jurisdiction_admin(jurisdiction_ids[0])
-    set_logged_in_user(client, UserType.JURISDICTION_ADMIN, DEFAULT_JA_EMAIL)
```
jonahkagan, line 276: this didn't work so well because there's no guaranteed order to the fixtures getting called

tests/conftest.py
```diff
@@ -129,8 +131,9 @@ def election_settings(client: FlaskClient, election_id: str) -> None:
 
 @pytest.fixture
 def manifests(client: FlaskClient, election_id: str, jurisdiction_ids: List[str]):
+    set_logged_in_user(client, UserType.JURISDICTION_ADMIN, DEFAULT_JA_EMAIL)
     rv = client.put(
-        f"/election/{election_id}/jurisdiction/{jurisdiction_ids[0]}/manifest",
+        f"/election/{election_id}/jurisdiction/{jurisdiction_ids[0]}/ballot-manifest",
```
benadida, line 136: I'm seeing a lot of these changes from `manifest` to `ballot-manifest` in the tests... but are you still testing the old endpoints for the single-jurisdiction flow? For the sake of high-level review, I trust that you will double check that, and otherwise I'm good :)

jonahkagan: The only places I am changing tests from the old to the new endpoint is in tests for the multi-jurisdiction flow - tests I've written since I started working on it. The existing tests for the single-jurisdiction flow still use the old endpoint. Just double checked to make sure.

PR #436, opened 2020-04-16
Return audit board passphrase in JA audit board list endpoint

jonahkagan:
**Description**

The passphrase is needed for the frontend to create the audit board login PDFs.

**Testing**

Updated existing tests.

**Progress**

Ready for review.


jonahkagan:
Kind of curious why all the PDFs are created by the frontend and not the backend. Since we already have the code there, I'm not suggesting we change it, but just wanting to understand the design decision.

benadida:
> Kind of curious why all the PDFs are created by the frontend and not the backend. Since we already have the code there, I'm not suggesting we change it, but just wanting to understand the design decision.

Tried to keep the backend mostly a data API, that's all :)

PR #440, opened 2020-04-17
Count votes and record results when ending a round

jonahkagan:
**Description**

Forgot to do this before since it's not something that was implemented yet in the single jurisdiction flow.

**Testing**

Added some checks to existing test cases.

**Progress**

Ready for review.

### DIFF
--- a/arlo_server/audit_boards.py
+++ b/arlo_server/audit_boards.py
@@ -19,6 +19,8 @@ from arlo_server.models import (
     SampledBallotDraw,
     BallotStatus,
     Batch,
+    RoundContestResult,
+    BallotInterpretation,
 )
 from arlo_server.errors import handle_unique_constraint_error
 from arlo_server.sample_sizes import cumulative_contest_results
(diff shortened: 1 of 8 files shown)


arlo_server/audit_boards.py
```diff
@@ -210,7 +211,27 @@ def calculate_risk_measurements(election: Election, round: Round):
         round_contest.is_complete = is_complete
 
 
+def count_audited_votes(election: Election, round: Round):
+    vote_counts = dict(
+        SampledBallotDraw.query.filter_by(round_id=round.id)
+        .outerjoin(SampledBallot)
+        .group_by(SampledBallot.vote)
+        .values(SampledBallot.vote, func.count())
+    )
```
benadida, line 220: is it necessary to do this join, or can you just select from `SampledBallot` ?

jonahkagan: Also, this query has another flaw: contest choices are not unique across contests, so I need to change this.

jonahkagan:
Updated to use the new BallotInterpretation data model

PR #447, opened 2020-04-23
Add BallotInterpretation model

jonahkagan:
**Description**

To support multiple contests, we need to be able to record an audit board's interpretation for more than one contest per ballot, since a given ballot might have multiple contests. This adds a new model `BallotInterpretation` to represent these interpretations, and adds a one-to-many relationship from `SampledBallot` to `BallotIntepretation`.

Note: this breaks the API for setting the audit board's interpretation of a ballot.

**Testing**

Updated existing tests to cover.

**Progress**

Ready for review, but can't merge until combined with a frontend PR (coming soon) to consume the new data model (integration tests fail).


arlo_server/models.py
```diff
@@ -278,6 +278,12 @@ class Round(BaseModel):
     audit_boards = relationship("AuditBoard", backref="round", passive_deletes=True)
 
 
+class BallotStatus(str, Enum):
+    NOT_AUDITED = "NOT_AUDITED"
+    AUDITED = "AUDITED"
+    SKIPPED = "SKIPPED"
```
jonahkagan, line 284: not strictly needed yet, but figured i'd throw it in now since we'll be using it later

arlo_server/routes.py
```diff
@@ -865,12 +872,11 @@ def ballot_set(election_id, jurisdiction_id, batch_id, ballot_position):
...
-
-    if "vote" in attributes:
-        ballot.vote = attributes["vote"]
-
-    if "comment" in attributes:
-        ballot.comment = attributes["comment"]
+    ballot.interpretations = [
+        deserialize_interpretation(ballot.id, interpretation)
+        for interpretation in attributes["interpretations"]
+    ]
```
jonahkagan, line 878: adding `delete-orphan` in the relationship config allows us to do the update this way. any `Interpretation` that was in the old array but not the new array will be orphaned and deleted automagically

arlo_server/routes.py
```diff
@@ -1088,6 +1091,18 @@ def audit_report(election_id):
     )
 
 
+# TODO make this work for multiple contests
+# TODO include comments
```
jonahkagan, line 1095: will come back to this when i update the audit report for multi-jurisdiction in general

arlo_server/models.py
benadida: I think this is partially redundant with the presence of the `BallotInterpretation` row, which means the data model could become inconsistent, which is not ideal.

Maybe consider a `SKIPPED` option in the `Interpretation` enum and then you can get rid of this field?

jonahkagan: I'm with you on having internally consistent data models. Super important.

I think the reason it's more appropriate this way for a few reasons:
- On the product side, audit boards skip a whole ballot, not individual contests.
- According to @mcchilders, an audit board might come across a ballot with no selected contests, in which case they "audit" but don't record any interpretations. I think this is a consequence of the fact that we have no way of knowing which contests are on which ballots when we sample.

Thoughts?
Are there ways we can help prevent this from becoming inconsistent? Constraints?

mcchilders: Could also do audited/not audited/not found (if that's clearer in code than "skipped").

PR #455, opened 2020-04-29
Initial setup for JA flow and Progress stage

MorganLove:
**Description**

- Tweaking the conditional rendering of components to support the other views. Just a small foundational PR to expand on in later PRs

**Testing**

- Just updating the old tests to reflect this new setup

**Progress**

- Updating tests.


arlo-client/src/components/Audit/Setup/Contests/index.tsx
```diff
@@ -74,7 +74,7 @@ const Contests: React.FC<IProps> = ({
     // TEST TODO
     /* istanbul ignore next */
     if (!response) return
-    nextStage.activate()
+    nextStage.activate!()
```
benadida, line 77: what's the reasoning for `activate` to be possibly null?

MorganLove: @benadida Because in the sidebar for the JA setup flow, there's literally only one screen, so only one item in the menu, therefore no toggling between them. Thus no `onClick` function that would ever do anything. Instead of doing a null function that we'd then have to test that it did nothing, just excluding it made more sense to me. We always know that it exists in the AA setup sidebar, though, so we can just tell TS that.

benadida: I think the problem here is that, if `activate` is null, this will compile okay, but will fail at runtime. The `!`, if I understand it correctly (and I'm relatively new to it as of last year), is just telling the compiler to shut up and ignore the potential null case, but it doesn't change the runtime.

So if `activate` can be null, we probably want this instead:

```
if (nextStage.activate) nextStage.activate()
```

arlo-client/src/components/Audit/index.tsx
```diff
@@ -82,55 +83,97 @@ const Audit: React.FC<{}> = () => {
...
+          <>
+            {progressSidebar}
+            <p>Progress view</p>
+          </>
+        )}
+        {viewMatch !== 'setup' && viewMatch !== 'progress' && (
+          <p>Round management view</p>
+        )}
+      </Wrapper>
+    )
```
jonahkagan, line 142: I'm a little worried by having data loading and rendering for the single-jurisdiction flow and the multi-jurisdiction flow in the same component. It's a bit confusing to read. I wonder if it would be simpler to have one high level conditional at the top of the component hierarchy that checks the logged in user type and delegates to one of three components:
- if anonymous user, render the single-jurisdiction flow
- if audit admin, render the audit admin view of the multi-jurisdiction flow
- if jurisdiction admin, render the jurisdiction admin view of the multi-jurisdiction flow

Each of these components would have its own routing and data loading internally.

arlo-client/src/components/Audit/index.tsx
```diff
@@ -1,5 +1,6 @@
 import React, { useState, useEffect, useCallback, useContext } from 'react'
 import { useRouteMatch, RouteComponentProps } from 'react-router-dom'
+// import { Spinner } from '@blueprintjs/core'
```
jonahkagan, line 3: is this needed?

arlo-client/src/components/Audit/Setup/Contests/index.tsx
MorganLove: I see your point. My concern is that the only time it would get a null value at runtime would be if someone accidentally coded it to pass in the wrong variable. In which case it *should* fail at runtime and get fixed, instead of being handled gracefully. If it simply doesn't call the function if the function doesn't exist, that would actually be a bug that could possibly be missed. Just my thought on it, though.

arlo-client/src/components/Audit/index.tsx
MorganLove: @jonahkagan I hear you, but that would also involve some massive file restructuring and would also involve splitting up and moving around a lot of tests and mocks. It would be cleaner in the end, and I do think that refactor would be a good thing to do, but right now I think it's a time commitment we don't need.

arlo-client/src/components/Audit/Setup/Contests/index.tsx
benadida: got it. My sense is that there might then be a better way to model it, but in the interest of moving forward, if this is meant to fail on null, then I can live with this.

Small suggestion: raise an explicit error that logs what happened?

MorganLove: I could do that, though in the interests of saving on writing tests, I'll add an istanbul ignore comment too, haha.

benadida: for sure. And to be clear I mean very little investment here, a UI if you already have it (toast?), or just a `console.log`.

PR #456, opened 2020-04-30
Add isAuditComplete field to round API

jonahkagan:
**Description**

The audit is complete if and only if the targeted contest meets its risk limit. This change adds a boolean field `isAuditComplete` to the round API that flags whether or not the audit ended at the end of that round. The field is `null` until the round ends.

**Testing**

Added a new test case and updated some existing test cases.

Also had to change some inconsistent test data. The test helper that was "completing" round 1 and setting up round 2 was auditing votes in a ratio that didn't actually require a second round. I changed this ratio to be correct, which also changed the round 2 sample size, so a number of expected values in the tests had to change. I'm confident that the new values are correct because I didn't change the code that computes them.

I also updated the code that splits the ballots between audit boards to key on `Batch.name` instead of `Batch.id`, which is non-deterministic and was making my tests fail due to things being in different orders on different runs. Why this didn't happen before, I don't know. Prob just got lucky.

**Progress**

Ready


jonahkagan:
cc @MorganLove

arlo_server/audit_boards.py
```diff
@@ -68,24 +69,25 @@ def assign_sampled_ballots(
         .filter_by(jurisdiction_id=jurisdiction.id)
         .join(SampledBallot.draws)
         .filter_by(round_id=round.id)
-        .order_by(SampledBallot.batch_id)  # group_by prefers a sorted list
+        .order_by(Batch.name)  # group_by prefers a sorted list
```
benadida, line 72: what's the reasoning for switching to name vs ID?

arlo_server/rounds.py
```diff
@@ -37,6 +38,7 @@ def serialize_round(r: Round) -> dict:
         "roundNum": r.round_num,
         "startedAt": isoformat(r.created_at),
         "endedAt": isoformat(r.ended_at),
+        "isAuditComplete": is_audit_complete(r.id),
```
benadida, line 41: flagging for later: this means that we're doing a query for each round that is serialized in a list of rounds, not ideal if there are a lot of rounds, probably not an issue right now.

arlo_server/audit_boards.py
jonahkagan: I explained in the PR description:

I also updated the code that splits the ballots between audit boards to key on Batch.name instead of Batch.id, which is non-deterministic and was making my tests fail due to things being in different orders on different runs. Why this didn't happen before, I don't know. Prob just got lucky.

arlo_server/rounds.py
jonahkagan: Yeah, I thought about that. For most list endpoints that have a computed field like this, I batched into one query, but I figured in this case it's so few rounds that it's a negligible difference.

PR #464, opened 2020-05-01
Refactor audit component

jonahkagan:
**Description**

Separate the `Audit` component into different components for diff audit types/user types.

**Testing**

Existing tests cover.

**Progress**

Ready

### DIFF
--- a/arlo-client/src/components/Audit/index.test.tsx
+++ b/arlo-client/src/components/Audit/index.test.tsx
@@ -1,15 +1,15 @@
-import React from 'react'
+import React, { useContext } from 'react'
 import { waitForElement, wait, fireEvent } from '@testing-library/react'
 import {
   BrowserRouter as Router,
   useRouteMatch,
   useParams,
 } from 'react-router-dom'
-import Audit from './index'
+import { SingleJurisdictionAudit, MultiJurisdictionAudit } from './index'
 import { statusStates, dummyBallots, auditSettings } from './_mocks'
 import * as utilities from '../utilities'
 import { asyncActRender } from '../testUtilities'
-import AuthDataProvider from '../UserContext'
+import AuthDataProvider, { AuthDataContext } from '../UserContext'
 import getJurisdictionFileStatus from './useSetupMenuItems/getJurisdictionFileStatus'
 import getRoundStatus from './useSetupMenuItems/getRoundStatus'
 import { contestMocks } from './Setup/Contests/_mocks'
@@ -73,7 +73,7 @@ describe('RiskLimitingAuditForm', () => {
     apiMock.mockImplementation(async () => statusStates.empty)
     const { container } = await asyncActRender(
       <Router>
-        <Audit />
+        <SingleJurisdictionAudit />
       </Router>
     )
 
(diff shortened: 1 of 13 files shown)


MorganLove:
@jonahkagan Interesting that the backend tests are failing. Huh.

But yeah, this looks like a great direction to go in!

jonahkagan:
Weird, I couldn't repro that backend test failure at all. I tried so many times on that commit, and on different branches. If you see it again, let me know.

arlo-client/src/components/Audit/index.test.tsx
```diff
@@ -207,9 +207,15 @@ describe('RiskLimitingAuditForm', () => {
 })
 
 describe('AA setup flow', () => {
+  // MultiJurisdictionAudit will only be rendered once the user is logged in, so
+  // we simulate that.
+  const MultiJurisdictionAuditWithAuth: React.FC = () => {
+    const { isAuthenticated } = useContext(AuthDataContext)
+    return isAuthenticated ? <MultiJurisdictionAudit /> : null
```
benadida, line 214: is `null` the right alternate return value here, or should it be an empty react fragment?

arlo-client/src/components/CreateAudit.test.tsx
```diff
@@ -20,7 +20,7 @@ const checkAndToastMock: jest.SpyInstance<
 checkAndToastMock.mockReturnValue(false)
 
 const routeProps: RouteComponentProps<ICreateAuditParams> = routerTestProps(
-  '/election/:electionId',
+  '/audit/:electionId',
```
benadida, line 23: wait, did i miss a URL change? I'm not seeing it

benadida: or is this just for testing? If so it's a tad confusing.

arlo-client/src/components/CreateAudit.tsx
```diff
@@ -70,7 +70,11 @@ const CreateAudit = ({ history }: RouteComponentProps<ICreateAuditParams>) => {
         return
       }
       const { electionId } = response
-      history.push(`/election/${electionId}/setup`)
+      if (isAuthenticated) {
+        history.push(`/election/${electionId}/setup`)
+      } else {
+        history.push(`/audit/${electionId}/setup`)
```
benadida, line 76: for the life of me I can't find this `/audit/` route... where is it?

arlo-client/src/components/Audit/index.test.tsx
jonahkagan: https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

arlo-client/src/components/CreateAudit.test.tsx
jonahkagan: https://github.com/votingworks/arlo/pull/464/files#diff-c4fc915a5c89a9458ba5320c7f2be3f8R72-R73

the single-jurisdiction flow is going under `/audit` so we can use routing to know which flow we're supposed to render

PR #466, opened 2020-05-01
added auth0 integration for superadmins

benadida:
Same approach we've taken for OAuth URLs, now for superadmins, with its own independent tenant. Difference is this manages a separate session variable and sends the user to a different URL, `/superadmin/`.

### DIFF
--- a/arlo_server/auth.py
+++ b/arlo_server/auth.py
@@ -53,7 +53,7 @@ def set_superadmin():
 
 
 def clear_superadmin():
-    del session["_superadmin"]  # pragma: no cover
+    session["_superadmin"] = False  # pragma: no cover
 
 
 def is_superadmin():
--- a/arlo_server/auth.py
+++ b/arlo_server/auth.py
@@ -53,7 +53,7 @@ def set_superadmin():
 
 
 def clear_superadmin():
-    session["_superadmin"] = False  # pragma: no cover
+    del session["_superadmin"]  # pragma: no cover
 
 
 def is_superadmin():
--- a/arlo_server/routes.py
+++ b/arlo_server/routes.py
@@ -21,6 +21,8 @@ from arlo_server.auth import (
     get_loggedin_user,
     require_audit_admin_for_organization,
     set_loggedin_user,
+    set_superadmin,
+    clear_superadmin,
     with_election_access,
 )
 from arlo_server.models import *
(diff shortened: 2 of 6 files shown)


arlo_server/routes.py
```diff
@@ -1246,6 +1267,30 @@ def logout():
...
+    return auth0_sa.authorize_redirect(redirect_uri=redirect_uri)
+
+
+@app.route(SUPERADMIN_OAUTH_CALLBACK_URL)
+def superadmin_login_callback():
+    auth0_sa.authorize_access_token()
+    resp = auth0_sa.get("userinfo")
+    userinfo = resp.json()
+
+    # we rely on the auth0 auth here, but check against a single approved domain.
```
jonahkagan, line 1282: why is this necessary?

tests/routes_tests/test_auth.py
```diff
@@ -93,6 +95,30 @@ def check_redirect_contains_redirect_uri(response, expected_url):
...
+
+    mock_response = Mock()
+    mock_response.json = MagicMock(return_value={"email": SA_EMAIL})
+    auth0_sa.get = Mock(return_value=mock_response)
+
+    rv = client.get("/auth/superadmin/callback?code=foobar")
+    assert rv.status_code == 302
+
+    with client.session_transaction() as session:  # type: ignore
+        assert session["_superadmin"]
```
jonahkagan, line 116: maybe some assertion about not messing with the rest of the session?

tests/routes_tests/test_auth.py
```diff
@@ -93,6 +95,30 @@ def check_redirect_contains_redirect_uri(response, expected_url):
...
+
+    assert auth0_sa.authorize_access_token.called
+    assert auth0_sa.get.called
+
+
```
jonahkagan, line 121: update tests for logout?

arlo_server/auth.py
```diff
@@ -53,7 +53,7 @@ def set_superadmin():
 
 
 def clear_superadmin():
-    del session["_superadmin"]  # pragma: no cover
+    session["_superadmin"] = False  # pragma: no cover
```
jonahkagan, line 56: why this change?

arlo_server/routes.py
benadida: defense-in-depth, I don't want to mistakenly get to a place where we forget to limit users in Auth0, and then anyone with a gmail address gets superadmin.

PR #467, opened 2020-05-01
Round management

MorganLove:
**Description**

- Adding the creation of audit boards and structure for the round management screens

**Testing**

- None as of yet

**Progress**

- The progress component needs to be created, and the offline data entry


arlo-client/src/components/Audit/RoundManagement/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+    return (
+      <Wrapper className="single-page">
+        <H2Title>
+          Congratulations! Your Risk-Limiting Audit is now complete.
+        </H2Title>
+      </Wrapper>
+    )
+  return (
+    <Wrapper className="single-page left">
+      <H2Title>Round 1 Audit Board Setup</H2Title>
```
benadida, line 101: I'm assuming this says round 1 for now and will be updated to the actual round number at some future point?

arlo-client/src/components/Audit/index.tsx
```diff
@@ -79,58 +81,105 @@ const Audit: React.FC<{}> = () => {
   const [menuItems, refresh] = useSetupMenuItems(stage, setStage, electionId)
 
   useEffect(() => {
-    refresh()
-  }, [refresh])
+    if (
+      isAuthenticated &&
+      viewMatch === 'setup' &&
+      meta!.type === 'audit_admin'
```
benadida, line 87: might we have an easier time if we fully separated the UX between audit admin and jurisdiction admin? I wonder if these internal if statements will be harder to manage over time.

MorganLove: @benadida Yes, that's actually being fixed here: https://github.com/votingworks/arlo/pull/464

PR #470, opened 2020-05-02
Adding progress bars

MorganLove:
**Description**

- Adding progress bars to the Round Progress view

**Testing**

- none yet

**Progress**

- Looks nice, imo!

### DIFF
--- a/arlo-client/src/components/Audit/RoundManagement/RoundProgress.tsx
+++ b/arlo-client/src/components/Audit/RoundManagement/RoundProgress.tsx
@@ -1,8 +1,62 @@
 import React from 'react'
+import { ProgressBar } from '@blueprintjs/core'
+import styled from 'styled-components'
 import H2Title from '../../Atoms/H2Title'
+import { IBallot, IAuditBoard, BallotStatus } from '../../../types'
+import { IRound } from '../useRoundsJurisdictionAdmin'
 
-const RoundProgress = () => {
-  return <H2Title>Round 1 Progress</H2Title>
+const MainBarWrapper = styled.div`
+  margin-bottom: 20px;
+  width: 500px;
+  font-size: 16px;
+`
+
+const SmallBarWrapper = styled.div`
+  margin-bottom: 10px;
+  width: 400px;
+`
+
+const RoundProgress = ({
+  rounds,
+  ballots,
+  auditBoards,
+}: {
+  rounds: IRound[]
+  ballots: IBallot[]
+  auditBoards: IAuditBoard[]
+}) => {
+  const completeBallots = ballots.filter(
+    b => b.status !== BallotStatus.NOT_AUDITED
+  ).length
+  const ballotsPerAB = auditBoards.map(a => {
+    const all = ballots.filter(b => b.auditBoard!.id === a.id)
+    return [all.filter(b => b.status).length, all.length]
+  })
+  return (
+    <>
+      <H2Title>Round {rounds.length} Progress</H2Title>
+      <MainBarWrapper>
+        <span>
+          {completeBallots} of {ballots.length} ballots audited{' '}
+        </span>
+        <ProgressBar
+          value={completeBallots / ballots.length}
+          animate={completeBallots < ballots.length}
+          intent="primary"
+        />
+      </MainBarWrapper>
+      {ballotsPerAB.map((v, i) => (
+        <SmallBarWrapper key={auditBoards[i].id}>
+          <span>{`${auditBoards[i].name}: ${v[0]} of ${v[1]} ballots audited `}</span>
+          <ProgressBar
+            value={v[0] / v[1]}
+            animate={v[0] < v[1]}
+            intent="primary"
+          />
+        </SmallBarWrapper>
+      ))}
+    </>
+  )
 }
 
 export default RoundProgress
--- a/arlo-client/src/components/Audit/RoundManagement/RoundProgress.tsx
+++ b/arlo-client/src/components/Audit/RoundManagement/RoundProgress.tsx
@@ -2,8 +2,8 @@ import React from 'react'
 import { ProgressBar } from '@blueprintjs/core'
 import styled from 'styled-components'
 import H2Title from '../../Atoms/H2Title'
-import { IBallot, IAuditBoard, BallotStatus } from '../../../types'
 import { IRound } from '../useRoundsJurisdictionAdmin'
+import { IAuditBoard } from '../useAuditBoards'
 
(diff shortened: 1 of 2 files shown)


arlo-client/src/components/Audit/RoundManagement/RoundProgress.tsx
```diff
@@ -1,8 +1,60 @@
...
+const RoundProgress = ({
+  rounds,
+  ballots,
+  auditBoards,
+}: {
+  rounds: IRound[]
+  ballots: IBallot[]
+  auditBoards: IAuditBoard[]
+}) => {
+  const completeBallots = ballots.filter(b => b.status).length
```
jonahkagan, line 28: `b.status` will always be defined. It will be either `'AUDITED'`, `'NOT_AUDITED'`, or `'NOT_FOUND'`. So in this case, I think you want `ballots.filter(b => b.status !== BallotStatus.NOT_AUDITED)` (assuming you make an enum)

jonahkagan, line 28: `b.status` will always be defined. It will be either `'AUDITED'`, `'NOT_AUDITED'`, or `'NOT_FOUND'`. So in this case, I think you want `ballots.filter(b => b.status !== BallotStatus.NOT_AUDITED)` (assuming you make an enum)

arlo-client/src/components/Audit/RoundManagement/RoundProgress.tsx
```diff
@@ -1,8 +1,60 @@
...
+  const ballotsPerAB = auditBoards.map(a => {
+    const all = ballots.filter(b => b.auditBoard!.id === a.id)
+    return [all.filter(b => b.status).length, all.length]
+  })
```
jonahkagan, line 32: these counts are already available on the `auditBoard` object at `auditBoard.currentRoundStatus.numSampledBallots` and `auditBoard.currentRoundStatus.numAuditedBallots` 😸

jonahkagan: Which means, if you'd like, the large status bar progress can be computed just by summing across the audit board stats - you don't even need the list of ballots

arlo-client/src/components/Audit/RoundManagement/RoundProgress.tsx
MorganLove: @jonahkagan The enum is already there, so that works. :grin:

arlo-client/src/components/Audit/RoundManagement/RoundProgress.tsx
MorganLove: @jonahkagan Huh, must have missed that. It's not in the types anywhere on this branch. I see a `currentRoundStatus` on the `jurisdictions` object, though it has `numBallotsSampled` and `numBallotsAudited` which are slightly different. Could you get me the full type for what's currently returned from `/election/:electionId/jurisdiction/:jurisdictionId/round/:roundId/audit-board`?

MorganLove: @jonahkagan Okay, rebased on #480 and cleaned this up a bit. Can you double check that I'm using `useAuditBoards` correctly?

arlo-client/src/components/Audit/RoundManagement/RoundProgress.tsx
```diff
@@ -1,8 +1,71 @@
...
+  electionId: string
+  jurisdictionId: string
+  round: IRound
+}) => {
+  const [auditBoards] = useAuditBoards(electionId, jurisdictionId, [round])
+  if (!auditBoards) return null
+  const [auditedBallots, sampledBallots] = auditBoards.reduce(
+    (a, v) => {
+      a[0] += v.currentRoundStatus.numAuditedBallots // eslint-disable-line no-param-reassign
+      a[1] += v.currentRoundStatus.numSampledBallots // eslint-disable-line no-param-reassign
```
benadida, line 33: I know it's all local code, but I'd still prefer `a.numAudited` over `a[0]`

jonahkagan: alternatively, could do something like:

```
const sum = (ns: number[]) => ns.reduce((a, b) => a+b)
const auditedBallots = sum(auditBoards.map(ab => ab.currentRoundStatus.numAuditedBallots))
const sampledBallots = sum(auditBoards.map(ab => ab.currentRoundStatus.numSampledBallots))
```

arlo-client/src/components/Audit/RoundManagement/RoundProgress.tsx
```diff
@@ -1,8 +1,71 @@
...
+        <span>
+          {auditedBallots} of {sampledBallots} ballots audited{' '}
+        </span>
+        <ProgressBar
+          value={auditedBallots / sampledBallots}
+          animate={auditedBallots < sampledBallots}
+          intent="primary"
+        />
+      </MainBarWrapper>
+      {auditBoards.map((v, i) => (
```
jonahkagan, line 51: maybe destructure at the top of this function to make things more concise?

```ts
auditBoards.map(({id, name, currentRoundStatus: {numAuditedBallots, numSampledBallots}) =>
```

arlo-client/src/components/Audit/RoundManagement/index.tsx
```diff
@@ -91,7 +91,11 @@ const RoundManagement = ({ round, auditBoards, createAuditBoards }: IProps) => {
             >
               Download Audit Board Credentials for Data Entry
             </FormButton>
-            <RoundProgress />
+            <RoundProgress
+              round={round}
+              electionId={electionId}
+              jurisdictionId={jurisdictionId}
```
jonahkagan, line 97: instead of calling `useAuditBoards` within `RoundProgres`, if you pass the `auditBoards` from this parent component in as a prop `auditBoards={auditBoards}`, then they will stay up to date, since it's flowing from the same centralized state

arlo-client/src/components/Audit/RoundManagement/RoundProgress.tsx
MorganLove: That does seem more self-documenting, thanks!

arlo-client/src/components/Audit/RoundManagement/index.tsx
MorganLove: Aha, I missed that it was already in there. I looked for it, haha!

PR #471, opened 2020-05-02
logging out on an empty session shouldn't fail

benadida:


arlo_server/auth.py
```diff
@@ -47,17 +47,20 @@ def clear_loggedin_user():
 ## to re-login
 ##
 
+SESSION_KEY_SUPERADMIN = "_superadmin"
```
jonahkagan, line 50: could this just be `_SUPERADMIN` to make things a bit more readable?

also can you make the user key a constant too while you're at it. i'd prefer to have consistency

PR #473, opened 2020-05-02
old-man HTML actions for super admins

benadida:
tests not fully written yet, waiting to see how horrified @jonahkagan is first :)

### DIFF
--- a/arlo_server/superadmin.py
+++ b/arlo_server/superadmin.py
@@ -1,18 +1,48 @@
-from flask import jsonify
+from flask import render_template, redirect, request
 
 from arlo_server import app
-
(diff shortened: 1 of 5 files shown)


benadida:
> I do think this incurs unnecessary debt. I don't think it would have taken much longer to implement in the frontend

It has the benefit of being done extremely quickly... likely a lot faster for me than doing it in React. But if you are worried about where it leaves us, I'm certainly okay to redo it in React. Sounds like you are worried, so I'll do it :)

jonahkagan:
My approval was sincere. I think it's done and yes, it was fast! I don't mind reimplementing it in React later another time.

PR #476, opened 2020-05-02
load testing plan

benadida:


docs/loadtesting.md
```diff
@@ -0,0 +1,42 @@
...
+
+We have superadmin features that allow for impersonation of any user,
+and which uses the same cookies as normal logins. Let's reuse that.
+
+We'll build a superadmin bypass using a special HTTP header. Because
+we don't ever want that code in production, we'll keep that code in a
+special branch, `loadtesting`, that we'll need to rebase every so
+often on `master`. We'll also require a special environment variable
+be set as an extra defense-in-depth measure, just in case we
+mistakenly merge that branch into `master`. Belt _and_ suspenders.
```
jonahkagan, line 24: we're gonna need this for our integration tests as well, so i think we'll need something we feel comfortable having on `master`

this overall approach seems mostly fine to me. my small concern is overloading the superadmin feature - it might need to change for other reasons and then we'd have to worry about preserving this

i was wondering if there's any way to mock auth0 to make this more straightforward. we could still do defense-in-depth

docs/loadtesting.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Tool
+
+One tool that some of us have used with success is
+[Artillery](https://artillery.io), and in particular [Serverless
+Artillery](https://github.com/Nordstrom/serverless-artillery), which
+combines the same test scripting language with AWS Lambda deployment
+for instant scaling without overhead of setting up machines.
```
jonahkagan, line 32: will we be able to reuse these as integration tests locally/on CI? or, put another way, can we use integration tests as the core script for load testing?

docs/loadtesting.md
```diff
@@ -0,0 +1,42 @@
...
+
+## User Modeling
+
+For a first test, we'll model one audit admin and 1,000 jurisdiction
+administrators, with the jurisdiction administrators arriving in a
+random fashion over the course of an hour to upload their manifests.
+
+As a second test, we'll model 1,000 jurisdiction administrators going
+through the ballot entry UI and completing the first stage of an
+audit, over the course of 2 hours.
```
jonahkagan, line 42: any reason to wait and do it over hours? seems like we could run it in just a few minutes as a first attempt - if the system doesn't fall over, it's an even stronger result

docs/loadtesting.md
benadida: I don't think load testing makes sense in integration testing. It would be very expensive and time-taking.

docs/loadtesting.md
benadida: Mocking auth0 in a running server? That is tricky and dangerous, effectively implementing another auth bypass in addition to superadmins, I don't think that's great.

docs/loadtesting.md
benadida: I think you're saying you want us to be 100x as performant as we need to be :) I mean, if it's easy sure, but seems like too high a requirement?

docs/loadtesting.md
jonahkagan: Ooh, sorry, I wasn't clear.

I just meant that the test scripts you're writing for load testing will be walking through the user flows in the exact same ways we want to walk through them in integration tests. So it would be nice if we could reuse those test scripts (but for integration tests, only running the flow once)

docs/loadtesting.md
jonahkagan: Yeah, I was just wondering if it would save you time to start with something quick, see if it works, then scale the test length up as necessary

docs/loadtesting.md
benadida: Aha I see. So that's an interesting conundrum, because that changes the requirements in a way that might increase risk. Let's discuss.

benadida:
after conversation with @jonahkagan, will add:
- turning on bypass only on `FLASK_ENV=loadtest`
- add an `exit 1` in CI so this branch can never be merged by mistake.

PR #477, opened 2020-05-02
auth hack for loadtesting

benadida:
a focused auth hack for load testing. PR against the `loadtesting` branch that should never be merged into master.


benadida:
image

yay, working as intended ^^

Issue #279, opened 2020-02-24
AA audit status box


mcchilders:
*User story:*
As an AA I want to be able to see the status of my audit at-a-glance, so that I can keep tabs on what needs to be done.

*Background:*
This grey box at the top is the "audit status" box - it is present on every page in both the AA and JA dashboards, and contains basic info about the audit. Principally, this info is about:
- Overall audit status (not started, in progress, complete)
- Whether the audit setup steps are complete (precursor for launching the audit)
- How many jurisdictions have completed their work in various audit phases (e.g. before the audit starts - how many have uploaded required files, during the audit - how many have completed the ballots assigned to them in the round

*Conditions of Acceptance:*
On the AA dashboard screens, this box contains:
- An indication of the audit status (e.g. "The audit has not started." or "Round 1 of the audit is in progress." or "The audit is complete.")
- Before the audit has started, and indication of whether the audit setup steps have been completed (e.g. "Audit setup is not complete." or "Audit setup is complete."
- Before the audit has started, an indication of how many of the participating jurisdictions have uploaded the files they need to upload for the audit to start (not shown here, requires data from this page to show, but adding for clarity).
- A button to download the audit report (hidden until the audit is started).

*Wireframe:*
Prior to audit launch:
![Screen Shot 2020-02-28 at 12.04.27 PM.png](https://zube.io/files/votingworks/abaf6d8c12fe8a8d1924bb5e79653f7f-screen-shot-2020-02-28-at-12-04-27-pm.png)
After audit launch (includes audit report download button):
![Screen Shot 2020-02-28 at 12.04.50 PM.png](https://zube.io/files/votingworks/2215d889db7649fdbdc11d7797d98a18-screen-shot-2020-02-28-at-12-04-50-pm.png)

mcchilders:
@MorganLove updated the wireframes to show new design. Here are the complete list of statuses for the AA dashboard:
Overall audit statuses ... "The audit has not started." "Round <X> of the audit is in progress." "The audit is complete."
Audit setup statuses (only shown prior to audit launch) ... "Audit setup is not complete." "Audit setup is complete."
Jurisdiction upload statuses (only shown prior to audit launch) ... "<X> of <total # of jurisdictions> jurisdictions have completed file uploads."
Jurisdiction progress statuses (only shown after audit launch & before completion) ...<X> of <total # of jurisdictions> jurisdictions have completed Round <Y>."

MorganLove:
[...]

PR #478, opened 2020-05-03
Audit admin status box

jonahkagan:
**Description**

Add the status box. Based on #279

**Testing**

Updated existing tests to pass.

**Progress**

Ready


arlo-client/src/components/Audit/useContests.ts
```diff
@@ -0,0 +1,102 @@
+import { useEffect, useState } from 'react'
```
jonahkagan, line 1: moved from `useContestsApi`, but git didn't pick up the change because there were also some small changes (e.g. adding `refreshId`)

arlo-client/src/components/Audit/StatusBox.tsx
```diff
@@ -0,0 +1,198 @@
...
+}
+
+export const isSetupComplete = (
+  jurisdictions: IJurisdiction[],
+  contests: IContest[],
+  auditSettings: IAuditSettings
+): boolean =>
+  jurisdictions.length > 0 &&
+  contests.some(c => c.isTargeted) &&
+  Object.entries(auditSettings).every(([, v]) => v !== null)
```
MorganLove, line 61: Not blocking, just curious, since I haven't used `Object.entries` before. But isn't this equivalent to `Object.values(auditSettings).every(v => v !== null)`?

jonahkagan: Oh dang right as I clicked the merge button! Yes, you're right, I'll swap that in

PR #480, opened 2020-05-04
Jurisdiction Admin Status Box

jonahkagan:
**Description**

Adds the Jurisdiction Admin status box. Hoists state that is depended on by both the status box and the other JA views into the parent component.

**Testing**

No new tests

**Progress**

Ready

### DIFF
--- a/arlo-client/src/components/Audit/RoundManagement/CreateAuditBoards.tsx
+++ b/arlo-client/src/components/Audit/RoundManagement/CreateAuditBoards.tsx
@@ -5,6 +5,7 @@ import styled from 'styled-components'
 import { generateOptions } from '../../Atoms/Form/_helpers'
 import FormButton from '../../Atoms/Form/FormButton'
 import Select from '../../Atoms/Form/Select'
+import { IAuditBoard } from '../useAuditBoards'
 
 const Wrapper = styled.div`
   margin: 20px 0;
@@ -16,16 +17,18 @@ interface IValues {
 
 interface IProps {
   createAuditBoards: (auditBoards: { name: string }[]) => Promise<boolean>
-  disabled: boolean
+  auditBoards: IAuditBoard[]
 }
 
-const CreateAuditBoards = ({ createAuditBoards, disabled }: IProps) => {
+const CreateAuditBoards = ({ createAuditBoards, auditBoards }: IProps) => {
   const submit = async ({ numAuditBoards }: IValues) => {
     const boards = [...Array(numAuditBoards).keys()].map(i => ({
-      name: `Audit Board #${i}`,
+      name: `Audit Board #${i + 1}`,
     }))
     createAuditBoards(boards)
   }
+
+  const disabled = auditBoards.length > 0
 
   return (
     <Wrapper>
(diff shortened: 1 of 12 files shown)


arlo-client/src/components/Audit/useBallotManifest.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+const useBallotManifest = (
+  electionId: string,
+  jurisdictionId: string
+): [IFileInfo | null, (csv: File) => Promise<boolean>] => {
+  const [ballotManifest, setBallotManifest] = useState<IFileInfo | null>(null)
+
+  const uploadBallotManifest = async (csv: File): Promise<boolean> => {
+    // TODO poll for result of upload
```
jonahkagan, line 49: I think there will be an elegant and straightforward way to put the polling into this hook. It's not a fully baked idea yet but my intuition is feeling good about it. @MorganLove would love to run it by you

arlo-client/src/components/Audit/RoundManagement/CreateAuditBoards.tsx
```diff
@@ -2,73 +2,44 @@ import React from 'react'
...
+
+interface IProps {
+  createAuditBoards: (auditBoards: { name: string }[]) => Promise<boolean>
+  disabled: boolean
+}
+
+const CreateAuditBoards = ({ createAuditBoards, disabled }: IProps) => {
+  const submit = async ({ numAuditBoards }: IValues) => {
+    const boards = [...Array(numAuditBoards).keys()].map(i => ({
+      name: `Audit Board #${i}`,
```
MorganLove, line 25: @jonahkagan I think this is a bug I found earlier but didn't get around to fixing. This should be `i + 1` so it's more user-friendly (not zero-indexed).

arlo-client/src/components/Audit/RoundManagement/CreateAuditBoards.tsx
```diff
@@ -2,73 +2,44 @@ import React from 'react'
...
+        Select the appropriate number of audit boards based upon the personnel
         available and the number of ballots assigned to your jurisdiction for
         this round of the audit. You will have the opportunity to adjust the
         number of audit boards before the next round of the audit, if another
         round is required.
       </p>
       <Formik
         enableReinitialize
-        initialValues={{ auditBoards: auditBoards.length || 1 }}
+        initialValues={{ numAuditBoards: 1 }}
```
MorganLove, line 42: @jonahkagan If I'm reading this right, the result will be that once the audit boards are created and the form is disabled, it will not show the correct number of audit boards in the locked form. On rerender it will default to 1, regardless of what the value actually is. Because we see this form even when we're done using it, it needs to default to the already-submitted value if it's already submitted. That's what the `enableReinitialize` option is for.

arlo-client/src/components/Audit/useBallotManifest.ts
MorganLove: @MorganLove Kind of like what I did here might work: https://github.com/votingworks/arlo/blob/master/arlo-client/src/components/Audit/useSetupMenuItems/index.ts

I definitely think it can be integrated here in an elegant fashion.

arlo-client/src/components/Audit/RoundManagement/CreateAuditBoards.tsx
jonahkagan: oooh nice catch thank you

arlo-client/src/components/Audit/RoundManagement/CreateAuditBoards.tsx
jonahkagan: oh yeah i'm now seeing that is what's happening. thank you

PR #481, opened 2020-05-04
require fresh logins from oauth providers

benadida:
This is better for security, and also solves the annoying issue where logging in as a jurisdiction admin that isn't in the database means we were stuck in a loop.

### DIFF
--- a/arlo_server/routes.py
+++ b/arlo_server/routes.py
@@ -1175,6 +1175,7 @@ auth0_sa = oauth.register(
     api_base_url=SUPERADMIN_AUTH0_BASE_URL,
     access_token_url=f"{SUPERADMIN_AUTH0_BASE_URL}/oauth/token",
     authorize_url=f"{SUPERADMIN_AUTH0_BASE_URL}/authorize",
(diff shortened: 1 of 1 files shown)


arlo_server/routes.py
```diff
@@ -1255,16 +1258,9 @@ def logout():
...
-    params = urllib.parse.urlencode({"returnTo": return_url})
-
-    base_url = (
-        AUDITADMIN_AUTH0_BASE_URL
-        if user_type == UserType.AUDIT_ADMIN
-        else JURISDICTIONADMIN_AUTH0_BASE_URL
-    )
-    return redirect(f"{base_url}/v2/logout?{params}")
+    # because we have max_age on the oauth requests now,
+    # we don't need to log out of Auth0 anymore.
```
jonahkagan, line 1262: This is the nittiest nitpick ever, but can you take out "now" and "anymore" from this comment? I have a (surprisingly) strong opinion that comments should be timeless since future readers won't know what came before

benadida: I want to frame this comment. ;)

PR #485, opened 2020-05-05
Allow JA to list ballots before audit boards created

jonahkagan:
**Description**

Previously, we just returned an empty list. But the "create audit boards" screen needs to show how many ballots were sampled for the jurisdiction.

**Testing**

Updated test case.

**Progress**

Ready


jonahkagan:
@MorganLove FYI

PR #489, opened 2020-05-05
Audit board signoff

jonahkagan:
**Description**

Added a signoff page and final confirmation page
Updated the backend to return `signedOffAt`

**Testing**

Updated existing tests

**Progress**

Ready

### DIFF
--- a/arlo-client/src/components/DataEntry/SignOff.tsx
+++ b/arlo-client/src/components/DataEntry/SignOff.tsx
@@ -1,0 +1,71 @@
+import React from 'react'
+import { Formik, Form } from 'formik'
+import { H1 } from '@blueprintjs/core'
+import FormSection from '../Atoms/Form/FormSection'
+import { LabelText, NameField } from './Atoms'
+import FormButton from '../Atoms/Form/FormButton'
+import { IAuditBoard } from '../../types'
+
+export interface IMemberNames {
+  memberName1: string
+  memberName2: string
+}
+
+export interface IProps {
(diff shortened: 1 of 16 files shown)


arlo-client/src/types.ts
```diff
@@ -88,6 +88,7 @@ export interface IAuditBoard {
   jurisdictionName: string
   roundId: string
   members: IAuditBoardMember[]
+  signedOffAt: string | null
```
benadida, line 91: should this be instead `signedOffAt?: string` so we are not mixing different approaches to nulls?

arlo-client/src/components/DataEntry/SignOff.tsx
```diff
@@ -0,0 +1,71 @@
...
+  memberName1: string
+  memberName2: string
+}
+
+export interface IProps {
+  auditBoard: IAuditBoard
+  submitSignoff: (memberNames: IMemberNames) => void
+}
+
+// TODO validate that names match
```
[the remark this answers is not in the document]
jonahkagan: just forgot to remove the comment after i implemented it. there's also a server side check. so sekure

arlo-client/src/types.ts
jonahkagan: `signedOffAt?: string` is equivalent to `signedOffAt: string | null | undefined`, which is a slightly different meaning (the field might not be on the object). Not the most important distinction in JS though...

PR #491, opened 2020-05-06
Jurisdiction Admin separate routes

jonahkagan:
**Description**

Move jurisdiction admin routes to `/election/:electionId/jurisdiction/:jurisdictionId` so that we can support cases where a JA is part of multiple audits.

Also clean up some other routing stuff and update a few tests that were failing.


arlo-client/src/components/Audit/CalculateRiskMeasurement.tsx
```diff
@@ -58,7 +58,7 @@ const QRs: React.FC<{ electionId: string; boardIds: string[] }> = ({
       {boardIds.map(id => (
         <span key={id} id={`qr-${id}`}>
           <QRCode
-            value={`${window.location.origin}/election/${electionId}/board/${id}`}
+            value={`${window.location.origin}/election/${electionId}/audit-board/${id}`}
```
benadida, line 61: Hmmm, maybe this slipped in in a prior change, but isn't this QR code supposed to link to `/auditboard/{passphrase}` ?

arlo-client/src/components/CreateAudit.tsx
```diff
@@ -79,6 +79,11 @@ const CreateAudit = ({ history }: RouteComponentProps<ICreateAuditParams>) => {
       toast.error(err.message)
     }
   }
+
+  if (isAuthenticated === null) return null // Still loading
```
benadida, line 83: I know this `isAuthenticated` variable isn't introduced in this PR, but it's a bit fragile to treat a boolean as a trinary value where `false` and `null` aren't the same thing. We should probably fix this at some point.

arlo-client/src/components/CreateAudit.tsx
```diff
@@ -79,6 +79,11 @@ const CreateAudit = ({ history }: RouteComponentProps<ICreateAuditParams>) => {
...
+  if (meta && meta.type === 'audit_board') {
+    return <div>404 Not Found</div>
```
benadida, line 85: is this a fake 404? :)

arlo-client/src/components/Audit/CalculateRiskMeasurement.tsx
jonahkagan: dunno, they end up at the same place though. should i change it?

arlo-client/src/components/CreateAudit.tsx
jonahkagan: good call. i actually think the variable is unnecessary. i explored refactoring a bit to having a user type that included anonymous users, and then just returning the user from this hook (`null` if still loading). but i agree it's not a priority

arlo-client/src/components/CreateAudit.tsx
jonahkagan: yeah, before this view (which is at the root path) was crashing if you were logged in as an audit board, so i just wanted to put in something slightly better

arlo-client/src/components/Audit/CalculateRiskMeasurement.tsx
benadida: so the URL you have should *not* authenticate the audit board user, I think? I wouldn't want someone seeing a URL on a screen be enough to log in as the audit board. The passphrase URL is the one that should set the cookie.

PR #492, opened 2020-05-06
docs: fe_sendauth issue can require IPv6 trust

nealmcb:
**Description**

When running `make resetdb` on a Chromebook Crostini virtual environment with
the default Debian 10.2 Linux, I ran into the `fe_sendauth: no password supplied`
error noted in the README. But changing to `trust` for just IPv4 in 
`/etc/postgresql/11/main/pg_hba.conf` didn't resolve the issue.
When I changed to `trust` for IPv6 also, the make succeeded.
So I guess connections are being made via IPv6 in some configurations.
Note also that this was with Postgresql version 11, default on Debian.

The troubleshooting tip now also describes one way to reload the postgres configs,
which is also necessary.

**Testing**
Documentation-only change


jonahkagan:
@nealmcb could you rebase your branch to `master` and we'll get this merged in?

nealmcb:
@jonahkagan I've rebased it. Thanks.

jonahkagan:
Looks like we had a setting in our CircleCI config that wasn't running the tests for PRs from forked repositories. Just changed it. Writing this comment to trigger a test run

PR #493, opened 2020-05-06
Fix jurisdiction url proxy patterns

jonahkagan:
**Description**

#491 introduced a bug because the new route it created for JAs `/election/:electionId/jurisdiction/:jurisdictionId` matched the proxy patterns. This updates the proxy patterns to exclude that route.

**Testing**

Manual test

**Progress**

Ready

### DIFF
--- a/arlo-client/src/setupProxy.js
+++ b/arlo-client/src/setupProxy.js
@@ -20,9 +20,9 @@ module.exports = function(app) {
   app.use(proxy('/election/*/sample-sizes', { target }))
   app.use(proxy('/election/*/contest', { target }))
   app.use(proxy('/election/*/round', { target }))
-  app.use(proxy('/election/*/jurisdiction', { target }))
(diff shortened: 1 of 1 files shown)


arlo-client/src/setupProxy.js
```diff
@@ -20,9 +20,9 @@ module.exports = function(app) {
   app.use(proxy('/election/*/sample-sizes', { target }))
   app.use(proxy('/election/*/contest', { target }))
   app.use(proxy('/election/*/round', { target }))
-  app.use(proxy('/election/*/jurisdiction', { target }))
```
MorganLove, line 23: This path is used in `useJurisdictions` for the status box: https://github.com/votingworks/arlo/blob/389526c5dc588da4808c13132574cecc2672de4d/arlo-client/src/components/Audit/useJurisdictions.ts#L49

So this breaks on the AA view.

PR #501, opened 2020-05-07
Fixing the download sheets

MorganLove:
**Description**

- Added in the invisible QR code element so it could be grabbed and injected into the sheet.
- Corrected the expected return from `/ballot-draws` to have `{ ballotDraws: [] }` instead of `{ ballots: [] }` so the other sheets would generate.

**Testing**

- none yet

**Progress**

- It works for me.

### DIFF
--- a/arlo-client/src/components/Audit/RoundManagement/QRs.tsx
+++ b/arlo-client/src/components/Audit/RoundManagement/QRs.tsx
@@ -1,0 +1,30 @@
+import React from 'react'
+import QRCode from 'qrcode.react'
+import styled from 'styled-components'
+
+const QRroot = styled.div`
+  display: none;
+`
+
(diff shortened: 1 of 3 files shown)


arlo-client/src/components/Audit/RoundManagement/QRs.tsx
```diff
@@ -0,0 +1,30 @@
...
+}: {
+  electionId: string
+  boardIds: string[]
+}) => {
+  return (
+    <QRroot id="qr-root">
+      {boardIds.map(id => (
+        <span key={id} id={`qr-${id}`}>
+          <QRCode
+            value={`${window.location.origin}/election/${electionId}/audit-board/${id}`}
```
jonahkagan, line 21: this should link to `/auditboard/${auditBoard.passphrase}` because hitting that url will initiate the audit board's session (which is needed to authorize requests to the audit board API)

PR #507, opened 2020-05-08
Confirm dialog and redirect for Review and Launch stage

MorganLove:
**Description**

- Added a confirmation dialog for the review and launch stage
- Made launching the audit redirect you to the progress view

**Testing**

- none

**Progress**

- works with manual testing

### DIFF
--- a/arlo-client/src/components/Audit/Setup/Review/index.tsx
+++ b/arlo-client/src/components/Audit/Setup/Review/index.tsx
@@ -20,7 +20,7 @@ import ContestsTable from './ContestsTable'
 import SettingsTable from './SettingsTable'
 import { isSetupComplete } from '../../StatusBox'
 import useJurisdictionFile from '../Participants/useJurisdictionFile'
-import useConfirmLaunch from './useConfirmLaunch'
+import ConfirmLaunch from './ConfirmLaunch'
 
 const percentFormatter = new Intl.NumberFormat(undefined, {
   style: 'percent',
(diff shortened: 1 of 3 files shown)


arlo-client/src/components/Audit/Setup/Review/index.tsx
```diff
@@ -108,12 +88,38 @@ const Review: React.FC<IProps> = ({ prevStage, locked, refresh }: IProps) => {
...
       }
       refresh()
+      history.push(`/election/${electionId}/progress`)
     } catch (err) /* istanbul ignore next */ {
       // TEST TODO
       toast.error(err.message)
     }
   }
 
+  const [openConfirmDialog, confirmDialog] = useConfirmLaunch(submit)
```
jonahkagan, line 98: I don't think this needs to be a hook to be abstracted out. I think a more common pattern would be to have the state controlled in this component:

```ts
const [isConfirmDialogOpen, setConfirmDialogOpen] = useState<boolean>(false)
```
and then rework the `useConfirmLaunch` hook into a component `ConfirmLaunchDialog` that takes props `isOpen: bool`, `onClose: () => void`, and `onLaunch: () => void`.

So you would render:

```ts
<ConfirmLaunchDialog isOpen={isConfirmDialogOpen} onLaunch={submit} onClose={() => setIsConfirmDialogOpen(false)} />
```

MorganLove: @jonahkagan True, that. I think I honestly accidentally made it into a hook, haha.

PR #512, opened 2020-05-08
Signoff and data entry completion styling fix

MorganLove:
**Description**

- Added a wrapper div inside the pre-existing wrapper div so that the flex doesn't make things go sideways.

**Testing**

- None.

**Progress**

- It looks right to me, and I only edited those two components, but @jonahkagan should check it since he wrote this part of the UI.


MorganLove:
@jonahkagan Hmm.. I was wondering that too. I'm honestly not that sure how many are vertical vs horizontal now, though. It started as all vertical, then horizontals started getting added as the exception. And then they became the most common one so verticals started needing the exceptions. I'm not sure if we've swung all the way back around yet. I'd say we take a look at it in a style refactor later.

PR #523, opened 2020-05-09
Jurisdiction detail dialog

MorganLove:
**Description**

- Adding 'minimal' option to FormButton
- Adding a dialog to the Progress screen that shows details and download link for jurisdictions

**Testing**

- Manual

**Progress**

LGTM! Very pretty

### DIFF
--- a/arlo-client/src/components/Audit/Progress/JurisdictionDetail.tsx
+++ b/arlo-client/src/components/Audit/Progress/JurisdictionDetail.tsx
@@ -1,0 +1,63 @@
+import React from 'react'
+import { Classes, Dialog } from '@blueprintjs/core'
+import FormButton from '../../Atoms/Form/FormButton'
+import { IJurisdiction } from '../useJurisdictions'
+
+export interface IDialogExampleState {
+  autoFocus: boolean
+  canEscapeKeyClose: boolean
+  canOutsideClickClose: boolean
+  enforceFocus: boolean
+  isOpen: boolean
+  usePortal: boolean
+}
+
+const JurisdictionDetail = ({
+  isOpen,
+  handleClose,
+  jurisdiction,
+  electionId,
+}: {
+  isOpen: boolean
+  handleClose: () => void
+  jurisdiction: IJurisdiction
+  electionId: string
+}) => {
+  if (!jurisdiction) return null // component still loading data
+  return (
+    <Dialog
+      icon="info-sign"
+      onClose={handleClose}
+      title={`${jurisdiction.name} Audit Information`}
+      isOpen={isOpen}
+    >
+      <div className={Classes.DIALOG_BODY}>
+        <p>
+          <strong>Ballot Manifest Upload Status: </strong>
+          {jurisdiction.ballotManifest.processing
+            ? jurisdiction.ballotManifest.processing.status
+            : 'N/A'}
+        </p>
+        <p>
+          <strong>Current Ballot Manifest File: </strong>
+          <a
+            href={`/election/${electionId}/jurisdiction/${jurisdiction.id}/ballot-manifest/csv`}
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            {jurisdiction.ballotManifest.file
+              ? jurisdiction.ballotManifest.file.name
+              : 'N/A'}
+          </a>
+        </p>
+      </div>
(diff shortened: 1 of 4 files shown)


arlo-client/src/components/Audit/Progress/JurisdictionDetail.tsx
```diff
@@ -0,0 +1,63 @@
...
+  handleClose,
+  jurisdiction,
+  electionId,
+}: {
+  isOpen: boolean
+  handleClose: () => void
+  jurisdiction: IJurisdiction
+  electionId: string
+}) => {
+  if (!jurisdiction) return null // component still loading data
```
jonahkagan, line 26: The type of `jurisdiction` says it won't be falsy, so I think this is unnecessary

arlo-client/src/components/Audit/Progress/JurisdictionDetail.tsx
```diff
@@ -0,0 +1,63 @@
...
+      onClose={handleClose}
+      title={`${jurisdiction.name} Audit Information`}
+      isOpen={isOpen}
+    >
+      <div className={Classes.DIALOG_BODY}>
+        <p>
+          <strong>Ballot Manifest Upload Status: </strong>
+          {jurisdiction.ballotManifest.processing
+            ? jurisdiction.ballotManifest.processing.status
+            : 'N/A'}
```
jonahkagan, line 39: I think this could be improved in a couple ways to be more useful to the AA:
- Prettify the statuses (maybe factor this bit out into a function and reuse it: https://github.com/votingworks/arlo/blob/jurisdiction-detail/arlo-client/src/components/Audit/Progress/index.tsx#L53-L63)
- In the error case, show the error message in `processing.error`

arlo-client/src/components/Audit/Progress/index.tsx
```diff
@@ -20,13 +22,28 @@ interface IProps {
 const Progress: React.FC<IProps> = ({ refreshId }: IProps) => {
   const { electionId } = useParams<{ electionId: string }>()
   const jurisdictions = useJurisdictions(electionId, refreshId)
+  const [isDetailOpen, setIsDetailOpen] = useState(false)
+  const [jurisdictionDetail, setJurisdictionDetail] = useState(jurisdictions[0])
```
jonahkagan, line 26: If you wanted, you could combine these into one state value of type `IJurisdiction | null` which would either be the jurisdiction to show the detail for or `null` when the detail modal is closed.

arlo-client/src/components/Audit/Progress/JurisdictionDetail.tsx
MorganLove: @jonahkagan I put it in because it errored on initial load. I think it's because I'm setting the initial value of the state hook to the 0th jurisdiction, which is undefined before `useJurisdictions` gets them, so it errors here. This looks to me like the most elegant solution, though I'm open to suggestions.

arlo-client/src/components/Audit/Progress/index.tsx
MorganLove: @jonahkagan :thinking: That sounds neat. I'll try that. I think it could fix your first comment as well.

arlo-client/src/components/Audit/useJurisdictions.ts
```diff
@@ -40,6 +40,18 @@ export interface IJurisdiction {
   } | null
 }
 
+export const prettifyStatus = (processing: IFileInfo['processing']): string => {
+  if (!processing) return 'No manifest uploaded'
+  switch (processing.status) {
+    case FileProcessingStatus.ERRORED:
+      return `Manifest upload failed: ${processing.error}`
```
jonahkagan, line 47: Have you tried this case out? I'm curious what the table looks like if the error is long. Does it render well?

MorganLove: @jonahkagan Yup, I threw in a massive long string, and the dialog shows it beautifully. In the table it simply truncates it with an ellipses at the end, so it works nicely there as well (since the same error will be in the details in the dialog box, so there's no need to dump the whole thing visibly in the table).

jonahkagan: Awesome

PR #524, opened 2020-05-09
Button to clear ballot manifest

jonahkagan:
**Description**

This one was a bit tricky. Explanation in the code.

**Testing**

Manual testing.

**Progress**

Ready

### DIFF
--- a/arlo-client/src/components/Audit/useBallotManifest.ts
+++ b/arlo-client/src/components/Audit/useBallotManifest.ts
@@ -39,10 +39,29 @@ const putBallotManifestFile = async (
   }
 }
 
+const deleteBallotManifestFile = async (
+  electionId: string,
+  jurisdictionId: string
+): Promise<boolean> => {
+  try {
+    return await api(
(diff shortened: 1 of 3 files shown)


arlo-client/src/components/Audit/useBallotManifest.ts
```diff
@@ -59,11 +78,23 @@ const useBallotManifest = (
...
   useEffect(() => {
     const isFinishedProcessing = (manifest: IFileInfo) =>
       !!(manifest.processing && manifest.processing.completedAt)
 
-    if (!ballotManifest || isFinishedProcessing(ballotManifest)) return
+    if (
+      !(ballotManifest && ballotManifest.file) ||
+      isFinishedProcessing(ballotManifest)
+    )
+      return
```
jonahkagan, line 97: This change fixes a bug where we were initiating polling after loading the initial ballot manifest state from the backend when there was no ballot manifest uploaded. That state looks like `{file: null, processing: null}`

PR #526, opened 2020-05-12
Refactor multi vs single

MorganLove:
**Description**

- Restructured the file system some

**Testing**

- Updated some of the tests to reflect recent changes

**Progress**

- We should do manual testing to make sure the refactor didn't break anything.

Note: This is mainly focused on prepping for getting the review branch back to 100% passing.


arlo-client/src/__snapshots__/App.test.tsx.snap
```diff
@@ -40,71 +40,6 @@ exports[`App renders properly 1`] = `
...
-        class="bp3-button bp3-fill bp3-large bp3-intent-primary sc-hqyNC sUzDH sc-bdVaJa gLVjYw"
-        type="button"
-      >
-        <span
-          class="bp3-button-text"
-        >
-          Log in as a Jurisdiction Admin
-        </span>
-      </button>
-    </div>
```
jonahkagan, line 107: This snapshot change looks suspicious. Do we not expect any of this to render anymore?

arlo-client/src/components/CreateAudit.test.tsx
```diff
@@ -50,7 +50,8 @@ describe('CreateAudit', () => {
     })
   })
 
-  it('calls the /election/new endpoint for nonauthenticated user', async () => {
+  it.skip('calls the /election/new endpoint for nonauthenticated user', async () => {
+    // we have hidden the create new audit feature for unauthenticated users currently
```
jonahkagan, line 54: actually, I just moved it to it's own path: `/audit`

arlo-client/src/components/MultiJurisdictionAudit/Setup/__snapshots__/index.test.tsx.snap
```diff
@@ -1669,1257 +1669,11 @@ exports[`Setup renders Opportunistic Contests stage with processing next stage 1
...
-            pathLength="280"
-            stroke-dasharray="280 280"
-            stroke-dashoffset="210"
-          />
-        </svg>
-      </div>
-    </div>
-  </form>
-</div>
-`;
```
jonahkagan, line 2922: all of these changes are also suspicious

arlo-client/src/components/__snapshots__/CreateAudit.test.tsx.snap
```diff
@@ -152,42 +110,6 @@ exports[`CreateAudit renders correctly 1`] = `
...
-    <button
-      class="bp3-button bp3-fill bp3-large bp3-intent-primary sc-iwsKbI kljBkO sc-bwzfXH eBfcXh"
-      type="button"
-    >
-      <span
-        class="bp3-button-text"
-      >
-        Create a New Audit
-      </span>
-    </button>
```
jonahkagan, line 190: these changes are also suspicious (just given the names of the tests)

arlo-client/src/components/CreateAudit.test.tsx
MorganLove: @jonahkagan Se we should be testing that path separately, right? Or do you mean I should change the comment to be more clear?

arlo-client/src/__snapshots__/App.test.tsx.snap
MorganLove: @jonahkagan It does look a bit suspicious. The create new audit button isn't supposed to be there, but the log in ones should still be there. I'm not super concerned about it yet, though. This PR is mainly just file reorganization with a few simple test fixes. I'm planning another one for updating more of the api stuff, and a few others. The snapshots will likely not get completely fixed until last, once everything else is done moving around. These updates are mostly due to me moving tests around in different files, with some side effects, I think.

arlo-client/src/components/MultiJurisdictionAudit/Setup/__snapshots__/index.test.tsx.snap
MorganLove: @jonahkagan See above. I'm kind of letting the snapshots fall where they fall right now, and then going back in later with a separate PR to fix them, if that makes sense.

arlo-client/src/components/CreateAudit.test.tsx
MorganLove: Actually, just saw that it's a completely different component that's on that path, so I updated the comment here accordingly.

PR #527, opened 2020-05-13
Multi-jurisdiction audit reports (AA + JA)

jonahkagan:
**Description**

New endpoints for the audit reports for AAs and JAs.

**Testing**

Added very basic snapshot tests.

**Progress**

Ready

### DIFF
--- a/arlo_server/reports.py
+++ b/arlo_server/reports.py
@@ -1,4 +1,4 @@
-import io, csv, json
+import io, csv
 from typing import Dict, List
 
 from arlo_server import app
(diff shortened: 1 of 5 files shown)


arlo_server/reports.py
```diff
@@ -0,0 +1,294 @@
...
+    }
+    return mapping.get(affiliation, "")
+
+
+def pretty_boolean(boolean: bool) -> str:
+    return "Yes" if boolean else "No"
+
+
+def pretty_targeted(is_targeted: bool) -> str:
+    return "Targeted" if is_targeted else "Opportunistic"
```
benadida, line 40: this feels either already reusable or soon to be. I can live with not refactoring it out but just a nudge in case it's easy.

jonahkagan: what do you mean "refactoring it out"? like putting it in its own module?

benadida: yeah, I mean putting it somewhere where it can be reused, but again, if you see no use case for that, leaving it as is is fine by me.

PR #530, opened 2020-05-14
Marking a ballot as "not found"

benadida:


arlo_server/routes.py
```diff
@@ -878,20 +882,37 @@ def ballot_set(election_id, jurisdiction_id, batch_id, ballot_position):
...
+    attributes = request.get_json()
+    ballot = get_ballot(election_id, jurisdiction_id, batch_id, ballot_position)
+
+    if not ballot:
         return (
             jsonify(
                 errors=[
                     {
-                        "message": f"Multiple ballots found with election_id={election_id}, jurisdiction_id={jurisdiction_id}, batch_id={batch_id}, ballot_position={ballot_position}",
-                        "errorType": "BadRequest",
```
benadida, line 887: This error can't occur given the database constraint.

arlo_server/routes.py
```diff
@@ -878,20 +882,37 @@ def ballot_set(election_id, jurisdiction_id, batch_id, ballot_position):
             ),
             404,
         )
-    elif len(ballots) > 1:
+
+    # explicitly remove existing interpretations in case this ballot was previously set.
+    ballot.interpretations = []
+    ballot.status = BallotStatus.NOT_FOUND
```
jonahkagan, line 888: I was imagining doing this within the same endpoint so we don't have to keep things consistent between diff endpoints. E.g. the ballot set endpoint would take a body with this type:

```
{
  status: BallotStatus
  interpretations: List[Interpretation]
}
```

Any reason you opted for a separate endpoint?

arlo-client/src/components/DataEntry/index.tsx
```diff
@@ -212,16 +230,25 @@ const DataEntry: React.FC<IProps> = ({
   }
 
   const submitBallot = (batchId: string, ballotPosition: number) => async (
-    interpretations: IBallotInterpretation[]
+    interpretations?: IBallotInterpretation[]
```
jonahkagan, line 233: this feels slightly tricky to me because there is a big difference in meaning between `interpretations` being set to `[]` vs `null` or `undefined`. is there a clearer interface?

benadida: in this case I wanted to reduce internal API complexity so felt this was the simplest smallest change.

arlo_server/routes.py
benadida: In this case I chose a separate endpoint because I am biased towards having frontend <--> backend APIs be the ones that can't change as often (as internal frontend ones like the other comment), and so wanted these two cases to be very different. Could do it your way but then that would change a lot of existing behavior and that felt riskier with not a ton of benefit (I don't see many other statuses being added.)

MorganLove:
@benadida Can you rebase this on `review` again so that I can review it?

PR #537, opened 2020-05-14
Show JA number of ballots to audit when creating audit boards

jonahkagan:
**Description**

Title says it all

Also fixed a bug that I introduced in #532 where we didn't actually load the ballots correctly.

**Testing**

Manually tested

**Progress**

Ready

### DIFF
--- a/arlo-client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
+++ b/arlo-client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
@@ -68,6 +68,8 @@ const RoundManagement = ({ round, auditBoards, createAuditBoards }: IProps) => {
         <CreateAuditBoards
           auditBoards={auditBoards}
           createAuditBoards={createAuditBoards}
+          numBallots={ballots.length}
+          roundNum={roundNum}
         />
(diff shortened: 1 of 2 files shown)


arlo-client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
```diff
@@ -31,17 +31,17 @@ const RoundManagement = ({ round, auditBoards, createAuditBoards }: IProps) => {
     jurisdictionId: string
   }>()
 
-  const [ballots, setBallots] = useState<IBallot[]>([])
+  const [ballots, setBallots] = useState<IBallot[] | null>(null)
```
benadida, line 34: is this `null` alternative necessary? Are we differentiating between an empty ballot list and `null`? Do we need to?

jonahkagan: `null` means we're loading the data. it's a pattern we've started to establish all across the frontend. empty lists might have diff meanings in diff places, so i think it's helpful to be consistent. (e.g. does an empty ballot list mean this jurisdiction wasn't assigned any ballots? or does it mean we're loading the data still?)

PR #539, opened 2020-05-15
Default to progress screen for AAs when audit setup complete

jonahkagan:
**Description**

In order to make this change, we hoist data loading up to the top level AuditAdminView component. Note that we don't pass data down into the `Setup` component because it has it's own way of doing things and I don't wanna mess with it in this PR.

**Testing**

Manually tested

**Progress**

Ready


jonahkagan:
@benadida I am not totally sure I understand your feedback, but your point about navigation does sound reasonable. I'm curious what you would recommend for this case.

benadida:
@jonahkagan likely worth a live discussion!

In general, I'm thinking that `/election/<election_id>` should be the URL for an election for audit-admins, no matter what state it's in. The URL should change only based on the screen you want to show. Ideally there would be a diff URL for the different options, e.g. "opportunitistic contests" woud have its own URL.

PR #545, opened 2020-05-16
Tweak wording

benadida:


benadida:
forgive me @MorganLove and @jonahkagan must merge this so we can test the fix ASAP.

jonahkagan:
My only real feedback on this is that I wonder if we should require the AB to select some option for every contest for every ballot. Right now they are only required to select an option for at least one contest per ballot (unless they select Not Found). It seems like with this label change it covers all of the cases, so we don't need to let them leave a contest blank (which previously meant "Contest not on ballot" in my understanding)

PR #547, opened 2020-05-17
clone election as superadmin

benadida:


arlo_server/superadmin.py
```diff
@@ -46,3 +50,238 @@ def superadmin_jurisdictionadmin_login():
...
+                processing_error=old_file.processing_error,
+            )
+            db.session.add(new_file)
+
+            new_jurisdiction.manifest_file_id = new_file.id
+
+        db.session.add(new_jurisdiction)
+
+        for old_ja in old_jurisdiction.jurisdiction_administrations:
+            new_user = clone_user_with_prefix(old_ja.user, "superadmin--")
```
jonahkagan, line 213: how will they be able to log in with this made up email address?

jonahkagan, line 213: how will they be able to log in with this made up email address?

arlo_server/superadmin.py
```diff
@@ -46,3 +50,238 @@ def superadmin_jurisdictionadmin_login():
...
+    if not new_user:
+        new_user = User(id=new_id(), email=new_email)
+        db.session.add(new_user)
+
+    return new_user
+
+
+@app.route("/superadmin/election-clone", methods=["POST"])
+@with_superadmin_access
+def superadmin_election_clone():
```
jonahkagan, line 76: i am worried that we will fail to keep this up to date when the data model changes

jonahkagan: maybe a simpler approach would be to add `election_id` to every table

arlo_server/superadmin.py
benadida: yeah this was for testing purposes, and I think it's worth a more thorough review of why we want this before we merge. In particular, I think prefixing email addresses should be an option, not a forced thing.

arlo_server/superadmin.py
benadida: for sure this will fail when data model changes. It needs at least a test that tries to do this and checks a few things... maybe the serialization of the election at least?

Could add `election_id` to every table, but that will make for some composite keys all over the place unless you want a potentially inconsistent data model. I don't mind composite keys but I hear you like them less ;)

jonahkagan:
@benadida do we need this PR or can we close?

PR #548, opened 2020-05-18
Update lint-staged to avoid file lossage

### LINKED ISSUE #543: Upgrade lint-staged so it doesn't delete your files
As discussed in https://github.com/okonet/lint-staged/issues/866 and https://github.com/okonet/lint-staged/issues/826, it appears that the currently-pinned version of lint-staged (10.0.8) has a tendency to lose work under some circumstances.
These tools are new to me, but it appears that they have changed their approach in more recent versions, which are probably worth looking into.

nealmcb:
**Description**
Before I figured out what was going on, on three different occasions, when I was just trying to `git commit` some files, the `lint-staged` pre-commit hook had an error and **threw away most files in an untracked subdirectory**.  Damn....
 
Fixes #543.

See also [Files lost after EISDIR: illegal operation on a directory, unlink - in version 10.0.8](https://github.com/okonet/lint-staged/issues/866)

**Testing**
I don't know of any git-related tests in this repo, but `git commit` stopped throwing away most of the files in my untracked subdirectory after I upgraded to lint-staged 10.2.2

**Progress**

I'm not sure what is needed to put this into effect, but I naively just manually did `npm up lint-staged`.  I was a bit surprised that it didn't seem to happen automagically, but I'm an npm newbie....

To get the current version, this seems to work: `./node_modules/.bin/lint-staged  --version`

I don't recall seeing any unwelcome changes in the release notes for lint-staged, but I encourage others who know more to check it out and decide if this is the right version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   },
   "devDependencies": {
     "husky": "^4.2.3",
-    "lint-staged": "^10.0.8",
+    "lint-staged": "^10.2.2",
     "prettier": "^1.19.1",
     "sort-package-json": "^1.40.0"
   }
--- a/yarn.lock
+++ b/yarn.lock
@@ -87,25 +87,30 @@
   resolved "https://registry.yarnpkg.com/@types/parse-json/-/parse-json-4.0.0.tgz#2f8bb441434d163b35fb8ffdccd7138927ffb8c0"
   integrity sha512-//oorEZjL6sbPcKUaCdIGlIUeH26mgzimjBB77G6XRgnDl/L5wOnpyBGRe/Mmf5CVW3PwEBE1NjiMZ/ssFh4wA==
 
-ansi-escapes@^3.0.0:
-  version "3.2.0"
-  resolved "https://registry.yarnpkg.com/ansi-escapes/-/ansi-escapes-3.2.0.tgz#8780b98ff9dbf5638152d1f1fe5c1d7b4442976b"
-  integrity sha512-cBhpre4ma+U0T1oM5fXg7Dy1Jw7zzwv7lt/GoCpr+hDQJoYnKVPLL4dCvSEFMmQurOQvSrwT7SL/DAlhBI97RQ==
+aggregate-error@^3.0.0:
+  version "3.0.1"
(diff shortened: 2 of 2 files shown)


eventualbuddha:
Yikes, that's frustrating. To properly upgrade the package, you need to run `yarn` in the project root. You can find instructions for installing `yarn` here: https://yarnpkg.com.

nealmcb:
Great - thanks. `yarn` ran smoothly and didn't seem to need to do anything.

jonahkagan:
Thanks for sending this @nealmcb. I tried checking out your branch and then ran `yarn` and it also updated `yarn.lock`. Did that not happen for you? Maybe that's what @eventualbuddha is getting at - that we need the updated lockfile in the PR as well. If it doesn't work, maybe one of us can try to add it.

nealmcb:
Good point - I didn't look for a diff after running yarn. Here is a diff:
 https://gist.github.com/nealmcb/675bfa8ea71f7ecace90014926631aec 

As you can see there are lots of differences. I'm happy to push it if that would help, but it made lots of updates, which makes me worried that other things might change and break things.

jonahkagan:
Yeah, go ahead and push it. I think they should be safe based on the versioning logic. They are all dev dependencies so worst case it just screws up the dev environment.

jonahkagan:
Also this branch needs a rebase

nealmcb:
OK, added yarn.lock and rebased on top of #492

jonahkagan:
Just merged #492 - looks like you'll need to rebase again against `master`

nealmcb:
Huh - looks like master at f05859c is attributed to GitHub and for some reason differs from my previous tip of fe_sendauth2, which is becc4e2. Perhaps it was merged instead of rebased or something. And even so, unclear why it would complain if it is going to do that sort of rewriting of the commit anyway.

At any rate, should be good now.

jonahkagan:
@nealmcb hate to do this to you but when I merged #492 it broke tests on `master` because Google Chrome released a new version (which broke our end to end tests). I've fixed that now but you'll need to rebase this branch again (hopefully the last time).

nealmcb:
I see the build-and-test failed, but can't see why without giving cirvleci full rights to impersonate me in too many places.

PR #551, opened 2020-05-18
resetdb.py: add --skip-table-creation flag

jonahkagan:
**Description**

When reloading a db snapshot exported with `pgdump`, you need to first reset the db but not recreate the tables. This is useful for quickly putting the database into a saved state for manual testing.

**Testing**

Ran it

**Progress**

Ready


benadida:
do we need this as part of the python script? It feels like

```
dropdb
createdb arlo
```

is fine in dev, and on Heroku it's different anyways. No?

jonahkagan:
Ok I can just write my own script to do it :)

PR #554, opened 2020-05-18
Toggle to show sample vs ballot counts on jurisdiction progress screen

jonahkagan:
JA and AB users only know about the ballot counts. AA users get to know about both the ballot counts and the sample counts.

This PR adds a toggle to the jurisdiction progress view to let the AA user switch which count they want to see.

Also updates the API to return both counts (previously just returned sample counts).

ezgif com-video-to-gif

### DIFF
--- a/server/api/jurisdictions.py
+++ b/server/api/jurisdictions.py
@@ -55,14 +55,38 @@ def round_status_by_jurisdiction(
         .values(AuditBoard.jurisdiction_id, func.count())
     )
 
-    sampled_ballot_count_by_jurisdiction = dict(
+    ballots_in_round = (
+        SampledBallot.query.join(SampledBallotDraw)
+        .filter_by(round_id=round.id)
+        .distinct(SampledBallot.id)
+        .subquery()
+    )
+    ballot_count_by_jurisdiction = dict(
+        SampledBallot.query.join(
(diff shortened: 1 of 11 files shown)


benadida:
@jonahkagan I imagine this PR got left behind, sorry about that. Can you rebase / bring it back to goodness and I will gladly review?

jonahkagan:
I will do that, but also will probably want to write tests now that we're back on `master` so it will take some time

jonahkagan:
@benadida tagging you in cuz there are backend changes

server/api/jurisdictions.py
```diff
@@ -55,14 +55,38 @@ def round_status_by_jurisdiction(
         .values(AuditBoard.jurisdiction_id, func.count())
     )
 
-    sampled_ballot_count_by_jurisdiction = dict(
+    ballots_in_round = (
+        SampledBallot.query.join(SampledBallotDraw)
+        .filter_by(round_id=round.id)
+        .distinct(SampledBallot.id)
+        .subquery()
```
benadida, line 62: you could probably do this with a `count(distinct sampled_ballot_id)` and just query `SampledBallotDraw`. Not critical, perf might be identical.

PR #558, opened 2020-05-19
Detect chromedriver version based on Chrome version

jonahkagan:
**Description**

This way we don't have to update the chromedriver version in
package.json every time Chrome releases a new version.

See https://www.npmjs.com/package/chromedriver#detect-chromedriver-version


jonahkagan:
Didn't quite figure out a solution that would work here, so closing for now. But at least the ideas I tried are recorded here

PR #564, opened 2020-05-20
Remove unnecessary generators from test fixtures

jonahkagan:
**Description**

For some reason, we all thought that fixtures had to yield their
results. But really they can return results - yielding is only necessary
if you want to do some teardown after the test runs.


umbernhard:
The only thing I’d say about this is we may need the generators if we need tear down code. But we can always add them back in later if needed.

PR #565, opened 2020-05-20
Fix test warnings

jonahkagan:
A number of warnings would show at the end of any test runs. This PR has a series of commits that fixes each warning.

### DIFF
--- a/tests/routes_tests/test_auth.py
+++ b/tests/routes_tests/test_auth.py
@@ -91,7 +91,7 @@ def check_redirect_contains_redirect_uri(response, expected_url):
     # which makes a double slash like 'http://localhost//authorize'
     # which won't work. So testing to make sure there is no '//'
     # other than '://'
-    assert re.search("[^:]\/\/", redirect_uri) is None
+    assert re.search("[^:]//", redirect_uri) is None
     assert expected_url in redirect_uri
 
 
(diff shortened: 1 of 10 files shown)


tests/routes_tests/test_auth.py
```diff
@@ -91,7 +91,7 @@ def check_redirect_contains_redirect_uri(response, expected_url):
     # which makes a double slash like 'http://localhost//authorize'
     # which won't work. So testing to make sure there is no '//'
     # other than '://'
-    assert re.search("[^:]\/\/", redirect_uri) is None
+    assert re.search("[^:]//", redirect_uri) is None
```
benadida, line 94: are we sure about this change? This doesn't turn this into a regexp matching?

jonahkagan: Seems to have the same behavior? But I'm not sure exactly what you mean by "turn this into a regexp matching"

```python
>>> import re
>>> re.search("[^:]//", "localhost://test")
>>> print(re.search("[^:]//", "localhost://test"))
None
>>> print(re.search("[^:]\/\/", "localhost://test"))
None
>>> print(re.search("[^:]//", "localhost://test//"))
<re.Match object; span=(15, 18), match='t//'>
>>> print(re.search("[^:]\/\/", "localhost://test//"))
<re.Match object; span=(15, 18), match='t//'>

```

PR #566, opened 2020-05-20
Cleanup db models

jonahkagan:
See commits for changes and reasoning.

### DIFF
--- a/arlo_server/models.py
+++ b/arlo_server/models.py
@@ -55,7 +55,7 @@ class Election(BaseModel):
         db.String(200), db.ForeignKey("organization.id", ondelete="cascade"),
     )
 
-    frozen_at = db.Column(db.DateTime(timezone=False))
+    frozen_at = db.Column(db.DateTime)
 
     jurisdictions = relationship(
         "Jurisdiction",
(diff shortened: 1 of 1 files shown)


arlo_server/models.py
```diff
@@ -35,13 +35,13 @@ class Election(BaseModel):
     # audit_name must be unique within each Organization
     audit_name = db.Column(db.String(200), nullable=False)
     # election_name can be the same across audits
-    election_name = db.Column(db.String(200), nullable=True)
-    state = db.Column(db.String(100), nullable=True)
-    election_date = db.Column(db.Date, nullable=True)
-    election_type = db.Column(db.String(200), nullable=True)
-    meeting_date = db.Column(db.Date, nullable=True)
-    risk_limit = db.Column(db.Integer, nullable=True)
-    random_seed = db.Column(db.String(100), nullable=True)
```
benadida, line 44: I'm not a fan of assuming the defaults on data models because one has to remember the default. I would suggest explicit is better. If you prefer relying on the default, I can live with it.

jonahkagan: Since we don't have a way to require that you include the `nullable` argument, I thought adding it in the cases where it's not the default actually created more confusion and thus more possibility for bugs.

E.g. I had the experience reading this file where I looked at a column that didn't specify it one way or the other and had no idea what the default was (so I had to go look it up).

With this approach, you can see by looking around that only `nullable=False` is ever set, so you can easily infer that it defaults to `True`.

jonahkagan: In general I agree with you and prefer explicitness.

jonahkagan: (A similar thing happened to me with the timezone argument - some columns had `db.DateTime(timezone=False)` and some had `db.DateTime`, so I inferred that the default was `timezone=True`, which is not the case!)

PR #567, opened 2020-05-20
Run pipenv directly (no more `python3.7 -m pipenv`)

jonahkagan:
`pipenv` knows what version of Python to run due to this line in Pipfile:

```
[requires]
python_version = "3.7"
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,3 @@
-PIPENV=python3.7 -m pipenv
-
 deps:
 	sudo apt install python3.7 python3-pip nodejs libpython3.7-dev libpq-dev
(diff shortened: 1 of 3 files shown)


eventualbuddha:
This seems to work, but I had to `pip install pipenv` and then add `export PATH="${PATH}:${HOME}/.local/bin"` to my shell config.

jonahkagan:
@eventualbuddha that's great! Did you use this line from the Makefile by chance? `python3.7 -m pip install pipenv`. That's what will be used to install pipenv automatically, so I'm wondering if you still need the `PATH` update in that case. If so, we might want to document in the README

PR #568, opened 2020-05-21
Lint more

jonahkagan:
Resets .pylintrc to pylint's default settings. Then disables a number of rules that don't serve us, keeping the rest. Also adds the pylint_flask_sqlalchemy in order to correctly lint the db models.

Lastly, fixes all the existing lint issues!


jonahkagan:
@eventualbuddha your dream has come true

PR #572, opened 2020-05-21
Add endpoint for audit board to audit ballot

jonahkagan:
This endpoint replaces the ballot_set/ballot_set_not_found endpoints
with a new endpoint that handles all of the requests that an audit board
needs to make to audit a ballot, remove the audit record for a ballot,
and mark a ballot not found.

Most importantly, this endpoint checks for authorization!

It also updates the frontend to use this new endpoint, and improves some styling on the audit board table view.

### DIFF
--- a/arlo-client/src/components/DataEntry/BoardTable.tsx
+++ b/arlo-client/src/components/DataEntry/BoardTable.tsx
@@ -1,6 +1,6 @@
 import React, { useState, useEffect } from 'react'
 import styled from 'styled-components'
-import { Table, Column, Cell } from '@blueprintjs/table'
+import { Table as BPTable, Column, Cell as BPCell } from '@blueprintjs/table'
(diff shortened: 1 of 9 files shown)


arlo-client/src/components/DataEntry/BoardTable.tsx
```diff
@@ -20,18 +20,21 @@ const RightWrapper = styled.div`
...
+  > div {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    height: 100%;
+  }
 `
 
-const GrowingTable = styled(Table)`
+const Table = styled(BPTable)`
```
jonahkagan, line 34: @MorganLove I was getting confused by all these custom named components that are really just styled versions of the BlueprintJS components. I would look at a component and have to track down the original BlueprintJS name in order to look up the props. So here's an experiment with a technique to let us use the original names. Let me know your thoughts.

PR #575, opened 2020-05-23
Fixes #574

### LINKED ISSUE #574: Sample can return colliding ticket numbers
The consistent sampler trims the least significant digits, defaulting to the 9 most significant digits for ticket numbers. With enough ballots, the odds that two tickets have the same number to 9 decimal places is non-trivial, so we should add more precision to the ticket numbers by invoking the `digits` field in the instantiation.

umbernhard:
Doubles the precision of the digits output by the consistent sampler, to avoid ticket number collisions. Also updates tests.


tests/test_app.py
```diff
@@ -533,6 +533,9 @@ def get_lines_from_retrieval_list(rv):
 
 def get_num_ballots_from_retrieval_list(rv):
     lines = get_lines_from_retrieval_list(rv)
+
+    for line in lines:
+        print(line)
```
jonahkagan, line 538: @umbernhard did you mean to leave these here?

umbernhard: Ope, no I did not.

umbernhard: Fixed in #576

PR #577, opened 2020-05-26
Don't require audit boards with no ballots assigned to sign off

jonahkagan:
Whenever an audit board signs off, we check if the round is complete. This change modifies the check to omit any audit boards that didn't have any ballots assigned to them.

We might also want to change the UI, but this is the basic change needed.

### DIFF
--- a/arlo_server/audit_boards.py
+++ b/arlo_server/audit_boards.py
@@ -254,6 +254,7 @@ def is_round_complete(election: Election, round: Round) -> bool:
     num_audit_boards_with_ballots_not_signed_off: int = (
         AuditBoard.query.filter_by(round_id=round.id, signed_off_at=None)
         .join(SampledBallot)
+        .group_by(AuditBoard.id)
         .count()
     )
     return (
--- a/arlo_server/audit_boards.py
+++ b/arlo_server/audit_boards.py
@@ -228,7 +228,7 @@ def end_round(election: Election, round: Round):
 
 
 def is_round_complete(election: Election, round: Round) -> bool:
-    num_jurisdictions_without_audit_boards_set_up = (
+    num_jurisdictions_without_audit_boards_set_up: int = (
         # For each jurisdiction...
         Jurisdiction.query.filter_by(election_id=election.id)
         # Where there are ballots that haven't been audited...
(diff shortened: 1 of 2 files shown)


tests/routes_tests/test_audit_boards.py
```diff
@@ -618,11 +618,24 @@ def run_audit_board_flow(jurisdiction_id: str, audit_board_id: str):
...
     rv = client.get(
         f"/election/{election_id}/jurisdiction/{jurisdiction_ids[1]}/round/{round_1_id}/audit-board"
     )
     audit_board = json.loads(rv.data)["auditBoards"][0]
 
+    # Create another audit board that doesn't have any ballots assigned. Even
+    # though this audit board doesn't sign off, it shouldn't stop the round
+    # from being completed.
+    audit_board_without_ballots = AuditBoard(
+        id="audit-board-without-ballots",
```
benadida, line 631: I'm pretty sure this won't work the second time around, since this ID has to be unique throughout the database. I'd just use a UUID here. The name is already good.

arlo_server/audit_boards.py
```diff
@@ -251,11 +251,15 @@ def is_round_complete(election: Election, round: Round) -> bool:
         .having(func.count(AuditBoard.id) == 0)
         .count()
     )
-    all_audit_boards_set_up = num_jurisdictions_without_audit_boards_set_up == 0
-    all_audit_boards_signed_off = all(
-        ab.signed_off_at is not None for ab in round.audit_boards
+    num_audit_boards_with_ballots_not_signed_off: int = (
+        AuditBoard.query.filter_by(round_id=round.id, signed_off_at=None)
+        .join(SampledBallot)
```
benadida, line 256: is this going to yield the right number of audit boards? I think it yields the number of *ballots* across those audit boards, unless SQLAlchemy does something unexpected.

tests/routes_tests/test_audit_boards.py
jonahkagan: Shouldn't matter because it's a test, right? But I'll fix it to make it obvious it's not an issue

benadida: ah yes, you're right, a lot less problematic than I thought. I got a little bit confused about the purpose of this audit board.

Yes, better to fix it I think, just to not make things fail if this runs twice, but not the problem I thought it was :)

arlo_server/audit_boards.py
jonahkagan: Good q. I thought about this one but got lazy. You're right that it will yield the number of ballots.

I'm pretty sure since we're just checking whether or not the count is zero, it's logically equivalent to ask:
- Are there any audit boards with at least one ballot assigned that didn't sign off?
- Are there any ballots assigned to an audit board that didn't sign off?

But it's confusing so I'll just be more precise with the query :)

benadida: 👍

PR #578, opened 2020-05-26
Endpoint to set audit board members

jonahkagan:
Adds a new endpoint that requires audit board authorization to set up
the audit board members (name and party affiliation). Updates the frontend to use this endpoint.

Also creates an Affiliation enum to lock in the values we expect.

### DIFF
--- a/arlo-client/src/components/DataEntry/index.tsx
+++ b/arlo-client/src/components/DataEntry/index.tsx
@@ -59,18 +59,19 @@ const loadBallots = async (
   return ballots
 }
 
-const saveMembers = async (
+const putMembers = async (
   electionId: string,
   jurisdictionId: string,
+  roundId: string,
   auditBoardId: string,
   members: IAuditBoardMember[]
 ) => {
   try {
     await api(
-      `/election/${electionId}/jurisdiction/${jurisdictionId}/audit-board/${auditBoardId}`,
+      `/election/${electionId}/jurisdiction/${jurisdictionId}/round/${roundId}/audit-board/${auditBoardId}`,
       {
-        method: 'POST',
-        body: JSON.stringify({ members }),
+        method: 'PUT',
+        body: JSON.stringify(members),
         headers: {
           'Content-Type': 'application/json',
         },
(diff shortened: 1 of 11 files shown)


arlo-client/src/components/DataEntry/index.tsx
```diff
@@ -59,18 +59,19 @@ const loadBallots = async (
...
 ) => {
   try {
     await api(
-      `/election/${electionId}/jurisdiction/${jurisdictionId}/audit-board/${auditBoardId}`,
+      `/election/${electionId}/jurisdiction/${jurisdictionId}/round/${roundId}/audit-board/${auditBoardId}`,
       {
-        method: 'POST',
-        body: JSON.stringify({ members }),
+        method: 'PUT',
+        body: JSON.stringify(members),
```
benadida, line 74: Did I miss that we're moving to JSON arrays? Not my cup of tea, though not gonna cry about it.

arlo_server/audit_boards.py
```diff
@@ -186,6 +187,70 @@ def list_audit_boards(
...
+    "items": MEMBER_SCHEMA,
+}
+
+
+def validate_members(members: List[JSONDict]):
+    # You can do all of these checks using JSON schema, but the resulting error
+    # messages aren't very good.
+    if len(members) == 0:
+        raise BadRequest("Must have at least one member.")
+    if len(members) > 2:
```
benadida, line 219: It's a bit odd that the check for members length is abstracted here while the code further down that calls into this check also assumes there are 1 or 2 members. Maybe a slight violation of DRY? I wonder if the length check should be in the set_audit_board_members function, in case we ever change it, that's where it belongs.

arlo-client/src/components/DataEntry/index.tsx
jonahkagan: Funny you mention it. I'm fairly convinced that the envelope pattern is unnecessary, and am starting to not use it.

However, in this case I think you're right that I was overzealous. Really this is only updating part of the audit board object, so the body should be `{members: [...]}`. I think technically the method should be `PATCH` also. Do you have thoughts about that?

arlo_server/audit_boards.py
jonahkagan: Hmm, I've been putting validation in its own function on every endpoint. But I am curious to understand your comment more deeply. Let's discuss

jonahkagan:
@benadida I made some of the changes we discussed based on your comments.

PR #581, opened 2020-05-27
Move dev deps to devDependencies

jonahkagan:
Will this break anything? Wasn't totally sure how to test that I got this right.


benadida:
to test:

```
rm -rf node_modules
yarn install
yarn start
```

MorganLove:
@jonahkagan It looks like this is passing CI, which I am curious about, since most of those were moved from devDeps to regular deps because it broke on CI when they weren't included there. It's cool and good that it does work now, so I'm happy with this, but I'm wondering what changed to make it so this works.

jonahkagan:
Ok passes @benadida's manual test.

@MorganLove that is so weird. I have nooo idea what changed. Hope it doesn't break again

PR #582, opened 2020-05-27
ESLint plugins for testing

jonahkagan:
Adds ESLint plugins for testing-library and jest-dom.


arlo-client/src/components/DataEntry/index.test.tsx
```diff
@@ -140,10 +140,10 @@ describe('DataEntry', () => {
       await waitFor(() => {
         expect(apiMock).toBeCalledTimes(3)
         expect(getByText('Audit Board #1: Ballot Cards to Audit')).toBeTruthy()
-        expect(getByText('Start Auditing').closest('a')).toBeTruthy()
+        expect(getByText('Start Auditing').closest('a')).toBeEnabled()
         expect(
           getByText('Auditing Complete - Submit Results').closest('a')
-        ).toHaveAttribute('disabled')
+        ).toHaveAttribute('disabled') // eslint-disable-line jest-dom/prefer-enabled-disabled
```
MorganLove, line 146: @jonahkagan Is `toBeDisabled` possible here?

jonahkagan: It wasn't working for some reason. When I printed out the element, it had `disabled=""`. Maybe the `disabled` property works differently for `a` tags and `toBeDisabled` doesn't support it?

jonahkagan: Ahh, from the Blueprint docs:

> AnchorButton uses the class .bp3-disabled because <a> tags do not support the disabled attribute. As a result, the AnchorButton component will prevent only the onClick handler when disabled but permit other events.

MorganLove: @jonahkagan Good find! That makes sense. Good thing to remember for future reference, haha. Perhaps a comment linking to the docs there?

jonahkagan: Hmm actually these tests don't really test anything if the `a` tag `disabled` property is meaningless. I think we need to test them in a different way - maybe trying to click and confirming whether or not something happens.

jonahkagan: I need to leave this for now because it sucked me into a rabbit hole, but we can revisit another time.

PR #590, opened 2020-05-30
100% Coverage of StatusBox

MorganLove:
**Description**

Increasing test coverage

**Testing**

- Covering all states of the status box and its functions

**Progress**

Done!

Note: Depending on if this or @jonahkagan 's PR with upgrading the libraries gets merged first, one or the other will need to be minorly updated.

### DIFF
--- a/arlo-client/package.json
+++ b/arlo-client/package.json
@@ -63,10 +63,10 @@
     ],
     "coverageThreshold": {
       "global": {
-        "statements": 79,
-        "branches": 78,
-        "lines": 79,
-        "functions": 78
+        "statements": 81,
+        "branches": 81,
+        "lines": 80,
+        "functions": 81
       }
     }
   },
--- a/arlo-client/package.json
+++ b/arlo-client/package.json
@@ -9,8 +9,8 @@
     "start": "react-scripts start",
     "test": "is-ci \"test:ci\" \"test:watch\"",
     "test:ci": "CI=true tsc && react-scripts test --maxWorkers 2 --env=jest-environment-jsdom-sixteen --coverage && yarn test:wdio",
-    "test:coverage": "react-scripts test --coverage --watchAll=false",
-    "test:update": "react-scripts test -u  --watchAll=false",
+    "test:coverage": "react-scripts test --coverage --watchAll=false --env=jest-environment-jsdom-sixteen",
+    "test:update": "react-scripts test -u  --watchAll=false --env=jest-environment-jsdom-sixteen",
     "test:watch": "react-scripts test --env=jest-environment-jsdom-sixteen",
     "test:wdio": "FLASK_ENV=test pipenv run python ../app.py & FLASK_ENV=test pipenv run python ../bgcompute.py & react-scripts start & wdio wdio.conf.js"
   },
--- a/arlo-client/package.json
+++ b/arlo-client/package.json
@@ -63,10 +63,10 @@
     ],
     "coverageThreshold": {
       "global": {
-        "statements": 81,
-        "branches": 81,
+        "statements": 80,
+        "branches": 80,
         "lines": 80,
-        "functions": 81
+        "functions": 80
       }
     }
   },
--- a/arlo-client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
+++ b/arlo-client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
@@ -1,0 +1,55 @@
+import React from 'react'
+import {
+  BrowserRouter as Router,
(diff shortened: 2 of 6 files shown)


arlo-client/package.json
```diff
@@ -63,10 +63,10 @@
...
     "coverageThreshold": {
       "global": {
-        "statements": 79,
-        "branches": 78,
-        "lines": 79,
-        "functions": 78
+        "statements": 81,
+        "branches": 81,
+        "lines": 80,
+        "functions": 81
```
jonahkagan, line 69: 🙌🏼

arlo-client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
```diff
@@ -0,0 +1,370 @@
+import React from 'react'
+import { toast } from 'react-toastify'
+import {
+  BrowserRouter as Router,
+  // Router as RegularRouter,
```
jonahkagan, line 5: remove if not using

arlo-client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
```diff
@@ -0,0 +1,370 @@
...
+            rounds={roundMocks.needAnother}
+            jurisdictions={jurisdictionMocks.twoComplete}
+            contests={contestMocks.filledTargeted.contests}
+            auditSettings={auditSettings.all}
+          />
+        </Router>
+      )
+      fireEvent.click(screen.getByRole('button', { name: 'Start Round 2' }), {
+        bubbles: true,
+      })
```
jonahkagan, line 159: i started using `userEvent` in place of `fireEvent`: https://github.com/testing-library/user-event

arlo-client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
```diff
@@ -0,0 +1,370 @@
...
+      )
+      screen.getByText('When you are ready, start Round 2')
+      screen.getByText('Start Round 2')
+    })
+
+    it('creates the next round', () => {
+      apiMock.mockImplementation(async () => ({
+        message: 'success',
+        ok: true,
+      }))
```
jonahkagan, line 146: the create round endpoint returns this: `{ "status": "ok" }`

arlo-client/src/components/MultiJurisdictionAudit/StatusBox/index.tsx
```diff
@@ -173,7 +173,7 @@ export const AuditAdminStatusBox: React.FC<IAuditAdminProps> = ({
   )
 }
 
-interface IJursidictionAdminProps {
```
jonahkagan, line 176: lol oops

arlo-client/src/components/MultiJurisdictionAudit/_mocks.ts
```diff
@@ -33,4 +41,252 @@ export const auditSettings: {
...
+    | 'empty'
+    | 'singleIncomplete'
+    | 'twoIncomplete'
+    | 'singleComplete'
+    | 'needAnother']: IRound[]
+} = {
+  empty: [],
+  singleIncomplete: [
+    {
+      endedAt: '',
```
jonahkagan, line 55: this should be `null` i believe. in general, I try to avoid using empty strings to represent empty values, so you can expect that the backend will return `null` for empty cases unless there's a good reason to do otherwise

arlo-client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
MorganLove: @jonahkagan would you recommend I install it here and use it, or should we do a separate PR to convert everything over to `userEvent` instead of `fireEvent`?

arlo-client/src/components/MultiJurisdictionAudit/_mocks.ts
jonahkagan: Oooh good point. Yes that should be `endedAt: string | null`

arlo-client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
jonahkagan: Good q. No strong opinion! I think either way would work

MorganLove: I think I'd prefer a separate PR so this one doesn't get too much scope creep, haha.

PR #591, opened 2020-06-02
Fix package.json/tsconfig for Heroku

jonahkagan:
Includes some packages needed for `yarn build` in the prod dependencies.
Updates tsconfig to ignore test files. Creates `tsconfig.wdio.json` to specifically include the types needed for wdio.

Also removes the pesky `noUnusedLocals` setting that prevents us from
running the app locally when there's an unused variable. The linter
should catch this in a less intrusive way.


arlo-client/yarn.lock
```diff
@@ -1851,11 +1851,6 @@
     "@types/history" "*"
     "@types/react" "*"
 
-"@types/react-tap-event-plugin@^0.0.30":
-  version "0.0.30"
-  resolved "https://registry.yarnpkg.com/@types/react-tap-event-plugin/-/react-tap-event-plugin-0.0.30.tgz#123f35080412f489b6770c5a65c159ff96654cb5"
-  integrity sha1-Ej81CAQS9Im2dwxaZcFZ/5ZlTLU=
-
```
jonahkagan, line 1858: this didn't seem necessary

jonahkagan:
I can't figure this out. Closing. Gonna revert #581

jonahkagan:
Oh wait nevermind, it can't be automatically reverted.

jonahkagan:
Ok I think this should work now

arlo-client/tsconfig.json
```diff
@@ -13,15 +13,20 @@
...
+    "jsx": "preserve"
   },
-  "include": ["src"]
+  "include": ["src"],
+  "exclude": [
+    "**/_mocks.ts",
+    "**/*.test.*",
+    "**/testUtilities.ts",
+    "src/test/**"
+  ]
```
MorganLove, line 31: @jonahkagan I just discovered that this breaks intellisense for VSCode in test files. It doesn't recognize `describe` et al until I remove these lines. Is there a way to keep them for development but not have them apply on heroku?

eventualbuddha: https://devcenter.heroku.com/articles/slug-compiler#ignoring-files-with-slugignore

jonahkagan: Ugh the problem that never ends!

PR #593, opened 2020-06-02
Simplify pytest command using pytest.ini

jonahkagan:
Adds a pytest.ini config file so that we don't have to put all the default settings in the Makefile command. This allows running `pipenv run pytest` if you want to add more flags to the command (`make` doesn't allow you to pass in more flags, so we had a hacky way to work around it before).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,21 +41,8 @@ test-client:
 	yarn --cwd arlo-client lint
 	yarn --cwd arlo-client test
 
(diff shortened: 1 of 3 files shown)


jonahkagan:
@umbernhard I took out your test shortcut commands because you should be able to easily use command line autocomplete now - e.g. `pipenv run pytest tests/audit_math_tests/`. I can add it back if you want it, but this way you can also pass flags to `pytest` more easily (dunno if that appeals to you)

PR #596, opened 2020-06-05
Sort/filter progress view

jonahkagan:
* Replaces the BlueprintJS table with react-table
* Adds sorting and filtering to the Jurisdiction Progress view for AAs


arlo-client/src/components/Atoms/Table.tsx
```diff
@@ -0,0 +1,145 @@
...
+            )
+          })}
+        </tbody>
+      </StyledTable>
+    </Wrapper>
+  )
+}
+
+export const sortByRank = <T extends object>(rank: (data: T) => number) =>
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
MorganLove, line 139: @jonahkagan Maybe document here why this is disabled? I'm presuming it's because you need to memoize the function passed to the sortBy feature, and the linter complains because this function is technically not a hook or a functional component?

jonahkagan: Yeah I think you nailed it.

jonahkagan:
Oh good call!

PR #597, opened 2020-06-08
Move ts excludes to .slugignore

jonahkagan:
Allow ts to typecheck test files locally but ignore them when building
on Heroku.


MorganLove:
@jonahkagan This seems to do the trick. I'm not seeing complaints in the IDE, and it successfully deployed so hopefully it isn't complaining on staging either.

PR #598, opened 2020-06-08
Ballot data entry allows multiple selections

MorganLove:
**Description**

- Supports the backend changes to the endpoints by sending an array of choices instead of single strings
- Changes the frontend to use checkboxes instead of radios to support this functionality

**Testing**

- Updating to pass

**Progress**

- Functional and green


arlo-client/src/components/DataEntry/BallotAudit.tsx
```diff
@@ -127,34 +122,48 @@ const BallotAuditContest = ({
...
+            interpretation: value,
+            choiceIds: [],
+          })
+        } else {
+          setInterpretation({
+            ...interpretation,
+            interpretation: null,
+            choiceIds: [],
+          })
+        }
```
jonahkagan, line 151: I got a little confused when I first read this because I couldn't tell what was different between these two cases lol. Maybe that's just me not reading closely.

I think it could be a bit more obvious what's different if you wrote it like this, but not a big deal

```ts
setInterpretation({
  ...interpretation,
  interpretation: checked ? value : null,
  choiceIds: [],
})
```

arlo-client/src/components/DataEntry/BallotAudit.tsx
```diff
@@ -169,32 +178,33 @@ const BallotAuditContest = ({
           () => undefined
         } // required by blueprintjs but we're implementing on BlockRadio instead
         selectedValue={
-          (isVote ? interpretation.choiceId : interpretation.interpretation) ||
-          undefined
+          (isVote
+            ? interpretation.choiceIds.length
```
jonahkagan, line 182: what does this do here? does it make sense to use RadioGroup now that we're using checkboxes?

arlo-client/src/components/DataEntry/BallotAudit.tsx
```diff
@@ -127,34 +122,48 @@ const BallotAuditContest = ({
     })
   }
 
-  // Since the radio button values must be strings, not full objects,
+  // Since the checkbox button values must be strings, not full objects,
   // we condense our data model for interpretations to strings:
   // - For buttons representing interpretation VOTE (one for each
   // contest choice), the value is interpretation.choiceId
```
jonahkagan, line 128: I think this line is slightly out of date now

arlo-client/src/components/DataEntry/BallotAudit.tsx
```diff
@@ -127,34 +122,48 @@ const BallotAuditContest = ({
...
   // - For buttons representing other interpretations (BLANK,
   // CANT_AGREE), the value is interpretation.interpretation
-  const radioProps = {
-    name: `interpretation-${contest.name}`,
+  const checkboxProps = (value: string) => ({
+    name: `interpretation-${contest.name}-${value}`,
```
jonahkagan, line 132: wondering if the `name` should be the same for all checkboxes in a contest or if each one needs its own unique name. not quite sure what `name` is for

arlo-client/src/components/DataEntry/BallotAudit.tsx
```diff
@@ -169,32 +178,33 @@ const BallotAuditContest = ({
...
         }
       >
         {contest.choices.map(c => (
-          <BlockRadio
+          <BlockCheckbox
             key={c.id}
-            {...radioProps}
-            checked={isVote && interpretation.choiceId === c.id}
+            {...checkboxProps(c.id)}
+            checked={isVote && interpretation.choiceIds.some(v => v === c.id)}
```
jonahkagan, line 190: could you do `interpretation.choiceIds.includes(c.id)`?

arlo-client/src/components/DataEntry/BallotReview.tsx
```diff
@@ -19,15 +19,17 @@ const LockedButton = styled(FormButton)`
...
   if (!interpretation) return <div />
   const label = (() => {
     switch (interpretation) {
       case Interpretation.VOTE: {
-        const choice = contest.choices.find(c => c.id === choiceId)
-        return choice!.name
+        const choices = contest.choices
+          .filter(c => choiceIds.some(v => v === c.id))
+          .map(c => c.name)
+        return choices.join(', ')
```
jonahkagan, line 32: for this screen, i imagined that there would be a `LockedButton` for each selected choice. not sure what @mcchilders has in mind

arlo-client/src/components/DataEntry/BallotReview.tsx
```diff
@@ -19,15 +19,17 @@ const LockedButton = styled(FormButton)`
...
       }
       case Interpretation.BLANK:
         return 'Overvote/Blank vote/Not on Ballot'
```
jonahkagan: take out Overvote

arlo-client/src/components/DataEntry/BlockCheckbox.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react'
+import styled from 'styled-components'
+
+const Block = styled.label`
+  &.bp3-control.bp3-checkbox {
```
jonahkagan, line 5: do the Blueprint classes do much for you here? seems like their checkbox styles are pretty different from what we want it to look like. wondering if their buttons styles would be more useful

arlo-client/src/components/DataEntry/BallotAudit.tsx
MorganLove: @jonahkagan Yeah, that's a much better way to do it, haha.

arlo-client/src/components/DataEntry/BallotAudit.tsx
MorganLove: @jonahkagan I initially left it in just in case removing it messed up styling or Formik, and then forgot to try removing it to see what happens. It appears to still work perfectly without it, though, so that's awesome.

arlo-client/src/components/DataEntry/BallotAudit.tsx
MorganLove: @jonahkagan `name` is normally used in this context for Formik to know which element is which, and so unique names is more important there. But since we're setting the whole interpretation at once, it's not needed here now, so I'm removing it.

arlo-client/src/components/DataEntry/BallotAudit.tsx
MorganLove: @jonahkagan Actually, it looks like we can remove the `value` field altogether.

arlo-client/src/components/DataEntry/BlockCheckbox.tsx
MorganLove: @jonahkagan We do overwrite a lot of the styles, but not all, and when I tried to change it out it screwed things up royally, haha.

MorganLove: @jonahkagan We do overwrite a lot, but it does a bunch behind the scenes that is necessary, evidently, since when I tried changing it out everything broke, haha.

arlo-client/src/components/DataEntry/BallotAudit.tsx
MorganLove: And upon things breaking because I was using that to define the testid used by the tests to find the actual element they need to interact with, I added it back in partially.

PR #599, opened 2020-06-09
ABs with no ballots

MorganLove:
**Description**

- Added generation of a single sheet to cover all audit boards without ballots to the PDF generation
- Swapped out the progress bar for 'no ballots to audit' when there aren't ballots to audit

**Testing**

- All green

**Progress**

- It looks like it works to me


MorganLove:
@jonahkagan Added tests for the pertinent component with some mocks, would appreciate a lookover to see if the data in the mocks make sense to you. Thanks!

client/src/components/MultiJurisdictionAudit/RoundManagement/_mocks.ts
```diff
@@ -0,0 +1,95 @@
+import { IAuditBoard } from '../useAuditBoards'
+
+export const auditBoardMocks: {
```
jonahkagan, line 3: could these mocks be combined with the mocks in `MultiJurisdictionAudit/_mocks`? I have been trying to follow your pattern of consolidating mocks there so we can reuse the same set more often

client/src/components/MultiJurisdictionAudit/RoundManagement/_mocks.ts
```diff
@@ -0,0 +1,95 @@
...
+    | 'noBallots'
+    | 'started'
+    | 'signedOff']: IAuditBoard[]
+} = {
+  empty: [],
+  single: [
+    {
+      id: 'audit-board-1',
+      name: 'Audit Board #01',
+      signedOffAt: '',
```
jonahkagan, line 17: i believe this is `null` until the audit board signs off

client/src/components/MultiJurisdictionAudit/RoundManagement/_mocks.ts
```diff
@@ -0,0 +1,95 @@
...
+        numSampledBallots: 30,
+        numAuditedBallots: 15,
+      },
+    },
+  ],
+  signedOff: [
+    {
+      id: 'audit-board-1',
+      name: 'Audit Board #01',
+      signedOffAt: 'a time in the temporal continuum',
```
jonahkagan, line 85: maybe use a real datetime string here?

jonahkagan: though i really like this haha

client/src/components/MultiJurisdictionAudit/RoundManagement/generateSheets.ts
```diff
@@ -105,8 +105,16 @@ export const downloadDataEntry = (auditBoards: IAuditBoard[]): void => {
...
+    } else if (board.currentRoundStatus.numSampledBallots === 0) {
+      auditBoardsWithoutBallots.push(board.name)
     }
   })
+  if (auditBoardsWithoutBallots.length) {
+    auditBoardCreds.addPage('letter')
+    auditBoardsWithoutBallots.forEach((name, i) => {
+      auditBoardCreds.text(`${name}: No ballots`, 20, i * 10 + 20)
+    })
+  }
```
jonahkagan, line 117: do you think there's any easy way to test the pdf generation? like maybe just an integration test where we download the pdf and click the link

client/src/components/MultiJurisdictionAudit/RoundManagement/_mocks.ts
MorganLove: @jonahkagan The type is `string` so should I update that to `string | null`?

client/src/components/MultiJurisdictionAudit/RoundManagement/_mocks.ts
MorganLove: @jonahkagan Haha, I suppooooose. :wink:

client/src/components/MultiJurisdictionAudit/RoundManagement/generateSheets.ts
MorganLove: @jonahkagan I've been digging into that. If you have any ideas, I have a thread on slack about it. I haven't found any easy ways yet, though. Or any hard ways either, for that matter. :stuck_out_tongue_closed_eyes:

client/src/components/MultiJurisdictionAudit/RoundManagement/_mocks.ts
jonahkagan: oh yeah good call. guess i messed up that type originally

PR #602, opened 2020-06-09
adding custom sample size

MorganLove:
**Description**

- The custom sample size feature got lost in the mix, and now it is added back in again.

**Testing**

- Snapshots updated

**Progress**

- lgtm, but haven't tested it exhaustively.

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
+++ b/client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
@@ -185,15 +185,15 @@ describe('Audit Setup > Review & Launch', () => {
         { jurisdictions: jurisdictionMocks.allManifests }
       )
     )
-    const { findByText, getAllByText } = render(
+    render(
       <Router>
         <Review locked={false} prevStage={prevStage} refresh={jest.fn()} />
       </Router>
     )
-    const launchButton = await findByText('Launch Audit')
+    const launchButton = await screen.findByText('Launch Audit')
     fireEvent.click(launchButton, { bubbles: true })
-    await findByText('Are you sure you want to launch the audit?')
-    const confirmLaunchButton = getAllByText('Launch Audit')[1]
+    await screen.findByText('Are you sure you want to launch the audit?')
+    const confirmLaunchButton = screen.getAllByText('Launch Audit')[1]
     fireEvent.click(confirmLaunchButton, { bubbles: true })
     await waitFor(() => {
       expect(apiMock).toHaveBeenCalledTimes(5)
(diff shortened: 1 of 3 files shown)


client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
```diff
@@ -246,4 +246,48 @@ describe('Audit Setup > Review & Launch', () => {
       })
     })
   })
+
+  it.only('accepts custom sample size', async () => {
```
jonahkagan, line 250: supposed to be `.only`?

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
```diff
@@ -246,4 +246,48 @@ describe('Audit Setup > Review & Launch', () => {
...
+    const { getByText, findByText, getAllByText } = render(
+      <Router>
+        <Review locked={false} prevStage={prevStage} refresh={jest.fn()} />
+      </Router>
+    )
+    const newSampleSize = await findByText(
+      'Enter your own sample size (not recommended)'
+    )
+    fireEvent.click(newSampleSize, { bubbles: true })
+    const customSampleSizeInput = await screen.findByRole('textbox')
```
jonahkagan, line 267: any reason you're using `screen` here but not in the rest of the test?

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
```diff
@@ -246,4 +246,48 @@ describe('Audit Setup > Review & Launch', () => {
...
+    fireEvent.change(customSampleSizeInput, { target: { value: '40' } })
+    fireEvent.blur(customSampleSizeInput)
+    await screen.findByText(
+      'Must be less than or equal to the total number of ballots in targeted contests'
+    )
```
jonahkagan, line 272: nice test

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
MorganLove: @jonahkagan I was just adding the new test in, without updating the others yet, so I used the new features for the new test. I could probably easily swap out that bit, though.

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
MorganLove: @jonahkagan Ah, thanks for the catch. I was doing that to speed up test development, forgot to take it back out again, haha.

PR #604, opened 2020-06-10
Reorganize server files

jonahkagan:
- Most everything server-related now under `server`
- API routes under `server/api`
- Renamed `arlo-client` to `client` to match
- Used relative imports for all the Python modules
- Scripts in the `scripts` folder
- Changed the python commands to use `python -m modulename` instead of `python modulename.py` because it wasn't working the old way


jonahkagan:
@umbernhard sorry this is prob gonna mess up your work

umbernhard:
You’re good, this might make parallel testing more feasible ¯\_(ツ)_/¯

PR #609, opened 2020-06-16
Jointly targeted contests

jonahkagan:
Task: #222 

Support jointly targeted contests. Multiple contests can be targeted,
provided they have the same contest universe (jurisdiction list).

Changes:
- We calculate the estimated samples sizes for all targeted contests, returning
the largest sizes
- The audit is only complete if all the targeted contests meet their
risk limit


server/api/contests.py
```diff
@@ -113,6 +113,20 @@ def validate_contests(contests: List[JSONDict]):
...
+    if any(
+        set(contest["jurisdictionIds"]) != set(contests[0]["jurisdictionIds"])
+        for contest in targeted_contests[1:]
+    ):
+        raise BadRequest("All targeted contests must have the same jurisdictions.")
+
+    if any(
+        contest["totalBallotsCast"] != contests[0]["totalBallotsCast"]
+        for contest in targeted_contests[1:]
+    ):
```
benadida, line 127: clean, I like it.

benadida: maybe small nit to, at some point, create constants of these field names.

server/api/rounds.py
```diff
@@ -38,14 +38,16 @@ def get_current_round(election: Election) -> Optional[Round]:
     return next(iter(rounds), None)
 
 
-def is_audit_complete(round_id: str):
-    targeted_round_contest = (
-        RoundContest.query.filter_by(round_id=round_id)
+def is_audit_complete(round: Round):
+    if not round.ended_at:
+        return None
```
benadida, line 43: I realize None acts like false, but why reason not to return actual `False`?

server/api/sample_sizes.py
```diff
@@ -23,33 +23,45 @@ def sample_size_options(election: Election, round_one=False) -> dict:
...
+    samples_sizes_for_targeted_contests = [
+        sample_sizes_for_contest(contest)
+        for contest in targeted_contests_that_havent_met_risk_limit
+    ]
+    # Choose the sample size options for the targted contest with the largest
+    # sample size, since that will cover the samples needed by the other
+    # targeted contests.
+    return max(
+        samples_sizes_for_targeted_contests,
+        key=lambda sample_sizes: sample_sizes["asn"]["size"],
```
benadida, line 58: only ASN? Not the other options?

server/api/contests.py
jonahkagan: will make a note to go back and do that everywhere at some point

server/api/rounds.py
jonahkagan: The way this value has functioned so far is to have three distinct states:
- `None` - a round is in progress
- `true` - the round is over and the audit is complete
- `false` the round is over and another round is needed

This change preserves this behavior. Open to feedback as to whether it's good to have all three states or not. I think when I originally designed it my thinking was that you should only ask whether the audit is complete once a round has ended, so there should be no answer during a round.

server/api/sample_sizes.py
jonahkagan: The contest with the largest ASN size will have the smallest diluted margin, so we're safe to key on that

PR #610, opened 2020-06-17
Activating multiple contests

MorganLove:
**Description**

- Uncommented and updated the code related to adding and removing contests on the multi jurisdiction flow.

**Testing**

- Updated snapshots and unskipped the test related to adding and removing contests

**Progress**

- The backend sorts the contests alphabetically by name from what I can tell, so contests don't come back in the same order they were entered in. This might confuse end users, if they put a contest as 'contest one' but then it comes back as the second contest. Not sure if this is an issue or not, @mcchilders ?

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/Setup/Contests/index.tsx
+++ b/client/src/components/MultiJurisdictionAudit/Setup/Contests/index.tsx
@@ -96,7 +96,7 @@ const Contests: React.FC<IProps> = ({
           >
             <FieldArray
               name="contests"
-              render={() => (
+              render={contestsArrayHelpers => (
                 <>
                   {values.contests.map((contest: IContest, i: number) => {
                     const jurisdictionOptions = jurisdictions.map(j => ({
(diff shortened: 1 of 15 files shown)


client/src/components/MultiJurisdictionAudit/Setup/Contests/index.tsx
```diff
@@ -108,22 +108,18 @@ const Contests: React.FC<IProps> = ({
...
                           <FormSection>
                             <hr />
                           </FormSection>
                         )}
                         <FormSection
                           label={`Contest ${
-                            /* istanbul ignore next */
                             values.contests.length > 1 ? i + 1 : ''
                           } Info`}
                           description="Enter the name of the contest that will drive the audit."
```
jonahkagan, line 119: I'm not sure this line of copy makes sense in the case where there are multiple targeted contests

client/src/components/MultiJurisdictionAudit/Setup/Contests/index.tsx
```diff
@@ -252,21 +240,28 @@ const Contests: React.FC<IProps> = ({
...
                           />
                         </FormSection>
+                        {values.contests.length > 1 && (
+                          <FormButton
+                            intent="danger"
+                            onClick={() => contestsArrayHelpers.remove(i)}
+                          >
+                            Remove Contest {i + 1}
+                          </FormButton>
+                        )}
```
jonahkagan, line 250: wonder if wrapping each contest in a card would make it easier to parse for the user. right now it's a lot of fields that aren't super separated visually

client/src/components/MultiJurisdictionAudit/Setup/Contests/index.tsx
MorganLove: @mcchilders Thoughts on this?

client/src/components/MultiJurisdictionAudit/Setup/Contests/index.tsx
MorganLove: @jonahkagan There's an `<hr/>`, but yeah, it could be more visually distinct. There's likely a satisfactory wrapping component in the blueprintjs uikit that would work.

MorganLove:
@jonahkagan I added cards and updated the test to work with not using the `asyncActRender` so there's some stuff for you to look over before I merge it.

client/src/components/MultiJurisdictionAudit/Setup/Contests/index.tsx
```diff
@@ -106,24 +107,20 @@ const Contests: React.FC<IProps> = ({
                     }))
                     return (
                       /* eslint-disable react/no-array-index-key */
-                      <React.Fragment key={i}>
+                      <Card key={i}>
```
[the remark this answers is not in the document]
MorganLove: @jonahkagan Great catches! I've fixed those now.

Issue #607, opened 2020-06-15
Error on re-audited ballots


mcchilders:
*Description:*
The error in the screenshot attached appears when you:
- Audit/review/submit a ballot w/ two votes in 1st contest
- Navigate back to the ballot list
- Re-audit the same ballot
- Unmark the two original votes, mark another vote in the other contest
- Review & attempt to submit (error shows once ballot is submitted on next ballot screen)

![Screenshot 2020-06-09 at 5.11.46 PM.png](https://zube.io/files/votingworks/90f23260457d0f0d00fdb7ee9541575d-screenshot-2020-06-09-at-5-11-46-pm.png)

PR #615, opened 2020-06-19
Overvote bug

jonahkagan:
Task: #607 

- Fixes a bug where deselecting a contest choice didn't reset the
interpretation to null, so you could accidentally send a `VOTE`
interpretation with no `choiceIds`.
- Adds some tests to start to cover the ballot auditing flow more
thoroughly
- Adds some new test helpers for rendering and mocking the api
(explained in more detail in the comments)


client/src/components/testUtilities.tsx
```diff
@@ -0,0 +1,115 @@
...
+  )
+
+  return {
+    ...render(ui, { wrapper: Wrapper }),
+    // Adding `history` to the returned utilities to allow us
+    // to reference it in our tests (just try to avoid using
+    // this to test implementation details).
+    history,
+  }
+}
```
jonahkagan, line 64: @MorganLove here's one new test helper I found in the docs. Seems like a good starting point for simplifying rendering components that rely on route params.

client/src/components/testUtilities.tsx
```diff
@@ -0,0 +1,115 @@
...
+      options,
+    }))
+    const expectedCalls = apiCalls.map(({ endpoint, options }) => ({
+      endpoint,
+      options,
+    }))
+    expect(actualCalls).toEqual(expectedCalls)
+    mock.mockReset()
+  }
+}
```
jonahkagan, line 94: @MorganLove here's a new idea for mocking api calls

client/src/components/DataEntry/BallotAudit.tsx
```diff
@@ -120,17 +120,14 @@ const BallotAuditContest = ({
...
       } else {
+        const choiceIds = checked
+          ? [...interpretation.choiceIds, value]
+          : interpretation.choiceIds.filter(v => v !== value)
         setInterpretation({
           ...interpretation,
-          interpretation: Interpretation.VOTE,
-          choiceIds: interpretation.choiceIds.filter(v => v !== value),
+          interpretation: choiceIds.length > 0 ? Interpretation.VOTE : null,
+          choiceIds,
```
jonahkagan, line 130: Here's the actual bug fix

client/src/components/DataEntry/index.test.tsx
```diff
@@ -265,5 +257,114 @@ describe('DataEntry', () => {
...
+            {
+              contestId: 'contest-id-2',
+              interpretation: 'CANT_AGREE',
+              choiceIds: [],
+              comment: null,
+            },
+          ],
+        }),
+        apiCalls.ballotsOneAudited,
+      ])
```
jonahkagan, line 284: Here's how that new mock helper gets used

client/src/components/DataEntry/index.test.tsx
```diff
@@ -265,5 +257,114 @@ describe('DataEntry', () => {
...
+      expect(screen.queryByText('Choice Three')).not.toBeInTheDocument()
+      expect(screen.queryByText('Choice Four')).not.toBeInTheDocument()
+
+      // Submit the ballot
+      userEvent.click(
+        screen.getByRole('button', { name: 'Submit & Next Ballot' })
+      )
+      await screen.findByText('Auditing ballot 3 of 27')
+
+      checkMockApi()
```
jonahkagan, line 325: and then you call this function to run the assertions at the end

client/src/components/DataEntry/index.test.tsx
MorganLove: Interesting! It looks more neat, but it's also not immediately apparent what's going on. I'll have to dig into it and play with it myself before I grok it completely.

MorganLove:
@jonahkagan Lots of changes! I like the test innovations. The test failure right now looks like it's some significant snapshot changes, so probably should look through them to make sure they are desired and not something broken.

jonahkagan:
@MorganLove ok I cleaned up this PR. Also changed the API mocking up a bit and added a comment to try to make it clearer. Let me know your thoughts

MorganLove:
@jonahkagan Okay, makes more sense now, thanks! So is the api mocking something that can be used to obsolete code like this: https://github.com/votingworks/arlo/blob/bb03d24b2c5f44e1a1d2adc60b38c060376ba24d/client/src/components/DataEntry/index.test.tsx#L114 ?

Because if we can use it to replace all those separate `mockImplementation`s that'd save sooo much time and effort I think.

jonahkagan:
Yes I think so! I should go back and update the rest of the tests in this file. Just wanted to make sure you were into it before I did that work. I'll be curious to hear your thoughts more as we try this new strategy - I'm sure we'll find out its strengths and weaknesses and improve upon it.

jonahkagan:
@MorganLove I went back and updated the tests to use the new API mocks. There were some race conditions happening with `jest.spyOn(utilities, 'api')` - the mocks were receiving calls from other tests that were running in parallel. I was able to solve it by mocking `window.fetch` instead, which I think is nice anyway because it's closer to the actual separation between the frontend and backend.

Can you give it another look before I merge?

MorganLove:
@jonahkagan Looking good! There's still a couple tests failing on CI, but I much prefer this new method of managing the api mocking. :slightly_smiling_face:

PR #616, opened 2020-06-22
Upgrade testing-library and fix all test warnings

jonahkagan:
Some package upgrades:
- @testing-library/react-hooks v3.3.0
- @testing-library/react v10.3.0

I started these upgrades to get some performance optimizations that I hoped would speed up some tests I was writing.

These broke some tests, specifically the ones using `asyncActRender`, since `act` no longer supports taking an async function (caused a type error).

Looking into this deeper, I discovered that since many of our components make some API calls when they render, we often need the tests to wait for those calls to get made before we can assert anything about the resulting DOM. I think `asyncActRender` was causing this to happen.

Reading some docs, I found out that testing-library's `render` already calls `act` internally, so that part wasn't necessary. So I replaced the calls to `asyncActRender` with calls to `render` (which broke a bunch of tests). To fix the tests, I made sure there was some `waitFor` (or other async helper) to ensure the API calls completed before checking assertions.

While I was at it, I realized that a similar issue was causing the warnings about certain calls not being wrapped in `act`. This warning was a result of API calls happening after the test had already completed (see https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning#how-to-fix-the-act-warning). So I applied a similar fix.

Now we are warning free and using the latest versions! And we know how to prevent those warnings in the future.

### DIFF
--- a/client/src/components/DataEntry/index.test.tsx
+++ b/client/src/components/DataEntry/index.test.tsx
@@ -1,10 +1,12 @@
 import React from 'react'
 import { render, waitFor, fireEvent } from '@testing-library/react'
 import { StaticRouter } from 'react-router-dom'
-import { routerTestProps, asyncActRender } from '../testUtilities'
+import { routerTestProps } from '../testUtilities'
 import DataEntry from './index'
 import { dummyBoards, dummyBallots, contest } from './_mocks'
 import * as utilities from '../utilities'
+
+window.scrollTo = jest.fn()
 
 const apiMock: jest.SpyInstance<
   ReturnType<typeof utilities.api>,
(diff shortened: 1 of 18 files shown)


client/src/components/DataEntry/index.test.tsx
```diff
@@ -1,11 +1,13 @@
 import React from 'react'
 import { render, waitFor, fireEvent } from '@testing-library/react'
 import { StaticRouter } from 'react-router-dom'
-import { routerTestProps, asyncActRender } from '../testUtilities'
+import { routerTestProps } from '../testUtilities'
 import DataEntry from './index'
 import { dummyBoards, dummyBallots, contest } from './_mocks'
 import * as utilities from '../utilities'
 
+window.scrollTo = jest.fn()
```
MorganLove, line 9: @jonahkagan Not blocking, just curious what this does.

jonahkagan: There was a warning about `window.scrollTo` being undefined (we use `window.scrollTo` to reset to the top of the screen when you transition between ballots). This is what the internet said would fix it, and it did :)

Issue #611, opened 2020-06-17
Ordering of contests/candidates should stay same btw setup/ballot display


mcchilders:
We should order the contests/choices on the data entry screens in the same order they are entered in the audit setup in the tool. Target contests first, opportunistic contests below them. The choices in all the contests should be ordered in the order entered at setup.

Note that editing a contest/choice should not change its order, but deleting/recreating would.

(Maybe implement using created date to order?)

PR #619, opened 2020-06-24
Preserve contest and choice order that audit admin enters

jonahkagan:
Task: #611 

In order to preserve the order that contests and choices are entered in the by the audit admin, we order by `created_at` when we pull them from the db.


jonahkagan:
Yeah I agree it's a bit of an overload on `created_at`. I think we will revisit this when we have CVRs soon anyway

PR #626, opened 2020-06-30
Order ballot draws in tests

jonahkagan:
The ordering of the ballot draws in the test helpers that audit the ballots to run a round will make the tests more deterministic (since sample sizes for future rounds depend on the audit results from previous rounds). Currently, the tests are deterministic from run to run, but are not resilient to changes in the code (e.g. if the way we sample ballots changes and results in a different internal database ordering). We have to make changes to the ballot sampling code, and in my exploration so far, it broke these tests even though I'm fairly sure the logic didn't change. So this PR sets us up to rely on these tests when changing that code.

Also moved the ballot retrieval lists to use snapshots while I was at it, since they were impacted.


jonahkagan:
Do you mean the expected sample sizes in the tests? They only changed for the second round.

That changed because the helper code that audits the ballots for a round now uses a (new) deterministic ordering of the samples. Some ballots get sampled multiple times, so with a new ordering some of those ballots that count for multiple votes changed which candidate they got awarded to. So the overall audit result for the first round changed slightly, which then changed the sample size for the second round.

PR #628, opened 2020-07-02
Count ballots not sample draws for opportunistic contests

jonahkagan:
Previously, we counted votes for targeted and opportunistic contests the
same way - using the sample draws. This was incorrect. For targeted
contests, we should count the sample draws. For opportunistic, we should
count distinct ballots.

This will make more a difference when we have multiple contests with
distinct sample draws that may refer to the same ballots.

### DIFF
--- a/server/api/audit_boards.py
+++ b/server/api/audit_boards.py
@@ -274,18 +274,28 @@ def calculate_risk_measurements(election: Election, round: Round):
         round_contest.is_complete = is_complete
 
 
-def count_audited_votes(election: Election, round: Round):
-    vote_counts = dict(
-        BallotInterpretation.query.filter_by(is_overvote=False)
-        .join(SampledBallot)
-        .join(SampledBallotDraw)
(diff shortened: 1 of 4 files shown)


server/api/audit_boards.py
```diff
@@ -274,18 +274,28 @@ def calculate_risk_measurements(election: Election, round: Round):
...
+            .join(SampledBallotDraw)
+            .filter_by(round_id=round.id)
+            .join(BallotInterpretation.selected_choices)
+            .group_by(ContestChoice.id)
+            .values(
+                ContestChoice.id,
+                func.count(
+                    SampledBallot.id
+                    if contest.is_targeted
+                    else SampledBallot.id.distinct()
```
benadida, line 294: you've checked the SQL coming out of this and it looks good?

jonahkagan: I think so - not quite sure what I should look for but the query looks straightforward in each case.

PR #629, opened 2020-07-02
Track which contest a ballot was sampled for

jonahkagan:
Task: #621 

For independently targeted contests, each targeted contest will need to have a separate sample drawn. Thus, we'll need to track which contest each ballot was sampled for. This change doesn't support independently targeted contests, but lays the groundwork while preserving existing functionality.

- Adds a field SampledBallotDraw.contest_id (plus a migration)
- Populates this field when drawing a sample with the targeted contest id
- For joinly targeted contests, instead of having one set of SampledBallotDraws that are used by all the targeted contests, gives each its own set (even though they correspond to the exact same sample).
- Updates the vote counting logic to count the SampledBallotDraws by targeted contest
- Creates a RoundContest object only for contests that will need to be audited that round, so we can tell which contests have not met their risk limit throughout the code (previously, we made one for every contest every round)

Also adds some types to the `group_by` utility function. I forget why I did that but it's a step in a more typed direction so leaving it in.


server/api/sample_sizes.py
```diff
@@ -41,7 +41,9 @@ def sample_sizes_for_contest(contest: Contest) -> dict:
             cumulative_results,
         )
 
-    targeted_contests = Contest.query.filter_by(is_targeted=True)
+    targeted_contests = Contest.query.filter_by(
+        election_id=election.id, is_targeted=True
+    )
```
jonahkagan, line 46: i think this was just a bug... yikes

server/tests/snapshots/snap_test_jointly_targeted_contests.py
```diff
@@ -0,0 +1,26 @@
...
+    "Contest 3 - candidate 3": 0,
+}
+
+snapshots["test_two_rounds 2"] = {
+    "Contest 2 - No": 440,
+    "Contest 2 - Yes": 1028,
+    "Contest 3 - candidate 1": 0,
+    "Contest 3 - candidate 2": 0,
+    "Contest 3 - candidate 3": 0,
+}
```
jonahkagan, line 26: pretty excited about this new way to use snapshot testing for numeric test constants (instead of storing them in variables by hand)

server/api/rounds.py
```diff
@@ -67,41 +68,100 @@ def validate_round(round: dict, election: Election):
...
-
-    # Compute the total number of ballot samples in all rounds leading up to
-    # this one. Note that this corresponds to the number of SampledBallotDraws,
-    # not SampledBallots.
-    num_previously_sampled = (
-        SampledBallotDraw.query.join(Round).filter_by(election_id=election.id).count()
+    # Figure out which contests still need auditing
+    last_round = get_current_round(election)
+    contests_that_havent_met_risk_limit = (
+        [
```
benadida, line 82: should there be guidance on the `round` argument passed here? I'm guessing it's a new round that's later than `last_round`?

server/api/sample_sizes.py
benadida: good catch. Let's add some tests that have other elections in there just hanging around to make sure we're truly not joining against them, ever.

server/api/audit_boards.py
```diff
@@ -260,41 +260,49 @@ def calculate_risk_measurements(election: Election, round: Round):
...
+            vote_counts = dict(
+                BallotInterpretation.query.filter_by(
+                    contest_id=contest.id, is_overvote=False
+                )
+                .join(SampledBallot)
+                .join(SampledBallotDraw)
+                .filter_by(round_id=round.id)
+                .join(BallotInterpretation.selected_choices)
+                .group_by(ContestChoice.id)
+                .values(ContestChoice.id, func.count(SampledBallot.id.distinct()))
```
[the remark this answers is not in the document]
jonahkagan: I had tried that before and it didn't work. Just tried it again to remember why: that query returns a count of 1 for each ContestChoice.id. If you can help me think through why that is, I'd welcome the help, because I didn't and still don't understand why it's happening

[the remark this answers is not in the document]
jonahkagan: I had tried that before and it didn't work. Just tried it again to remember why: that query returns a count of 1 for each ContestChoice.id. If you can help me think through why that is, I'd welcome the help, because I didn't and still don't understand why it's happening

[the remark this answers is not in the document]
benadida: if you've got it working with your query, I'm good with it. Worth a longer debug on the other one at some point, but don't want to block you today.

server/api/audit_boards.py
```diff
@@ -260,41 +260,49 @@ def calculate_risk_measurements(election: Election, round: Round):
...
+            vote_counts = dict(
+                BallotInterpretation.query.filter_by(
+                    contest_id=contest.id, is_overvote=False
+                )
+                .join(SampledBallot)
+                .join(SampledBallotDraw)
+                .filter_by(round_id=round.id, contest_id=contest.id)
+                .join(BallotInterpretation.selected_choices)
+                .group_by(ContestChoice.id)
+                .values(ContestChoice.id, func.count(SampledBallot.id))
```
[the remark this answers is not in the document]
jonahkagan: done

server/api/audit_boards.py
```diff
@@ -260,41 +260,49 @@ def calculate_risk_measurements(election: Election, round: Round):
...
+        # sampled for.
+        else:
+            vote_counts = dict(
+                BallotInterpretation.query.filter_by(
+                    contest_id=contest.id, is_overvote=False
+                )
+                .join(SampledBallot)
+                .join(SampledBallotDraw)
+                .filter_by(round_id=round.id)
+                .join(BallotInterpretation.selected_choices)
```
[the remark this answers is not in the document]
jonahkagan: done

server/api/rounds.py
jonahkagan: changed this up to be more clear

server/api/sample_sizes.py
jonahkagan: So, I've been thinking about this problem and it's feeling kind of tricky. In order to do this well, we need to make sure that *every* query filters by election, which means we either need to write a new test for every query, or add the extra elections to the db for every test. Writing a new test for every query seems like a lot of work, and also a practice we'd likely fail to keep up. Adding extra elections to the db for every test seems easier and more future-proof.

However, even adding an extra election to the db in this case wouldn't necessarily make the test fail. The extra election would have to have a targeted contest that yielded a larger sample size than the one the test is already using - in other words, the data design has to have knowledge of the logic to reliably cause a failure. We could design a test for this bug in particular, but I'm not confident this will make for a robust general approach for unknown future bugs.

That makes me think about potentially building in some sort of check into the SQLAlchemy models to ensure that every query for a model includes the election_id field. For models that don't currently have an election_id, we could either add one (good for sharding in the future, though prob not necessary) or we could make those models check for a different foreign key that has the same logical effect (e.g. for a Batch, check that jurisdiction_id is included, since Jurisdictions are election-specific).

Thoughts?

This would have been a pretty gnarly bug and we got pretty damn lucky to catch it so I'm definitely inclined to try to do something to prevent it from happening again.

benadida: I agree 100% with your point, that if we can build something that forces a where clause by election ID, we should. Very much worth the investment. Every query should have one, except where there is an explicit opt out named with an appropriately scary name.

migrations/versions/5b7164faeda5_sampledballotdraw_contest_id.py
```diff
@@ -0,0 +1,54 @@
+# pylint: disable=invalid-name
+"""SampledBallotDraw.contest_id
```
jonahkagan, line 2: @benadida I added this migration. Can you review?

migrations/versions/5b7164faeda5_sampledballotdraw_contest_id.py
```diff
@@ -0,0 +1,62 @@
...
+    op.execute(
+        """
+        UPDATE sampled_ballot_draw
+        SET contest_id = contest.id
+        FROM contest
+        JOIN election ON election.id = contest.election_id
+        JOIN round ON election.id = round.election_id
+        WHERE round.id = sampled_ballot_draw.round_id
+        AND contest.is_targeted
+        """
```
benadida, line 42: this works even if there's more than one targeted contest?

jonahkagan: at this point in time, we know there's only one targeted contest per election in the prod db

benadida: but is that happenstance? Could the data model have more than one? So maybe `limit 1` just to be safe?

PR #589, opened 2020-05-28
Overvotes

jonahkagan:
Task: #585 

- Modifies the `BallotInterpretation` model to allow multiple selected choices per contest per ballot
- Adds an `is_overvote` field to `BallotInterpretation` to cache whether the selected choices are greater than the number of allowed votes for that contest
- Updates affected API endpoints (e.g. auditing a ballot, retrieving ballot interpretations)
- Updates existing tests
- Adds new tests for overvotes

Does not update the audit report to include explicit overvote info - coming in a later PR.

PR #632, opened 2020-07-07
DB migrations with Alembic

jonahkagan:
Adds Alembic (https://alembic.sqlalchemy.org/), a tool that integrates with SQLAlchemy to auto-generate and run db migrations. Ooooh it's soo cool

Tests it out by creating a migration from the last prod deploy (branch `rhode-island`) to `main`, which covers the db model changes from #589. Ran it manually against a database backup from prod

Questions:
- Do we want any automated testing for migrations? Someone somewhere on the internet said they ran all their migrations to set up their test db as a way of ensuring that the migrations worked correctly


Pipfile
```diff
@@ -31,6 +32,7 @@ requests = "*"
 sqlalchemy-utils = "*"
 uuid = "*"
 xkcdpass = "*"
+sqlalchemy = "*"
```
jonahkagan, line 35: realized we were getting away with using sqlalchemy as a dep of some other package

alembic.ini
```diff
@@ -0,0 +1,87 @@
...
+script_location = migrations
+
+# template used to generate migration files
+# file_template = %%(rev)s_%%(slug)s
+
+# timezone to use when rendering the date
+# within the migration file as well as the filename.
+# string value is passed to dateutil.tz.gettz()
+# leave blank for localtime
+timezone = UTC
```
jonahkagan, line 14: autogenerated except these two lines

migrations/env.py
```diff
@@ -0,0 +1,89 @@
...
+
+from sqlalchemy import engine_from_config
+from sqlalchemy import pool
+
+from alembic import context
+
+# Modify PYTHONPATH so that we can import the Arlo config
+# https://stackoverflow.com/a/58891735
+sys.path = ["", ".."] + sys.path[1:]
+from server.config import DATABASE_URL
```
jonahkagan, line 13: my changes

migrations/env.py
```diff
@@ -0,0 +1,89 @@
...
+
+# this is the Alembic Config object, which provides
+# access to the values within the .ini file in use.
+config = context.config
+
+# Set the db url to use based on the Arlo config
+# https://stackoverflow.com/a/37891036
+config.set_main_option("sqlalchemy.url", DATABASE_URL)
```
jonahkagan, line 21: also my change

migrations/env.py
```diff
@@ -0,0 +1,89 @@
...
+fileConfig(config.config_file_name)
+
+# add your model's MetaData object here
+# for 'autogenerate' support
+# from myapp import mymodel
+# target_metadata = mymodel.Base.metadata
+from server.database import Base
+import server.models  # pylint: disable=unused-import
+
+target_metadata = Base.metadata
```
jonahkagan, line 34: aaand one more change

migrations/env.py
```diff
@@ -0,0 +1,89 @@
+# pylint: disable=no-member,wrong-import-position
```
jonahkagan, line 1: autogenerated file

migrations/script.py.mako
```diff
@@ -0,0 +1,25 @@
+# pylint: disable=invalid-name
```
jonahkagan, line 1: autogenerated

server/database.py
```diff
@@ -9,8 +9,18 @@
...
 
+meta = MetaData(
+    naming_convention={
+        "ix": "%(column_0_N_label)s_idx",
+        "uq": "%(table_name)s_%(column_0_N_name)s_key",
+        "ck": "%(table_name)s_%(constraint_name)s_check",
+        "fk": "%(table_name)s_%(column_0_N_name)s_fkey",
+        "pk": "%(table_name)s_pkey",
+    }
+)
```
jonahkagan, line 20: ok so this is a tricky thing. this configures sqlalchemy to use these templates for naming constraints instead of using the postgresql auto-naming. this is necessary for alembic to be able to come up with constraint names that match the actual constraint names in the database when autogenerating migration code. i tried to match the postgresql constraint naming conventions as well as I could

more here: https://alembic.sqlalchemy.org/en/latest/naming.html

Pipfile
benadida: wasn't it `flask-sqlalchemy` which you removed? ;)

server/database.py
benadida: totally sane reason for naming things a particular way :)

PR #636, opened 2020-07-07
Migration tests

jonahkagan:
Adds some basic tests to ensure our migrations run successfully and yield the same resulting db schema as our models.

How we do it:
- Create an initial migration from the empty database to the database before our current first migration
- Leverage pytest-alembic, a plugin that runs these basic tests for you

This PR also moves the `migrations` directory inside `server` and adds some documentation to it.


server/migrations/env.py
```diff
@@ -55,6 +55,9 @@ def run_migrations_offline():
     context.configure(
         url=url,
         target_metadata=target_metadata,
+        compare_type=True,
+        compare_server_default=True,
+        include_schemas=True,
```
jonahkagan, line 60: these are suggested configs from the pytest-alembic docs

server/migrations/env.py
```diff
@@ -70,11 +73,14 @@ def run_migrations_online():
...
-        poolclass=pool.NullPool,
-    )
+    connectable = config.attributes.get("connection", None)
+
+    if connectable is None:
+        connectable = engine_from_config(
+            config.get_section(config.config_ini_section),
+            prefix="sqlalchemy.",
+            poolclass=pool.NullPool,
+        )
```
jonahkagan, line 83: change required by pytest-alembic

PR #642, opened 2020-07-13
RoundManagement coverage

MorganLove:
**Description**

More tests

**Testing**

Covering RoundManagement

**Progress**

done!

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/RoundManagement/index.test.tsx
+++ b/client/src/components/MultiJurisdictionAudit/RoundManagement/index.test.tsx
@@ -1,5 +1,6 @@
 import React from 'react'
-import { render, screen, fireEvent } from '@testing-library/react'
+import { render, screen } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
 import { useParams } from 'react-router-dom'
 import RoundManagement from './index'
 import { roundMocks, auditBoardMocks } from '../_mocks'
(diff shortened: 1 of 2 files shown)


client/src/components/MultiJurisdictionAudit/RoundManagement/index.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+    const expectedCalls = [apiCalls.getBallots]
+    await withMockFetch(expectedCalls, async () => {
+      render(
+        <RoundManagement
+          round={roundMocks.singleIncomplete[0]}
+          auditBoards={auditBoardMocks.unfinished}
+          createAuditBoards={jest.fn()}
+        />
+      )
+      fireEvent.click(
```
jonahkagan, line 86: use userEvent instead of fireEvent?

client/src/components/MultiJurisdictionAudit/RoundManagement/index.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+        await screen.findByText(
+          'Download Aggregated Ballot Retrival List for Round 1'
+        ),
+        { bubbles: true }
+      )
+      fireEvent.click(
+        await screen.findByText(
```
jonahkagan, line 93: i think you only need to use `await screen.findByText` on the first call after render (to allow for the API calls to happen). afterwards you should be able to use `screen.getByText`

client/src/components/MultiJurisdictionAudit/RoundManagement/index.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+window.open = jest.fn()
+
+const apiCalls = {
+  getBallots: {
+    url: '/api/election/1/jurisdiction/1/round/round-1/ballots',
+    response: dummyBallots,
+  },
+}
+
+describe('RoundManagement', () => {
```
jonahkagan, line 43: do we need a test for the flow where the JA creates the audit boards?

MorganLove: @jonahkagan That is tested in the CreateAuditBoards component directly.

client/src/components/MultiJurisdictionAudit/RoundManagement/index.test.tsx
MorganLove: Oh right, I haven't tried that yet.

Issue #622, opened 2020-06-25
Add logic for targeting multiple contests independently


mcchilders:
When multiple target contests are present and the UI is toggled to target them independently:

- use contest data entered on the front end for each targeted contest to determine an independent set of samples for each target contest, from the universe of that particular target contest (which may or may not be the same as the other contests)
- combine and de-dupe those independent samples to create the global ballot retrieval list for a given round, being careful to retain the target contests for which a ballot was sampled even after the samples are combined
- if entering ballot-by-ballot data online, allow AB's to input votes from each ballot for all targeted contests that include that ballot's jurisdiction in their universe
- use the vote totals from only the ballots selected for each targeted contest to determine whether the risk limit has been met (p-value < risk limit) contest by contest
- the audit completes when all targeted contests have met the risk limit.
-- if some targeted contests have not met the risk limit at the end of the auditing round, the contests that have met the risk limit drop out (complete), and only those that haven't (incomplete) are used to escalate in further rounds.

PR #643, opened 2020-07-14
Independently targeted contests

jonahkagan:
Task: #622

Adds support for multiple independently targeted contests by drawing a separate sample for each targeted contest.

Removes support for jointly targeted contests because we decided we don't need to implement it right now.

Breaking changes to two APIs:
- GET /sample-sizes now returns a map from targeted contest ids to sample size options
- POST /round now requires a map from targeted contest id to sample size, rather than just one sample size

Also fixes an issue in the audit report where we were reporting a sample size for opportunistic contests. This doesn't make sense when there are multiple targeted contests, because the number of ballots we count for the opportunistic contests doesn't necessarily match the sample size. Instead, we don't report a sample size for opportunistic contests.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -30,10 +30,13 @@ jsonschema = "*"
 numpy = "*"
 psycopg2-binary = "*"
 requests = "*"
-sqlalchemy-utils = "*"
+# sqlalchemy-utils 0.36.8 introduced a bug
+# Can probably reset this to * once the next patch comes out
+sqlalchemy-utils = "==0.36.7"
 uuid = "*"
 xkcdpass = "*"
 sqlalchemy = "*"
+typing-extensions = "*"
 
 [requires]
 python_version = "3.7"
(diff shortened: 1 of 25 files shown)


jonahkagan:
@MorganLove check out these API changes. I think you can make a PR based on this one to update the frontend sample size UI. I will wait for that to merge this into `main` so we don't break staging. Let me know if you have questions/want me to review changes to your type defs.

Pipfile
```diff
@@ -30,10 +30,13 @@ jsonschema = "*"
 numpy = "*"
 psycopg2-binary = "*"
 requests = "*"
-sqlalchemy-utils = "*"
+# sqlalchemy-utils 0.36.8 introduced a bug
+# Can probably reset this to * once the next patch comes out
+sqlalchemy-utils = "==0.36.7"
```
benadida, line 35: let's file a ticket so we remember to check on this in a few weeks.

mcchilders: @ben yep - this is a simplification @jonahkagan came up with per this thread: https://votingworks.slack.com/archives/CKCVA0F9S/p1593121185202400?thread_ts=1593119993.195400&cid=CKCVA0F9S

PR #644, opened 2020-07-14
Report: add column for each contest's ticket numbers

jonahkagan:
When there are multiple targeted contests, we draw a separate sample for each contest. That means a ballot might get sampled for multiple contests. In the report, we want to be able to see which contests each ballot was sampled for, so we create a column of ticket numbers corresponding to each contest (previously, there was just one column for all ticket numbers).


server/tests/snapshots/snap_test_multiple_targeted_contests.py
```diff
@@ -41,3 +41,870 @@
...
+1,Contest 1,Targeted,191,Yes,1.84402281147622e-05,DATETIME,DATETIME,candidate 1: 132; candidate 2: 59\r
+1,Contest 2,Targeted,485,No,1.0,DATETIME,DATETIME,Yes: 0; No: 0\r
+1,Contest 3,Opportunistic,,No,1.0,DATETIME,DATETIME,candidate 1: 0; candidate 2: 0; candidate 3: 0\r
+2,Contest 2,Targeted,1468,Yes,8.27580207807509e-24,DATETIME,DATETIME,Yes: 1028; No: 440\r
+2,Contest 3,Opportunistic,,No,1.0,DATETIME,DATETIME,candidate 1: 0; candidate 2: 0; candidate 3: 0\r
+\r
+######## SAMPLED BALLOTS ########\r
+Jurisdiction Name,Batch Name,Ballot Position,Audited?,Ticket Numbers: Contest 1,Ticket Numbers: Contest 2,Audit Result: Contest 1,Audit Result: Contest 2,Audit Result: Contest 3\r
+J1,1,4,AUDITED,,"Round 1: 0.324042377098893634, Round 2: 0.781169515578102262",candidate 1,Yes,\r
+J1,1,9,AUDITED,Round 1: 0.132689946437387284,Round 1: 0.132689946437387284,candidate 1,,\r
```
jonahkagan, line 75: @umbernhard would we expect a ballot that got sampled once in each of the targeted contests to have the same ticket number for each contest like we see here? The contests have the same universe but diff margins (and thus diff sample sizes). Also note similar behavior on line 77 below.

benadida: Is that right? Are we using the same seed for each contest, no contest index involved?

PR #645, opened 2020-07-15
Supporting sample sizes for multiple independently targeted contests

MorganLove:
**Description**

- Updated the types to reflect api changes
- Updated the review and launch stage to have multiple sample size selection boxes for the targeted contests
- Updated launch process to use the format expected by the api changes

**Testing**

- None yet, waiting for review

**Progress**

- It looks like it works, need to double check with @jonahkagan that it all makes sense.

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
+++ b/client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import userEvent from '@testing-library/user-event'
-import { render, screen, fireEvent, waitFor } from '@testing-library/react'
-import { BrowserRouter as Router, useParams } from 'react-router-dom'
+import { screen, fireEvent, waitFor } from '@testing-library/react'
+import { useParams } from 'react-router-dom'
 import { toast } from 'react-toastify'
 import relativeStages from '../_mocks'
 import Review from './index'
@@ -10,7 +10,8 @@ import useAuditSettings from '../../useAuditSettings'
 import { settingsMock, sampleSizeMock } from './_mocks'
 import { contestMocks } from '../Contests/_mocks'
 import { jurisdictionMocks } from '../../_mocks'
-import { withMockFetch } from '../../../testUtilities'
+import { withMockFetch, renderWithRouter } from '../../../testUtilities'
+import { ISampleSizes } from './useSampleSizes'
 
 const auditSettingsMock = useAuditSettings as jest.Mock
 
@@ -19,12 +20,12 @@ const apiCalls = {
     url: '/api/election/1/sample-sizes',
     response: sampleSizeMock,
   },
-  putDefaultSampleSize: {
+  postRound: (sampleSizes: ISampleSizes) => ({
     url: '/api/election/1/round',
     response: { status: 'ok' },
     options: {
       body: JSON.stringify({
-        sampleSizes: { 'contest-id': 46 },
+        sampleSizes,
         roundNum: 1,
       }),
       headers: {
@@ -32,35 +33,7 @@ const apiCalls = {
       },
       method: 'POST',
     },
-  },
-  putNondefaultSampleSize: {
-    url: '/api/election/1/round',
-    response: { status: 'ok' },
-    options: {
-      body: JSON.stringify({
-        sampleSizes: { 'contest-id': 67 },
-        roundNum: 1,
-      }),
-      headers: {
-        'Content-Type': 'application/json',
-      },
-      method: 'POST',
-    },
-  },
-  putCustomSampleSize: {
-    url: '/api/election/1/round',
-    response: { status: 'ok' },
-    options: {
-      body: JSON.stringify({
-        sampleSizes: { 'contest-id': 5 },
-        roundNum: 1,
-      }),
-      headers: {
-        'Content-Type': 'application/json',
-      },
-      method: 'POST',
-    },
-  },
+  }),
   getEmptyJurisdictions: {
     url: '/api/election/1/jurisdiction',
     response: { jurisdictions: [] },
@@ -106,6 +79,14 @@ const { prevStage } = relativeStages('Review & Launch')
 
 const refreshMock = jest.fn()
 
+const renderView = () =>
+  renderWithRouter(
+    <Review locked={false} prevStage={prevStage} refresh={refreshMock} />,
+    {
+      route: '/election/1/setup',
+    }
+  )
+
 beforeEach(() => {
   refreshMock.mockClear()
   toastSpy.mockClear()
@@ -124,11 +105,7 @@ describe('Audit Setup > Review & Launch', () => {
       apiCalls.getSampleSizeOptions,
     ]
     await withMockFetch(expectedCalls, async () => {
-      const { container } = render(
-        <Router>
-          <Review locked={false} prevStage={prevStage} refresh={refreshMock} />
-        </Router>
-      )
+      const { container } = renderView()
       await screen.findByText('Review & Launch')
       expect(container).toMatchSnapshot()
     })
@@ -143,11 +120,7 @@ describe('Audit Setup > Review & Launch', () => {
       apiCalls.getSampleSizeOptions,
     ]
     await withMockFetch(expectedCalls, async () => {
-      const { container } = render(
-        <Router>
-          <Review locked={false} prevStage={prevStage} refresh={refreshMock} />
-        </Router>
-      )
+      const { container } = renderView()
       await screen.findByText('Review & Launch')
       expect(container).toMatchSnapshot()
     })
@@ -162,11 +135,7 @@ describe('Audit Setup > Review & Launch', () => {
       apiCalls.getSampleSizeOptions,
     ]
     await withMockFetch(expectedCalls, async () => {
-      const { container } = render(
-        <Router>
-          <Review locked={false} prevStage={prevStage} refresh={refreshMock} />
-        </Router>
-      )
+      const { container } = renderView()
       await screen.findByText('Review & Launch')
       expect(container).toMatchSnapshot()
     })
@@ -181,11 +150,7 @@ describe('Audit Setup > Review & Launch', () => {
       apiCalls.getSampleSizeOptions,
     ]
     await withMockFetch(expectedCalls, async () => {
-      const { container } = render(
-        <Router>
-          <Review locked={false} prevStage={prevStage} refresh={refreshMock} />
-        </Router>
-      )
+      const { container } = renderView()
       await screen.findByText('Review & Launch')
       expect(container).toMatchSnapshot()
     })
@@ -200,11 +165,7 @@ describe('Audit Setup > Review & Launch', () => {
       apiCalls.getSampleSizeOptions,
     ]
     await withMockFetch(expectedCalls, async () => {
-      const { container } = render(
-        <Router>
-          <Review locked={false} prevStage={prevStage} refresh={refreshMock} />
-        </Router>
-      )
+      const { container } = renderView()
       await screen.findByText('Review & Launch')
       expect(container).toMatchSnapshot()
     })
@@ -217,21 +178,17 @@ describe('Audit Setup > Review & Launch', () => {
       apiCalls.getJurisdictionFile,
       apiCalls.getTargetedContests,
       apiCalls.getSampleSizeOptions,
-      apiCalls.putDefaultSampleSize,
+      apiCalls.postRound({ 'contest-id': 46 }),
       apiCalls.getSampleSizeOptions,
     ]
     await withMockFetch(expectedCalls, async () => {
-      render(
-        <Router>
-          <Review locked={false} prevStage={prevStage} refresh={refreshMock} />
-        </Router>
-      )
+      renderView()
       await screen.findByText('Review & Launch')
       const launchButton = screen.getByText('Launch Audit')
-      userEvent.click(launchButton, { bubbles: true })
+      userEvent.click(launchButton)
       await screen.findByText('Are you sure you want to launch the audit?')
       const confirmLaunchButton = screen.getAllByText('Launch Audit')[1]
-      userEvent.click(confirmLaunchButton, { bubbles: true })
+      userEvent.click(confirmLaunchButton)
       await waitFor(() => expect(refreshMock).toHaveBeenCalled())
     })
   })
@@ -243,24 +200,20 @@ describe('Audit Setup > Review & Launch', () => {
       apiCalls.getJurisdictionFile,
       apiCalls.getTargetedContests,
       apiCalls.getSampleSizeOptions,
-      apiCalls.putNondefaultSampleSize,
+      apiCalls.postRound({ 'contest-id': 67 }),
       apiCalls.getSampleSizeOptions,
     ]
     await withMockFetch(expectedCalls, async () => {
-      render(
-        <Router>
-          <Review locked={false} prevStage={prevStage} refresh={refreshMock} />
-        </Router>
-      )
+      renderView()
       const newSampleSize = await screen.findByText(
         '67 samples (70% chance of reaching risk limit and completing the audit in one round)'
       )
-      userEvent.click(newSampleSize, { bubbles: true })
+      userEvent.click(newSampleSize)
       const launchButton = await screen.findByText('Launch Audit')
-      userEvent.click(launchButton, { bubbles: true })
+      userEvent.click(launchButton)
       await screen.findByText('Are you sure you want to launch the audit?')
       const confirmLaunchButton = screen.getAllByText('Launch Audit')[1]
-      userEvent.click(confirmLaunchButton, { bubbles: true })
+      userEvent.click(confirmLaunchButton)
       await waitFor(() => expect(refreshMock).toHaveBeenCalled())
     })
   })
@@ -272,19 +225,15 @@ describe('Audit Setup > Review & Launch', () => {
       apiCalls.getJurisdictionFile,
       apiCalls.getTargetedContests,
       apiCalls.getSampleSizeOptions,
-      apiCalls.putCustomSampleSize,
+      apiCalls.postRound({ 'contest-id': 5 }),
       apiCalls.getSampleSizeOptions,
     ]
     await withMockFetch(expectedCalls, async () => {
-      render(
-        <Router>
-          <Review locked={false} prevStage={prevStage} refresh={refreshMock} />
-        </Router>
-      )
+      renderView()
       const newSampleSize = await screen.findByText(
         'Enter your own sample size (not recommended)'
       )
-      userEvent.click(newSampleSize, { bubbles: true })
+      userEvent.click(newSampleSize)
       const customSampleSizeInput = await screen.findByRole('textbox')
       userEvent.type(customSampleSizeInput, '40')
       fireEvent.blur(customSampleSizeInput)
@@ -303,10 +252,10 @@ describe('Audit Setup > Review & Launch', () => {
         ).toBeNull()
       )
       const launchButton = await screen.findByText('Launch Audit')
-      userEvent.click(launchButton, { bubbles: true })
+      userEvent.click(launchButton)
       await screen.findByText('Are you sure you want to launch the audit?')
       const confirmLaunchButton = screen.getAllByText('Launch Audit')[1]
-      userEvent.click(confirmLaunchButton, { bubbles: true })
+      userEvent.click(confirmLaunchButton)
       // evidently this is calling handleSubmit on line 293, but onSubmit on line 198 is not getting called.
       // The error text is vanishing (I think), so it should be passing validation, but that's the only thing that I know of to block submission.
       await waitFor(() => expect(refreshMock).toHaveBeenCalled())
--- a/client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
+++ b/client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
@@ -126,6 +126,21 @@ describe('Audit Setup > Review & Launch', () => {
     })
   })
 
+  it('renders full state with offline setting', async () => {
+    auditSettingsMock.mockReturnValue(settingsMock.offline)
+    const expectedCalls = [
+      apiCalls.getJurisdictions,
+      apiCalls.getJurisdictionFile,
+      apiCalls.getTargetedContests,
+      apiCalls.getSampleSizeOptions,
+    ]
+    await withMockFetch(expectedCalls, async () => {
+      const { container } = renderView()
+      await screen.findByText('Review & Launch')
+      expect(container).toMatchSnapshot()
+    })
+  })
+
   it('renders full state with jurisdictions on opportunistic contest', async () => {
     auditSettingsMock.mockReturnValue(settingsMock.full)
     const expectedCalls = [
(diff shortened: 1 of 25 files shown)


client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
```diff
@@ -9,28 +9,38 @@ import { ISidebarMenuItem } from '../../../Atoms/Sidebar'
...
   style: 'percent',
 })
 
-interface IStringSampleSize {
-  size: string
-  type: string | null
-  prob: number | null
+interface IStringSampleSizes {
+  [key: string]: {
+    size: string
```
jonahkagan, line 36: why use strings here? seems like numbers would be easier, since that's the API uses

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
```diff
@@ -126,6 +153,15 @@ const Review: React.FC<IProps> = ({ prevStage, locked, refresh }: IProps) => {
       j.ballotManifest.processing.status === FileProcessingStatus.PROCESSED
   ).length
 
+  const initialValues = targetedContests.length
+    ? targetedContests.map(c => ({
+        [c.id]: sampleSizeOptions[c.id] ? sampleSizeOptions[c.id][0].size : '',
```
jonahkagan, line 158: the backend guarantees that the keys of `sampleSizeOptions` will include all the targeted contest ids

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
```diff
@@ -134,141 +170,165 @@ const Review: React.FC<IProps> = ({ prevStage, locked, refresh }: IProps) => {
...
+          )
           setIsConfirmDialogOpen(true)
         }}
       >
         {({
           values,
           handleSubmit,
           setFieldValue,
-        }: FormikProps<{ sampleSize: string }>) => (
+        }: FormikProps<{ sampleSizes: { [key: string]: string }[] }>) => (
```
jonahkagan, line 270: why a list of objects? i think life might be simpler with a mapping from contest id to sample size, so `{ [contestId: string]: number }`

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
```diff
@@ -47,23 +57,31 @@ const Review: React.FC<IProps> = ({ prevStage, locked, refresh }: IProps) => {
...
   const history = useHistory()
   const [isConfirmDialogOpen, setIsConfirmDialogOpen] = useState(false)
 
   useEffect(() => {
     ;(async () => {
       try {
-        const { sampleSizes }: { sampleSizes: ISampleSizeOption[] } = await api(
+        const {
+          sampleSizes: options,
+        }: { sampleSizes: ISampleSizeOptions } = await api(
```
jonahkagan, line 71: what do you think about pulling this out into a hook and defining the type there? like the pattern we've been using for other API data sets

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
MorganLove: @jonahkagan Strings are what the form converts everything to, so it has a separate type for that so it's easier to track the conversion to strings going into the form and then remembering to convert back to numbers before sending it to the api.

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
MorganLove: @jonahkagan I might have been overcautious, but my concern here was that the call to the api to get the sample sizes will not have returned yet on the first run through this component. So `sampleSizeOptions` will only be the default `{}`. We're getting data from two asynchronous sources here, for the targeted contests and the sample sizes, and it needs both in order to create the `initialValues` value. Hence the double-layered ternary and fallback data.

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
MorganLove: @jonahkagan My concern was that we don't know the keys (the contest ids being the keys), and so having it in an array would make it more palatable for the form. But I ended up having to do weird stuff to get at it anyway, so I don't think it ended up simplifying it in the end. I'm going to try if it will work with just the object.

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
MorganLove: Probably a good idea. It would clean things up significantly and keep in pattern with everything else. Not sure why this one wasn't pulled out already, actually.

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
```diff
@@ -11,26 +11,25 @@ import useAuditSettings from '../../useAuditSettings'
 import useContests from '../../useContests'
 import { IContest, ISampleSizeOption } from '../../../../types'
 import useJurisdictions, { FileProcessingStatus } from '../../useJurisdictions'
-import { api, checkAndToast, testNumber } from '../../../utilities'
+import { testNumber } from '../../../utilities'
```
jonahkagan, line 14: Screen Shot 2020-07-20 at 1 07 23 PM

was playing around and triggered this error message (I think by typing a number and then deleting it). i think it's a pretty confusing message to a user

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
```diff
@@ -126,6 +93,22 @@ const Review: React.FC<IProps> = ({ prevStage, locked, refresh }: IProps) => {
...
 
+  const initialValues: IFormOptions = sampleSizeOptions
+    ? Object.keys(sampleSizeOptions).reduce(
+        (a, contestId) => ({
+          ...a,
+          [contestId]: sampleSizeOptions[contestId][0].size,
+        }),
+        {}
+      )
+    : {}
```
jonahkagan, line 104: what happens if two of the sample size options have the same size number? i think it would break the form.

this isn't a new issue - I think the same issue was there previous to this PR. makes me think that the data structure for sample sizes that the backend gives is not great. each sample size option should have some sort of key or id that you can use to uniquely identify it - the size number is not guaranteed to be uniquely identifying.

thoughts?

jonahkagan: another potential advantage to that approach would be that you could add in an option with a `custom` key so that you wouldn't need a separate state value to hold the custom size

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
```diff
@@ -136,139 +119,166 @@ const Review: React.FC<IProps> = ({ prevStage, locked, refresh }: IProps) => {
...
+            Object.keys(sizes).reduce((a, contestId) => {
+              if (sizes[contestId] === 'custom') {
+                return {
+                  ...a,
+                  [contestId]: customSampleSizes[contestId],
+                }
+              }
+              return { ...a, [contestId]: sizes[contestId] }
+            }, {})
+          )
```
jonahkagan, line 209: i notice that you use this pattern a few times. I've seen it abstracted out into a helper function called `mapValues` in previous codebases I've worked on, since it comes up a lot. that would look like this:

```ts
const mapValues = (obj, fn) =>
  Object.keys(obj).reduce((a, key) => ({
    ...a,
    [key]: fn(obj[key], key)
  }), {})

// or this might be a faster implementation, not that it matters so much
const mapValues = (obj, fn) =>
  Object.fromEntries(Object.entries(obj).map((key, val) => fn(val, key)))
```

and then this bit could just be:

```ts
setSampleSizes(mapValues(sizes, (size, contestId) =>
  sizes[contestId] === 'custom' ? customSampleSizes[contestId] : sizes[contestId]
))
```

client/src/components/MultiJurisdictionAudit/Setup/Review/useSampleSizes.ts
```diff
@@ -0,0 +1,107 @@
...
+      }),
+      {}
+    )
+  } catch (err) {
+    toast.error(err.message)
+    return null
+  }
+}
+
+const putSampleSizes = async (
```
jonahkagan, line 47: i think this should actually be called `postRound` and go in `useRounds`

client/src/components/MultiJurisdictionAudit/Setup/Review/useSampleSizes.ts
```diff
@@ -0,0 +1,107 @@
...
+): [
+  IStringSampleSizeOptions | null,
+  (sampleSizes: ISampleSizes) => Promise<boolean>
+] => {
+  const [
+    sampleSizeOptions,
+    setSampleSizeOptions,
+  ] = useState<IStringSampleSizeOptions | null>(null)
+
+  const uploadSampleSizes = async (sizes: ISampleSizes): Promise<boolean> => {
```
jonahkagan, line 89: similar to comment above, i think this makes more sense as `createRound` as part of `useRounds`

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
jonahkagan: i'm thinking something like this:
```ts
interface ISampleSizeOption {
  key: string
  size: number | null
  prob: number | null
}
```
a few examples:
```ts
{ key: "asn", size: 20, prob: 0.57 }
{ key: "prob90", size: 56, prob: 0.9 }
{ key: "custom", size: null, prob: null } // before the user sets a custom size
{ key: "custom", size: 43, prob: null } // after the user sets a custom size
```

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
```diff
@@ -1,53 +1,87 @@
...
+      body: JSON.stringify({
+        sampleSizes: { 'contest-id': 5 },
+        roundNum: 1,
+      }),
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      method: 'POST',
+    },
+  },
```
jonahkagan, line 63: one pattern i played with is making these API calls functions that take in whatever data you want to customize between tests. e.g. these three could become:

```ts
postRound: (sampleSizes) => ({
    url: '/api/election/1/round',
    response: { status: 'ok' },
    options: {
      body: JSON.stringify({
        sampleSizes,
        roundNum: 1,
      }),
      headers: {
        'Content-Type': 'application/json',
      },
      method: 'POST',
    },
  })
```

and then in the test you can say:
```ts
apiCalls.postRound({ 'contest-id': 46 })
```

dunno if it makes a big difference. curious what you think

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
```diff
@@ -81,213 +117,199 @@ beforeEach(() => {
...
-    await waitFor(() => expect(apiMock).toHaveBeenCalled())
-    expect(container).toMatchSnapshot()
+    const expectedCalls = [
+      apiCalls.getJurisdictions,
+      apiCalls.getJurisdictionFile,
+      apiCalls.getTargetedContests,
+      apiCalls.getSampleSizeOptions,
+    ]
+    await withMockFetch(expectedCalls, async () => {
+      const { container } = render(
```
jonahkagan, line 127: could potentially use the `renderWithRouter` test helper

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
```diff
@@ -81,213 +117,199 @@ beforeEach(() => {
...
-          roundNum: 1,
-        }),
-        headers: {
-          'Content-Type': 'application/json',
-        },
-        method: 'POST',
-      })
+      await screen.findByText('Review & Launch')
+      const launchButton = screen.getByText('Launch Audit')
+      userEvent.click(launchButton, { bubbles: true })
```
jonahkagan, line 231: i don't think you need `{ bubbles: true }` anymore

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
```diff
@@ -46,51 +45,19 @@ const Review: React.FC<IProps> = ({ prevStage, locked, refresh }: IProps) => {
...
-          roundNum: 1,
-        }),
-        headers: {
-          'Content-Type': 'application/json',
-        },
-      })
-      // checkAndToast left here for consistency and reference but not tested since it's vestigial
-      /* istanbul ignore next */
-      if (checkAndToast(result)) {
+      if (uploadSampleSizes(sampleSizes)) {
```
jonahkagan, line 55: missing `await` here?

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
MorganLove: @jonahkagan Hmmm... we should look at how we're handling that in the other validations. I think it crops up more than here, and will require a solution across the board.

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
MorganLove: @jonahkagan That has merit, yeah. I can picture how that would work on the frontend, and it would definitely clear up that particular bug. Would the `type` field possibly also be marked as 'custom'? It might make things easier on the frontend for generating the details of the different options, but I don't know if that would make it more complex on the backend.

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
MorganLove: @jonahkagan Oh yes, I remember that now. Good catch. :slightly_smiling_face:

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
jonahkagan: I was thinking of getting rid of the `type` field since currently it's only getting used for `asn`. I was thinking you could use the `key` field instead. If you're into this, I'll update my PR with the modified data format

MorganLove: @jonahkagan Oh right, I missed that you moved the `asn` to the key. That makes perfect sense. Go for it! :slightly_smiling_face:

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
MorganLove: @jonahkagan I find the ES6 format a bit more self-documenting and readable, but that's probably just me. And this will likely change a bit when you redo the api. I'll look at it again when I refactor for that.

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
```diff
@@ -136,139 +121,163 @@ const Review: React.FC<IProps> = ({ prevStage, locked, refresh }: IProps) => {
...
+              <tr key={c.id}>
+                <td>{c.name}</td>
+                <td>{c.jurisdictionIds.join(', ')}</td>
+              </tr>
+            ))}
+          </tbody>
+        </ContestsTable>
+      </ElevatedCard>
+      <br />
+      <H4>Estimated Sample Sizes</H4>
```
jonahkagan, line 193: I feel like this header is a bit misleading. The sample sizes are the actual sizes that are going to be used, they aren't just estimated. I feel like it should say "Select Sample Size" or just "Sample Size" or something like that. @mcchilders

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
```diff
@@ -136,139 +121,163 @@ const Review: React.FC<IProps> = ({ prevStage, locked, refresh }: IProps) => {
...
       <Formik
         initialValues={{
-          sampleSize: sampleSizeOptions.length ? sampleSizeOptions[0].size : '',
-          customSampleSize: '',
+          sampleSizes: initialValues,
```
jonahkagan, line 196: does this need to be nested, or could it just be `intialValues={initialValues}` now?

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
```diff
@@ -279,7 +288,9 @@ const Review: React.FC<IProps> = ({ prevStage, locked, refresh }: IProps) => {
                   locked ||
                   !isSetupComplete(jurisdictions, contests, auditSettings)
                 }
-                onClick={handleSubmit}
+                onClick={() => {
+                  handleSubmit()
+                }}
```
jonahkagan, line 293: unnecessary change?

client/src/components/MultiJurisdictionAudit/Setup/Review/useSampleSizes.ts
```diff
@@ -0,0 +1,123 @@
...
+import { ISampleSizeOption } from '../../../../types'
+
+export interface IStringSampleSizeOption {
+  size: string
+  key: string
+  prob: number | null
+}
+
+export interface IStringSampleSizeOptions {
+  [key: string]: IStringSampleSizeOption[]
```
jonahkagan, line 13: i think it's kind of confusing to call this `key` because `key` refers to something different in `IStringSampleSizeOption`. I think `contestId` would be clearer. same feedback for the other types below

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
```diff
@@ -1,53 +1,60 @@
...
+    },
+  }),
+  getEmptyJurisdictions: {
+    url: '/api/election/1/jurisdiction',
+    response: { jurisdictions: [] },
+  },
+  getJurisdictions: {
+    url: '/api/election/1/jurisdiction',
+    response: { jurisdictions: jurisdictionMocks.allManifests },
+  },
```
jonahkagan, line 44: could combine these like you did with `postRound`. same with the `getContests` calls below

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
```diff
@@ -81,213 +98,198 @@ beforeEach(() => {
...
-        body: JSON.stringify({
-          sampleSize: 46,
-          roundNum: 1,
-        }),
-        headers: {
-          'Content-Type': 'application/json',
-        },
-        method: 'POST',
-      })
+    auditSettingsMock.mockReturnValue(settingsMock.full)
```
jonahkagan, line 190: any reason this is mocked while all the other data sources use the mock API?

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
```diff
@@ -81,213 +98,198 @@ beforeEach(() => {
...
+    const expectedCalls = [
+      apiCalls.getJurisdictions,
+      apiCalls.getJurisdictionFile,
+      apiCalls.getTargetedContests,
+      apiCalls.getSampleSizeOptions,
+      apiCalls.postRound({ 'contest-id': 46 }),
+      apiCalls.getSampleSizeOptions,
```
jonahkagan, line 197: surprised that there's another call to `getSampleSizeOptions` after launching the round

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
```diff
@@ -81,213 +98,198 @@ beforeEach(() => {
...
+    ]
+    await withMockFetch(expectedCalls, async () => {
+      renderView()
+      await screen.findByText('Review & Launch')
+      const launchButton = screen.getByText('Launch Audit')
+      userEvent.click(launchButton)
+      await screen.findByText('Are you sure you want to launch the audit?')
+      const confirmLaunchButton = screen.getAllByText('Launch Audit')[1]
+      userEvent.click(confirmLaunchButton)
+      await waitFor(() => expect(refreshMock).toHaveBeenCalled())
```
jonahkagan, line 207: another option would be to find something on the next screen that indicates you were successfully redirected/the audit got launched. i suggest that because it's more in line with the philosophy of testing from the user's perspective

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
```diff
@@ -81,213 +98,198 @@ beforeEach(() => {
...
+      apiCalls.getSampleSizeOptions,
+    ]
+    await withMockFetch(expectedCalls, async () => {
+      renderView()
+      const newSampleSize = await screen.findByText(
+        'Enter your own sample size (not recommended)'
+      )
+      userEvent.click(newSampleSize)
+      const customSampleSizeInput = await screen.findByRole('textbox')
+      fireEvent.change(customSampleSizeInput, { target: { value: '40' } }) // userEvent has a problem with this field due to the lack of an explicit value field: https://github.com/testing-library/user-event/issues/356
```
jonahkagan, line 273: that's annoying. why doesn't it have a value field? isn't it a regular `input`?

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
MorganLove: @jonahkagan Formik needs a root object to name things, and if we made that the undeterministic contest ids, that would make things more confusing for referencing within the code, so I think this is the best way to do it.

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
MorganLove: @jonahkagan Oh good catch. I was adding in a console.log there to check that it was submitting when I was testing things, but forgot to fix it all the way when I removed the log.

client/src/components/MultiJurisdictionAudit/Setup/Review/useSampleSizes.ts
MorganLove: @jonahkagan I actually forgot you could use something other than `key` there, haha.

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
MorganLove: @jonahkagan It's wrapped a bunch of times, but ultimately it's the [input group element from Blueprint](https://blueprintjs.com/docs/#core/components/text-inputs). And from examining how they handle things, it looks like the fields don't have a value attribute until you start entering data into them. Which would explain why userEvent is getting confused.

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
MorganLove: @jonahkagan It was either check that the mock for `refresh` was called, or that the mock for `history.push` was called. The former just seemed simpler. Both happen when it is successfully launched, though.

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
MorganLove: @jonahkagan That's due to me copy-pasting the pattern from one of the other api hooks, where you reload the data from the server after posting the update. In this context, though, I don't think that's necessary, so should change that probably.

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
MorganLove: @jonahkagan I think that's a holdover from the old tests, before mocking the api was simplified. The settings were tested directly, and adding those mocks into the mix would have made a horrible mess. We could integrate the settings into the sample size api mocks now, I think, though this is still cleaner imo.

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
jonahkagan: oh interesting, `withMockFetch` is supposed to fail if its given expected API calls that don't occur

client/src/components/MultiJurisdictionAudit/Setup/Review/index.test.tsx
jonahkagan: oh right, forgot there were no other routes to navigate to in this test setup

Issue #631, opened 2020-07-06
Require all db queries filter by election


jonahkagan:
See https://github.com/votingworks/arlo/pull/629#discussion_r449295041

mcchilders:
This is for test cases that will prevent this from happening (uncaught) again

jonahkagan:
Closing in favor of votingworks/arlo#650

PR #646, opened 2020-07-15
Check db queries filter by election

jonahkagan:
Task: #631 

Here's a first attempt at ensuring that all db queries filter by election. We use a SQLAlchemy event hook that runs every time a query is going to be compiled to check that the query has either:
- a filter on a primary key
- a filter on election id
- a filter on some foreign key to another object that is scoped to an election

This solution feels brittle for a few reasons:
- it relies heavily on SQLAlchemy internals that I don't understand well
- I ran into a few edge cases with the queries we have already, which makes me think we'll have to come back and address more edge cases in the future
- we have half a permissions scheme already in place with the route decorators that check and populate election/jurisdiction/audit board objects based on the user's permissions - this outsources the rest of the permission checking (that we correctly use those objects in the rest of the code) to a totally unrelated piece of code. it seems like something might slip through the cracks

In summary, I think this could work, but the process of implementing it has made me realize that really this is a piece of a larger permissions problem, and there is probably a more robust solution. I'm thinking about creating an abstraction layer on top of the database that provides a "view" into the database from a particular user's perspective, offering some interface similar to SQLAlchemy models that enforces proper scoping in its design.

### DIFF
--- a/server/check_query_scope.py
+++ b/server/check_query_scope.py
@@ -1,0 +1,139 @@
+from typing import Dict
+from sqlalchemy import event, sql
+from sqlalchemy.orm import Query
+from . import models
+from .models import *  # pylint: disable=wildcard-import
+
+# Every database query the app makes should be scoped to the audit for the
+# logged-in user. This module registers an event listener with SQLAlchemy to
+# check every database query to ensure that it is properly scoped.
+#
+# It checks that each query filters by either:
+# - a primary key
+# - election_id
+# - a foreign key to a model that is related to Election
+#
+# In order to opt out of this check, a query can specify:
+# .execution_options(query_across_elections=True)
+
+# This dict defines which columns satisfy the requirement for each model. Note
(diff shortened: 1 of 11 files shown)


server/check_query_scope.py
```diff
@@ -0,0 +1,139 @@
...
+if checked_models != all_models:
+    # pylint: disable=invalid-name
+    missing_models = ",".join(m.__name__ for m in all_models - checked_models)
+    raise Exception(f"Missing models from check_query_scope: {missing_models}")
+
+# Basic global variable lock - Python's threading locks don't seem to work
+# easily here. We need to make sure only one event handler can operate at a
+# time because when we try to print query.statement, it triggers the handler
+# again, leading to an infinite loop.
+check_query_scope_lock = False  # pylint: disable=invalid-name
```
benadida, line 72: is there no way to do this other than with a global?

jonahkagan:
Closing in favor of #648

PR #648, opened 2020-07-20
View layer for db permissions

jonahkagan:


server/api/rounds.py
```diff
@@ -226,6 +227,7 @@ def list_rounds_jurisdiction_admin(
 @api.route("/election/<election_id>/round", methods=["POST"])
 @with_election_access
 def create_round(election: Election):
+    view = ElectionView(election.id)
```
jonahkagan, line 230: if we went forward, this view would be populated by the decorator

benadida:
Overall, this is a much more intrusive approach than the other one. I worry that when it comes to the complex queries, this might get tricky. What do you think?

Issue #595, opened 2020-06-02
Initial MVP Cypress proof of concept for End-to-end tests for multi-jurisdiction flow


jonahkagan:
Test plan: https://docs.google.com/document/d/1nrVd45l5PcG19d4nWoHKtEe-s3fkZRNGj0JMuvMSDTY/edit#

MorganLove:
@Jonah-Kagan Just renamed this based off the convo on slack. If you could create the initial PR with noauth set up we can tag team on creating a short happy path through creating an audit or something like that to make sure we like it. How does that sound?

jonahkagan:
Sounds great, I'll prioritize setting up the basics today

PR #653, opened 2020-07-22
Cypress set up

jonahkagan:
Task: #595 

Sets up a very basic Cypress test to run locally and on CI.


README.md
```diff
@@ -179,3 +160,18 @@ For testing:
   require a password by editing `/etc/postgresql/10/main/pg_hba.conf` and
   changing `md5` to `trust` for both the IPv4 and IPv6 local connections
   settings, and then restart postgres via `sudo systemctl restart postgresql`.
+
+### Testing
```
MorganLove, line 164: @jonahkagan Perhaps add in somewhere in here that [the dependencies for cypress might need to be installed](https://docs.cypress.io/guides/getting-started/installing-cypress.html#System-requirements).

client/cypress/seed-test-db.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+export FLASK_ENV=test
+trap 'kill 0' SIGINT SIGHUP
+pushd ..
```
MorganLove, line 5: This wouldn't work for me unless I changed it to `../..`. It couldn't find the scripts modules because it was looking relative to the /client folder instead of /arlo.

client/run-cypress-tests.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+export ARLO_AUDITADMIN_AUTH0_BASE_URL=https://votingworks-noauth.herokuapp.com
+export ARLO_JURISDICTIONADMIN_AUTH0_BASE_URL=https://votingworks-noauth.herokuapp.com
+
+trap 'kill 0' SIGINT SIGHUP
+./cypress/seed-test-db.sh
+pushd ..
+FLASK_ENV=test ./run-dev.sh &
```
MorganLove, line 9: This leaves `run-dev.sh` running after cypress has finished, which makes it fail on the second run since the 3000 port is still occupied by the previous process. I have to run `run-dev.sh` again, which fails but gives me the pid of the blocking process, then kill that process, and then I'm free to run it again. Perhaps add `EXIT` to the trap here? (I tried that and it worked from what I could tell.)

README.md
jonahkagan: was there something in particular you needed to install?

client/run-cypress-tests.sh
jonahkagan: Hmm yeah I was struggling with this a lot. I think I remember that I had EXIT in there but then it was giving me problems because it was failing on CI (for some reason trapping EXIT made Circle think the test step failed). I'll mess with it some more

README.md
MorganLove: I didn't notice what it was, but it installed something when I ran it when I was troubleshooting. So I actually don't know if it was part of what caused me problems or not, haha. I just thought it would likely be handy eventually for someone.

client/run-cypress-tests.sh
MorganLove: Ah yeah, I wondered if it might do that. Perhaps trapping a EXIT on a different script?

client/cypress/seed-test-db.sh
jonahkagan: Ok I think i fixed this by making sure each script first `cd`s to the directory it lives in

PR #659, opened 2020-07-28
Superadmin: add button to delete election

jonahkagan:
Motivation: clean up the staging database. Could be useful in production someday too

### DIFF
--- a/server/superadmin.py
+++ b/server/superadmin.py
@@ -1,4 +1,5 @@
 from flask import render_template, redirect, request, Blueprint
+from werkzeug.exceptions import Forbidden
 
 from .models import *  # pylint: disable=wildcard-import
 from .database import db_session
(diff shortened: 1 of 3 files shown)


jonahkagan:
Good call, I will limit it to non-prod envs for now

server/superadmin.py
```diff
@@ -47,3 +50,14 @@ def superadmin_jurisdictionadmin_login():
     user_email = request.form["email"]
     set_loggedin_user(UserType.JURISDICTION_ADMIN, user_email)
     return redirect("/")
+
+
+@superadmin.route("/superadmin/delete-election/<election_id>", methods=["POST"])
+@with_superadmin_access
+def superadmin_delete_election(election_id: str):
+    if FLASK_ENV == "production":  # pragma: no cover
+        raise Forbidden("Can't delete audits in production")
```
jonahkagan, line 59: @benadida how does this look?

benadida: looks great!

Issue #652, opened 2020-07-22
API for multi-jurisdiction offline ballot data entry


mcchilders:
If you're using offline data entry:
- Audit board only retrieves and tallies, does not do online data entry.
- Don't need to generate credentials
- Can't require AB signoff to complete a round
Also need to:
- Make sure audit report shows aggregates only, not ballot by ballot data. (should just be leaving the ballot by ballot data out?)

jonahkagan:
```
POST /election/:electionId/jurisdiction/:jurisdictionId/round/:roundId/results
{
  "contest-id-1": {
    "contest-1-choice-id-1": 10,
    "contest-1-choice-id-2": 15,
    ...
  },
  "contest-id-2": ...
}
```

PR #662, opened 2020-07-29
Record offline results (backend)

jonahkagan:
Task: #652 

Create a new API endpoint to record offline audit results by jurisdiction:
`/election/<election_id>/jurisdiction/<jurisdiction_id>/round/<round_id>/results`

Changes:
- Add a new JurisdictionResult model to record the result for each jurisdiction+round+contest
- Create a migration for this change
- Add the new endpoint handler
- Modify code that checks if a round is over/aggregates vote counts to handle offline audits
- Add a basic integration test for an offline audit and unit tests for the new route
- Updates the audit report


jonahkagan:
@benadida I've got some more to do on this (see PR description), but would love a first pass review. There's some stuff in here that feels important to get correct so I'd love to know if you see a need to test any specific cases now, since I'm going back to add more tests soon.

server/api/audit_boards.py
```diff
@@ -255,105 +256,6 @@ def set_audit_board_members(
...
-        .join(SampledBallot)
-        .group_by(AuditBoard.id)
-        .count()
-    )
-    return (
-        num_jurisdictions_without_audit_boards_set_up == 0
-        and num_audit_boards_with_ballots_not_signed_off == 0
-    )
-
-
```
jonahkagan, line 356: moved to a diff file

server/api/rounds.py
```diff
@@ -1,41 +1,19 @@
...
 
 from . import api
 from ..database import db_session
 from ..models import *  # pylint: disable=wildcard-import
 from ..auth import with_election_access, with_jurisdiction_access
-from .sample_sizes import sample_size_options
+from .sample_sizes import sample_size_options, cumulative_contest_results
 from ..util.isoformat import isoformat
 from ..util.group_by import group_by
-from ..audit_math import sampler
```
jonahkagan, line 14: reorganized this file a bit

server/api/rounds.py
```diff
@@ -62,25 +40,151 @@ def is_audit_complete(round: Round):
...
+            # sampled for.
+            else:
+                vote_counts = dict(
+                    interpretations_query.values(
+                        ContestChoice.id, func.count(SampledBallot.id.distinct())
+                    )
+                )
+
+        # For offline audits, sum the JurisdictionResults
+        else:
```
jonahkagan, line 77: this is the only new case in this function

server/api/rounds.py
```diff
@@ -62,25 +40,151 @@ def is_audit_complete(round: Round):
...
+            .count()
+        )
+        return (
+            num_jurisdictions_without_audit_boards_set_up == 0
+            and num_audit_boards_with_ballots_not_signed_off == 0
+        )
+
+    # For offline audits, check that we have results recorded for every
+    # jurisdiction that had ballots sampled
+    else:
```
jonahkagan, line 161: this is the only new case in this function

server/migrations/versions/22c615fe67ab_jurisdictionresult.py
```diff
@@ -0,0 +1,67 @@
+# pylint: disable=invalid-name
```
jonahkagan, line 1: this was completely auto-generated. i didn't have to change anything!

server/models.py
```diff
@@ -499,6 +500,28 @@ class RoundContestResult(BaseModel):
...
+    )
+    contest_choice_id = Column(
+        String(200),
+        ForeignKey("contest_choice.id", ondelete="cascade"),
+        nullable=False,
+    )
+    result = Column(Integer, nullable=False)
+
+    __table_args__ = (
+        PrimaryKeyConstraint("round_id", "jurisdiction_id", "contest_choice_id"),
```
jonahkagan, line 521: @benadida do you think it's worth having `contest_id` in here as well? I noticed that you didn't include it in the primary key for `RoundContestResult`, so was following that pattern, but part of me feels like it would be clearer to say that it is part of the primary key

nealmcb:
This looks very apropos of what I've been wondering about. I ran an audit from the main branch and it asked for individual ballots (as if it was in online mode) even though I used the default of offline mode. I was about to enter an issue on it.

So was offline mode not working for some period of time?
And this code will fix that?
Thanks.

server/models.py
benadida: I think you can go either way, but if you include in the primary key some keys that effectively have redundant info, I think it's important to have those keys foreign-reference the table that ensures they are internally consistent.

jonahkagan: Ok makes sense. I'll prob leave it out.

But your comment made me think: is it worth adding a constraint on `contest_id` and `jurisdiction_id` to make sure they're related via the `contest_jurisdiction` table?

server/api/rounds.py
```diff
@@ -62,25 +40,151 @@ def is_audit_complete(round: Round):
     return all(c.is_complete for c in targeted_round_contests)
 
 
-# Raises if invalid
-def validate_round(round: dict, election: Election):
-    validate(round, CREATE_ROUND_REQUEST_SCHEMA)
+def count_audited_votes(election: Election, round: Round):
+    for round_contest in round.round_contests:
```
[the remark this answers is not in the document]
jonahkagan: this function is called from two routes (audit board signoff, record offline results). The routes should already validate this. But I think I missed that validation in the new record offline results route, so I'll add it

server/api/reports.py
```diff
@@ -237,22 +237,27 @@ def sampled_ballot_rows(election: Election, jurisdiction: Jurisdiction = None):
...
-                list(ballot.interpretations), list(election.contests)
-            ),
+            + (
+                [ballot.status]
+                + pretty_interpretations(
+                    list(ballot.interpretations), list(election.contests)
+                )
+                if election.online
+                else []
+            )
```
jonahkagan, line 260: new changes since last review

server/models.py
```diff
@@ -499,6 +501,33 @@ class RoundContestResult(BaseModel):
...
+        ForeignKeyConstraint(
+            ["contest_id", "jurisdiction_id"],
+            ["contest_jurisdiction.contest_id", "contest_jurisdiction.jurisdiction_id"],
+            ondelete="cascade",
+        ),
```
jonahkagan, line 527: new constraint since last review

server/tests/test_offline_data_entry.py
```diff
@@ -0,0 +1,292 @@
...
+            for contest in contests
+        },
+    )
+    assert rv.status_code == 400
+    assert json.loads(rv.data) == {
+        "errors": [{"errorType": "Bad Request", "message": "Invalid contest ids"}]
+    }
+
+
+def test_offline_cant_create_audit_boards(
```
jonahkagan, line 274: new test since last review

benadida:
> So was offline mode not working for some period of time?

Hi Neal: offline was not working in multi-jurisdiction mode, because we had never made it work until now. But it was always working in single-jurisdiction pilot mode.

PR #665, opened 2020-08-03
Offline data entry [FE]

MorganLove:
**Description**

- Renders the RoundDataEntry component instead of the RoundProgress component when the audit is offline
- Added the forms for data entry

**Testing**

- So far just the bare basics, needs more

**Progress**

- Doesn't work as of yet because this requires the JA to have access to the settings and contests apis, which they didn't need on the online flow.

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/RoundManagement/RoundDataEntry.tsx
+++ b/client/src/components/MultiJurisdictionAudit/RoundManagement/RoundDataEntry.tsx
@@ -1,14 +1,20 @@
 import React from 'react'
+import { toast } from 'react-toastify'
 import { useParams } from 'react-router-dom'
 import { H5 } from '@blueprintjs/core'
 import { Field, Formik, Form, FormikProps } from 'formik'
+import styled from 'styled-components'
 import H2Title from '../../Atoms/H2Title'
 import useContestsJurisdictionAdmin from './useContestsJurisdictionAdmin'
 import { IRound } from '../useRoundsJurisdictionAdmin'
 import Card from '../../Atoms/SpacedCard'
 import FormField from '../../Atoms/Form/FormField'
 import FormButton from '../../Atoms/Form/FormButton'
-import { testNumber } from '../../utilities'
+import { testNumber, api } from '../../utilities'
+
+const BottomButton = styled(FormButton)`
+  margin: 30px 0;
+`
 
 interface IValues {
   results: {
@@ -41,8 +47,38 @@ const RoundDataEntry = ({ round }: IProps) => {
       )
     : null
 
-  const submit = (values: IValues) => {
-    console.log(values)
+  const submit = async ({ results: r }: IValues) => {
+    const body = JSON.stringify(
+      Object.keys(r).reduce(
+        (a, contestId) => ({
+          ...a,
+          [contestId]: Object.keys(r[contestId]).reduce(
+            (b, choiceId) => ({
+              ...b,
+              [choiceId]: Number(r[contestId][choiceId]),
+            }),
+            {}
+          ),
+        }),
+        {}
+      )
+    )
+    try {
+      await api(
+        `/election/${electionId}/jurisdiction/${jurisdictionId}/round/${round.id}/results`,
+        {
+          method: 'PUT',
+          headers: {
+            'Content-Type': 'application/json',
+          },
+          body,
+        }
+      )
+      return true
+    } catch (err) {
+      toast.error(err.message)
+      return false
+    }
   }
 
   if (!results) return null
@@ -77,10 +113,9 @@ const RoundDataEntry = ({ round }: IProps) => {
                 ))}
               </Card>
             ))}
-          <br />
-          <FormButton type="submit" intent="primary" onClick={handleSubmit}>
+          <BottomButton type="submit" intent="primary" onClick={handleSubmit}>
             Submit Data for Round {round.roundNum}
-          </FormButton>
+          </BottomButton>
         </Form>
       )}
     </Formik>
--- a/client/src/components/MultiJurisdictionAudit/RoundManagement/RoundDataEntry.tsx
+++ b/client/src/components/MultiJurisdictionAudit/RoundManagement/RoundDataEntry.tsx
@@ -3,16 +3,12 @@ import { useParams } from 'react-router-dom'
 import { H5 } from '@blueprintjs/core'
 import { Field, Formik, Form, FormikProps } from 'formik'
 import H2Title from '../../Atoms/H2Title'
-import useContests from '../useContests'
+import useContestsJurisdictionAdmin from './useContestsJurisdictionAdmin'
 import { IRound } from '../useRoundsJurisdictionAdmin'
 import Card from '../../Atoms/SpacedCard'
 import FormField from '../../Atoms/Form/FormField'
 import FormButton from '../../Atoms/Form/FormButton'
 import { testNumber } from '../../utilities'
-
-interface IParams {
-  electionId: string
-}
 
 interface IValues {
   results: {
@@ -27,8 +23,11 @@ interface IProps {
 }
 
 const RoundDataEntry = ({ round }: IProps) => {
-  const { electionId } = useParams<IParams>()
-  const [contests] = useContests(electionId)
+  const { electionId, jurisdictionId } = useParams<{
+    electionId: string
+    jurisdictionId: string
+  }>()
+  const contests = useContestsJurisdictionAdmin(electionId, jurisdictionId)
 
   const results = contests
     ? contests.reduce(
--- a/client/src/components/MultiJurisdictionAudit/RoundManagement/RoundDataEntry.tsx
+++ b/client/src/components/MultiJurisdictionAudit/RoundManagement/RoundDataEntry.tsx
@@ -1,11 +1,25 @@
 import React from 'react'
 import { useParams } from 'react-router-dom'
+import { H5 } from '@blueprintjs/core'
+import { Field, Formik, Form, FormikProps } from 'formik'
 import H2Title from '../../Atoms/H2Title'
 import useContests from '../useContests'
 import { IRound } from '../useRoundsJurisdictionAdmin'
+import Card from '../../Atoms/SpacedCard'
+import FormField from '../../Atoms/Form/FormField'
+import FormButton from '../../Atoms/Form/FormButton'
+import { testNumber } from '../../utilities'
 
 interface IParams {
   electionId: string
+}
+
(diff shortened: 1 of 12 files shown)


client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
```diff
@@ -95,16 +97,22 @@ const RoundManagement = ({ round, auditBoards, createAuditBoards }: IProps) => {
...
-            <FormButton
-              verticalSpaced
-              onClick={() => downloadDataEntry(auditBoards)}
-            >
-              Download Audit Board Credentials for Data Entry
-            </FormButton>
-            <QRs passphrases={auditBoards.map(b => b.passphrase)} />
-            <RoundProgress auditBoards={auditBoards} round={round} />
-            {/* {online ? <RoundProgress /> : <RoundDataEntry />} */}
+            {online && (
```
jonahkagan, line 100: Also offline audits shouldn't have to create audit boards.

client/src/components/MultiJurisdictionAudit/RoundManagement/RoundDataEntry.tsx
```diff
@@ -1,8 +1,90 @@
...
+      )
+    : null
+
+  const submit = (values: IValues) => {
+    console.log(values)
+  }
+
+  if (!results) return null
+  return (
+    <Formik initialValues={{ results }} enableReinitialize onSubmit={submit}>
```
jonahkagan, line 51: for validation, the backend requires that you enter a result for every choice for every contest. dunno if that's the right thing to do, but it seemed like the simplest place to start. might want to require that on the frontend as well

client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
MorganLove: @jonahkagan that's what I initially thought, but @mcchilders updated that part, unless I'm misinterpreting her.

client/src/components/MultiJurisdictionAudit/RoundManagement/RoundDataEntry.tsx
MorganLove: @jonahkagan It tests each field for being a number, which also requires them to be filled.

client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
MorganLove: For now, just so I can test the behavior, I've sidestepped audit board creation.

MorganLove:
Had to upgrade `styled-components` because of this: https://github.com/microsoft/TypeScript/issues/37597

client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
```diff
@@ -56,22 +61,27 @@ const RoundManagement = ({ round, auditBoards, createAuditBoards }: IProps) => {
     auditBoards,
   ])
 
-  // const [{ online }] = useAuditSettings(electionId!)
+  const settings = useAuditSettingsJurisdictionAdmin(electionId, jurisdictionId)
+
+  const online = settings ? settings.online : false
```
jonahkagan, line 66: maybe `return null` if settings are still loading (like we do with ballots below)?

MorganLove: I changed it to do something similar.

MorganLove:
@jonahkagan Okay, this one is based off #677 now and I abstracted out the endpoints into a new hook `useResults` and was able to clean it up nicely. There's some new refactoring of the types and data transformation functions in that hook now too.

client/src/components/MultiJurisdictionAudit/RoundManagement/RoundDataEntry.tsx
```diff
@@ -1,8 +1,103 @@
...
+    jurisdictionId: string
+  }>()
+  const contests = useContestsJurisdictionAdmin(electionId, jurisdictionId)
+  const [results, updateResults] = useResults(
+    electionId,
+    jurisdictionId,
+    round.id
+  )
+  const alreadySubmittedResults =
+    !!results &&
```
jonahkagan, line 40: what's the `!!` for here?

jonahkagan: potentially could move L47 up above this and cut out this check

client/src/components/MultiJurisdictionAudit/RoundManagement/RoundDataEntry.tsx
```diff
@@ -1,8 +1,103 @@
...
+                    htmlFor={`results[${contest.id}][${choice.id}]`}
+                  >
+                    Votes for {choice.name}:
+                    {alreadySubmittedResults ? (
+                      results.results[contest.id][choice.id]
+                    ) : (
+                      <Field
+                        id={`results[${contest.id}][${choice.id}]`}
+                        name={`results[${contest.id}][${choice.id}]`}
+                        disabled={round.endedAt || alreadySubmittedResults}
```
jonahkagan, line 78: this won't render if `alreadySubmittedResults` is `true`, so maybe not necessary here? or you could take out the conditional rendering on L72 and just render a disabled field

jonahkagan: also every JA has to submit results for the round to end, so it might be sufficient to just check `alreadySubmittedResults` here

client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
```diff
@@ -56,22 +61,28 @@ const RoundManagement = ({ round, auditBoards, createAuditBoards }: IProps) => {
...
   ])
 
-  // const [{ online }] = useAuditSettings(electionId!)
+  const { online } = useAuditSettingsJurisdictionAdmin(
+    electionId,
+    jurisdictionId
+  )
 
-  if (!ballots) return null // Still loading
+  if (!ballots || online === null) return null // Still loading
```
jonahkagan, line 69: nice

client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
```diff
@@ -56,22 +61,28 @@ const RoundManagement = ({ round, auditBoards, createAuditBoards }: IProps) => {
...
-        <H2Title>Round {roundNum} Audit Board Setup</H2Title>
-        <CreateAuditBoards
-          auditBoards={auditBoards}
-          createAuditBoards={createAuditBoards}
-          numBallots={ballots.length}
-          roundNum={roundNum}
-        />
-        {auditBoards.length > 0 && (
+      <PaddedWrapper className="single-page left">
+        {auditBoards.length === 0 ? (
```
jonahkagan, line 75: what's this change for?

client/src/components/MultiJurisdictionAudit/RoundManagement/useAuditSettingsJurisdictionAdmin.ts
```diff
@@ -0,0 +1,45 @@
...
+const useAuditSettingsJurisdictionAdmin = (
+  electionId: string,
+  jurisdictionId: string
+): IAuditSettings => {
+  const [settings, setSettings] = useState<IAuditSettings>(defaultValues)
+
+  useEffect(() => {
+    ;(async () => {
+      const newSettings = await getSettings(electionId, jurisdictionId)
+      setSettings(newSettings || defaultValues)
```
jonahkagan, line 38: why return default values here as opposed to just whatever the response is like you do in the `useContestsJurisdictionAdmin` hook below?

client/src/components/MultiJurisdictionAudit/RoundManagement/useResults.ts
```diff
@@ -0,0 +1,92 @@
...
+const getResults = async (
+  electionId: string,
+  jurisdictionId: string,
+  roundId: string
+): Promise<IResults | null> => {
+  try {
+    const response: IResultValues = await api(
+      `/election/${electionId}/jurisdiction/${jurisdictionId}/round/${roundId}/results`
+    )
+    return { results: reformatResults(response, false) }
```
jonahkagan, line 47: any reason that you're wrapping the results in `{ results: ... }` here? seems like it just makes for more unwrapping work in the component that consumes this value

client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
MorganLove: @jonahkagan The `PaddedWrapper` is created above to add a bottom margin: https://github.com/votingworks/arlo/pull/665/files/3e594ad5164b6998640979359b547715cdf67ac8#diff-4684ee384072240a8164a08c77c1e3c7R23

client/src/components/MultiJurisdictionAudit/RoundManagement/useAuditSettingsJurisdictionAdmin.ts
MorganLove: @jonahkagan Because I'm destructuring the `online` value from it to use, so it makes things simpler in the other component to always have that value on the object, even if it's defined as `null` to indicate it's still loading.

client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
jonahkagan: oops, should have been more specific, I meant L75 - checking `auditBoards.length`

client/src/components/MultiJurisdictionAudit/RoundManagement/useAuditSettingsJurisdictionAdmin.ts
jonahkagan: that makes sense. might be simpler overall to follow the pattern of the other hooks and destructure the `online` value after you check to see if the whole response is `null`. that way a `null` value for a setting doesn't get confused with it still loading. it's a hypothetical issue for now so not a big deal

client/src/components/MultiJurisdictionAudit/RoundManagement/useResults.ts
MorganLove: @jonahkagan It needs to be there for the form eventually, but I tried moving the wrapping and typing around a bit again and was able to localize that closer to the form itself, which did indeed clean things up more everywhere else. Thanks!

PR #675, opened 2020-08-04
Parallel tests

jonahkagan:
Uses pytest-xdist to run the server tests in parallel. We use one database throughout most of the tests, which also simulates the real-world a bit better and helps ensure we are scoping data correctly.

In working on this PR, I found a race condition in bgcompute - when computing sample sizes, we can accidentally sample ballots twice for the same contest (for round > 1). So I copied the locking pattern from `process_file` to lock the RoundContest object.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ jobs:
       - run:
           name: server tests
           command: |
-            make test-server-coverage
+            pipenv run pytest -n 2 --cov=.
       - run:
           name: client tests
           command: |
--- a/server/bgcompute.py
+++ b/server/bgcompute.py
@@ -1,4 +1,5 @@
 import time
+from sqlalchemy import update
 
 from server.database import db_session
 from server.models import *  # pylint: disable=wildcard-import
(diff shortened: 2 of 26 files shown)


.circleci/config.yml
```diff
@@ -45,7 +45,7 @@ jobs:
       - run:
           name: server tests
           command: |
-            make test-server-coverage
+            pipenv run pytest -n 2 --cov=.
```
umbernhard, line 48: Is there a reason this is set to 2 and not `auto`?

server/tests/conftest.py
```diff
@@ -29,21 +30,43 @@
...
+
+
 @pytest.fixture
 def client() -> Generator[FlaskClient, None, None]:
     app.config["TESTING"] = True
     client = app.test_client()
 
-    reset_db()
+    # TODO run this once at the beginning of all tests?
+    # reset_db()
```
umbernhard, line 60: We should probably remove these two lines ?

.circleci/config.yml
jonahkagan: Apparently on CircleCI it thinks there are 32 CPUs 🤦 so it runs way too many workers and a bunch of them crash and the tests fail. Currently, our CircleCI setup only provides 2 CPUs, so I hardcoded to 2

jonahkagan: See: https://ideas.circleci.com/ideas/CCI-I-578

server/tests/conftest.py
jonahkagan: Good call. I was actually able to simplify this fixture a lot now that I look at it: 260f44e

server/bgcompute.py
```diff
@@ -32,6 +33,21 @@ def bgcompute_compute_round_contests_sample_sizes():
...
+            # to an "in progress" value This acts like a lock on the
+            # RoundContest so it can only get processed once (since this
+            # process has a side effect of sampling ballots in some cases,
+            # which we only want to happen once).
+            result = db_session.execute(
+                update(RoundContest.__table__)  # pylint: disable=no-member
+                .where(RoundContest.contest_id == round_contest.contest_id)
+                .where(RoundContest.round_id == round_contest.round_id)
+                .where(RoundContest.sample_size_options.is_(None))
+                .values(sample_size_options="CALCULATION_IN_PROGRESS")
```
benadida, line 46: a better way to do this would be with a `select for update` that would be cleared automatically if an error occurs. I'm okay with this for now, but worth further improvement later.

jonahkagan: ok, hopefully this code will not exist much longer! it's single-jurisdiction flow stuff

Issue #670, opened 2020-08-04
Backend: Audit type field


jonahkagan:
Should accept "ballot polling" or "batch comparison"

- Set in create audit API
- Returned in election settings API call or /me (depends on where it's needed)

PR #676, opened 2020-08-05
Audit type

jonahkagan:
Task: #670 

- Remove unused Election fields election_date, election_type, meeting_date
- Add Election.audit_type field
- Add auditType field to the new election endpoint
- Migration for Election.audit_type
- Return audit type in election settings endpoint


Pipfile
```diff
@@ -14,10 +14,10 @@ pylint = "*"
 # for now (or enable pre-releases for all packages, which we don't want)
 black = "==19.10b0"
 snapshottest = "*"
-alembic = "*"
 pytest-alembic = "*"
 
 [packages]
+alembic = "*"
```
jonahkagan, line 20: needed for running migrations on staging/prod

MorganLove:
@jonahkagan The `PUT /api/election/<electionId>/settings` endpoint needs to accept the `auditType` value as well. The way it works currently, it has to pull the old settings object, merge with the new object (which is never the whole object, just slices of it), and submit the new merged object. The hook returns the whole object, so it's accessible in the different stages in whatever form they need. Until this is fixed I have to exclude it from the api call and add it back in again for the state update.

Issue #657, opened 2020-07-28
Document Auth0 configuration steps for tenants, apps etc.


nealmcb:
Arlo now relies exclusively on https://auth0.com/ for login and authentication.
Setting up an tenant, configuring its Auth0 Dashboard, connecting it with the application, etc are tricky but important steps.
They need to be documented.

#656 is a start.

These steps should be included (and corrected if necessary):

* Create an app for Arlo as a single page application(?)
* Copy down the Domain, Client ID and Client Secret from the app.
* Configure the app's `Allowed Callback URLs
` with something like `http://localhost:3000/auth/auditadmin/callback`.

What is the relationship between an "organization" as documented in the README and a "Jurisdiction" like a county within a State, and a "Jurisdiction" as listed in a `jurisdiction_filesheet.csv` as uploaded by an Arlo administrator?

Do the email addresses used in `create-admin` have to match tenant Auth0 ids, or are the generated UUIDs used when matching somehow or what?

For testing purposes, can the same email address and/or Auth0 tenant be both an Arlo admin and a Jurisdiction admin?

More config is evidently necessary to get around the 404 Not Found errors I'm getting when I click on
`Log in as a Jurisdiction Admin` and get to
http://localhost:3000/authorize?response_type=code&client_id=&redirect_uri=http%3A%2F%2Flocalhost%3A3000%2Fauth%2Fju..

benadida:
@nealmcb Auth0 is not required, just any OAuth server is. Auth0 is a good one, though. I wouldn't want to add Auth0-specific documentation to this repository when Auth0 does a fine job of explaining their own product.

Agreed that we could use more documentation on what needs to be configured, though it may take us a little bit of time to get to it.

nikhilb4a:
Closing as not planned due to age.

PR #680, opened 2020-08-09
Update and clarify README.md and auth0.md

nealmcb:
**Description**

As noted in #657, documentation is lacking for the new authentication and authorization approach
in Arlo, which now relies on OAuth 2 and/or Auth0.

This is my attempt to clarify what seems to work at this point. It should be reviewed by someone who can confirm these changes.

**Testing**

Only documentation is changed.

**Progress**

It would probably make sense to put the Organizations section before the Configuration section, to introduce the use of administrators and Auth0.

This PR should be helpful and self-contained, but before the issue can be closed, more detail on how to configure Auth0 should be added.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,8 +31,9 @@ dev-environment: deps initdevdb install-development resetdb
 typecheck-server:
 	pipenv run mypy server scripts
 
+# Note that this reformats ALL the Python files in the given directories, in-place
 format-server:
-	pipenv run black .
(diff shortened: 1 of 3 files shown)


nealmcb:
Also has a tweak to Makefile to help prevent unexpected modifications by black
to Python files stashed away in various places in the current directory.

jonahkagan:
Closed this to clean up longstanding open PRs

PR #683, opened 2020-08-11
adding front end support for batch comparison audit creation

### LINKED ISSUE #638: Add audit type options to Create New Audit page
User Story:
As an audit admin I want to be able to choose between a ballot polling audit and a batch comparison audit when I create a new audit.

Conditions of Acceptance:
- An "audit type" section with radio buttons appears on the AA Create New/View Existing page a la the attached mockup
- "Ballot polling" and "Batch comparison" are the two options right now (we will eventually add "Batch comparison" and "Hybrid" models, and probably others in the future)
- The ballot polling option is selected by default
- You must select an option in order to create a new audit

### LINKED ISSUE #639: Remove data entry setting from to audit settings page, hardcode to offline
User Story:
As an AA I don't want to have to fill in any data that isn't necessary, so I can save time

Conditions of acceptance:
- When the audit type is a batch comparison audit, the Audit Settings page should not show options for the data entry, and instead should only allow offline data entry as indicated in the attached screenshot.

### LINKED ISSUE #640: Add audit type to AA Audit Setup Review page
User Story:
As an AA I want to be able to see all the relevant data on the review page before I launch my audit, so that I can make sure everything is correct

Conditions of Acceptance:
- An "audit type" field is shown on the audit information page and includes the type of audit for all audits, a la the screenshot attached

MorganLove:
**Description**

Closes #640 
Closes #639 
Closes #638 

**Testing**

- none yet

**Progress**

- Functional, just need to update tests.

### DIFF
--- a/client/src/components/CreateAudit.tsx
+++ b/client/src/components/CreateAudit.tsx
@@ -107,14 +107,13 @@ const CreateAudit = ({ history }: RouteComponentProps<ICreateAuditParams>) => {
                   Audit type:
                   <LeftRadioGroup
                     name="auditType"
-                    data-testid="auditType"
                     onChange={e =>
                       setFieldValue('auditType', e.currentTarget.value)
                     }
                     selectedValue={values.auditType}
                   >
-                    <Radio value="BALLOT_POLLING">Ballot polling</Radio>
-                    <Radio value="BATCH_COMPARISON">Batch comparison</Radio>
+                    <Radio value="BALLOT_POLLING">Ballot Polling</Radio>
+                    <Radio value="BATCH_COMPARISON">Batch Comparison</Radio>
                   </LeftRadioGroup>
                 </label>
               </FormSection>
--- a/client/src/components/CreateAudit.tsx
+++ b/client/src/components/CreateAudit.tsx
@@ -2,10 +2,11 @@ import React, { useState } from 'react'
 import styled from 'styled-components'
 import { toast } from 'react-toastify'
 import { RouteComponentProps, Link } from 'react-router-dom'
+import { RadioGroup, Radio } from '@blueprintjs/core'
 import { Formik, FormikProps, Field } from 'formik'
 import FormButton from './Atoms/Form/FormButton'
 import { api } from './utilities'
-import { ICreateAuditParams } from '../types'
+import { ICreateAuditParams, IAuditSettings } from '../types'
 import { useAuthDataContext } from './UserContext'
 import FormSection from './Atoms/Form/FormSection'
 import FormField from './Atoms/Form/FormField'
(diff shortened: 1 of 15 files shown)


MorganLove:
I think this one has conflicts because the base branch does with `main`? Not sure.

client/src/components/CreateAudit.tsx
```diff
@@ -91,6 +102,22 @@ const CreateAudit = ({ history }: RouteComponentProps<ICreateAuditParams>) => {
                   />
                 </label>
               </FormSection>
+              <FormSection>
+                <label htmlFor="auditType">
+                  Audit type:
+                  <LeftRadioGroup
+                    name="auditType"
+                    data-testid="auditType"
```
jonahkagan, line 110: do you need this?

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
```diff
@@ -128,6 +134,14 @@ const Review: React.FC<IProps> = ({ prevStage, locked, refresh }: IProps) => {
               <td>Election Name:</td>
               <td>{electionName}</td>
             </tr>
+            <tr>
+              <td>Audit Type:</td>
+              <td>
+                {auditType === 'BALLOT_POLLING'
+                  ? 'Ballot Polling'
+                  : 'Batch Comparison'}
```
jonahkagan, line 142: I noticed that here you have "Ballot Polling" while above you have "Ballot polling" - standardize?

client/src/components/CreateAudit.tsx
```diff
@@ -38,22 +39,29 @@ const CenterField = styled(FormField)`
...
-  const onSubmit = async ({ auditName }: IValues) => {
+  const onSubmit = async ({ auditName, auditType }: IValues) => {
     try {
       setLoading(true)
       const response: { electionId: string } = await api('/election/new', {
         method: 'POST',
         body: JSON.stringify({
           organizationId: meta!.organizations[0].id,
           auditName,
+          auditType,
```
jonahkagan, line 64: i think you might also need to add this to `CreateSingleJurisdictionAudit`

client/src/components/MultiJurisdictionAudit/Setup/Review/index.tsx
MorganLove: @jonahkagan They're on different views, and I just went with what was on the mockup. But it could be both capitalized on the first view too.

client/src/components/CreateAudit.tsx
MorganLove: @jonahkagan Nope, good catch.

MorganLove:
@jonahkagan cleaned up and tests updated, ready for another pass. :)

PR #686, opened 2020-08-12
Batch tallies file upload

jonahkagan:
- Adds a new endpoint PUT /election/{election_id}/jurisdiction/{jurisdiction_id}/batch-tallies for JAs to upload the batch tallies file, as well as corresponding GET and DELETE endpoints (copying the ballot manifest upload endpoints)
- Requires that this file be uploaded after the ballot manifest
- Validates the tallies against the ballot manifest/contest settings
- Stores the tallies as a JSON blob in Jurisdiction.batch_tallies in the format expected by the audit_math.macro module, since we always pass that module the tallies for all the batches at once
- Creates a db migration for the new Jurisdiction fields
- Adds the batch tallies file info to the AA's jurisdiction status endpoint, as well as an endpoint to download the CSV, so that we can add that to the AA jurisdiction progress screen (again copying the ballot manifest pattern)


server/api/batch_tallies.py
```diff
@@ -0,0 +1,184 @@
...
+from ..util.csv_download import csv_response
+from ..util.csv_parse import decode_csv_file, parse_csv, CSVValueType, CSVColumnType
+
+BATCH_NAME = "Batch Name"
+
+
+def process_batch_tallies_file(
+    session: Session, jurisdiction: Jurisdiction, file: File
+):
+    def process():
```
benadida, line 26: Not a huge deal, but I wonder if these internal function definitions are less efficient because they get redefined on every run of the outer function. Not a blocker here, just a thing to lookg into at some point.

server/api/jurisdictions.py
```diff
@@ -12,23 +12,25 @@
 
 
 def serialize_jurisdiction(
-    jurisdiction: Jurisdiction, round_status: Optional[JSONDict]
+    election: Election, jurisdiction: Jurisdiction, round_status: Optional[JSONDict]
 ) -> JSONDict:
-    return {
+    json_jurisdiction = {
```
benadida, line 17: should there be a simple status field that indicates the type of audit? Or is that assumed to be loaded from the election metadata?

server/bgcompute.py
```diff
@@ -97,6 +99,25 @@ def bgcompute_update_ballot_manifest_file() -> int:
...
+    )
+
+    for file in files:
+        try:
+            jurisdiction = Jurisdiction.query.filter_by(
+                batch_tallies_file_id=file.id
+            ).one()
+            process_batch_tallies_file(db_session, jurisdiction, file)
+        except Exception:
+            print("ERROR updating batch tallies file")
```
benadida, line 116: worth logging the jurisdiction ID for debugging?

server/api/ballot_manifest.py
```diff
@@ -126,13 +126,10 @@ def upload_ballot_manifest(
 def get_ballot_manifest(
     election: Election, jurisdiction: Jurisdiction  # pylint: disable=unused-argument
 ):
-    if jurisdiction.manifest_file:
-        return jsonify(
-            file=serialize_file(jurisdiction.manifest_file),
-            processing=serialize_file_processing(jurisdiction.manifest_file),
-        )
-    else:
-        return jsonify(file=None, processing=None)
```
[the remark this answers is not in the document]
jonahkagan: no, i just moved the `None` checks into the serialization helpers cuz they were repeated everywhere. see https://github.com/votingworks/arlo/pull/686/files#diff-6c2f50d2d8bb927324ff7c7714f65dc6R51-R67

server/api/batch_tallies.py
jonahkagan: https://stackoverflow.com/questions/7839632/is-there-an-overhead-when-nesting-functions-in-python

this answer in particular was compelling: https://stackoverflow.com/a/50164012

> we can see that it [defining an inner function] takes less time than creating an empty dict (the fast way), so if you're doing anything non-trivial, this probably does not matter at all.

server/api/jurisdictions.py
jonahkagan: it's loaded in the election settings already

server/bgcompute.py
jonahkagan: this is a great question. in general I was confused by the pattern of swallowing the exception in this file, but I followed it assuming there was some thinking behind it.

my gut says we should log some context (like the jurisdiction id) and the exception itself. thoughts?

benadida: yeah that sounds good.

server/api/jurisdictions.py
benadida: 👍

server/api/batch_tallies.py
benadida: great! today I learned.

server/bgcompute.py
jonahkagan: coming back to this in a separate PR

PR #687, opened 2020-08-13
Batch tallies upload front end

MorganLove:
**Description**

- refactored the main hook for handling the endpoints so it could support both varieties of file uploads
- duplicated over the ballot manifest form and changed to reflect the batch tallies language
- made the batch tallies form only show up once the ballot manifest has finished processing

**Testing**

- still to do

**Progress**

- Haven't manually tested it yet, so not sure if it works, but I wanted to get it up so that @jonahkagan could double check that this approach will work at all.


client/src/components/MultiJurisdictionAudit/index.tsx
```diff
@@ -135,6 +141,15 @@ export const JurisdictionAdminView: React.FC = () => {
...
           />
+          {ballotManifest.processing &&
+            ballotManifest.processing.status ===
+              FileProcessingStatus.PROCESSED && (
+              <BatchTallies
+                batchTallies={batchTallies}
+                uploadBatchTallies={uploadBatchTallies}
+                deleteBatchTallies={deleteBatchTallies}
+              />
+            )}
```
jonahkagan, line 152: it might be clearer for the user if the batch tallies form is always displayed but disabled until the ballot manifest is uploaded. I'm not sure, but I have a hunch that they will want to know that they need to generate both files at the beginning, rather than generate one and then find out they have to generate the other

client/src/components/MultiJurisdictionAudit/index.tsx
```diff
@@ -114,13 +115,18 @@ export const JurisdictionAdminView: React.FC = () => {
     uploadBallotManifest,
     deleteBallotManifest,
   ] = useBallotManifest(electionId, jurisdictionId)
+  const [
+    batchTallies,
+    uploadBatchTallies,
+    deleteBatchTallies,
+  ] = useBatchTallies(electionId, jurisdictionId)
```
jonahkagan, line 122: not sure where the right place to do it is, but somewhere you might want to condition on the audit type being `BATCH_COMPARISON`

client/src/components/MultiJurisdictionAudit/useCSV.ts
```diff
@@ -0,0 +1,130 @@
...
+export const useBatchTallies = (
+  electionId: string,
+  jurisdictionId: string
+): [
+  IFileInfo | null,
+  (csv: File) => Promise<boolean>,
+  () => Promise<boolean>
+] => useCSV(electionId, jurisdictionId, 'batch-tallies')
+
+export default useCSV
```
jonahkagan, line 130: this feels like a useful abstraction. it's definitely some complicated logic so good to reuse it. wondering if it's also possible/worthwhile to abstract the CSV upload form component as well (maybe too soon for that)

also noting that I didn't abstract any of this on the backend lol and now I'm like hmm 🤔

client/src/components/MultiJurisdictionAudit/index.tsx
MorganLove: @jonahkagan Oh interesting, yeah. I cannot make the hook itself conditional of course, so I'd have to do that on whether I render the form or not. And then like you said earlier have it just be disabled at first.

client/src/components/MultiJurisdictionAudit/useCSV.ts
MorganLove: Yeah I was thinking of abstracting that too, but it looked like it might end up being too much with how many little places need tweaked between the two in order to get the language right. I might still have a go of it though.

client/src/components/MultiJurisdictionAudit/CSVForm/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+                  Click &quot;Browse&quot; to choose the appropriate Ballot
+                  Manifest file from your computer. This file should be a
+                  comma-separated list of all the ballot boxes/containers used
+                  to store ballots for this particular election, plus a count of
+                  how many ballot cards (individual pieces of paper) are stored
+                  in each container.
+                  <br />
+                  <br />
+                  <a
+                    href={`${window.location.origin}/sample_ballot_manifest.csv`}
```
benadida, line 82: why do we need `window.location.origin` here? Won't just `/sample_ballot_manifest.csv` do?

client/src/components/MultiJurisdictionAudit/CSVForm/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+                  Click &quot;Browse&quot; to choose the appropriate Candidate
+                  Totals by Batch file from your computer. This file should be a
+                  comma-separated list of all the ballot boxes/containers used
+                  to store ballots for this particular election, plus a count of
+                  how many votes were counted for each candidate in each of
+                  those containers.
+                  <br />
+                  <br />
+                  <a
+                    href={`${window.location.origin}/sample_batch_tallies.csv`}
```
benadida, line 103: same comment here, don't think we need `window.location.origin`.

client/src/components/MultiJurisdictionAudit/CSVForm/index.tsx
MorganLove: Huh, good catch. Not sure why that got put in there, haha.

PR #689, opened 2020-08-19
Misc test coverage

MorganLove:
**Description**

tests

**Testing**

- going through and patching up holes in the coverage here and there

**Progress**

ongoing wip


client/src/App.test.tsx
```diff
@@ -1,15 +1,177 @@
...
+        aaApiCalls.getSettings,
+        aaApiCalls.getJurisdictionFile,
+        aaApiCalls.getRounds,
+        aaApiCalls.getRounds,
+        aaApiCalls.getJurisdictions,
+        aaApiCalls.getContests,
+        aaApiCalls.getSettings,
+        aaApiCalls.getSettings,
+        aaApiCalls.getJurisdictionFile,
+      ]
```
jonahkagan, line 169: this seems like so many API calls just to render one screen! do you know what's going on?

client/src/App.test.tsx
```diff
@@ -1,15 +1,177 @@
...
+    jurisdictions: [],
+  },
+  aa: {
+    name: 'Leia Organa',
+    email: 'princess@rebelalliance.com',
+    type: 'audit_admin',
+    organizations: [],
+    jurisdictions: [],
+  },
+}
```
jonahkagan, line 31: could you use `aaApiCalls` and `jaApiCalls` instead of these?

client/src/App.test.tsx
```diff
@@ -1,15 +1,177 @@
...
 import App from './App'
-import * as utilities from './components/utilities'
+import { withMockFetch, renderWithRouter } from './components/testUtilities'
+import { IUserMeta } from './types'
+import { dummyBoards } from './components/DataEntry/_mocks'
+import {
+  // jaApiCalls,
+  aaApiCalls,
+} from './components/MultiJurisdictionAudit/_mocks'
+// import { manifestMocks } from './components/MultiJurisdictionAudit/useSetupMenuItems/_mocks'
```
jonahkagan, line 12: don't forget to remove

client/src/App.test.tsx
```diff
@@ -1,15 +1,177 @@
...
-    const { container } = render(<App />)
-    await screen.findByAltText('Arlo, by VotingWorks')
-    expect(container).toMatchSnapshot()
+  describe('/', () => {
+    it('renders unauthenticated properly', async () => {
+      const expectedCalls = [apiMocks.failedAuth]
+      await withMockFetch(expectedCalls, async () => {
+        const { container } = renderView('/')
+        await screen.findByAltText('Arlo, by VotingWorks')
+        expect(container).toMatchSnapshot()
```
jonahkagan, line 61: i have been finding these kinds of tests a bit difficult to maintain and review. I'm not totally sure exactly what they are testing for. the snapshots are so big that I can't really read them, so I don't know exactly what we're expecting (or at least what are the relevant things we're looking for). also it's a bit annoying because the snapshots change whenever we change styles (though that's a separate issue)

here are a few of my guesses for what these tests are about and suggestions for how they might be improved (but let me know if I'm misunderstanding):
- we're trying to test that it doesn't render empty - suggestion: remove the snapshots and just check for one relevant element to be rendered
- we're trying to test that it renders exactly the UI that we want - suggestion: check for the relevant pieces of the UI directly by finding them in the screen and using them - this would make it much easier to read the test and know what we're looking for
- we're trying to get code coverage up - not really sure what to do in this case, but might be worth thinking harder about

client/src/index.tsx
```diff
@@ -1,11 +1,17 @@
...
+import { BrowserRouter as Router } from 'react-router-dom'
 import './index.css'
 import App from './App'
 import * as serviceWorker from './serviceWorker'
 
-ReactDOM.render(<App />, document.getElementById('root'))
+ReactDOM.render(
+  <Router>
+    <App />
+  </Router>,
```
jonahkagan, line 12: why this change?

client/src/components/testUtilities.tsx
```diff
@@ -89,7 +93,12 @@ export const withMockFetch = async (
       expectedRequest.url === url &&
       (expectedRequest.skipBody || equal(expectedRequest.options, options))
     ) {
-      return new Response(JSON.stringify(expectedRequest.response))
+      return expectedRequest.error
+        ? new Response(
+            JSON.stringify(expectedRequest.response),
+            expectedRequest.error
+          )
+        : new Response(JSON.stringify(expectedRequest.response))
```
jonahkagan, line 101: nice extension, seems very useful

MorganLove: @jonahkagan Yeah, we can test error responses now, which is super handy.

I ran into an issue with comparing the formData objects because they'd never match the mock (literally impossible, since the code creates and sends a new one each run), so I was wondering if passing in a matcher function that defaults to the current behavior could resolve that, but I bypassed it for now with just a skip flag.

client/src/index.tsx
MorganLove: @jonahkagan Because in order to test the route implementation I needed to change the routes, which is only possible with passing in different routes to the Router component. I couldn't do that if the Router component was included in the component I was testing.

client/src/App.test.tsx
MorganLove: @jonahkagan Good questions. Here are some of my thoughts.

The code coverage aspect doesn't change at all regardless of whether we use snapshots or not. As long as it renders, it's counted as covered.

However, testing each individual significant piece is far more verbose than just capturing it all as a snapshot. We aren't just testing specific pieces, we're testing to make sure nothing changes that we didn't expect. There's been many times where something changed in one part of the codebase that resulted in changes elsewhere that were unanticipated (and undesired) which was only noticed because the snapshots changed and sounded the alarm.

As far as readability goes, there are plugins in VS Code which can help with that I think. I haven't used them yet, though. Maybe they'd help?

client/src/index.tsx
jonahkagan: Makes sense. It actually mirrors the example in the docs! https://testing-library.com/docs/example-react-router

client/src/components/testUtilities.tsx
jonahkagan: Hmm what's an example test where that's happening? I'm curious to check it out

client/src/App.test.tsx
MorganLove: @jonahkagan This is the status box, and it has a bunch of api calls, yeah. :stuck_out_tongue_closed_eyes:

client/src/App.test.tsx
MorganLove: @jonahkagan Good idea actually. I think I wrote these before I set up `aaApiCalls` and `jaApiCalls`, haha.

client/src/index.tsx
MorganLove: Yup!

client/src/components/testUtilities.tsx
MorganLove: Here's where I needed to use it: https://github.com/votingworks/arlo/blob/main/client/src/components/MultiJurisdictionAudit/index.test.tsx#L373

Basically, the file upload process creates a new formData object when it's submitting the form, and while I could create another formData object using the same file in the test, it's two different objects and they won't equal each other.

client/src/App.test.tsx
jonahkagan: Whoops sorry I let this convo slip. Thanks for explaining - helpful to understand that it's not so much about coverage but more about trying to lock in the content of the UI.

Personally, I think being verbose and explicit would make for more useful and readable tests. But this might just be a matter of opinion and style, since it sounds like the snapshot tests have been useful for you.

client/src/App.test.tsx
jonahkagan: I'm actually kind of confused by the tests in this `describe` block. I'm not sure what the URL `/election/1/jurisdiction-id-1/1` corresponds to. It looks like the pattern is `/election/:electionId/:jurisdictionId/:anotherParam`, but I can't find that in the code.

I think this is an example where the snapshot tests make it hard for me to figure out what's supposed to be going on.

client/src/components/MultiJurisdictionAudit/useSetupMenuItems/index.ts
```diff
@@ -55,8 +55,7 @@ function useSetupMenuItems(
       setContests('processing')
       const condition = async () => {
         const fileProcessing = await getJurisdictionFileStatus(electionId)
-        if (!fileProcessing) return false
-        const { status, error } = fileProcessing
+        const { status, error } = fileProcessing!
```
jonahkagan, line 58: how are you sure this is never `null`?

jonahkagan, line 58: how are you sure this is never null?

client/src/components/testUtilities.tsx
jonahkagan: Is it possible to detect when the `body` is a `FormData` object and then either skip it or, better, pull out the actual data as a string and check it?

client/src/App.test.tsx
MorganLove: @jonahkagan Actually, that's a bug, haha. That's not what the url is supposed to be. :stuck_out_tongue_closed_eyes: It's supposed to be `/election/1/jurisdiction/jurisdiction-id-1` but my copy and paste got screwed up at some point. Good catch on that.

Ironically, when checking on this and trying to figure out if I got the mocked data right again just now, I used the snapshots heavily to know what was being rendered given the various configurations.

client/src/components/testUtilities.tsx
MorganLove: @jonahkagan Would you advocate doing that inside your test function? Or passing a function into the test function that itself handles that kind of conditional management?

jonahkagan: It makes sense to me to have it be part of `withMockFetch`, since form data is a common kind of request body

client/src/components/MultiJurisdictionAudit/useSetupMenuItems/index.ts
MorganLove: I might be overassuming, but since we only reach this point in the code if it's already processing the file, the endpoint should never return null, if I understand it right. It only returns null for the processing object on the first call. If something goes wrong after that it returns with an object with an error and so on, but it won't return null.

client/src/components/testUtilities.tsx
MorganLove: @jonahkagan Fair enough. We only ever use the `FormData` object when we're uploading a file, though, which we only do in a couple places. I think if it's possible to extract the data of both and compare those, though, it'd be valuable to check for that and do it automatically.

client/src/components/MultiJurisdictionAudit/useSetupMenuItems/index.ts
jonahkagan: ah i didn't read the context of the rest of the function. yeah I think your reasoning is correct

client/src/components/testUtilities.tsx
MorganLove: @jonahkagan So we can detect that it's a `FormData` object, but extracting the data is an asynchronous process, and I think that's why it's not working. I've tried a bunch of different ways, but I think it's something we'll have to return to later, or if one of us gets a stroke of inspiration. We can use the `skipBody` option for now without concern, I think, since comparing the content of the file is not actually testing any of the functionality of the app.

jonahkagan (approved):
i notice there are a bunch of places where `apiMock` is change to return `true` or `false` instead of returning an object. i don't quite understand the change - I feel like an object or `null` would be a more realistic response. but maybe there's a reason

other than that seems good to go

MorganLove:
@jonahkagan Actually, that's a really good catch. It should reflect what `api` actually returns, which is now `null` instead of `false`, I'd just forgotten to update that. Fixed now!

PR #703, opened 2020-08-25
Use snapshots in test_sampler.py

jonahkagan:
Should make it easier when we tweak the code.


jonahkagan:
@umbernhard are there any other audit math tests that could benefit from snapshots?

Issue #672, opened 2020-08-04
Backend: Record batch results


jonahkagan:
- New db model: BatchResult
- Endpoint for the JA to know which batches to record results for
- Endpoint for the JA to record batch results
- Use the batch results to calculate the risk measurement

PR #709, opened 2020-08-27
JA get batches and batch retrieval list endpoints

jonahkagan:
Task: #672 

Two endpoints that are needed for the JA for batch audits.

### DIFF
--- a/server/api/batches.py
+++ b/server/api/batches.py
@@ -1,0 +1,81 @@
+import io, csv
+from flask import jsonify
+
+from . import api
+from ..auth import with_jurisdiction_access
+from ..models import *  # pylint: disable=wildcard-import
+from ..util.csv_download import csv_response, election_timestamp_name
+from ..util.jsonschema import JSONDict
+
+
+@api.route(
+    "/election/<election_id>/jurisdiction/<jurisdiction_id>/round/<round_id>/batches/retrieval-list",
+    methods=["GET"],
+)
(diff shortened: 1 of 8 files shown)


server/api/batches.py
```diff
@@ -0,0 +1,81 @@
...
+from ..models import *  # pylint: disable=wildcard-import
+from ..util.csv_download import csv_response, election_timestamp_name
+from ..util.jsonschema import JSONDict
+
+
+@api.route(
+    "/election/<election_id>/jurisdiction/<jurisdiction_id>/round/<round_id>/batches/retrieval-list",
+    methods=["GET"],
+)
+@with_jurisdiction_access
```
benadida, line 15: when I haven't reviewed the code in a few days, I keep asking myself "does this call check that it has jurisdiction *and* election access?" I'm wondering if for more casual readers of the code, calling this `@with_jurisdication_and_election_access` might be helpful.

Not required, just musing.

jonahkagan: yeah, i hear that. i think these would actually be better described according to user type. i would call this `with_jurisdiction_admin_access`

MorganLove:
@jonahkagan I'm getting a `ZeroDivisionError: float division by zero` error on the `GET /api/election/:electionId/sample-sizes` endpoint on this branch. I did a resetdb but it still is erroring. Ideas?

PR #711, opened 2020-09-01
Fixes bug in ASN probability computation. Fixed #701. 

### LINKED ISSUE #701: Tweak expected_prob to allow for irrelevant ballots
expected_prob, which is used to estimate BRAVO completion percentages for a specified first-round sample size (typically the ASN), presently assumes that there are no irrelevant ballots/cards. The estimates therefore are inflated. A quick patch is to change `n = asn` to `n = asn * p_wr`. (Typically n will not be an integer, which is fine for the math used here.)

umbernhard:
**Description**
Fixes bug with ASN probability calculation that was artificially inflating the probabilities, per #701. 

**Testing**

Updated snapshots and tests. 

**Progress**
N/A


server/audit_math/bravo.py
```diff
@@ -251,7 +251,7 @@ def expected_prob(
     if threshold <= 0:
         return 0
 
-    n = asn
+    n = asn * p_w
```
jonahkagan, line 254: I noticed that #701 recommended multiplying by `p_wr` not `p_w` - was this intentional?

jonahkagan:
@umbernhard would you mind giving your PR a more descriptive name just so it's easier to look back on and understand

server/audit_math/bravo.py
umbernhard: ooooh good catch. Will fix.

PR #712, opened 2020-09-01
scripts.create-admin: support existing audit admins

jonahkagan:
Sometimes we want users to be able to be audit admins for multiple orgs,
so we modify the create-admin script to find or create the user, then
make them an admin for the org.

### DIFF
--- a/scripts/create-admin.py
+++ b/scripts/create-admin.py
@@ -10,13 +10,15 @@ if __name__ == "__main__":
         sys.exit(1)
 
     org_id, email = sys.argv[1:]  # pylint: disable=unbalanced-tuple-unpacking
-    u = User(id=str(uuid.uuid4()), email=email, external_id=email)
-    audit_admin = AuditAdministration(user_id=u.id, organization_id=org_id)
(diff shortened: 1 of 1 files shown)


scripts/create-admin.py
```diff
@@ -10,13 +10,15 @@
...
 
     org_id, email = sys.argv[1:]  # pylint: disable=unbalanced-tuple-unpacking
-    u = User(id=str(uuid.uuid4()), email=email, external_id=email)
-    audit_admin = AuditAdministration(user_id=u.id, organization_id=org_id)
-    db_session.add(u)
+    user = User.query.filter_by(email=email).first()
+    if not user:
+        user = User(id=str(uuid.uuid4()), email=email, external_id=email)
+    audit_admin = AuditAdministration(user_id=user.id, organization_id=org_id)
+    db_session.add(user)
```
benadida, line 17: should this be inside the if?

jonahkagan: yes, ill move it, though it works fine this way as well

PR #713, opened 2020-09-01
Changed the sigfigs for p-values in audit report

umbernhard:
**Description**
Forces audit reports to use at most 10 significant digits for p-values. Also disallows scientific notation, and changes the type to string to avoid annoying excel issues. 

**Testing**

Updated snapshots. 

**Progress**

N/A


server/tests/api/snapshots/snap_test_reports.py
```diff
@@ -9,141 +9,14 @@
...
-J2,4,18,Round 1: 0.038013103949978453,NOT_AUDITED,,\r
-J2,4,25,Round 1: 0.077777948601238256,NOT_AUDITED,,\r
-J2,4,26,"Round 1: 0.000710191532006945, 0.117496458878296268",NOT_AUDITED,,\r
-J2,4,29,Round 1: 0.122889862183711840,NOT_AUDITED,,\r
-J2,4,30,Round 1: 0.035663218359419393,NOT_AUDITED,,\r
+] = """{
+  "errors": [
+    {
+      "errorType": "Internal Server Error", 
+      "message": "\'<\' not supported between instances of \'NoneType\' and \'float\'"
```
umbernhard, line 16: hmm, this seems bad.

server/api/reports.py
```diff
@@ -28,6 +28,16 @@ def pretty_targeted(is_targeted: bool) -> str:
     return "Targeted" if is_targeted else "Opportunistic"
 
 
+def pretty_pvalue(value: float) -> str:
+
+    if value is None:
+        return ""
+    elif value < 10 ** -10:
+        return "<0.0000000001"
```
jonahkagan, line 36: i know this is the most accurate, but wondering if it will make it more difficult for consumers of the audit report to use this value, since now it's not a number. thoughts on just removing the `<`?

server/tests/snapshots/snap_test_multiple_targeted_contests.py
```diff
@@ -48,11 +63,11 @@
 \r
 ######## ROUNDS ########\r
 Round Number,Contest Name,Targeted?,Sample Size,Risk Limit Met?,P-Value,Start Time,End Time,Audited Votes\r
-1,Contest 1,Targeted,191,Yes,1.84402281147622e-05,DATETIME,DATETIME,candidate 1: 132; candidate 2: 59\r
-1,Contest 2,Targeted,485,No,1.0,DATETIME,DATETIME,Yes: 0; No: 0\r
-1,Contest 3,Opportunistic,,No,1.0,DATETIME,DATETIME,candidate 1: 0; candidate 2: 0; candidate 3: 0\r
-2,Contest 2,Targeted,1468,Yes,8.27580207807509e-24,DATETIME,DATETIME,Yes: 1028; No: 440\r
-2,Contest 3,Opportunistic,,No,1.0,DATETIME,DATETIME,candidate 1: 0; candidate 2: 0; candidate 3: 0\r
+1,Contest 1,Targeted,191,Yes,0.0000184402,DATETIME,DATETIME,candidate 1: 132; candidate 2: 59\r
+1,Contest 2,Targeted,485,No,1.0000000000,DATETIME,DATETIME,Yes: 0; No: 0\r
```
jonahkagan, line 67: this seems a little silly. is there an easy way to say "print as many digits as necessary up to 10" instead of "always print 10 digits"? if not, no biggie

server/api/reports.py
umbernhard: @mcchilders this seems like a Q for you

server/tests/snapshots/snap_test_multiple_targeted_contests.py
umbernhard: This one too, @mcchilders. It should be possible.

server/api/reports.py
mcchilders: In this one instance, where the number we’re trying to represent is smaller than what we can show with our 10 significant digits, I think the “less than” needs to be there. For the numbers that we can represent accurately in 10 decimals, we won’t use it. Does that make sense? This just covers the “we limited our digits and this is smaller than a number we can show.

server/tests/snapshots/snap_test_multiple_targeted_contests.py
mcchilders: Since these are percentages represented as decimals, I think preserving all 10 is fine (we won’t ever have 1.0000000000 unless you didn’t actually audit any ballots, so that’s a not-very-likely edge case.)

umbernhard: I've fixed it so it strips all trailing zeros except in the `p = 1.0` case.

PR #714, opened 2020-09-02
Record batch results

jonahkagan:
Task: #672 
- Add a new model BatchResult to record the audit results for a
batch+contest choice 
- Db migration to add BatchResult
- Add an endpoint for recording batch results for a jurisdiction for a
given round
- When results are recorded, check if the round is over, and if so,
calculate the risk measurements
- Add the recorded batch results to the batch list endpoint under `results` for each batch


server/api/batches.py
```diff
@@ -51,11 +70,17 @@ def get_batch_retrieval_list(
 
 def serialize_batch(batch: Batch) -> JSONDict:
     audit_board = batch.audit_board
+    results = (
+        {r.contest_choice_id: r.result for r in batch.results}
+        if len(list(batch.results)) > 0
```
benadida, line 75: I believe this will perform the query twice, since `batch.results` is lazy. Recommend stashing `list(batch.results)` once in a variable, and then using it multiple times.

server/api/rounds.py
```diff
@@ -120,6 +204,18 @@ def end_round(election: Election, round: Round):
...
 
 def is_round_complete(election: Election, round: Round) -> bool:
+    # For batch audits, check that all sampled batches have recorded results
+    if election.audit_type == AuditType.BATCH_COMPARISON:
+        num_batches_without_results: int = (
+            Batch.query.join(SampledBatchDraw)
+            .filter_by(round_id=round.id)
+            .outerjoin(BatchResult)
+            .group_by(Batch.id)
+            .having(func.count(BatchResult.batch_id) == 0)
```
benadida, line 214: this count does the right thing of not counting null batch results in the outer join?

benadida (approved):

server/api/batches.py
jonahkagan: It actually doesn't perform the query again until the objects are expired from the session (e.g. on `commit`). Check this out:

```python
>>> from server.models import *
>>> b = Batch.query.first()
2020-09-03 10:32:34,713 INFO sqlalchemy.engine.base.Engine BEGIN (implicit)
2020-09-03 10:32:34,714 INFO sqlalchemy.engine.base.Engine SELECT batch.created_at AS batch_created_at, batch.updated_at AS batch_updated_at, batch.id AS batch_id, batch.jurisdiction_id AS batch_jurisdiction_id, batch.name AS batch_name, batch.num_ballots AS batch_num_ballots, batch.storage_location AS batch_storage_location, batch.tabulator AS batch_tabulator, batch.audit_board_id AS batch_audit_board_id 
FROM batch 
 LIMIT %(param_1)s
2020-09-03 10:32:34,714 INFO sqlalchemy.engine.base.Engine {'param_1': 1}
>>> b.results
2020-09-03 10:32:38,836 INFO sqlalchemy.engine.base.Engine SELECT batch_result.created_at AS batch_result_created_at, batch_result.updated_at AS batch_result_updated_at, batch_result.batch_id AS batch_result_batch_id, batch_result.contest_choice_id AS batch_result_contest_choice_id, batch_result.result AS batch_result_result 
FROM batch_result 
WHERE %(param_1)s = batch_result.batch_id
2020-09-03 10:32:38,836 INFO sqlalchemy.engine.base.Engine {'param_1': '3082a3e7-55d2-4afa-a246-c9e8f61698c3'}
[]
>>> b.results
[]
```

jonahkagan: That being said, we could use an eager loading technique to avoid having a separate `SELECT` query for each batch's results. It's probably the primary optimization we could be making across arlo, and very easy to do (just add an option to the original db query), but I've held off on doing it anywhere really because I didn't know if it was needed.

server/api/rounds.py
jonahkagan: well, i hope so! do the tests convince you that it does? if not, i'd like to add more tests

PR #715, opened 2020-09-02
Adding jurisdiction names to header for JA views

### LINKED ISSUE #693: Add jurisdiction name to JA dashboards
When jurisdictions log in, we want their dashboard to be labeled with their jurisdiction name so it feels personalized

MorganLove:
**Description**

Closes #693 

@mcchilders It looks like we currently support multiple jurisdictions under the same login, so I made it so that it will display all the jurisdiction names associated with the current login.

**Testing**

- Updated snapshots

**Progress**

- LGTM
- I think this part of the code might be covered in my wip coverage PR, so instead of redoing it all over here, I'd rather merge this then update that PR to reflect this relatively simple change.

### DIFF
--- a/client/src/components/Header.tsx
+++ b/client/src/components/Header.tsx
@@ -23,7 +23,7 @@ const Nav = styled(Navbar)`
 
   .bp3-navbar-heading img {
     height: 35px;
-    padding-top: 8px;
+    padding-top: 5px;
   }
 
   .bp3-navbar-divider {
@@ -55,6 +55,11 @@ const Header: React.FC<{}> = () => {
               <img src="/arlo.png" alt="Arlo, by VotingWorks" />
             </Link>
           </NavbarHeading>
+          {isAuthenticated && meta!.type === 'jurisdiction_admin' && (
+            <NavbarHeading>
+              Jurisdictions: {meta!.jurisdictions.map(j => j.name).join(', ')}
+            </NavbarHeading>
+          )}
         </NavbarGroup>
         <NavbarGroup align={Alignment.RIGHT}>
           {isAuthenticated && electionId && meta!.type === 'audit_admin' && (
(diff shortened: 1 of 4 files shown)


client/src/components/Header.tsx
```diff
@@ -38,13 +38,16 @@ const InnerBar = styled(Inner)`
...
 interface TParams {
   electionId: string
+  jurisdiction?: string
+  jurisdictionId?: string
 }
 
 const Header: React.FC<{}> = () => {
   const match: RouteComponentProps<TParams>['match'] | null = useRouteMatch(
-    '/election/:electionId'
+    '/election/:electionId/:jurisdiction?/:jurisdictionId?'
```
jonahkagan, line 47: this feels kinda tricky, since `jurisdiction` isn't really a parameter, but rather just the string `jurisdiction`. maybe add a comment or just do two separate matches, one for the election path and one for the jurisdiction path

jonahkagan, line 47: this feels kinda tricky, since `jurisdiction` isn't really a parameter, but rather just the string `jurisdiction`. maybe add a comment or just do two separate matches, one for the election path and one for the jurisdiction path

client/src/components/Header.tsx
```diff
@@ -55,6 +58,15 @@ const Header: React.FC<{}> = () => {
               <img src="/arlo.png" alt="Arlo, by VotingWorks" />
             </Link>
           </NavbarHeading>
+          {isAuthenticated && meta!.type === 'jurisdiction_admin' && (
+            <NavbarHeading>
+              {jurisdictionId &&
+                `Jurisdiction: ${
+                  meta!.jurisdictions.filter(j => j.id === jurisdictionId)[0]
```
jonahkagan, line 65: can use `.find` instead of `.filter`

jonahkagan (approved):

client/src/components/Header.tsx
MorganLove: I hadn't even thought of doing two matches, that actually cleaned things up nicely. :slightly_smiling_face:

PR #716, opened 2020-09-02
Added cache to CircleCi config

umbernhard:
**Description**
Adds dependency caching and parallelism to CircleCI to speed up testing. 

**Testing**
N/A

**Progress**
Ready to go.


umbernhard:
@jonahkagan this is ready for review. I'll squash merge it once it's approved. Speeds up testing by about a minute, and that will increase the more tests we add.

.circleci/config.yml
```diff
@@ -12,12 +12,43 @@ jobs:
...
+      - restore_cache:
+          keys:
+            - pip-packages-v4-{{ .Branch }}-{{ checksum "Pipfile.lock" }}
+            - pip-packages-v4-{{ .Branch }}-
+            - pip-packages-v4-
+      - restore_cache:
+          keys:
+            - npm-packages-v6-{{ .Branch }}-{{ checksum "package-lock.json" }}
       - run:
+          parallelism: 4
```
jonahkagan, line 32: is it possible to configure this at a higher level in the file, or does it have to be set on each command?

.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2.1
 orbs:
   cypress: cypress-io/cypress@1
 jobs:
-  build-and-test:
+  dependencies:
```
jonahkagan, line 5: what does this change do?

.circleci/config.yml
```diff
@@ -12,12 +12,43 @@ jobs:
...
+          keys:
+            - yarn-packages-v3-{{ .Branch }}-{{ checksum "yarn.lock" }}
+      - restore_cache:
+          keys:
+            - pip-packages-v4-{{ .Branch }}-{{ checksum "Pipfile.lock" }}
+            - pip-packages-v4-{{ .Branch }}-
+            - pip-packages-v4-
+      - restore_cache:
+          keys:
+            - npm-packages-v6-{{ .Branch }}-{{ checksum "package-lock.json" }}
```
jonahkagan, line 30: is it possible to name these steps so that we can tell which one is which? this is what I see now: 
Screen Shot 2020-09-08 at 2 44 19 PM

.circleci/config.yml
```diff
@@ -12,12 +12,43 @@ jobs:
...
+          key: pip-packages-v4-{{ .Branch }}-{{ checksum "Pipfile.lock" }}
+          paths:
+            - "~/.local"
+            - "~/.cache"
+      - save_cache:
+          key: npm-packages-v6-{{ .Branch }}-{{ checksum "package-lock.json" }}
+          paths:
+            #- "/usr/local/lib/node_modules" # location depends on npm version
+            - "node_modules"
+            - "client/node_modules"
```
jonahkagan, line 51: same question about naming as above

.circleci/config.yml
```diff
@@ -12,12 +12,43 @@ jobs:
...
+
+      - save_cache:
+          key: pip-packages-v4-{{ .Branch }}-{{ checksum "Pipfile.lock" }}
+          paths:
+            - "~/.local"
+            - "~/.cache"
+      - save_cache:
+          key: npm-packages-v6-{{ .Branch }}-{{ checksum "package-lock.json" }}
+          paths:
+            #- "/usr/local/lib/node_modules" # location depends on npm version
```
jonahkagan, line 49: do we need this line still?

.circleci/config.yml
```diff
@@ -30,9 +61,14 @@ jobs:
           command: |
             make resetdb
       - run:
+          parallelism: 4
           name: lint server (pylint)
           command: |
             make lint-server
+      - run:
+          name: line client (yarn)
```
jonahkagan, line 69: "lint"

.circleci/config.yml
```diff
@@ -41,20 +77,22 @@ jobs:
           name: check formatting
           command: |
             make format-server
-            git diff-index --quiet HEAD -- || (echo "Found unexpected changes!" && git diff && exit 1)
       - run:
+          parallelism: 4
           name: server tests
           command: |
-            pipenv run pytest -n 2 --cov=.
+            circleci tests glob server/tests/**.py | circleci tests split --split-by=timings | xargs pipenv run pytest -n auto
```
jonahkagan, line 84: what does this do? also we lost test coverage I believe, which is actually one of the more important things to run on CI cuz ppl (me) don't often run it locally

Makefile
```diff
@@ -35,7 +35,7 @@ format-server:
 	pipenv run black .
 
 lint-server:
-	pipenv run pylint server scripts
+	pipenv run pylint -j 0 server scripts
```
jonahkagan, line 38: ooo exciting. pylint is so slow!

.circleci/config.yml
umbernhard: My understanding is that if you use `parallelism > 1` for sequential steps, it runs those steps that many times. So yes, we could but from my testing that made it slower, as random stuff would cause one run to be way slower than the others.

.circleci/config.yml
umbernhard: nothing, it's a legacy of my moving stuff into workflows. Will fix :smile:

.circleci/config.yml
umbernhard: I'll have to check.

.circleci/config.yml
umbernhard: same answer about naming as above :stuck_out_tongue:

.circleci/config.yml
umbernhard: Will remove.

.circleci/config.yml
umbernhard: CircleCI requires some special juju to actually run tests in parallel, you basically have to let it handle the chunking of tests to different nodes. 

Re: coverage I will have to see what I can do about that.

umbernhard:
@jonahkagan okay I think this is ready to go.

.circleci/config.yml
```diff
@@ -2,65 +2,271 @@ version: 2.1
...
+            - ~/.local
+            - ~/.cache
+      - run:
+          name: create data model
+          command: |
+            make resetdb
+      - run:
+          name: test-server # server tests
+          command: |
+            pipenv run pytest --cov=.
```
jonahkagan, line 100: `make test-server-coverage`?

.circleci/config.yml
```diff
@@ -2,65 +2,271 @@ version: 2.1
...
+          POSTGRES_USER: root
+          POSTGRES_DB: arlo-test
+    resource_class: xlarge
+    steps:
+      - checkout
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - run: sudo chown -R circleci:circleci /usr/local/bin
+      - run: sudo chown -R circleci:circleci /usr/local/lib/python3.7/site-packages
+      - restore_cache:
```
jonahkagan, line 22: maybe pull out the different caching steps as reusable commands? https://circleci.com/docs/2.0/reusing-config/

.circleci/config.yml
```diff
@@ -2,65 +2,271 @@ version: 2.1
...
+    docker:
+      - image: circleci/python:3.7.3-node-browsers
+        environment:
+          DATABASE_URL: postgresql://root@localhost:5432/arlo-test
+          FLASK_ENV: test
+      - image: circleci/postgres:9.6.3-alpine-ram
+        environment:
+          POSTGRES_USER: root
+          POSTGRES_DB: arlo-test
+    resource_class: xlarge
```
jonahkagan, line 15: how did you choose resource_class? do all the jobs benefit from this class? also do they cost diff amounts of money or is it all the same?

.circleci/config.yml
```diff
@@ -2,65 +2,271 @@ version: 2.1
...
+            - client/node_modules
+
       - run:
           name: Install Chrome
           command: |
             curl -L -o google-chrome.deb https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
             sudo dpkg -i google-chrome.deb
             sudo sed -i 's|HERE/chrome\"|HERE/chrome\" --disable-setuid-sandbox|g' /opt/google/chrome/google-chrome
             rm google-chrome.deb
+
```
jonahkagan, line 154: do we actually need chrome for the frontend tests?

.circleci/config.yml
```diff
@@ -2,65 +2,271 @@ version: 2.1
...
+          keys:
+            - pip-packages-v5-{{ .Branch }}-{{ checksum "Pipfile.lock" }}
+            - pip-packages-v5-{{ .Branch }}-
+            - pip-packages-v5-
       - run:
-          name: client tests
+          name: install
           command: |
-            make test-client
+            make install
```
jonahkagan, line 247: surprised that we don't want `make install-development` here...

.circleci/config.yml
```diff
@@ -2,65 +2,271 @@ version: 2.1
...
+
       - run:
-          name: wdio end-to-end tests
+          name: Install Chrome
           command: |
-            yarn --cwd client test:wdio
+            curl -L -o google-chrome.deb https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
+            sudo dpkg -i google-chrome.deb
+            sudo sed -i 's|HERE/chrome\"|HERE/chrome\" --disable-setuid-sandbox|g' /opt/google/chrome/google-chrome
+            rm google-chrome.deb
```
jonahkagan, line 255: do we need to install chrome separately for the cypress tests (given that the docker image has `browsers` in the name)?

client/package.json
```diff
@@ -8,7 +8,7 @@
     "lint:fix": "yarn lint --fix",
     "start": "react-scripts start",
     "test": "is-ci \"test:ci\" \"test:watch\"",
-    "test:ci": "CI=true tsc && react-scripts test --maxWorkers 2 --env=jest-environment-jsdom-sixteen --coverage",
+    "test:ci": "CI=true tsc && react-scripts test --maxWorkers 4 --env=jest-environment-jsdom-sixteen --coverage",
```
jonahkagan, line 11: this command isn't getting used anymore as far as I can tell

client/run-cypress-tests.sh
```diff
@@ -3,7 +3,7 @@
 export ARLO_AUDITADMIN_AUTH0_BASE_URL=https://votingworks-noauth.herokuapp.com
 export ARLO_JURISDICTIONADMIN_AUTH0_BASE_URL=https://votingworks-noauth.herokuapp.com
 
-trap 'kill 0' SIGINT SIGHUP EXIT
+trap 'exit 0' SIGINT SIGHUP EXIT
```
jonahkagan, line 6: im pretty sure the `kill 0` was there to kill all the child processes as well. have you tried running this locally? does it leave zombie processes?

.circleci/config.yml
```diff
@@ -2,65 +2,271 @@ version: 2.1
...
+      - run: sudo chown -R circleci:circleci /usr/local/bin
+      - run: sudo chown -R circleci:circleci /usr/local/lib/python3.7/site-packages
+      - restore_cache:
+          name: Restore Yarn Cache
+          keys:
+            - yarn-packages-v3-{{ .Branch }}-{{ checksum "yarn.lock" }}
+      - restore_cache:
+          name: Restore Node Cache
+          keys:
+            - npm-packages-v8-{{ .Branch }}-{{ checksum "package-lock.json" }}
```
jonahkagan, line 126: why don't we use yarn for everything? there's a `yarn.lock` file in `arlo` and in `arlo/client`

.circleci/config.yml
umbernhard: That runs it in parallel, which I don't want to do because it breaks stuff.

.circleci/config.yml
umbernhard: from what I can tell the larger the resource class the faster the tests run. We can probably knock down some of the already fast steps down. Going from medium -> xlarge saves us about a minute of compute time. And it uses more credits, but I don't think we're coming close to our alotted limit that would require us to pay more (except when I'm testing these configs :sweat_smile:)

.circleci/config.yml
umbernhard: I can check, but I tried removing it and it broke stuff. The docker we're loading actually has browsers in it, but yarn complained about their version so the manual install seems necessary.

.circleci/config.yml
jonahkagan: oh really? that's kind of concerning. i run them in parallel locally. are we throwing out that parallelism?

.circleci/config.yml
jonahkagan: cool, sounds fine then

.circleci/config.yml
umbernhard: Done

.circleci/config.yml
umbernhard: @jonahkagan let me know if you have thoughts on this.

client/package.json
umbernhard: isn't it called when `yarn --cwd client test` is called? (see the line right above it)

.circleci/config.yml
umbernhard: Fixed this.

.circleci/config.yml
jonahkagan: i thought it would at least need to install cypress, which is a dev dependency

client/package.json
jonahkagan: oh lol true 🤦

client/run-cypress-tests.sh
jonahkagan: that's weird, because i dont think it did that before, did it? i.e. cypress tests have been passing on CI

.circleci/config.yml
jonahkagan: WOW it looks so good with this change

jonahkagan (approved):
i think we still have a few outstanding conversations (e.g. the cypress thing) but this is looking amazing. approving pending we resolve those last bits

umbernhard:
@jonahkagan for some reason the cypress script works fine now, so I restored it to the version on main. Otherwise I think we're good to go here, but happy to answer other questions.

jonahkagan (approved):
SO READY FOR THIS

PR #720, opened 2020-09-05
Batch offline data entry

MorganLove:
**Description**

- Adding a new component and hook for batch data entry.

**Testing**

- Duplicated the tests from the regular data entry, and added a couple more to both to help with the coverage cause.

**Progress**

- This seems to work just fine with the first round, but the second round is displaying some really weird behavior. But I think that's from the backend.
- There's a react warning because the source of truth for the results (the batches endpoint) returns null instead of the expected structure of batch ids and choice ids with null values, which means the form is switching from uncontrolled to controlled components when we first interact with the fields. Ideally, the results object would contain the expected structure instead of null, and this would clean up the FE code quite a bit.

- Also noting that this doesn't support multiple contests, and this is not reflected during the AA setup process yet. We should add constraints there if it's a batch comparison audit to prevent unexpected surprises.

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/AASetup/Contests/index.tsx
+++ b/client/src/components/MultiJurisdictionAudit/AASetup/Contests/index.tsx
@@ -31,6 +31,7 @@ interface IProps {
   nextStage: ISidebarMenuItem
   prevStage: ISidebarMenuItem
   locked: boolean
+  isBatch?: boolean
 }
 
 const Contests: React.FC<IProps> = ({
@@ -38,6 +39,7 @@ const Contests: React.FC<IProps> = ({
   nextStage,
   prevStage,
   locked,
+  isBatch,
 }) => {
   const contestValues: IContest[] = [
     {
@@ -255,17 +257,19 @@ const Contests: React.FC<IProps> = ({
                       </Card>
                     )
                   })}
-                  <FormButtonBar>
-                    <FormButton
-                      type="button"
-                      onClick={() =>
-                        contestsArrayHelpers.push({ ...contestValues[0] })
-                      }
-                    >
-                      Add another {isTargeted ? 'targeted' : 'opportunistic'}{' '}
-                      contest
-                    </FormButton>
-                  </FormButtonBar>
+                  {!isBatch && ( // TODO support multiple contests in batch comparison audits
+                    <FormButtonBar>
+                      <FormButton
+                        type="button"
+                        onClick={() =>
+                          contestsArrayHelpers.push({ ...contestValues[0] })
+                        }
+                      >
+                        Add another {isTargeted ? 'targeted' : 'opportunistic'}{' '}
+                        contest
+                      </FormButton>
+                    </FormButtonBar>
+                  )}
                 </>
               )}
             />
--- a/client/src/components/MultiJurisdictionAudit/AASetup/Contests/index.tsx
+++ b/client/src/components/MultiJurisdictionAudit/AASetup/Contests/index.tsx
@@ -71,6 +71,7 @@ const Contests: React.FC<IProps> = ({
   if (!contests) return null // Still loading
   const filteredContests = contests.filter(c => c.isTargeted === isTargeted)
 
+  /* istanbul ignore next */
   if (isBatch && !isTargeted && nextStage.activate) nextStage.activate() // skip to next stage if on opportunistic contests screen and during a batch audit (until batch audits support multiple contests)
 
   const submit = async (values: { contests: IContest[] }) => {
--- a/client/src/components/MultiJurisdictionAudit/AASetup/Contests/index.tsx
+++ b/client/src/components/MultiJurisdictionAudit/AASetup/Contests/index.tsx
@@ -71,6 +71,8 @@ const Contests: React.FC<IProps> = ({
   if (!contests) return null // Still loading
   const filteredContests = contests.filter(c => c.isTargeted === isTargeted)
 
+  if (isBatch && !isTargeted && nextStage.activate) nextStage.activate() // skip to next stage if on opportunistic contests screen and during a batch audit (until batch audits support multiple contests)
+
   const submit = async (values: { contests: IContest[] }) => {
     const response = await updateContests(values.contests)
     // TEST TODO
--- a/client/src/components/MultiJurisdictionAudit/RoundManagement/BatchRoundDataEntry.test.tsx
+++ b/client/src/components/MultiJurisdictionAudit/RoundManagement/BatchRoundDataEntry.test.tsx
@@ -1,0 +1,96 @@
+import React from 'react'
+import { screen, fireEvent } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { useParams } from 'react-router-dom'
+import BatchRoundDataEntry from './BatchRoundDataEntry'
+import { roundMocks, contestMocks } from '../useSetupMenuItems/_mocks'
+import { batchesMocks, batchResultsMocks, INullResultValues } from './_mocks'
+import { renderWithRouter, withMockFetch } from '../../testUtilities'
+import { IContest } from '../../../types'
+import { IBatch } from './useBatchResults'
+
+jest.mock('react-router-dom', () => ({
+  ...jest.requireActual('react-router-dom'), // use actual for all non-hook parts
+  useRouteMatch: jest.fn(),
+  useParams: jest.fn(),
+}))
+const paramsMock = useParams as jest.Mock
+paramsMock.mockReturnValue({
+  electionId: '1',
+  jurisdictionId: '1',
+})
+
+const apiCalls = {
+  getJAContests: (response: { contests: IContest[] }) => ({
+    url: `/api/election/1/jurisdiction/1/contest`,
+    response,
+  }),
+  getBatches: (response: { batches: IBatch[] }) => ({
+    url: '/api/election/1/jurisdiction/1/round/round-1/batches',
+    response,
+  }),
+  putResults: (results: INullResultValues) => ({
+    url: '/api/election/1/jurisdiction/1/round/round-1/batches/results',
+    options: {
+      method: 'PUT',
+      body: JSON.stringify(results),
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    },
+    response: { status: 'ok' },
+  }),
+}
+
+describe('offline round data entry', () => {
+  it('renders', async () => {
+    const expectedCalls = [
+      apiCalls.getJAContests({ contests: contestMocks.oneTargeted }),
+      apiCalls.getBatches(batchesMocks.emptyInitial),
+    ]
+    await withMockFetch(expectedCalls, async () => {
+      const { container } = renderWithRouter(
+        <BatchRoundDataEntry round={roundMocks.singleIncomplete[0]} />,
+        {
+          route: '/election/1/jurisdiction/1',
+        }
+      )
+      await screen.findByText('Round 1 Data Entry')
+      expect(container).toMatchSnapshot()
+    })
+  })
+
+  it('submits', async () => {
+    const expectedCalls = [
+      apiCalls.getJAContests({ contests: contestMocks.oneTargeted }),
+      apiCalls.getBatches(batchesMocks.emptyInitial),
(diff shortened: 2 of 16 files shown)


MorganLove:
I just disabled the ability to add in multiple contests if it's a batch audit and marked where those changes were made so it's easy to undo when we start supporting that. It was pretty simple and doesn't break anything, so figured it was better safe than sorry.

client/src/components/MultiJurisdictionAudit/RoundManagement/BatchRoundDataEntry.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+    await withMockFetch(expectedCalls, async () => {
+      const { container } = renderWithRouter(
+        <BatchRoundDataEntry round={roundMocks.singleIncomplete[0]} />,
+        {
+          route: '/election/1/jurisdiction/1',
+        }
+      )
+      await screen.findByText('Round 1 Data Entry')
+      ;[0, 1, 2].forEach(batch => {
+        fireEvent.change(
```
jonahkagan, line 95: could use `userEvent.type`

client/src/components/MultiJurisdictionAudit/RoundManagement/BatchRoundDataEntry.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+jest.mock('react-router-dom', () => ({
+  ...jest.requireActual('react-router-dom'), // use actual for all non-hook parts
+  useRouteMatch: jest.fn(),
+  useParams: jest.fn(),
+}))
+const paramsMock = useParams as jest.Mock
+paramsMock.mockReturnValue({
+  electionId: '1',
+  jurisdictionId: '1',
+})
```
jonahkagan, line 21: do these all need to be mocked? i would have thought they would work when using `renderWithRouter`

client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
```diff
@@ -88,11 +89,15 @@ const RoundManagement = ({ round, auditBoards, createAuditBoards }: IProps) => {
...
+                  `/election/${electionId}/jurisdiction/${jurisdictionId}/round/${
+                    round.id
+                  }/${
+                    auditType === 'BALLOT_POLLING' ? 'ballots' : 'batches'
+                  }/retrieval-list`
                 )
               }
             >
-              Download Aggregated Ballot Retrival List for Round {roundNum}
+              Download Aggregated Ballot Retrieval List for Round {roundNum}
```
jonahkagan, line 100: this should say `Batch Retrieval List`

client/src/components/MultiJurisdictionAudit/AASetup/Contests/index.tsx
```diff
@@ -69,6 +71,9 @@ const Contests: React.FC<IProps> = ({
   if (!contests) return null // Still loading
   const filteredContests = contests.filter(c => c.isTargeted === isTargeted)
 
+  /* istanbul ignore next */
+  if (isBatch && !isTargeted && nextStage.activate) nextStage.activate() // skip to next stage if on opportunistic contests screen and during a batch audit (until batch audits support multiple contests)
```
jonahkagan, line 75: why is this necessary if you already removed opportunistic contests from the menu items?

MorganLove: @jonahkagan Because when you click the Next button on the Targeted Contests stage it goes to the Opportunistic Contests stage even if the menu item isn't displayed. Removing it entirely from the whole menu items code would have been waaaay more work than just skipping by it invisibly.

client/src/components/MultiJurisdictionAudit/RoundManagement/BatchRoundDataEntry.test.tsx
MorganLove: @jonahkagan I tried not mocking them but it broke. :shrug:

client/src/components/MultiJurisdictionAudit/RoundManagement/BatchRoundDataEntry.test.tsx
MorganLove: @jonahkagan I think the other bug with `fireEvent` applies here as well, it broke when I swapped it out.

client/src/components/MultiJurisdictionAudit/RoundManagement/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const [results, batches, updateResults] = useBatchResults(
+    electionId,
+    jurisdictionId,
+    round.id
+  )
+
+  if (!results || !allContests || !batches) return null
+
+  // batch comparison audits only support a single contest for now
+  const contest = allContests.filter(c => c.isTargeted)[0]
```
jonahkagan, line 47: this can just be `allContests[0]` if you want

client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
```diff
@@ -117,6 +124,8 @@ const RoundManagement = ({ round, auditBoards, createAuditBoards }: IProps) => {
                 <QRs passphrases={auditBoards.map(b => b.passphrase)} />
                 <RoundProgress auditBoards={auditBoards} round={round} />
               </>
+            ) : auditType === 'BATCH_COMPARISON' ? (
+              <BatchRoundDataEntry round={round} />
```
jonahkagan, line 128: this conditional seems a bit sketchy cuz it relies on `online` always being `false` for batch audits. maybe just move this case to the beginning of the condition, so that we only test `online` once we know we're not in a batch audit

client/src/components/MultiJurisdictionAudit/RoundManagement/useBatchResults.ts
```diff
@@ -0,0 +1,124 @@
...
+  IResultValues | null,
+  IBatch[] | null,
+  (arg0: IResultValues) => Promise<boolean>
+] => {
+  const [results, setResults] = useState<IResultValues | null>(null)
+  const [batches, setBatches] = useState<IBatch[] | null>(null)
+
+  const updateResults = async (newResults: IResultValues): Promise<boolean> => {
+    /* istanbul ignore next */
+    if (!results) return false
```
jonahkagan, line 91: not sure what this line is for

client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
MorganLove: @jonahkagan We hard code it to `online: false` if it's in a batch audit, though, so that's a given, right?

client/src/components/MultiJurisdictionAudit/RoundManagement/useBatchResults.ts
```diff
@@ -0,0 +1,124 @@
...
+          [choiceId]: numberify
+            ? Number(r[batchId][choiceId])
+            : stringifyPossibleNull(r[batchId][choiceId]),
+        }),
+        {}
+      ),
+    }),
+    {}
+  )
+}
```
jonahkagan, line 39: i've noticed this pattern and i have an idea about a simpler way to do it - if we convert to strings when you plug the data into the actual DOM components, and then convert back to numbers in the onChange event handler, I think it will be a lot more concise. what do you think?

client/src/components/MultiJurisdictionAudit/RoundManagement/useBatchResults.ts
MorganLove: @jonahkagan Oh dear, haha. I keep meaning to ask you what it's there for as well. I've taken it out and put it back in and I've seen zero functionality change, but I've just figured I was missing something. :stuck_out_tongue_closed_eyes:

client/src/components/MultiJurisdictionAudit/RoundManagement/BatchRoundDataEntry.test.tsx
jonahkagan: dang

client/src/components/MultiJurisdictionAudit/RoundManagement/useBatchResults.ts
MorganLove: @jonahkagan We'd be replacing the native Formik event handler which is conveniently handled magically currently. It's doable, but better to do it inside our form field wrapper across the board, and that's a relatively huge refactor. I'm not averse, just I think it's out of scope for this.

client/src/components/MultiJurisdictionAudit/RoundManagement/BatchRoundDataEntry.test.tsx
jonahkagan: don't necessarily need to dig into this too far, but DataEntry/index.test.tsx seems to not have this issue. maybe worth checking out

client/src/components/MultiJurisdictionAudit/AASetup/Contests/index.tsx
jonahkagan: gotcha, makes sense

client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
jonahkagan: yes, but reading this bit of code I didn't know that, and there's no other code that I know of that makes that assumption so if that ever changed this would just quietly break. it's only hard-coded to false because the default value in the db is already set to false - I don't think it was an intentional decision

client/src/components/MultiJurisdictionAudit/RoundManagement/useBatchResults.ts
jonahkagan: uhhh i have no idea. reading it, it just seems to disable updating before loading. i don't remember writing it (did i?)

client/src/components/MultiJurisdictionAudit/RoundManagement/BatchRoundDataEntry.test.tsx
MorganLove: @jonahkagan :monocle_face:

client/src/components/MultiJurisdictionAudit/RoundManagement/useBatchResults.ts
MorganLove: I thought you did, but now I'm not sure, haha. I'll take it out, though.

client/src/components/MultiJurisdictionAudit/RoundManagement/useBatchResults.ts
jonahkagan: ahhh right that makes sense. i have to say that so far I don't really understand what Formik does for us besides automatically running validation and inserting those error messages. the magic state updates don't seem that helpful, and I've often struggled to figure out how to integrate the formik components with the blueprint ones. are there things formik does that I'm just not appreciating?

agree it's not in scope to refactor everything here, though if you wanted to just try it out once, I think even replacing the formik event handler would be more concise haha

jonahkagan (approved):

client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
MorganLove: @jonahkagan Okay, had to look to find it, haha, but here's the original decision about this: https://github.com/votingworks/arlo/issues/639

Batch comparison audits are designed to only support offline. The tests are a bit sparse around this section, so I think we should include some tests to verify that when it's a batch audit it automatically submits it as offline during the settings stage. I seriously thought I'd already covered that test case somewhere but I can't find it anywhere, which is annoying. In the meantime I've added comments to clarify what's going on.

client/src/components/MultiJurisdictionAudit/RoundManagement/useBatchResults.ts
MorganLove: I'll re-take a look at what all Formik is providing us and refresh my memory as well. :)

PR #721, opened 2020-09-08
Refactor api utility to include toast functionality

MorganLove:
**Description**

- Moved the try catch blocks and the toasting to inside the api utility itself. This involved a lot of refactoring across the codebase, but it's a lot cleaner and easier to test now.

**Testing**

- A ton of test refactors (again, making it cleaner).

**Progress**

- The coverage didn't change much mostly because we had the branches for the old catch blocks and such being ignored by istanbul. They aren't ignored anymore and are covered now, so it's a huge leap forward in coverage even though the actual reported percentage is a tiny bit lower.


client/src/components/DataEntry/index.tsx
```diff
@@ -150,14 +148,17 @@ const DataEntry: React.FC = () => {
...
       auditBoard!.jurisdictionId,
       auditBoard!.roundId,
       auditBoardId,
       members
     )
-    setAuditBoard(await loadAuditBoard())
+    if (!response1) return
+    const response2 = await loadAuditBoard()
+    if (!response2) return
+    setAuditBoard(response2)
```
jonahkagan, line 161: this pattern seems a little tedious. before, the rejected promises short-circuited the function wherever there was a failure. now we have to manually do it ourselves. seems like we lose some of the zen of promises. I can't think of any way around it except letting the exception bubble up and trying to catch it at a higher level in the app (maybe with a React error boundary?)

alternatively we could just not short circuit on failure

jonahkagan (commented):
nice, this is exciting. just a few notes

client/src/components/utilities.ts
```diff
@@ -13,18 +13,23 @@ const tryJson = (responseText: string) => {
...
+      const error =
+        errors && errors.length ? errors[0] : { message: response.statusText }
+      const errorData = { ...error, responseText, response }
+      console.error(responseText) // eslint-disable-line no-console
+      throw errorData
+    }
+    return response.json() as Promise<T>
+  } catch (err) {
+    toast.error(err.message)
+    return false
```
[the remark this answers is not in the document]
MorganLove: @jonahkagan Fair enough. I don't think that would change its implementations at all, so should be easy to tweak.

client/src/components/DataEntry/index.tsx
MorganLove: @jonahkagan True, it did seem rather tedious. There are quite a few places where we handle an api failure in different ways, so it was kinda handy there. Trapping the error inside the api call for the toast is generally cleaner, but I also like the idea of using error boundaries. :thinking:

jonahkagan: I've never used error boundaries before, so can't really speak to whether they would be a good fit here. Since you already have this working, I'd say we just stick with it

PR #722, opened 2020-09-08
Implements more intelligent subsequent round MACRO sample size computation

umbernhard:
**Description**

Modifies the MACRO math to make more informed sample size estimates based on previously drawn samples. Notably, this doesn't change the conservative nature of the statistics, since the new computation will always require sampling at least one batch. Leaves the conservative math up to p-value computations. 

**Testing**

Adds test for `get_sample_sizes` which we really should have had anyways. 

**Progress**

Ready to go.


server/audit_math/macro.py
```diff
@@ -187,9 +194,12 @@ def get_sample_sizes(
...
     assert risk_limit < 1, "The risk-limit must be less than one!"
 
-    # TODO: actually use past batch results
+    # Computing U with the max error for already sampled batches knocked out
+    # to try to provide a sense of "how close" the audit is to finishing.
+    U = compute_U(reported_results, sample_results, contest)
 
-    U = compute_U(reported_results, contest)
+    if not U:
+        return 1
```
jonahkagan, line 202: what's this case about?

server/tests/audit_math/test_macro.py
```diff
@@ -136,6 +136,76 @@ def test_max_error(contests, batches):
...
+    for contest in contests:
+        computed = macro.get_sample_sizes(
+            RISK_LIMIT, contests[contest], batches, sample
+        )
+
+        assert (
+            expected_first_round[contest] == computed
+        ), "First round sample expected {}, got {}".format(
+            expected_first_round[contest], computed
+        )
```
jonahkagan, line 156: in my experience, the assertion error messages are pretty clear without writing a custom error message (they have really good diffs and line numbers attached)

jonahkagan (approved):

server/tests/audit_math/test_macro.py
jonahkagan: if it's useful to you, no prob

PR #725, opened 2020-09-08
Update batch results

jonahkagan:
- Add a separate endpoint to get batch results
- When no results are recorded, return full results object (with choice
ids) instead of null
- Filter out batches that were sampled in previous rounds in batches and
results endpoints

### DIFF
--- a/server/api/batches.py
+++ b/server/api/batches.py
@@ -126,6 +126,7 @@ def validate_batch_results(
         Batch.query.filter_by(jurisdiction_id=jurisdiction.id)
         .join(SampledBatchDraw)
         .filter_by(round_id=round.id)
+        .filter(Batch.id.notin_(already_audited_batch_ids(jurisdiction, round)))
         .order_by(Batch.name)
         .all()
     )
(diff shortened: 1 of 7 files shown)


jonahkagan:
@umbernhard I made a more realistic test case! Would love your review for that

@benadida tagging you to review the db queries specifically. I feel like they could be improved but I'm not sure how

umbernhard (approved):

MorganLove:
@jonahkagan This is giving me an `Invalid batch ids` error on the second round. I rebased #720 on this so you can try it out and see what might be going on.

benadida (approved):
Two possible improvements to the queries:

- instead of pulling the data out into the python code, you could create a subquery that you use in the NOT IN clause. It's a tad cleaner and it's one query instead of two.

- you might get some benefit from querying with an "IN" instead of "NOT IN", but ... I'm not 100% sure and since I imagine the number of audited batches is smaller, it may not be a benefit.

These are things to potentially look into, but I don't think they block merging. This kind of optimization is for later once we find actual performance bottlenecks.

jonahkagan:
Made the subquery change: 1b3be7e. very easy!

PR #731, opened 2020-09-10
using 3.8 instead of 3.7

MorganLove:
**Description**

- upgrading to 3.8

**Testing**

- shouldn't affect tests

**Progress**

- yey!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ orbs:
 jobs:
   build-and-test:
     docker:
-      - image: circleci/python:3.7.3-node-browsers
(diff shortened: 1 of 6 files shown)


benadida:
I have no objections but since I'm not in the code day-in day-out, I want to let @jonahkagan approve this one.

MorganLove:
And I'm not able to grok what the error is on CI, so I'll let @jonahkagan take a gander at that as well, haha.

jonahkagan:
I was able to upgrade my dev env like so:

```
make deps
pipenv --rm
make install-development
```

And then the tests pass!

jonahkagan (approved):

Issue #684, opened 2020-08-11
Backend: only allow one targeted contest for batch audits


jonahkagan:

nealmcb:
Is this just a temporary change? We audited multiple contests in parallel in Boulder for many years via batch auditing.
So supporting audits of multiple contests in a batch audit should be a goal, even if not top priority.

PR #732, opened 2020-09-10
Batch audit guards

jonahkagan:
Task: #684 

- Validate batch tallies files
  - Check that all jurisdictions have uploaded a file before computing
sample sizes/starting a round
  - Check that the total tallies for each choice don't exceed the
reported vote count for that choice
- Only allow one contest

### DIFF
--- a/server/api/rounds.py
+++ b/server/api/rounds.py
@@ -1,6 +1,6 @@
 import uuid
 from collections import defaultdict
-from typing import Optional, NamedTuple, List, Tuple, Dict
+from typing import Optional, NamedTuple, List, Tuple, Dict, cast as typing_cast
 from datetime import datetime
 from flask import jsonify, request
 from jsonschema import validate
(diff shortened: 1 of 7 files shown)


server/api/rounds.py
```diff
@@ -122,13 +122,36 @@ def cumulative_contest_results(contest: Contest) -> Dict[str, int]:
 BatchTallies = Dict[Tuple[str, str], Dict[str, Dict[str, int]]]
 
 
-def batch_tallies(election: Election,) -> BatchTallies:
+def batch_tallies(election: Election) -> BatchTallies:
+    # We only support one contest for batch audits
+    contest = list(election.contests)[0]
```
benadida, line 127: should we throw here if `len()>1` just to prevent future bugs?

benadida (approved):

server/api/rounds.py
jonahkagan: hmm, not sure what you mean - we prevent that from happening in this PR

if you're suggesting a backup (cuz i know you like multiple layers of safety), then I'm open to it, but also we do this like 10 places in the code, so we'd have to do it everywhere. personally I'm not that worried about it - but maybe I'm just lazy. I also wrote a comment every time haha, not that it helps. i guess I could replace the comment with an `assert`...

benadida: yeah to be clear the only thing I'm suggesting here is considering an `assert` or the like so that, when you change the *other* code, it doesn't magically look like it's working the first time you try a two-contest audit.

PR #733, opened 2020-09-11
Ensure all batch tallies files are processed

### LINKED ISSUE #717: Require jurisdiction data to calculate # of batches that need to be audited
Enforce that all jurisdiction uploads must be complete before "# of batches to be sampled" options are shown on the AA Audit Setup - Audit Review screen.
 
Until uplaods are finished, show a message "Waiting for jurisidiction data to calculate required number of batches ....

(backend: also validate that batch tallies add up to vote total from contest settings)

MorganLove:
**Description**

Closes #717 

**Testing**

- updated mocks, tests, snapshots

**Progress**

- ready to go!


client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
```diff
@@ -205,111 +214,118 @@ const Review: React.FC<IProps> = ({ prevStage, locked, refresh }: IProps) => {
...
-                                  {`${option.size} samples`}
-                                  {option.prob
-                                    ? ` (${percentFormatter.format(
-                                        option.prob
-                                      )} chance of reaching risk limit and completing the audit in one round)`
-                                    : ''}
-                                </Radio>
+      {auditType !== 'BATCH_COMPARISON' || talliesUploadsCompleted ? (
+        <p>
+          Waiting for jurisdiction data to calculate required number of batches
```
jonahkagan, line 219: what about something like "All jurisdiction files must be uploaded to calculate the sample size. View jurisdiction upload progress." with "View jurisdiction upload progress" linking to the progress view

jonahkagan (commented):
looks good so far

client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
MorganLove: This was the text @mcchilders provided, but I like what you suggested. @mcchilders thoughts?

client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
```diff
@@ -197,110 +207,118 @@ const Review: React.FC<IProps> = ({ prevStage, locked, refresh }: IProps) => {
...
-      </Formik>
+                          </RadioGroup>
+                          {currentOption && currentOption.key === 'custom' && (
+                            <Field
+                              component={FormField}
+                              name={`sampleSizes[${contest.id}][size]`}
+                              type="text"
+                              validate={testNumber(
+                                Number(contest.totalBallotsCast),
+                                `Must be less than or equal to: ${contest.totalBallotsCast} (the total number of ballots in this targeted contest)`
```
jonahkagan, line 295: this validation probably needs to change for batch audits

jonahkagan: on the backend we check that it's no greater than the total number of batches, so you could also just remove the validation and let the backend handle it if that's easier

client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
```diff
@@ -197,110 +207,118 @@ const Review: React.FC<IProps> = ({ prevStage, locked, refresh }: IProps) => {
...
+                  intent="primary"
+                  disabled={
+                    locked ||
+                    !isSetupComplete(jurisdictions, contests, auditSettings)
+                  }
+                  onClick={handleSubmit}
+                >
+                  Launch Audit
+                </FormButton>
+              </FormButtonBar>
```
jonahkagan, line 317: does this hide the launch audit button if the tallies files haven't been uploaded? maybe it would be better to just disable it

MorganLove: @jonahkagan It hides it currently. The reason (from what I remember) is because the information for rendering the form is dependent on the return from the sample size option calculation, which can't be called if the tallies files haven't been uploaded. The launch audit button is also the submit button for the form, so it's dependent on the form being there.

client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
MorganLove: @jonahkagan In general I think it's safest to mirror backend validation on the frontend as much as possible, so users don't have to do things over again if they mess up. Much more friendly to let them know right away.

@mcchilders What's your take on this one?

client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
jonahkagan: could you conditionally render the content of the form that depends on the sample size options and leave the button?

client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
mcchilders: Ok that language change is fine @MorganLove!

client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
mcchilders: @MorganLove yes, let's mirror if possible.

MorganLove: Okay so @jonahkagan what's the exact math that's used on the backend here? Do you mean the total number of ballots in all the batches?

jonahkagan: Omg I just realized we don't have validation of custom sample sizes on the backend after all (for any audit type). I got confused cuz there's so much other validation.

jonahkagan: The sample size for batch audits is the number of batches to be sampled, so we should validate that it's no greater than the total number of batches across the jurisdictions included in the audit.

jonahkagan: Just created #740 to implement the backend validation

client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
MorganLove: @jonahkagan Success! I was able to make that work, and along the way I found some bugs that I was able to fix as well, haha.

client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
MorganLove: @jonahkagan just changed the validation of the custom sample size for batch comparison audits to just check that it's a number, since evidently the math doesn't require that it have the batch number validation. From what I can tell adding in that validation will add a significant amount of complexity, so for MVP we can do without.

client/src/components/HomeScreen.tsx
```diff
@@ -283,7 +283,7 @@ const CreateAudit: React.FC = () => {
                 onChange={e =>
                   setFieldValue('auditType', e.currentTarget.value)
                 }
-                selectedValue={values.auditType}
+                selectedValue={values.auditType || 'BALLOT_POLLING'}
```
jonahkagan, line 286: what's this change for?

client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
```diff
@@ -53,7 +53,19 @@ const Review: React.FC<IProps> = ({ prevStage, locked, refresh }: IProps) => {
...
+    jurisdictions.every(
+      j =>
+        j.batchTallies &&
+        j.batchTallies.processing &&
+        j.batchTallies.processing.status === FileProcessingStatus.PROCESSED
+    )
+  const [sampleSizeOptions, uploadSampleSizes] = useSampleSizes(
+    electionId,
+    auditType === 'BALLOT_POLLING' || talliesUploadsCompleted // only fetch sample sizes for ballot polling audits or if all tallies files are uploaded
+  )
```
jonahkagan, line 68: ok i should have thought of this earlier but the idea just came to me. something seems a little weird to me about having to tell the `useSampleSizes` hook whether it should fetch or not. instead, what if you pulled out the sample size form into its own component that calls the `useSampleSizes` hook? then you could conditionally render the component instead of putting the condition into the hook

client/src/types.ts
```diff
@@ -204,5 +204,5 @@ export interface IAuditSettings {
   online: boolean | null
   randomSeed: string | null
   riskLimit: number | null
-  auditType: 'BALLOT_POLLING' | 'BATCH_COMPARISON'
+  auditType: 'BALLOT_POLLING' | 'BATCH_COMPARISON' | null
```
jonahkagan, line 207: I actually don't think this can ever be null, since it gets set when the audit is created. it's not allowed to be null in the db

client/src/components/HomeScreen.tsx
MorganLove: @jonahkagan See comment below.

client/src/types.ts
MorganLove: @jonahkagan This is why I needed to update the `selectedValue={values.auditType || 'BALLOT_POLLING'}` bit. Defaulting to 'BALLOT_POLLING' was breaking the conditions since you only care about waiting for the batch tallies on batch audits. And before the apis were returning, it was causing lots of issues. So I defaulted the setting to null like the rest of the settings, and just set the initial field to default to 'BALLOT_POLLING' if the value was null. But of course that needed to add null to the type like the rest. So basically it's only null before it hears back from the api.

client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
MorganLove: @jonahkagan Yeah, that could conceivably work, but again that runs into issues with the submit button. If you want to show a disabled submit button instead of nothing, we'll need to have them both in this component.

client/src/types.ts
jonahkagan: hmm, what if the entire settings state was `null` before the API returns. would that solve this problem?

client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
jonahkagan: oh, right. maybe just the form section then?

client/src/types.ts
MorganLove: @jonahkagan That would break several things, actually. You need the structure there to be able to destructure from it. I think this right now is the least complex solution.

client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
MorganLove: @jonahkagan I just looked, but we need to declare the initial values outside of the form, and that requires the sample size options. So extracting out the `useSampleSizes` hook elsewhere would not work without adding complexity. I think passing in the toggle for whether the hook should fetch or not is the simplest solution at this point.

client/src/types.ts
jonahkagan: you can check for null before you destructure. something like this:

```ts
const [auditSettings, updateAuditSettings] = useAuditSettings()

if (!auditSettings) return null

const {auditType} = auditSettings
```

the reason I suggest this is because it's the pattern that all the other API hooks use, and I think being consistent will make it easier to read/maintain the code

client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
jonahkagan: ah i see what you mean

jonahkagan (approved):

client/src/types.ts
MorganLove: @jonahkagan Okay, trying this is actually... really involved and changes a bunch of things across the board. So I'm merging this and introducing the refactor in a separate PR.

PR #735, opened 2020-09-15
Adding auto-refresh to the progress view

### LINKED ISSUE #696: Add auto-refresh and countdown timer to audit progress on AA dashboard
Right now you have to manually refresh the audit progress screen on the AA dashboard to see updated counts - we should add an auto-refresh function and an onscreen timer so that you can see how recently it was updated

MorganLove:
**Description**

Closes #696 

- Adds a `useInterval` custom hook for managing `setInterval` and `clearInterval` in React

**Testing**

- none yet

**Progress**

- It polls every 5 minutes successfully and without any problems, very clean.
- It displays in a very human friendly fashion how long it's been since the last refresh

- Need to test it all

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/index.test.tsx
+++ b/client/src/components/MultiJurisdictionAudit/index.test.tsx
@@ -217,6 +217,40 @@ describe('AA setup flow', () => {
       })
     })
   })
+
+  describe.skip('timers', () => {
+    beforeEach(() => {
+      jest.useFakeTimers()
+    })
+    afterEach(() => {
+      jest.runOnlyPendingTimers()
+      jest.useRealTimers()
+    })
+
+    it('refreshes every five minutes on progress', async () => {
+      paramsMock.mockReturnValue({
+        electionId: '1',
+        view: 'progress',
+      })
+      const expectedCalls = [aaApiCalls.getUser, ...loadEach, ...loadEach]
+      await withMockFetch(expectedCalls, async () => {
+        render(
+          <AuthDataProvider>
+            <Router>
+              <AuditAdminViewWithAuth />
+            </Router>
+          </AuthDataProvider>
+        )
+        jest.advanceTimersByTime(5000)
+
+        await screen.findByText('Refreshed just now')
+        jest.advanceTimersByTime(180000)
+        await screen.findByText('Refreshed 3 minutes ago')
+        jest.advanceTimersByTime(120005)
+        await screen.findByText('Refreshed just now')
+      })
+    })
+  })
 })
 
 describe('JA setup', () => {
--- a/client/src/components/MultiJurisdictionAudit/index.test.tsx
+++ b/client/src/components/MultiJurisdictionAudit/index.test.tsx
@@ -218,13 +218,17 @@ describe('AA setup flow', () => {
     })
   })
 
-  describe.skip('timers', () => {
-    beforeEach(() => {
-      jest.useFakeTimers()
-    })
-    afterEach(() => {
-      jest.runOnlyPendingTimers()
-      jest.useRealTimers()
+  describe('timers', () => {
+    const j = (function* idMaker() {
+      let index = 0
+      while (true) yield (index += 180000)
+    })()
+    const dateSpy = jest
+      .spyOn(Date, 'now')
+      .mockImplementation(() => j.next().value)
+
+    afterAll(() => {
+      dateSpy.mockRestore()
     })
 
     it('refreshes every five minutes on progress', async () => {
@@ -241,12 +245,12 @@ describe('AA setup flow', () => {
             </Router>
           </AuthDataProvider>
         )
-        jest.advanceTimersByTime(5000)
+        // jest.advanceTimersByTime(5000)
 
         await screen.findByText('Refreshed just now')
-        jest.advanceTimersByTime(180000)
+        // jest.advanceTimersByTime(180000)
         await screen.findByText('Refreshed 3 minutes ago')
-        jest.advanceTimersByTime(120005)
+        // jest.advanceTimersByTime(120005)
         await screen.findByText('Refreshed just now')
       })
     })
--- a/client/src/components/MultiJurisdictionAudit/index.test.tsx
+++ b/client/src/components/MultiJurisdictionAudit/index.test.tsx
@@ -6,7 +6,11 @@ import {
   Router as RegularRouter,
   useParams,
 } from 'react-router-dom'
-import { AuditAdminView, JurisdictionAdminView } from './index'
+import {
+  AuditAdminView,
+  JurisdictionAdminView,
+  prettifyRefreshStatus,
+} from './index'
 import {
   manifestMocks,
   talliesMocks,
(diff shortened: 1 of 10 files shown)


client/src/components/utilities.ts
```diff
@@ -125,3 +126,31 @@ export const checkAndToast = (
...
+  }, [callback])
+
+  // Set up the interval.
+  useEffect(/* eslint consistent-return: "off" */
+  () => {
+    function tick() {
+      savedCallback.current!()
+    }
+    if (callImmediately) tick()
+    if (delay !== null) {
```
jonahkagan, line 151: impossible for `delay` to be `null` according to the type

jonahkagan (commented):
nice!

client/src/components/utilities.ts
MorganLove: Good catch, I'll fix that.

client/src/components/MultiJurisdictionAudit/index.tsx
```diff
@@ -50,9 +66,21 @@ export const AuditAdminView: React.FC = () => {
...
+  useInterval(
+    () => {
+      setRefreshTime(0)
+      refresh()
+    },
+    300000,
+    true
+  )
+  // update the refresh time tracker every second
+  useInterval(incrementRefreshTime, 1000, false)
```
jonahkagan, line 82: how much do you trust JS intervals? in my experience, they are so-so at actually doing things on time over time. In other words, I don't trust that this refresh time will actually add up correctly. see: https://stackoverflow.com/a/8332678/1472662

I'm wondering if it would be more reliable to store the last refresh time as a timestamp and then have an interval that checks how long it's been. something like this:

```ts
const [lastRefreshTime, setLastRefreshTime] = useState(Date.now())
useInterval(() => {
  if (Date.now() - lastRefreshTime > 60 * 60 * 5) {
    refresh()
    setLastRefreshTime(Date.now())
  }
})

const refreshStatus = prettifyRefreshStatus(Date.now() - lastRefreshTime)
```

no idea if this works haha, just sketching

client/src/components/MultiJurisdictionAudit/index.tsx
```diff
@@ -111,6 +142,9 @@ export const AuditAdminView: React.FC = () => {
               auditSettings={auditSettings}
             />
           </Inner>
+          <RefreshStatusWrapper>
+            <Tag>{refreshStatus}</Tag>
+          </RefreshStatusWrapper>
```
jonahkagan, line 147: Since the progress table can be arbitrarily long (depending on the number of jurisdictions), I think this might go better at the top of the page so it is always visible. Here are a few places I think it might work well

Screen Shot 2020-09-17 at 2 50 18 PM

client/src/components/MultiJurisdictionAudit/index.tsx
```diff
@@ -23,11 +24,26 @@ import useRoundsAuditAdmin from './useRoundsAuditAdmin'
...
 
 const VerticalInner = styled(Inner)`
   flex-direction: column;
 `
 
+const RefreshStatusWrapper = styled.div`
+  display: flex;
+  justify-content: center;
+  margin-top: 40px;
+`
```
jonahkagan, line 45: since the content inside changes length, might be nice to give this (or the tag) a fixed width. not totally sure which would look better, just a random thought

client/src/components/MultiJurisdictionAudit/index.tsx
MorganLove: @jonahkagan Good point, and yeah, I had some concerns about that as well. This does look like a more stable way of doing it. I'll see what I can put together!

client/src/components/MultiJurisdictionAudit/index.tsx
MorganLove: @jonahkagan I think the best option would be to modify the status box to accept children and display them below the regular content. Putting it in the right would conflict with the round launch buttons and so forth. I'll look into that.

client/src/components/MultiJurisdictionAudit/index.tsx
MorganLove: @jonahkagan Agreed!

MorganLove:
Testing these timers is a royal pain, haha. And they kept messing with the other tests and I couldn't figure out why they were affecting each other at all, so I ended up moving them into a separate file, at least for now. I'm still not happy with how the tests are behaving, even though they're passing.

jonahkagan (commented):
still working on figuring out the tests but submitting these comments for now

client/src/components/MultiJurisdictionAudit/index.tsx
```diff
@@ -50,9 +66,23 @@ export const AuditAdminView: React.FC = () => {
...
+  // poll the apis every 5 minutes
+  // TODO figure out how to test timer-related code
+  /* istanbul ignore next */
+  useInterval(() => {
+    if (time - lastRefreshTime > 1000 * 60 * 5) {
+      refresh()
+      setLastRefreshTime(time)
+    }
+  }, 500)
+  useInterval(updateTime, 500) // have to force rerender on a regular basis so the clocks update regularly
```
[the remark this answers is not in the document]
MorganLove: @jonahkagan Oh yes, that does look much better. And yeah, I originally had it as just a second, but I changed to half a second during the testing to see if it made a difference and forgot to change it back.

client/src/components/MultiJurisdictionAudit/index.test.tsx
```diff
@@ -328,3 +332,28 @@ describe('JA setup', () => {
...
+  })
+  it('handles a single minute', () => {
+    expect(prettifyRefreshStatus(60001)).toBe('Refreshed 1 minute ago')
+    expect(prettifyRefreshStatus(119999)).toBe('Refreshed 1 minute ago')
+  })
+  it('handles minute increments', () => {
+    expect(prettifyRefreshStatus(120000)).toBe('Refreshed 2 minutes ago')
+    expect(prettifyRefreshStatus(180000)).toBe('Refreshed 3 minutes ago')
+    expect(prettifyRefreshStatus(240000)).toBe('Refreshed 4 minutes ago')
+    expect(prettifyRefreshStatus(300000)).toBe('Refreshed 5 minutes ago')
```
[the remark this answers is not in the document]
MorganLove: @jonahkagan :thinking: I actually like that, it seems more useful, even if I'm not sure if I've seen it done that way before.

client/src/components/utilities.ts
MorganLove: @jonahkagan This is fixed by adding `null` to the type. The point of this is so that you can have a state variable controlling the delay, and change it to `null` to pause the loop. It's a feature of the original function that I thought could be useful at some point and so I left it in. The original didn't have types, and I forgot to include that when I added them. So this isn't an error check, it's actually a feature.

jonahkagan: ah cool!

client/src/components/MultiJurisdictionAudit/index.test.tsx
```diff
@@ -328,3 +329,25 @@ describe('JA setup', () => {
...
+  })
+
+  it('handles minute increments', () => {
+    expect(prettifyRefreshStatus(60000)).toBe('Will refresh in 4 minutes')
+    expect(prettifyRefreshStatus(120000)).toBe('Will refresh in 3 minutes')
+    expect(prettifyRefreshStatus(180000)).toBe('Will refresh in 2 minutes')
+  })
+
+  it('handles ten second increments', () => {
+    expect(prettifyRefreshStatus(240000)).toBe('Will refresh in 60 seconds')
```
jonahkagan, line 346: small nit: it's more common to say "1 minute" than "60 seconds"

jonahkagan (approved):
looks good!

PR #742, opened 2020-09-17
Multiple user types per route

jonahkagan:
Reworks the Flask route decorators that control permissions to allow multiple user types per route. This will be useful since we want to start letting AAs access more of the JA routes.

Also updates a ton of tests, which were lazily using an election that wasn't tied to an organization to circumvent the need to have the correct AA user logged in.

### DIFF
--- a/server/auth/lib.py
+++ b/server/auth/lib.py
@@ -1,8 +1,9 @@
 import functools
 import enum
-from typing import Callable, Optional, Tuple, Union
+from typing import Callable, Tuple, Union, List
 from flask import session
 from werkzeug.exceptions import Forbidden, Unauthorized
+from sqlalchemy.orm import Query
 
 from ..models import *  # pylint: disable=wildcard-import
 
@@ -40,7 +41,6 @@ def clear_loggedin_user():
     session[_USER] = None
 
 
-##
 ## The super admin bit lets a user impersonate any other user
 ## Having the bit only grants access to the superadmin functionality
 ## that enables becoming any other user, and then taking action as them.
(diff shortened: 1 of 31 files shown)


server/auth/lib.py
```diff
@@ -65,211 +62,118 @@ def is_superadmin():
...
+                election = get_or_404(Election, kwargs.pop("election_id"))
+                kwargs["election"] = election
+
+            jurisdiction = None
+            if "jurisdiction_id" in kwargs:
+                jurisdiction = find_or_404(
+                    Jurisdiction.query.filter_by(
+                        id=kwargs.pop("jurisdiction_id"), election_id=election.id
+                    )
+                )
```
benadida, line 136: this will fail if the path has `jurisdiction_id` but not `election_id`, right? Maybe fail explicitly in that case?

server/auth/lib.py
```diff
@@ -65,211 +62,118 @@ def is_superadmin():
...
+                kwargs["jurisdiction"] = jurisdiction
+
+            round = None
+            if "round_id" in kwargs:
+                round = find_or_404(
+                    Round.query.filter_by(
+                        id=kwargs.pop("round_id"), election_id=election.id
```
benadida, line 143: same coment as above.

server/auth/lib.py
```diff
@@ -65,211 +62,118 @@ def is_superadmin():
...
+                )
+                kwargs["round"] = round
+
+            audit_board = None
+            if "audit_board_id" in kwargs:
+                audit_board = find_or_404(
+                    AuditBoard.query.filter_by(
+                        id=kwargs.pop("audit_board_id"),
+                        round_id=round.id,
+                        jurisdiction_id=jurisdiction.id,
```
benadida, line 154: same with `jurisdiction_id` missing.

benadida (approved):

server/auth/lib.py
jonahkagan: ea499a5

PR #744, opened 2020-09-21
Adding total ballots in manifest to progress screen

### LINKED ISSUE #695: Add JA manifest totals to Audit Progress view
Add a column with number of ballots total in each county's manifest to the AA dashboard Audit Progress table so the AA can sanity-check that the various manifests at least have the right number of ballots in them. We've been doing that manually, so let's make that an on-demand feature in the UI.

MorganLove:
**Description**

- Added a column displaying the data from `jurisdiction.ballotManifest.numBallots`

Closes #695 

**Testing**

- Snapshots updated

**Progress**

- Done, unless that's the wrong value to use, haha.


jonahkagan (approved):
So easy! It looks like there are also some tests in Progress/index.test.tsx that should also be updated

PR #747, opened 2020-09-22
Restyle/refactor JA RoundManagement screen

jonahkagan:
- Ensure there's always a heading at the top of the page
- Pull the num ballots message out of the CreateAuditBoards component so that
it still shows during data entry
- Remove unnecessary props from CreateAuditBoards
- Put the buttons in a ButtonGroup and add icons so they look nice (this
will be useful when we reuse those buttons in the AA Audit Progress
view)
- Shrink the page width so that lines of text aren't super long
- Remove the special-case styles from the Wrapper class, since they are
only used in this page and the single-jurisdiction flow

I separated out the snapshot updates into a separate commit because there were so many of them.

Screen Shot 2020-09-22 at 11 22 32 AM
Screen Shot 2020-09-22 at 11 23 22 AM
Screen Shot 2020-09-22 at 11 29 06 AM

### DIFF
--- a/client/src/__snapshots__/App.test.tsx.snap
+++ b/client/src/__snapshots__/App.test.tsx.snap
@@ -6,7 +6,7 @@ exports[`App / renders aa logged in properly 1`] = `
     class="Toastify"
   />
   <div
-    class="sc-lhVmIH daMHBT"
+    class="sc-elJkPf jKMEUh"
   >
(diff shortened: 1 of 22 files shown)


MorganLove (approved):
Looks good! I don't see any issues. :slightly_smiling_face:

PR #750, opened 2020-09-23
Interactive configuration and launching of Arlo locally

umbernhard:
**Description**

A script to interactively configure the Arlo dev environment to help me (and other contributors) spin up arlo locally. Also updates the README accordingly.

**Testing**
N/A

**Progress**
Ready to go.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ We recommend Ubuntu 18.0.4.
 
 #### Automatic configuration and setup
 
-If you would just like to run Arlo and do not wish to setup a custom configuration, you can run `pipenv run python -m scripts.setup-dev`, which provides interactive configuration. The script installs VotingWorks' fork of nOAuth locally, runs it, and configures Arlo to use it. It creates the necessary audit administrator and jurisdiction administrator credentials discussed above, and launches a dev instance of Arlo. Once you have navigated to `localhost:3000` in your broswer, you should be able to log in as an audit admin using the credentials you configured earlier in the script. 
+If you would just like to run Arlo and do not wish to setup a custom configuration, you can run `pipenv run python -m scripts.setup-dev`, which provides interactive configuration. The script optionally installs VotingWorks' [nOAuth](https://github.com/votingworks/nOAuth) locally, runs it, and configures Arlo to use it. It creates the necessary audit administrator and jurisdiction administrator credentials discussed above, and launches a dev instance of Arlo. Once you have navigated to `localhost:3000` in your broswer, you should be able to log in as an audit admin using the credentials you configured earlier in the script. 
 
 #### Troubleshooting
 
--- a/README.md
+++ b/README.md
@@ -105,6 +105,8 @@ Arlo is configured mostly through environment variables:
 - `ARLO_AUDITADMIN_AUTH0_BASE_URL`, `ARLO_AUDITADMIN_AUTH0_CLIENT_ID`, `ARLO_AUDITADMIN_AUTH0_CLIENT_SECRET`: base url, client id, and client secret for the auth0 app used for audit admins.
 - `ARLO_JURISDICTIONADMIN_AUTH0_BASE_URL`, `ARLO_JURISDICTIONADMIN_AUTH0_CLIENT_ID`, `ARLO_JURISDICTIONADMIN_AUTH0_CLIENT_SECRET`: base url, client id, and client secret for the auth0 app used for jurisdiction admins.
 
+Rather than manually config the environment, you can also run the setup script discussed below.
+
 ### Creating Organizations and Administrators
 
 Organizations are, for example, the State of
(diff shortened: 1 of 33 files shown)


README.md
```diff
@@ -151,6 +153,10 @@ We recommend Ubuntu 18.0.4.
 - `make dev-environment` or, if you prefer, look at individual make tasks like `deps`, `initdevdb`, `install-development`, and `resetdb`
 - `bash ./run-dev.sh`
 
+#### Automatic configuration and setup
+
+If you would just like to run Arlo and do not wish to setup a custom configuration, you can run `pipenv run python -m scripts.setup-dev`, which provides interactive configuration. The script installs VotingWorks' fork of nOAuth locally, runs it, and configures Arlo to use it. It creates the necessary audit administrator and jurisdiction administrator credentials discussed above, and launches a dev instance of Arlo. Once you have navigated to `localhost:3000` in your broswer, you should be able to log in as an audit admin using the credentials you configured earlier in the script.
```
jonahkagan, line 158: we invented nOAuth, it's not a fork

scripts/setup-dev.py
```diff
@@ -0,0 +1,146 @@
...
+
+try:
+    owd = os.getcwd()
+
+    d = ""
+    noAuthPath = ""
+
+    print("Running script to setup local Arlo instance...")
+
+    res = input("Do you have an instance of nOAuth installed? [y/N]")
```
jonahkagan, line 17: any reason not just to use the VotingWorks nOAuth instance that's hosted on Heroku? the only reason I can think of is running Arlo offline, but it seems unlikely that someone would need to do that

scripts/setup-dev.py
```diff
@@ -0,0 +1,146 @@
...
+    )
+
+    os.environ["ARLO_AUDITADMIN_AUTH0_CLIENT_ID"] = "test"
+    os.environ["ARLO_JURISDICTIONADMIN_AUTH0_CLIENT_ID"] = "test"
+
+    os.environ["ARLO_AUDITADMIN_AUTH0_CLIENT_SECRET"] = "secret"
+    os.environ["ARLO_JURISDICTIONADMIN_AUTH0_CLIENT_SECRET"] = "secret"
+
+    print("Running Arlo. Happy auditing!")
+    subprocess.run(["./run-dev.sh"], check=True)
```
jonahkagan, line 138: do you need to set up the db first by running `make resetdb`?

scripts/setup-dev.py
MorganLove: Maybe offer an option for doing either? If someone wants to start a completely new setup from scratch it could make sense for them to want to have all the pieces under their control. Also from a dev perspective, it's nice to be able to have the option to code even when on the road or wherever.

README.md
umbernhard: Good note, will fix!

scripts/setup-dev.py
umbernhard: I'll mess around with it. I just wrote this after the way I use nOAuth. I didn't know we had web instance :sweat_smile:

scripts/setup-dev.py
umbernhard: I'm not sure. The install steps will do that, and I'm assuming here that if you've gotten this far you've already done it. The upside to doing it that way is that subsequent runs of the script don't erase credentials, so if you run the script twice with the same credentials you don't have to log in again if your browser saves the session. The downside is that if we monkey with the data model and don't explicitly `make resetdb`, it breaks. Thoughts?

jonahkagan: another option is to run the latest db migrations (if there are any). we do this on every heroku release in order to keep the db up to date while preserving the existing data: https://github.com/votingworks/arlo/blob/main/Procfile#L1

umbernhard: ah, that sounds like the right answer!

umbernhard:
@jonahkagan @MorganLove okay, I've implemented your feedback. The script now defaults to using the heroku nOAuth, but asks the user if they want to install it locally. I added database migration, and I changed the README.

scripts/setup-dev.py
```diff
@@ -0,0 +1,162 @@
...
+
+    res = input(
+        "Would you like to install the Arlo dev environment?\n"
+        + "If this is your first run of Arlo after install, you should do this. [y/N]"
+    )
+
+    if res == "y":
+        subprocess.run(["make dev-environment"], check=True, shell=True)
+
+    print("Migrating the database to the newset data model")
```
jonahkagan, line 120: newset -> newest

scripts/setup-dev.py
```diff
@@ -0,0 +1,162 @@
...
+
+try:
+    owd = os.getcwd()
+
+    d = ""
+    noAuthPath = ""
+
+    print("Running script to setup local Arlo instance...")
+
+    res = input("Do you have an instance of nOAuth configured? [y/N]")
```
jonahkagan, line 17: if you are using Arlo for the first time, I would imagine you don't have any idea what nOAuth is or whether it's better to use our instance or your own. maybe add a small explanation or recommendation?

jonahkagan (approved):

scripts/setup-dev.py
umbernhard: Good call. Will add.

PR #752, opened 2020-09-24
Save table state on refresh

jonahkagan:
Two problems were happening:

1. Every time the useInterval that checks if it's time to refresh
ticked, the whole Progress component (including the table) was getting
re-rendered. (The React DevTools profiler is amazing and highlights each
component as it renders, so it was easy to figure out). I solved this by
isolating the state for the refresh tag in its own component.

2. When the Progress component re-rendered, the Table reset its sort
state. This was originally happening every second, but after fixing the
above problem, was still happening when the data actually refreshed
every 5 minutes. I fixed this by passing `autoResetSortBy: false` to the
Table config.


MorganLove (approved):
LGTM! And this might actually help with extracting out some of the tests and making them more stable as well.

PR #749, opened 2020-09-23
Settings defaults to null before api returns object

MorganLove:
**Description**

- Trying to refactor this to be more consistent in api usage

**Testing**

- It works, but there's one test that's failing and I'm not sure how it's even related.

**Progress**

- Still puzzled by this.

PR #754, opened 2020-09-26
Refresh onmount

MorganLove:
**Description**

- fixes the sidebar, since it needs to refresh after everything is mounted so it will accurately reflect the proper state. This refresh got removed in another PR, this just puts it back in.

**Testing**

- updated

**Progress**

- ready to go

### DIFF
--- a/client/src/components/HomeScreen.test.tsx
+++ b/client/src/components/HomeScreen.test.tsx
@@ -80,8 +80,6 @@ const setupScreenCalls = [
   aaApiCalls.getJurisdictions,
   aaApiCalls.getContests,
   aaApiCalls.getSettings(auditSettings.blank),
-  aaApiCalls.getSettings(auditSettings.blank),
-  aaApiCalls.getJurisdictionFile,
 ]
 
 const renderView = (route: string) => renderWithRouter(<App />, { route })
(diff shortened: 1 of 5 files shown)


MorganLove:
@jonahkagan This is only based on #749 to avoid hassles with the types. This can be merged to main after the other is merged.

client/src/components/HomeScreen.test.tsx
```diff
@@ -117,8 +115,18 @@ describe('Home screen', () => {
         isMultiJurisdiction: true,
       }),
       ...setupScreenCalls,
+      aaApiCalls.getJurisdictionFile,
+      aaApiCalls.getRounds,
+      ...setupScreenCalls,
+      aaApiCalls.getSettings(auditSettings.blank),
+      aaApiCalls.getJurisdictionFile,
```
jonahkagan, line 122: any way this chunk can go inside the definition `setupScreenCalls`? It looks like it's repeated three times in this PR

jonahkagan (approved):

client/src/components/HomeScreen.test.tsx
MorganLove: @jonahkagan I was waffling on that. Here's why:

```
      ...setupScreenCalls,
      aaApiCalls.getSettings(auditSettings.blank),
      aaApiCalls.getJurisdictionFile,
```
You see this pattern about as much as you see this pattern:

```
      ...setupScreenCalls,
      aaApiCalls.getJurisdictionFile,
```

You also may or may not have the `getRounds` bit after, too. So basically, if I'm going to include even one more line, whether it's `getSettings` or `getJurisdictionFile`, I'd need to create two separate versions of `setupScreenCalls` or have some other complexity to include or exclude the settings endpoint. This looked to me like the simplest solution while still avoiding as much duplication as possible.

jonahkagan: cool, seems like this is fine then. hopefully we can reduce the number of API calls as a better solution :)

PR #771, opened 2020-09-28
Add ballot comparision audit type

jonahkagan:
Task: #762 

Add a new type to the db enum: AuditType.BALLOT_COMPARISON, plus a
migration.


server/migrations/versions/3edc260ab0b1_audittype_batch_comparison.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def upgrade():
+    # Alembic doesn't support changing enums, so we have to do with with raw
+    # SQL. You can't alter an enum inside a transaction block (which this
+    # function is wrapped in by Alembic), so first we have to commit the
+    # transaction.
+    # https://github.com/sqlalchemy/alembic/issues/123#issuecomment-441882887
+    op.execute("COMMIT")
+    op.execute("ALTER TYPE audittype ADD VALUE 'BALLOT_COMPARISON'")
```
benadida, line 26: Yeah, this is one reason I haven't always enjoyed enums, but alright.

Is it worth doing an `op.execute("BEGIN TRANSACTION")` afterwards so the rest is under a new transaction?

benadida (approved):

server/migrations/versions/3edc260ab0b1_audittype_batch_comparison.py
jonahkagan: what do you mean "the rest"?

benadida: maybe I misunderstood, is it one transaction around each upgrade file, or around all of them at once? I meant the rest of the upgrade files. If it's one transaction per upgrade file, then ignore my comment.

jonahkagan: ah i figured out what you mean. i didn't realize the entire set of migrations was wrapped in one transaction. given that, I agree that this is not a great approach. found a better one: bb48e0d

benadida: much nicer!

benadida (approved):

PR #774, opened 2020-09-30
Ballot comparison sample size

jonahkagan:
- Parse the contest metadata out of Jurisdiction.cvr_contests_metadata
where we stashed it
- Use that contest metadata to call supersimple.get_sample_sizes to get
the estimated sample sizes for ballot comparison audits

Still todo in a future PR: come back and compute the sample results for
previous rounds so we can draw samples for later rounds.

### DIFF
--- a/server/api/cvrs.py
+++ b/server/api/cvrs.py
@@ -41,7 +41,7 @@ def process_cvr_file(session: Session, jurisdiction: Jurisdiction, file: File):
         for contest_header in contest_headers:
             match = re.match(r"^(.+) \(Vote For=(\d+)\)$", contest_header)
             contest_names.append(match[1])
-            contest_votes_allowed.append(match[2])
+            contest_votes_allowed.append(int(match[2]))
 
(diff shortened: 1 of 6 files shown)


server/api/cvrs.py
```diff
@@ -53,7 +53,7 @@ def process():
         for column, (contest_name, votes_allowed, contest_choice) in enumerate(
             interpretation_headers
         ):
-            contests_metadata[contest_name]["votes_allowed"] = votes_allowed
+            contests_metadata[contest_name]["votes_allowed"] = int(votes_allowed)
```
benadida, line 56: why is it not an integer to start with?

benadida (approved):

server/api/cvrs.py
jonahkagan: well it gets parsed on L44 above - I guess I could move the `int` up there, if that's what you're suggesting

jonahkagan: moved it

Issue #760, opened 2020-09-28
BE - Send audited votes for selected ballot with CVR votes for the same ballot


mcchilders:
- Find the appropriate CVR records and audited votes and send those along with matching IDs to ComputeRisk and GetSampleSizes endpoints
(File is called SuperSimple.py)

PR #775, opened 2020-10-01
Ballot comparison calculate risk

jonahkagan:
Task: #760 

Calculate the risk measurement for ballot comparison audits using `supersimple.calculate_risk`


server/api/cvrs.py
```diff
@@ -20,6 +21,37 @@
...
+                    id=str(uuid.uuid4()),
+                    contest_id=contest.id,
+                    name=choice_name,
+                    num_votes=0,
+                )
+                for choice_name in contest_metadata["choices"]
+            ]
+
+        contest.total_ballots_cast += contest_metadata["total_ballots_cast"]
+        contest.votes_allowed = contest_metadata["votes_allowed"]
```
benadida, line 51: does this get overwritten as the loop goes over the various jurisdictions?

server/api/rounds.py
```diff
@@ -196,28 +198,141 @@ def cumulative_batch_results(election: Election) -> BatchTallies:
...
+            # pick out the interpretation for each contest choice. We saved the
+            # column index for each choice when we parsed the CVR.
+            interpretations = interpretations_str.split(",")
+            for choice_name, choice_metadata in choices_metadata.items():
+                interpretation = interpretations[choice_metadata["column"]]
+                # If the interpretations are empty, it means the contest wasn't
+                # on the ballot, so we should skip this contest entirely for
+                # this ballot.
+                if interpretation == "":
+                    cvrs[ballot_key] = {}
```
benadida, line 249: do you mean to zero the whole CVR, not just the value for `contest.id` ?

benadida (approved):

server/api/cvrs.py
jonahkagan: yeah, the other option i could think of was just to take the first jurisdiction's value and trust that - didn't seem any different from taking the last jurisdiction's value (by overwriting it each iteration). thoughts?

server/api/rounds.py
jonahkagan: this is what the audit math expects /shrug

jonahkagan: ohhh wait i see what you're saying

jonahkagan: oh no wait it does work because there's only ever one contest (because the function - and the audit math) operates on a contest by contest basis

PR #776, opened 2020-10-02
Ballot comparison frontend

### LINKED ISSUE #755: FE - Require CVR file uploads from JAs
Need to require jurisdictions upload a second file (in addition to the Ballot Manifest) for Ballot Comparison-type audits, much like we did the Batch Totals file for Batch Comparison audits. This time the file is for Cast Vote Records, or CVRs:

### LINKED ISSUE #756: FE - Add Ballot Comparison audit type option to Create New Audit page

### LINKED ISSUE #763: FE - AA uploads Standardized Contests file
- Sidebar label needs to change
- Page header label needs to change

### LINKED ISSUE #764: FE - Add contest selector from standardized contest names to Target Contests page

### LINKED ISSUE #765: FE - Add contest selector from standardized contest list to Opportunistic Contests List

MorganLove:
**Description**

- All the things.

Closes #756 
Closes #763 
Closes #764 
Closes #765 
Closes #755 

**Testing**

- Snapshots updated as needed

**Progress**

WIP


benadida (approved):
A couple of small questions inline.

Overall this looks like a good direction, though  given my reduced context, you may want to wait until Jonah returns. That said, if you're trying to move quickly, I'm not seeing a major issue, so approving.

client/src/components/DataEntry/MemberForm.tsx
```diff
@@ -50,7 +50,7 @@ const MemberForm: React.FC<IProps> = ({
           values,
           handleSubmit,
         }: FormikProps<[IAuditBoardMember, IAuditBoardMember]>) => (
-          <Form>
+          <form>
```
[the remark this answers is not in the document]
MorganLove: @benadida There's a glitch with types in Formik and React that makes `Form` fail unless you add `translate="true"` to it, which is... kind of a ridiculous solution imo. So since we aren't actually using the magic that `Form` provides, I opted to just swap it out for the regular html version and avoid the issue.

client/src/components/MultiJurisdictionAudit/AASetup/Contests/ContestForm.tsx
```diff
@@ -0,0 +1,304 @@
...
+                          } Info`}
+                          description="Enter the name of the contest that will drive the audit."
+                        >
+                          <br />
+                          <label htmlFor={`contests[${i}].name`}>
+                            Contest {values.contests.length > 1 ? i + 1 : ''}{' '}
+                            Name
+                            <Field
+                              id={`contests[${i}].name`}
+                              name={`contests[${i}].name`}
```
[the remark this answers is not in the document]
MorganLove: @benadida No, this is how Formik handles objects and arrays of form data. Every field has a unique name attached to it which is a string constructed from its location in the state, like this.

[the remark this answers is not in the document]
benadida: I'm still confused, cause you've got `values.contests.map((contest, ..)` above, which makes me think `contests[${i}]` is the same as `contest`.. but that's not the case? If so, that's confusing, maybe different variable names would help?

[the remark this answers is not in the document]
MorganLove: You are correct that the value of `contest` and `contests[i]` is the same. The string ``contests[${i}].name`` though is what is required for Formik to know which form element goes with which piece of its internal state. The strings passed to `id` and `name` are basically used as arguments for a getter function inside of Formik. If you passed in the string `'contest.name'` it would look on the root state object for a key called `contest` and things would get very confused, haha.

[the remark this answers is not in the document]
benadida: oh! Duh. These are strings. My bad. I get it now.

client/src/components/MultiJurisdictionAudit/AASetup/Participants/index.tsx
```diff
@@ -28,55 +30,70 @@ interface IProps {
   locked: boolean
 }
 
+type IFileSubmitStatus = 'submit' | 'success' | 'failure' | null
```
jonahkagan, line 33: I can see why the changes to this form were a tough challenge, these file uploads have so many states and trying to upload two from one form multiplies the states

One idea I had that might make it simpler: make the `FileForm` component controlled by the `Participants` component

Here's a rough sketch:

```ts
const Participants = (props) => {
  const { electionId } = useParams<{ electionId: string }>()
  const [jurisdictionFile, uploadJurisdictionFile] = useJurisdictionFile(electionId)
  const [standardizedContestsFile, uploadStandardizedContestsFile] = useStandardizedContestsFile(electionId)

  const submit = (values) => {
    // error handling left out for brevity
    await updateSettings({ state: values.state })
    await updateJurisdictionFile(values.jurisdictionFile)
    await updateStandardizedContestsFile(values.standardizedContestsFile)
    // if all the requests succeed, move to the next stage
    if (nextStage.activate) nextStage.activate()
  }

  return (
    <Formik
      initialValues={{
        state: auditSettings.state || '',
        jurisdictionFile: null,
        standardizedContestsFile: null,
      }}
    >
      ... form section for State ...
      <FileForm
        file={jurisdictionFile}
        onFileChange={(file) => setFieldValue("jurisdictionFile", file)}
        ... other props ...
       />
      <FileForm
        file={standardizedContestsFile}
        onFileChange={(file) => setFieldValue("standardizedContestsFile", file)}
        ... other props ...
       />
      // potentially only enable the submit button if both files have been selected
      ... submit button ...
    </Formik>
}
```

client/src/components/MultiJurisdictionAudit/AASetup/Participants/useContestFile.ts
```diff
@@ -0,0 +1,54 @@
...
+    `/election/${electionId}/standardized-contests/file`,
+    {
+      method: 'PUT',
+      body: formData,
+    }
+  )
+  return !!response
+}
+
+const useContestFile = (
```
jonahkagan, line 31: recommend `useStandardizedContestsFile` for consistency with backend naming

client/src/components/MultiJurisdictionAudit/AASetup/Participants/useContestFile.ts
```diff
@@ -0,0 +1,54 @@
...
+  electionId: string
+): Promise<IFileInfo | null> => {
+  const response = await api<IFileInfo>(
+    `/election/${electionId}/standardized-contests/file`
+  )
+  if (!response) return null
+  return response
+}
+
+const putContestFileFile = async (
```
jonahkagan, line 15: double `FileFile` here haha

client/src/components/MultiJurisdictionAudit/CSVForm/index.tsx
```diff
@@ -109,6 +111,17 @@ const CSVFileForm = ({
...
+            {filePurpose === 'cvrs' && (
+              <FormSection>
+                <H4>Cast Vote Records</H4>
+                <FormSectionDescription>
+                  Click &quot;Browse&quot; to choose the appropriate Cast Vote
+                  Records (CVR) file from your computer. This file should be a
+                  comma-separated list of all the ballots counted by your
+                  tabulator, in order.
+                </FormSectionDescription>
+              </FormSection>
```
jonahkagan, line 123: maybe just pull these out into a `title` and `description` prop to replace the `filePurpose` prop?

client/src/components/MultiJurisdictionAudit/index.tsx
```diff
@@ -58,9 +58,22 @@ export const AuditAdminView: React.FC = () => {
...
+        return !isBallotComparison
+      case 'Participants & Contests':
+        return !!isBallotComparison
+      default:
+        return true
+    }
+  })
+
+  if (isBallotComparison && stage === 'Participants')
+    setStage('Participants & Contests')
```
jonahkagan, line 76: could you solve this issue by adding another field `id` to the `ISidebarMenuItem` type? that way you could change `title` but keep `id` the same. e.g. for the participants stage, the `id` could be `Participants` but the title could be `Participants & Contests` (when it's a ballot comparison audit)

client/src/components/MultiJurisdictionAudit/AASetup/Participants/index.tsx
MorganLove: @jonahkagan That's actually where I went with it initially, but here's the issue I ran into:

We're using the same component for ballot comparison audits and non-ballot-comparison audits, meaning that the standardized contests section may or may not be present at all. This means that the `useStandardizedContests` hook throws a wrench into things, since you cannot ever use a hook conditionally. We can't just say `if (isBallotComparison) useStandardizedContests` for example. We have to conditionally render the standardized contests upload form, and have the hooks related to it managed by that sub component.

So basically, either we duplicate the entire component and have one version include standardized contests uploading and the other not, or we refactor in some fashion similar to how I've got it. The resulting status tracking is... awkward, to say the least, but it's the best I could come up with in order to integrate with the triple form submissions. If you have another idea I'm open to it, though, particularly if there's a bug I haven't found with this setup, haha.

jonahkagan: Didn't you come up with a solution where we pass a flag into the hook to tell it whether to make the request or not? We could potentially do that with `useStandardizedContests`

MorganLove: I was trying to avoid that, but it might actually be the simplest solution here. I'll see what the refactor looks like.

PR #777, opened 2020-10-07
Exposes discrepancies from math code. 

### LINKED ISSUE #761: BE - Expose any found discrepancies for audit report
Expose the found discrepancies (which ballot, expected vote from CVR & actual vote from auditors)

umbernhard:
**Description**
Exposes discrepancies between the sampled ballots and the CVRs via the `supersimple.find_discrepancies` function. Also refactors a bit to take advantage of the new function call. Closes #761.

**Testing**

Adds two new test cases to verify the new discrepancy finding is correct and doesn't monkey with the p-value calculations.

**Progress**

Ready.

### DIFF
--- a/server/audit_math/supersimple.py
+++ b/server/audit_math/supersimple.py
@@ -1,6 +1,6 @@
 # pylint: disable=invalid-name
 import math
-from typing import Dict, Tuple, Union
+from typing import Dict, Tuple, Union, TypedDict
 
 from .sampler_contest import Contest
 
@@ -14,6 +14,17 @@ u1: float = 0.001
 # This sets the expected two-vote misstatements at 1 in 10000
 o2: float = 0.0001
 u2: float = 0.0001
+
+# { ballot_id: { contest_id: { choice_id: 0 | 1 }}}
+CVRS = Dict[str, Dict[str, Dict[str, int]]]
+CVR = Dict[str, Dict[str, int]]
+
+
+class Discrepancy(TypedDict):
+    counted_as: int
+    weighted_error: float
+    cvr: CVR
+    sample_cvr: CVR
 
 
 def nMin(
@@ -38,6 +49,82 @@ def nMin(
             / contest.diluted_margin
         ),
     )
+
+
+def find_discrepancies(
+    contest: Contest, cvrs: CVRS, sample_cvr: CVRS,
+) -> Dict[str, Discrepancy]:
+    """
+    Iterates through a given sample and returns the discrepancies found.
+
+    Inputs:
+        contests       - the contests and results being audited
+        cvrs           - mapping of ballot_id to votes:
+                {
+                    'ballot_id': {
+                        'contest': {
+                            'candidate1': 1,
+                            'candidate2': 0,
+                            ...
+                        }
+                    ...
+                }
+
+        sample_cvr - the CVR of the audited ballots
+                {
+                    'ballot_id': {
+                        'contest': {
+                            'candidate1': 1,
+                            'candidate2': 0,
+                            ...
+                        }
+                    ...
+                }
+
+    Outputs:
+        discrepancies   - A mapping of ballot ids to their discrepancies. This
+                          includes entries for all ballots in the sample, even those without
+                          discrepancies.
+                {
+                    'ballot_id': {
+                        'counted_as': -1, # The maximum value for a discrepancy
+                        'weighted_error': -0.0033 # Weighted error used for p-value calculation
+                        'cvr': <CVR for this ballot>,
+                        'sample_cvr': <Sample CVR for this ballot>
+                    }
+    """
+
+    discrepancies: Dict[str, Discrepancy] = {}
+    for ballot in sample_cvr:
+        e_r = 0.0
+        e_int = 0
+
+        if contest.name not in sample_cvr[ballot]:
+            continue
+        for winner in contest.winners:
+            for loser in contest.losers:
+                v_w = cvrs[ballot][contest.name][winner]
+                a_w = sample_cvr[ballot][contest.name][winner]
+
+                v_l = cvrs[ballot][contest.name][loser]
+                a_l = sample_cvr[ballot][contest.name][loser]
+
+                V_wl = contest.candidates[winner] - contest.candidates[loser]
+
+                e = (v_w - a_w) - (v_l - a_l)
+                e_weighted = e / V_wl
+                if e_weighted > e_r:
+                    e_r = e_weighted
+                    e_int = e
+
+        discrepancies[ballot] = Discrepancy(
+            counted_as=e_int,
+            weighted_error=e_weighted,
+            cvr=cvrs[ballot],
+            sample_cvr=sample_cvr[ballot],
+        )
+
+    return discrepancies
 
 
 def get_sample_sizes(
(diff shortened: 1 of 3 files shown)


server/audit_math/supersimple.py
```diff
@@ -40,6 +51,82 @@ def nMin(
     )
 
 
+def find_discrepancies(
```
benadida, line 54: nit: `compute_discrepancies` is probably more accurate than `find_...`

benadida (approved):

server/audit_math/supersimple.py
```diff
@@ -15,6 +15,17 @@
 o2: float = 0.0001
 u2: float = 0.0001
 
+# { ballot_id: { contest_id: { choice_id: 0 | 1 }}}
+CVRS = Dict[str, Dict[str, Dict[str, int]]]
```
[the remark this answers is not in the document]
umbernhard: It's better as an integer because it lets us do math with the votes without having to cast. Also, in the future if we support RCV then it will be necessary.

server/audit_math/supersimple.py
```diff
@@ -40,6 +51,82 @@ def nMin(
...
+                        'weighted_error': -0.0033 # Weighted error used for p-value calculation
+                        'cvr': <CVR for this ballot>,
+                        'sample_cvr': <Sample CVR for this ballot>
+                    }
+    """
+
+    discrepancies: Dict[str, Discrepancy] = {}
+    for ballot in sample_cvr:
+        e_r = 0.0
+        e_int = 0
```
jonahkagan, line 100: my understanding is that `e_r` and `e_int` are used to compute maximum error values for the ballot across the different winner/loser combos. if that's the case, should their initial values be the minimum possible error value (which I believe is `-2` for `e_int`, not sure what it would be for `e_r`)?

server/audit_math/supersimple.py
```diff
@@ -15,6 +15,17 @@
...
+# { ballot_id: { contest_id: { choice_id: 0 | 1 }}}
+CVRS = Dict[str, Dict[str, Dict[str, int]]]
+CVR = Dict[str, Dict[str, int]]
+
+
+class Discrepancy(TypedDict):
+    counted_as: int
+    weighted_error: float
+    cvr: CVR
+    sample_cvr: CVR
```
jonahkagan, line 27: what are these two fields for?

server/audit_math/supersimple.py
```diff
@@ -15,6 +15,17 @@
 o2: float = 0.0001
 u2: float = 0.0001
 
+# { ballot_id: { contest_id: { choice_id: 0 | 1 }}}
+CVRS = Dict[str, Dict[str, Dict[str, int]]]
+CVR = Dict[str, Dict[str, int]]
+
+
+class Discrepancy(TypedDict):
+    counted_as: int
```
jonahkagan, line 24: maybe name this `discrepancy` or `simple_error`?

server/tests/audit_math/test_supersimple.py
```diff
@@ -60,6 +60,62 @@ def test_compute_diluted_margin(contests):
...
+        )
+
+        assert discrepancies[0]["counted_as"] == 0
+        assert discrepancies[0]["weighted_error"] == 0
+        assert discrepancies[0]["cvr"] == discrepancies[0]["sample_cvr"]
+
+
+def test_find_one_discrepancy(contests, cvrs):
+
+    # Test no discrepancies
```
jonahkagan, line 88: this comment doesn't apply. maybe copy-paste error?

server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison.py
```diff
@@ -41,8 +41,8 @@
 \r
 ######## ROUNDS ########\r
 Round Number,Contest Name,Targeted?,Sample Size,Risk Limit Met?,P-Value,Start Time,End Time,Audited Votes\r
-1,Contest 1,Targeted,11,No,1.1043717961,DATETIME,DATETIME,Choice 1-1: 4; Choice 1-2: 5\r
-1,Contest 2,Opportunistic,,No,0.4183843254,DATETIME,DATETIME,Choice 2-1: 4; Choice 2-2: 4; Choice 2-3: 0\r
+1,Contest 1,Targeted,11,No,0.7413124393,DATETIME,DATETIME,Choice 1-1: 4; Choice 1-2: 5\r
+1,Contest 2,Opportunistic,,No,0.3142300101,DATETIME,DATETIME,Choice 2-1: 4; Choice 2-2: 4; Choice 2-3: 0\r
```
jonahkagan, line 45: why did the p-values change here?

server/audit_math/supersimple.py
umbernhard: In order to remain as conservative as possible, we're choosing to ignore negative errors, so that they don't cancel out positive errors.

server/audit_math/supersimple.py
umbernhard: Yeah, I wasn't sure if I should follow the class naming convention or not.

server/audit_math/supersimple.py
umbernhard: So that there's an easy way to grab both ballots and look at what the actual discrepancy was

server/tests/audit_math/test_supersimple.py
umbernhard: Yep :sweat_smile:

server/audit_math/supersimple.py
jonahkagan: what's the class naming convention?

server/audit_math/supersimple.py
jonahkagan: Oh that's super interesting. Can you add a comment explaining that?

Also, does that mean that in the audit report we'll end up showing a discrepancy of 0 for ballots that upon examination would appear to have a negative discrepancy? Will that be confusing for people? Or is that expected behavior?

umbernhard: Will add a comment. @mcchilders what do you think about this?

server/audit_math/supersimple.py
umbernhard: If it's a class, it gets a capital first letter.

server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison.py
umbernhard: Hmmmmm that's a really good question.

server/audit_math/supersimple.py
jonahkagan: oh yeah def follow that. i was referring to changing the name of `counted_as`

server/audit_math/supersimple.py
mcchilders: We "ignore" them only for the purposes of calculating the p-value (we still record them all as discrepancies, and list them out that way in the audit report, etc. Each discrepancy has its own row - it's not a combined discrepancy count per ballot, where the combination might hide individual errors, it's literally a list of each discrepancy and which ballot it appeared on.

umbernhard: Hmm, that means either I need to be doing more work in the math code, or @jonahkagan will need to check for the discrepancies when writing the audit report. Which of those sounds better to you, Jonah?

server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison.py
umbernhard: Looks like I forgot to add testing of the p-value calculations. I will work on that and continue to chase this down.

umbernhard: Okay, this issue should be resolved. There was a bug in the p-value calculations that is now fixed, so the snapshot will still change.

jonahkagan:
Can you `git rebase -i` and remove my accidental PR commit (029bef7) from this PR?

umbernhard:
Ah! Yes!

umbernhard:
@jonahkagan lemme know when you've had a chance to look this over again. I think it's good to go.

jonahkagan (approved):

PR #784, opened 2020-10-12
Fixes precision issues that cause bugs like #778

### LINKED ISSUE #778: OverflowError: Numerical result out of range in BRAVO given third candidate with low total
When I plugged in the results from just the top 2 candidates in [Colorado's 2016 election](https://en.wikipedia.org/wiki/2016_United_States_presidential_election_in_Colorado)
into a new ballot-polling audit on the latest version in main (d6181d3a), along with a third candidate reported to have just 10 votes, I got this result

```
127.0.0.1 - - [07/Oct/2020 12:43:03] "PUT /api/election/1badc9a8-a59b-40c0-ab01-080e9c081837/jurisdiction/c1c13994-54b4-40d6-a064-290c192fe4c9/round/0272f4ad-56bd-43f1-83d8-48a352db68ae/results HTTP/1.1" 500 -

  OverflowError: (34, 'Numerical result out of range')
    Traceback (most recent call last):
      File ".../site-packages/flask/app.py", line 2464, in __call__
	return self.wsgi_app(environ, start_response)
      File ".../site-packages/werkzeug/middleware/proxy_fix.py", line 169, in __call__
	return self.app(environ, start_response)
      File ".../site-packages/flask/app.py", line 2450, in wsgi_app
	response = self.handle_exception(e)
      File ".../site-packages/flask/app.py", line 1867, in handle_exception
	reraise(exc_type, exc_value, tb)
      File ".../site-packages/flask/_compat.py", line 39, in reraise
	raise value
      File ".../site-packages/flask/app.py", line 2447, in wsgi_app
	response = self.full_dispatch_request()
      File ".../site-packages/flask/app.py", line 1952, in full_dispatch_request
	rv = self.handle_user_exception(e)
      File ".../site-packages/flask/app.py", line 1821, in handle_user_exception
	reraise(exc_type, exc_value, tb)
      File ".../site-packages/flask/_compat.py", line 39, in reraise
	raise value
      File ".../site-packages/flask/app.py", line 1950, in full_dispatch_request
	rv = self.dispatch_request()
      File ".../site-packages/flask/app.py", line 1936, in dispatch_request
	return self.view_functions[rule.endpoint](**req.view_args)
      File ".../arlo/server/auth/lib.py", line 176, in wrapper
	return route(*args, **kwargs)
      File ".../arlo/server/api/offline_results.py", line 113, in record_offline_results
	end_round(election, round)
      File ".../arlo/server/api/rounds.py", line 343, in end_round
	calculate_risk_measurements(election, round)
      File ".../arlo/server/api/rounds.py", line 315, in calculate_risk_measurements
	p_values, is_complete = bravo.compute_risk(
      File ".../arlo/server/audit_math/bravo.py", line 415, in compute_risk
	T = get_test_statistics(contest.margins, sample_results)
      File ".../arlo/server/audit_math/bravo.py", line 109, in get_test_statistics
	T[(cand, loser)] *= (winners[cand]["swl"][loser] / 0.5) ** votes
    OverflowError: (34, 'Numerical result out of range')

```

That overflow is easy to reproduce with these even more common parameters: 10% margin, plus a third candidate with a 1% vote share. This sort of contest requires thousands of samples, and along the way produces a vanishing risk level (i.e. a huge test statistic) for the third candidate.

```
over = Contest("Overflow", {"ballots": 20000, "numWinners": 1,"votesAllowed": 1, "a": 10000, "b": 9000, "c": 200})
margins = over.margins
margins
{'winners': {'a': {'p_w': 0.5,
   's_w': 0.5208333333333334,
   'swl': {'b': 0.5263157894736842, 'c': 0.9803921568627452}}},
 'losers': {'b': {'p_l': 0.45, 's_l': 0.46875},
  'c': {'p_l': 0.01, 's_l': 0.010416666666666666}}}
0, 0, 
bravo_sample_sizes(0.1, margins["winners"]["a"]["p_w"], margins["losers"]["b"]["p_l"], 0, 0, 0.9)
5426
get_test_statistics(over.margins, {"a": 1200, "b": 1000, "c": 10})
---------------------------------------------------------------------------
OverflowError                             Traceback (most recent call last)
<ipython-input-107-ee79987286e4> in <module>
----> 1 get_test_statistics(over.margins, {"a": 1200, "b": 1000, "c": 10})

/srv/s/electionaudits/arlo/arlo/server/audit_math/bravo.py in get_test_statistics(margins, sample_results)
    107         if cand in winners:
    108             for loser in losers:
--> 109                 T[(cand, loser)] *= (winners[cand]["swl"][loser] / 0.5) ** votes
    110         elif cand in losers:
    111             for winner in winners:

OverflowError: (34, 'Numerical result out of range')
```

If we only draw a few hundred ballots, we see how high the test statistic for ('a', 'c') is getting already:

```
get_test_statistics(over.margins, {"a": 120, "b": 100, "c": 1})
{('a', 'b'): 2.1137702461183197, ('a', 'c'): 4.8421550053990614e+33}
```

umbernhard:
**Description**

* Changes from `float` to `Decimal` in relevant places
* Retypes inputs where applicable
* Updates code that depends on the math modules accordingly
* Updates tests where rounding errors caused wrong results or overflows.
* Fixes #778

**Testing**

Adds new BRAVO test case to ensure we don't overflow. Also modifies places where rounding errors caused problems. 

**Progress**

Ready to go.


server/api/rounds.py
```diff
@@ -334,7 +335,7 @@ def calculate_risk_measurements(election: Election, round: Round):
                 sampled_ballot_interpretations_to_cvrs(contest),
             )
 
-        round_contest.end_p_value = p_value
+        round_contest.end_p_value = float(p_value)
```
jonahkagan, line 338: Could the `risk_limit` passed to `compute_risk` remain a `float` and get computed into a `Decimal` internally? It doesn't seem like the abstraction needs to change

Similarly, could the resulting p-value be a `float` as well, since it just gets converted immediately after being returned from `compute_risk`?

(Also, you can extend this question to all functions that are part of the external interface presented by the audit math)

server/audit_math/bravo.py
```diff
@@ -185,21 +198,21 @@ def bravo_sample_sizes(
     # max here handles cases where, due to rounding error,
     # the base (content) of the radical is trivially
     # negative for p_completion very close to 0.5.
-    radical = math.sqrt(max(0, q_b ** 2 - 4 * q_a * q_c))
+    radical = math.sqrt((q_b ** 2 - 4 * q_a * q_c).max(0))
```
jonahkagan, line 201: why change `max` but not `math.sqrt`? in the decimal docs it seems like they use the builtin `max` but use the decimal's `sqrt` method (see examples a few paragraphs down in https://docs.python.org/3/library/decimal.html#quick-start-tutorial)

jonahkagan: in general i'm confused about when something needs to be a decimal and when it doesn't

server/tests/audit_math/test_bravo.py
```diff
@@ -305,16 +328,19 @@ def test_compute_risk(contests):
...
         "test12": True,
+        "test_small_third_candidate": True,
     }
 
     for contest in contests.values():
         sample = round1_sample_results[contest.name]
         T, decision = bravo.compute_risk(RISK_LIMIT, contest, sample)
         expected_T = expected_Ts[contest.name]
+
+        print(T)
```
jonahkagan, line 339: remove

server/api/rounds.py
umbernhard: I changed `risk_limit` because it was super cumbersome to have to cast it in every function. This way we're just casting it once. 

`p_value` needs to be computed as a Decimal to avoid losing precision. I could return it as a float, that's fair.

server/audit_math/bravo.py
umbernhard: Ah, in this particular instance I suppose it doesn't matter. Generally, things need to be Decimal when an operation is being performed with another Decimal number, and/or we don't want to lose precision.

Here, `radical` needs to be a float because it is passed into some stats code that expects a float. I could cast it later on, but I found I was casting it a bunch of times later on.

PR #785, opened 2020-10-13
Pass discrepancies into sample size calculator

jonahkagan:
In ballot comparison audits, for rounds after the first round, the
sampler needs to know the results from the previous round in terms of
the discrepancies found.

### DIFF
--- a/server/api/sample_sizes.py
+++ b/server/api/sample_sizes.py
@@ -67,6 +67,11 @@ def sample_size_options(
             set_contest_metadata_from_cvrs(contest)
             contest_for_sampler = sampler_contest.from_db_contest(contest)
 
+            num_previous_samples = (
+                SampledBallotDraw.query.join(Round)
+                .filter_by(election_id=election.id)
+                .count()
+            )
             discrepancies = supersimple.compute_discrepancies(
                 contest_for_sampler,
                 rounds.cvrs_for_contest(contest),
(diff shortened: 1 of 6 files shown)


server/api/sample_sizes.py
```diff
@@ -64,18 +65,26 @@ def sample_sizes_for_contest(contest: Contest):
...
+            discrepancies = supersimple.compute_discrepancies(
+                contest_for_sampler,
+                rounds.cvrs_for_contest(contest),
+                rounds.sampled_ballot_interpretations_to_cvrs(contest),
+            )
+            discrepancy_groups = group_by(
+                discrepancies.values(), lambda d: d["counted_as"]
+            )
+            discrepancy_counts = {
+                "sample_size": len(discrepancies),  # TODO is this the right value?
```
umbernhard, line 79: This isn’t the right value. Can you get the sample size from earlier in this function?

umbernhard: Or I could reconfigure the discrepancy function to work this way, whichever is easier.

jonahkagan: what is the right number? is it the total number of samples drawn in the audit thus far? if so, probably need to query the db for it - not a big deal

umbernhard: Yeah, the total number of samples drawn so far

jonahkagan: ok how does it look now? c2d0026

umbernhard (approved):

server/api/sample_sizes.py
umbernhard: Perfect :)

PR #789, opened 2020-10-15
Tabulator/Container in manifest for ballot comparison audits

jonahkagan:
Product changes:
- For ballot comparison audits, batches are not uniquely identified by Batch Name alone, but rather by Tabulator+Batch Name (this is how they are recorded in the CVR). So we require the Tabulator column in the ballot manifest and use Tabulator+Batch Name as the unique identifier for batches throughout the audit.
- For ballot comparison audits, multiple batches of ballots might be stored in one physical container. We allow an optional Container column in the ballot manifest. If Container is provided, when divvying up the sampled ballots between audit boards, we ensure that all ballots in the same Container are given to the same audit board.
- These changes may both be useful in other types of audits as well, so we make this functionality available (though not required) for other audit types.

Implementation notes:
- Removed the Batch.storage_location field (unused) and replaced it with Batch.container
- If Tabulator or Container is specified, we require that it must be specified for every batch (all rows in the manifest)
- Changed the batch key for sampling ballots to `(batch.tabulator, batch.name)`. If `batch.tabulator` is `None`, then this is logically equivalent to using `batch.name` as the batch key, so it should work both when tabulator is or isn't given. However, this does change the output of the sampler for existing test cases, since the sampler hashes the given batch keys, so we do expect many of the test snapshots to change, but only in number of samples, not in other logical results.
- Added Tabulator to the audit report in order to make sure each ballot row was uniquely identified.


server/api/rounds.py
```diff
@@ -256,13 +256,13 @@ def cvrs_for_contest(contest: Contest) -> supersimple.CVRS:
 
 def sampled_ballot_interpretations_to_cvrs(contest: Contest) -> supersimple.CVRS:
     def interpretation_to_cvr_value(interpretation: Interpretation) -> int:
-        if interpretation == Interpretation.BLANK:
-            return 0
-        elif interpretation == Interpretation.VOTE:
+        if interpretation == Interpretation.VOTE:
             return 1
+        elif interpretation == Interpretation.BLANK:
+            return 0  # TODO make this a vote for the loser?
```
umbernhard, line 262: No. this is right. We're saying that this wasn't a vote, therefore the most it can contribute is a one-vote misstatement. If we count it as a vote for the loser, then it potentially becomes a two-vote misstatement....

server/api/rounds.py
```diff
@@ -256,13 +256,13 @@ def cvrs_for_contest(contest: Contest) -> supersimple.CVRS:
...
         else:
             assert interpretation == Interpretation.CANT_AGREE
-            return 0  # TODO what to do here?
+            return 0  # TODO make this a vote for the loser?
```
umbernhard, line 265: This is a more interesting case. @mcchilders how has this been handled in the past?

server/tests/ballot_comparison/test_ballot_comparison_manifests.py
```diff
@@ -0,0 +1,207 @@
...
+import pytest
+
+from ...models import *  # pylint: disable=wildcard-import
+from ..helpers import *  # pylint: disable=wildcard-import
+from ...bgcompute import bgcompute_update_ballot_manifest_file
+
+
+# In one jurisdiction, add the Container column to the manifest. In this
+# jurisdiction, ballots should be divvied up between audit boards by container,
+# not the usual tabulator+batch name.
```
umbernhard, line 14: I assume the tests that already exist test the case where there isn't a container/batch difference?

umbernhard (approved):
Looks good. There are a few questions, but we can always address them in subsequent PRs.

server/tests/ballot_comparison/test_ballot_comparison_manifests.py
jonahkagan: there are existing tests for ballot manifests with no container or tabulator column. this file tests what happens when you include container and tabulator.

umbernhard: dope.

server/api/audit_boards.py
```diff
@@ -59,25 +59,30 @@ def assign_sampled_ballots(
         .filter_by(jurisdiction_id=jurisdiction.id)
         .join(SampledBallot.draws)
         .filter_by(round_id=round.id)
-        .order_by(Batch.name)  # group_by prefers a sorted list
         .options(contains_eager(SampledBallot.batch))
```
benadida, line 62: should this be ordered by a sampled ballot ID or something else, just to make sure there is a consistent ordering whenever you query this more than once?

benadida (approved):

server/api/audit_boards.py
jonahkagan: hmm yeah ill figure something out

jonahkagan: actually the order doesn't matter because the next thing that happens is it gets run through `group_by`

PR #791, opened 2020-10-16
Ballot comparison audit report

jonahkagan:
Add Container, Imprinted ID, and Discrepancy columns to audit report


umbernhard (approved):

benadida (approved):

jonahkagan:
@umbernhard I found a bug in the way I was computing the sample CVRs to pass into the audit math. I was skipping over some ballots that were audited as blank. I only found it because I was spot checking the discrepancies in the audit report. The fix is in server/api/rounds.py. Could you take a look and also maybe check that the discrepancies look correct now?

server/api/reports.py
```diff
@@ -72,28 +74,57 @@ def pretty_batch_ticket_numbers(batch: Batch, round_id_to_num: Dict[str, int]) -
...
+def pretty_discrepancy(
+    ballot: SampledBallot,
+    contest: Contest,
+    contest_discrepancies: Dict[str, supersimple.Discrepancy],
+) -> str:
+    ballot_key = (
+        list(ballot.draws)[0].ticket_number if contest.is_targeted else ballot.id
+    )
+    if ballot_key in contest_discrepancies:
+        return str(contest_discrepancies[ballot_key]["counted_as"])
```
umbernhard, line 125: Just to be clear, this is just converting an integer to a string. Is that the intended behavior?

jonahkagan: yeah cuz we are printing it in the audit report!

umbernhard: I'm just trying to clarify, this will literally just be like "2" in the discrepancies field. Do we want something a little more information-dense than that? @mcchilders?

jonahkagan: ah good question. i think in the context of the report it's all the info you need, because there are other columns with the CVR/audit result. check out one of the snapshot reports. but def open to feedback

Issue #758, opened 2020-09-28
FE - Include CVRs in Jurisdiction File Upload Status portions of AA dashboard


mcchilders:
- This is the count of jurisidictions that have uploaded files that shows on the confirmation modal before starting the audit, the grey status bar indicator at the top of the screen, and the Audit Progress screen.

PR #798, opened 2020-10-20
CVR file in status box

jonahkagan:
Task: #758 

Status box
- Includes the CVR file when computing the status messages
- Also includes the batch tallies file (which we should have included before)

Review screen
- Add checks for all CVR/batch tallies files being uploaded to isSetupComplete
- Use isSetupComplete to decide when we load sample sizes as well as when we enable the launch button
- In the confirmation popup, only show a warning about how many jurisdictions uploaded ballot manifests for ballot polling audits (since in other audit types, all jurisdictions will have to have uploaded their manifests to get to this point)
- Change isSetupComplete and that warning to only factor in participating jurisdictions (i.e. ones that are part of a selected contest universe)

Misc
- Fix everywhere that we checked for `BALLOT_POLLING` audit type to check for `BATCH_COMPARISON` - that way we can treat `BALLOT_POLLING` and `BALLOT_COMPARISON` the same in all those cases, which is often what we want to do

Waiting to fix the failing tests until upstream branches fix their failing tests

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
+++ b/client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
@@ -47,35 +47,21 @@ const Review: React.FC<IProps> = ({ prevStage, locked, refresh }: IProps) => {
   const history = useHistory()
   const [isConfirmDialogOpen, setIsConfirmDialogOpen] = useState(false)
 
-  const talliesUploadsCompleted =
-    !!jurisdictions.length &&
+  const shouldShowSampleSizes =
     !!contests &&
-    jurisdictions.every(
-      j =>
-        contests.every(contest => !contest.jurisdictionIds.includes(j.id)) || // don't worry about this jurisdiction if it's not in the contest universe
-        (j.batchTallies &&
-          j.batchTallies.processing &&
-          j.batchTallies.processing.status === FileProcessingStatus.PROCESSED)
-    )
-  const cvrsUploadsCompleted =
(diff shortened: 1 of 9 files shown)


client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
```diff
@@ -159,9 +149,13 @@ const Review: React.FC<IProps> = ({ prevStage, locked, refresh }: IProps) => {
...
-                {auditType === 'BALLOT_POLLING'
-                  ? 'Ballot Polling'
-                  : 'Batch Comparison'}
+                {
+                  {
+                    BALLOT_POLLING: 'Ballot Polling',
+                    BATCH_COMPARISON: 'Batch Comparison',
+                    BALLOT_COMPARISON: 'Ballot Comparison',
+                  }[auditType]
+                }
```
MorganLove, line 158: @jonahkagan Really cool way of doing this!!

MorganLove (approved):
Looks good to me!

client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
jonahkagan: yeah i like this pattern cuz it makes every case explicit (as opposed to a series of ternary expressions, where the last case is implicit). still wishing that any of the languages we used had a multi-arm conditional (i.e. a `switch` that's an expression, not a statement)

PR #799, opened 2020-10-21
Set Election.online based on audit type

jonahkagan:
Ballot polling audits are the only ones that have the configurable
online setting. This change sets online to the correct value for ballot
comparison and batch comparison audits on audit creation. It removes
setting of the online value from the frontend, except in the case of
ballot polling audits.

### DIFF
--- a/server/models.py
+++ b/server/models.py
@@ -44,7 +44,9 @@ class Election(BaseModel):
     risk_limit = Column(Integer)
     random_seed = Column(String(100))
 
-    # an election is "online" if every ballot is entered online, vs. offline in a tally sheet.
+    # An audit is "online" if each ballot's audit results are entered in Arlo
+    # individually, vs. written in a tally sheet and then totaled before
+    # submitting to Arlo.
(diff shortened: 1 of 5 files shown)


server/models.py
```diff
@@ -45,7 +45,7 @@ class Election(BaseModel):
     random_seed = Column(String(100))
 
     # an election is "online" if every ballot is entered online, vs. offline in a tally sheet.
-    online = Column(Boolean, nullable=False, default=False)
```
benadida, line 48: shouldn't this come with an alembic migration?

benadida (approved):
one question, otherwise good.

server/models.py
jonahkagan: i thought so, but apparently the default value is handled by SQLAlchemy, not postgres, so there's nothing to migrate

jonahkagan: there's another option `server_default` that does it at the db level

benadida: ooooh, bad SQLAlchemy, bad. Anyways, that's fine then!

PR #813, opened 2020-10-21
Data entry tweaks

MorganLove:
**Description**

- moving imprinted ID to last place

**Testing**

- none yet

**Progress**

- moved the imprinted ID, adjusted order of columns

### DIFF
--- a/client/src/components/DataEntry/BoardTable.tsx
+++ b/client/src/components/DataEntry/BoardTable.tsx
@@ -55,11 +55,12 @@ interface IProps {
   url: string
 }
 
-const KEYS: ('position' | 'tabulator' | 'batch' | 'status')[] = [
+const KEYS: ('position' | 'tabulator' | 'batch' | 'status' | 'container')[] = [
   'batch',
   'position',
   'tabulator',
   'status',
+  'container',
 ]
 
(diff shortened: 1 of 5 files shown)


jonahkagan (approved):

client/src/components/DataEntry/BoardTable.tsx
```diff
@@ -171,13 +178,14 @@ const BoardTable: React.FC<IProps> = ({ boardName, ballots, url }: IProps) => {
         columnWidths={cols}
         enableRowHeader={false}
       >
+        <Column key="container" name="Container" cellRenderer={renderCell} />
+        <Column key="tabulator" name="Tabulator" cellRenderer={renderCell} />
```
jonahkagan, line 182: should we make these two columns conditional? i.e. only show them when there is data for them? (if we have data in one of these fields for one ballot, we are guaranteed to have it in all ballots)

MorganLove: @jonahkagan Oh yeah, I was going to ask specifically that but forgot, haha. If you have it in one field it's in all fields that would be relatively easy I think, yeah.

Issue #805, opened 2020-10-21
Validation in audit math


jonahkagan:
The audit math has various expectations about the input data it receives. We should make those expectations explicit and fail with a clear error message when those expectations aren't met.

For example, `supersimple.compute_discrepancies` expects that every record in `sample_cvrs` that has a CVR for the given contest is also in `cvrs` and has a CVR record for that contest

nikhilb4a:
Closing as outdated, mostly completed. 

Claude analysis: 

Issue #805 is mostly outdated — the one concrete defect it named was fixed in 2021, and what remains is an open-ended aspiration rather than a trackable work item. I'd close it, with one caveat below.

The issue (Jonah Kagan, Oct 2020) asked that audit math modules make their input expectations explicit and fail with clear errors, citing one example: supersimple.compute_discrepancies would break if a sampled ballot's CVR wasn't present in the cvrs dict for that contest.

What I found:

The cited example is resolved. Jonah's own PR #1096, "Factor out discrepancy calculation" (March 2021), rewrote it so a missing CVR entry is no longer an unstated expectation — cvrs.get(ballot) feeds a discrepancy() function that explicitly defines the missing cases: ballot absent from CVRs or not found by the audit board counts as a two-vote overstatement, and a contest absent from a ballot's CVR counts as (0,0) votes ([supersimple.py:114-158](vscode-webview://0k4k6l6jq1j000v6br37se94obt1aph4jrn6rfigbb226lgpt5dc/server/audit_math/supersimple.py#L114-L158)). That's arguably better than the issue's ask — the edge cases became defined audit semantics instead of validation errors.
The general ask is partially true and inherently unfinishable as written. There's meaningful validation with clear messages in the math modules today (bravo.py has 7 raise/assert guards — "The risk-limit must be less than one!", "Cannot compute ASN for a tied contest" — suite.py 7, macro.py 3, others fewer), and in Arlo's architecture the heavy input validation deliberately lives at the API layer, with audit math receiving server-constructed data rather than raw user input. "Make all expectations explicit everywhere" has no completion criterion, which is presumably why the ticket sat assigned-but-untouched for five years.
[...]

PR #816, opened 2020-10-22
Makes audit math robust to ties and recounts

### LINKED ISSUE #812: Float division by 0 error on AB signoff

umbernhard:
**Description**

Updates the three implemented classes of audit math to handle tied elections and full hand recounts correctly. Fixes #812, starts to address #805. 

**Testing**
Adds new tests for tied election for each audit class. Updates BRAVO to handle ties more appropriately. 

**Progress**

Ready.


server/audit_math/macro.py
```diff
@@ -57,6 +57,8 @@ def compute_error(
             a_lp = sampled_results[contest.name][loser]
 
             V_wl = contest.candidates[winner] - contest.candidates[loser]
+            if not V_wl:
```
jonahkagan, line 60: small nit: when dealing with numbers, i prefer explicit: `if V_wl == 0:`

server/audit_math/macro.py
```diff
@@ -200,6 +205,9 @@ def get_sample_sizes(
 
     if not U:
         return 1
+    elif U < 0:
+        # This means we have a tie
+        return len(reported_results)
```
jonahkagan, line 210: interesting, so in a tie we do a full recount. i was thinking we'd just fail but i guess this is a more graceful answer

server/audit_math/macro.py
```diff
@@ -240,6 +248,10 @@ def compute_risk(
     """
     assert risk_limit < 1, "The risk-limit must be less than one!"
 
+    # We've done a full hand recount
+    if len(sample_results) == len(reported_results):
+        return 0, True
```
jonahkagan, line 253: we don't get a p-value in a full hand recount? what if the recount results don't match the reported results?

server/audit_math/macro.py
```diff
@@ -252,7 +264,10 @@ def compute_risk(
 
         taint = e_p / u_p
 
-        p *= (1 - 1 / U) / (1 - taint)
+        if taint == 1:
+            p *= 1
```
jonahkagan, line 268: ok maybe im missing something but what's the point of multiplying by 1?

server/audit_math/supersimple.py
```diff
@@ -139,7 +139,13 @@ def compute_discrepancies(
                 if e:
                     # we found a discrepancy!
                     found = True
-                e_weighted = e / V_wl
+
+                if not V_wl:
+                    # In this case the error is undefined
+                    e_weighted = -1.0
```
jonahkagan, line 145: can you explain this case? i dont understand it. when would `V_wl` be 0 and why do you then set `e_weighted` to -1.0?

server/audit_math/supersimple.py
```diff
@@ -270,11 +274,22 @@ def compute_risk(
...
 
-        denom = (2 * gamma) / V
-        p_b = (1 - 1 / U) / (1 - (e_r / denom))
+        if contest.diluted_margin:
+            U = 2 * gamma / contest.diluted_margin
+            denom = (2 * gamma) / V
+            p_b = (1 - 1 / U) / (1 - (e_r / denom))
+        else:
+            # If the contest is a tie, this step results in 1 - 1/(infinity)
+            # divided by 1 - e_r/infinity, i.e. 1
```
jonahkagan, line 283: had to think about this lol, but i think i agree!

server/tests/audit_math/test_bravo.py
```diff
@@ -379,6 +379,44 @@ def test_compute_risk_empty(contests):
...
+
+    # Now do a full hand recount
+    sample_results = {
+        "cand1": 501,
+        "cand2": 499,
+    }
+
+    computed_p, res = bravo.compute_risk(RISK_LIMIT, contest, sample_results)
+
+    assert not computed_p[("cand1", "cand2")]
```
jonahkagan, line 416: similar nit as to above - are you asserting that it's `None` or `0`?

server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison.py
```diff
@@ -100,7 +100,7 @@
 Round Number,Contest Name,Targeted?,Sample Size,Risk Limit Met?,P-Value,Start Time,End Time,Audited Votes\r
 1,Contest 1,Targeted,22,No,0.1128631733,DATETIME,DATETIME,Choice 1-1: 5; Choice 1-2: 1\r
 1,Contest 2,Opportunistic,,No,0.1888821229,DATETIME,DATETIME,Choice 2-1: 4; Choice 2-2: 1; Choice 2-3: 0\r
-2,Contest 1,Targeted,41,Yes,0.0008560347,DATETIME,DATETIME,Choice 1-1: 16; Choice 1-2: 6\r
+2,Contest 1,Targeted,41,Yes,<0.0000000001,DATETIME,DATETIME,Choice 1-1: 16; Choice 1-2: 6\r
```
jonahkagan, line 103: i know im always doing this, but do you have an idea of what changed here? i dont think this test case should have hit any of the edge cases this PR is accounting for

server/audit_math/macro.py
umbernhard: Yeah. I don't expect Arlo will ever be used in this way (except for opportunistic contests, maybe?) but regardless it hedges against oddities like the funky CVRs @mcchilders was using.

server/audit_math/macro.py
umbernhard: So in a recount p-values no longer mean much. If you've counted all of your ballots by hand, then it is necessarily the case that there is zero chance that a computer malfunction could have caused a miscount (since no computers were used to do the count). While we could compute a p-value, it would be a little like estimating the chance it is raining while it is actively raining.

server/audit_math/macro.py
umbernhard: if the taint is one, then the p-value is undefined since the KM denominator goes to zero. I suppose the correct thing to do would be to use Decimal's `inf` value, because the p-value goes to infinity in this case. Essentially we're trying to say that if we ever observe a batch that obtains its maximum possible error, then something has truly gone horribly wrong and we need to do a full hand recount.

umbernhard: Multiplying p by 1 here is approximately equivalent to the p-value blowing up, but not exactly. I can change it appropriately.

server/audit_math/supersimple.py
umbernhard: If V_wl is 0, the race was tied, meaning that the denominator of the weighted error is 0 and the weighted error approaches infinity. I'm using -1 here to code that this has occurred, so later on we can gracefully handle some of the math. I suppose I should pick a better value like `inf`.

server/tests/audit_math/test_bravo.py
umbernhard: 0. I will change it to a `==`

server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison.py
umbernhard: Ah, I think this may have been a corner case identified above where the weighted error was improperly handled. I'll investigate further.

server/audit_math/macro.py
jonahkagan: might be nice to include that explanation as a comment for future readers

server/audit_math/macro.py
jonahkagan: cool, that all makes sense to me except why multiplying by 1 blows up the p-value. in math, doesn't multiplying by 1 usually not change the number?

server/audit_math/supersimple.py
jonahkagan: ah that makes sense

jonahkagan: agreed, inf would be better

server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison.py
umbernhard: Revisiting this, this contest reports 24 ballots total but has 16 votes for the winner and 10 votes for the loser for a total of 26 votes. That should never happen. 

I dug into it some more, and it appears that the p-value really should be that low. I think the previous result was being affected by the precision bug!

server/tests/audit_math/test_supersimple.py
```diff
@@ -297,6 +297,106 @@ def test_compute_risk(contests, cvrs):
...
+def test_snapshot_test():
+    contest_data = {
+        "winner": 16,
+        "loser": 10,
+        "ballots": 26,
+        "numWinners": 1,
+        "votesAllowed": 1,
+    }
+
+    contest = Contest("Jonah Test", contest_data)
```
jonahkagan, line 358: lol i dont need to be immortalized

server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison.py
jonahkagan: shouldn't this have changed with the precision PR then, not this PR?

Pipfile
```diff
@@ -4,7 +4,7 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [dev-packages]
-mypy = "*"
+mypy = "==0.780"
```
jonahkagan, line 7: i upgraded to the latest mypy version on `main` and it didn't cause any problems. do you remember what happened when you tried?

jonahkagan: oh weird i made a new branch to check and it does fail on CI: https://app.circleci.com/pipelines/github/votingworks/arlo/3144/workflows/4410b01b-67c4-4273-8c44-1297e83a0f04/jobs/4966

umbernhard: It was complaining about types in `server/util/group_by.py` and somewhere else. See here: https://app.circleci.com/pipelines/github/votingworks/arlo/3137/workflows/3be6d301-e3cf-46c3-aa56-8eb77e560549/jobs/4936

server/audit_math/macro.py
umbernhard: I fixed this by setting p to `Decimal('inf')`

server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison.py
umbernhard: Oh, I see what's happening here. Because we have more ballots in the CVR than the `total_ballots` field, the audit thinks a full-hand recount has happened in the second round, so it sets the p value to 0.

umbernhard: So yeah, this test case shouldn't really correspond to real-world, ever. We'd never sample 24 ballots 41 times :joy:

jonahkagan: ok that makes sense. kind of surprised it didn't just print a p-value of 0 though. but maybe we don't expect that to actually happen so it doesn't matter?

Pipfile
jonahkagan: fixing this on #849

jonahkagan: so you can take this out

server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison.py
umbernhard: Yeah. When we changed the report to just replace any values less than 0.00000001 with that string, we didn't special case p-values of 0. I highly doubt we'll ever be in this situation though.

jonahkagan (approved):

PR #818, opened 2020-10-22
Sorting audits alphabetically

MorganLove:
**Description**

- Just what the title says

**Testing**

- No changes needed.

**Progress**

- Done!

### DIFF
--- a/client/src/components/HomeScreen.tsx
+++ b/client/src/components/HomeScreen.tsx
@@ -130,7 +130,7 @@ const ListAuditsAuditAdmin: React.FC = () => {
 
   return (
     <ListAuditsWrapper>
-      {meta.organizations.map(organization => (
+      {sortBy(meta.organizations, o => o.name).map(organization => (
         <div key={organization.id}>
           <h2>Audits - {organization.name}</h2>
           {organization.elections.length === 0 ? (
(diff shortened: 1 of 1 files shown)


client/src/components/HomeScreen.tsx
```diff
@@ -138,17 +138,19 @@ const ListAuditsAuditAdmin: React.FC = () => {
...
-                to={`/election/${election.id}`}
-                intent="primary"
-                large
-                fill
-              >
-                {election.auditName}
-              </AuditLink>
-            ))
+            organization.elections
+              .sort((a, b) => (a.auditName > b.auditName ? 1 : -1))
```
jonahkagan, line 142: two things:
- do you think we should sort the organizations as well?
- i think i used `sortBy` below - do you want to use that here? i think the reason i did that was because `sort` modifies the array in place, so it's generally a bit dangerous to use. and also so you don't have to define your own comparator function. but i don't think the in-place modification is a big risk here, so maybe doesn't matter

jonahkagan (approved):

client/src/components/HomeScreen.tsx
MorganLove: @jonahkagan Oh yes, I'm tracking what you did there now. Sure, no problem. I swapped to `sortBy` and sorted both organizations and audits now.

jonahkagan (approved):

PR #819, opened 2020-10-22
Making container and tabulator columns conditional

MorganLove:
**Description**

Created in response to https://github.com/votingworks/arlo/pull/813#discussion_r509744529

**Testing**

- Updating snapshots and mocks slightly

**Progress**

- Done

PR #820, opened 2020-10-24
Refactor data entry table

MorganLove:
**Description**

- trying a refactor of the data entry table from blueprintjs to react-table

**Testing**

- snapshots updated
- a couple tests are not passing, not sure why yet

**Progress**

- this breaks the responsiveness of the table to width changes, making it harder to use on mobile. Haven't looked into the solutions for that with react-table like we did with blueprintjs yet.
- unclear on if this will be simpler or easier than blueprintjs yet. The initial motivation was to make it possible to drop the container and tabulator columns instead of doing 'N/A' if their values aren't present, but this is looking like a lot of overkill for a relatively minor improvement.

Alternative PR to #819

### DIFF
--- a/client/src/components/Atoms/Wrapper.tsx
+++ b/client/src/components/Atoms/Wrapper.tsx
@@ -11,7 +11,13 @@ export const Inner = styled.div`
   margin-right: auto;
   margin-left: auto;
   width: 1020px;
-  min-width: 1020px;
+  padding: 0 30px;
+`
+
+export const ThinnerInner = styled.div`
+  display: flex;
+  margin-right: auto;
+  margin-left: auto;
   padding: 0 30px;
 `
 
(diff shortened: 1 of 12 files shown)


client/src/components/Atoms/Wrapper.tsx
```diff
@@ -11,7 +11,13 @@ export const Inner = styled.div`
   margin-right: auto;
   margin-left: auto;
   width: 1020px;
-  min-width: 1020px;
+  padding: 0 30px;
+`
```
jonahkagan, line 15: I was playing around with this a bit and figured out that you can just change Inner to these styles and everything seems to work:
```ts
export const Inner = styled.div`
  display: flex;
  margin-right: auto;
  margin-left: auto;
  width: 100%;
  max-width: 1020px;
  padding: 0 30px;
`
```

an advantage here is that the main panel width for the audit board view will then match the width of the Header, which also uses Inner

jonahkagan (approved):

PR #829, opened 2020-10-26
Reimplementing precision fix on a clean branch

umbernhard:
**Description**
This is a copy of #784 that fixes some precision bugs like #778.

**Testing**

New case for a distant third party in BRAVO. 

**Progress**
Ready?


server/api/sample_sizes.py
```diff
@@ -20,11 +20,14 @@ def sample_size_options(
...
-    if not election.risk_limit:
+    if election.risk_limit is None:
         raise BadRequest("Cannot compute sample sizes until risk limit is set")
-    risk_limit = float(election.risk_limit) / 100
 
     def sample_sizes_for_contest(contest: Contest):
+        # For some reason putting this here works, but not if it's in the outer
+        # function
+        if election.risk_limit is None:
+            raise BadRequest("Cannot compute sample sizes until risk limit is set")
```
jonahkagan, line 30: maybe just `assert election.risk_limit is not None` then, since we know it's ok at this point

server/audit_math/macro.py
```diff
@@ -254,7 +259,7 @@ def compute_risk(
 
         p *= (1 - 1 / U) / (1 - taint)
 
-        if p < risk_limit:
-            return p, True
+        if p <= alpha:
+            return float(p), True
 
-    return p, p < risk_limit
+    return float(p), p <= alpha
```
jonahkagan, line 265: does `float(p)` pretty much round the way we want?

server/api/rounds.py
```diff
@@ -291,30 +291,29 @@ def sampled_ballot_interpretations_to_cvrs(contest: Contest) -> supersimple.CVRS
 def calculate_risk_measurements(election: Election, round: Round):
     if not election.risk_limit:  # Shouldn't happen, we need this for typechecking
         raise Exception("Risk limit not defined")  # pragma: no cover
```
jonahkagan, line 293: since this shouldn't happen, maybe switch to `assert election.risk_limit is not None`. im realizing that's a better pattern than having a custom exception

server/audit_math/macro.py
umbernhard: yeah, since the issues with floats are way out at the end of the decimal places this should be fine. And really, those errors don't usually matter, except when they stack up when doing multiple multiplications.

jonahkagan: great!

jonahkagan (approved):
really clean!

PR #830, opened 2020-10-27
Remove single jurisdiction backend

jonahkagan:
- Remove Election.is_multi_jurisdiction and other unused fields
- Also changes the /election/new endpoint to just /election
- Removes the parallelism from the backend tests - for some reason it was starting to cause a lot more intermittent failures after these changes - will investigate separately


client/src/components/HomeScreen.test.tsx
```diff
@@ -16,7 +16,7 @@ const apiCalls = {
     },
   },
   postNewAudit: (body: {}) => ({
-    url: '/api/election/new',
+    url: '/api/election',
```
benadida, line 19: In this pattern, I'm used to a trailing slash for the URL to indicate a collection of elections that you're adding to.

benadida (approved):

client/src/components/HomeScreen.test.tsx
jonahkagan: i know, but none of the other collections do that. eventually i want to change all of them to be plural

Makefile
```diff
@@ -42,10 +42,10 @@ test-client:
 	yarn --cwd client test
 
 test-server:
-	pipenv run pytest -n 2
+	pipenv run pytest
 
 test-server-coverage:
-	pipenv run pytest -n 2 --cov=.
+	pipenv run pytest --cov=.
```
jonahkagan, line 48: @umbernhard heads up - had to remove the parallelism because it's causing intermittent failures again

PR #833, opened 2020-10-28
Adding headers to ballot comparison audit screen

### LINKED ISSUE #821: Add sub-labels to the Participants and Contests file upload for AAs
Now that there are two file uploads in the Participants/Contests tab of the AA Audit Setup workflow, we need to add labels above each so you know which file goes where:
"Participants File" and "Standardized Contests File"

Thank you to Ginny for the mockup!

MorganLove:
**Description**

Closes #821 

**Testing**

- snapshots updated

**Progress**

done


jonahkagan (approved):
Seems like it would be simpler to have a header for State, Participants File, and Standardized Contests File (if applicable) all the time, instead of doing it conditionally. But this is fine too

MorganLove:
@jonahkagan yeah, I was thinking that too, but if we want to do that it's easy to change it later. Up to @mcchilders as far as design goes.

PR #834, opened 2020-10-28
Adding clickable link to pdf for AB credentials

### LINKED ISSUE #826: Make the AB URL in the PDF linked
Jonah posted this: https://stackoverflow.com/questions/25501024/how-to-create-hyperlink-in-pdf-using-jspdf-js-library/60821851#60821851

MorganLove:
**Description**

Closes #826 

Just added an extra line with a clickable direct link. Straightforward and simple.

**Testing**

- snapshots

**Progress**

done

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/RoundManagement/generateSheets.ts
+++ b/client/src/components/MultiJurisdictionAudit/RoundManagement/generateSheets.ts
@@ -118,22 +118,18 @@ export const downloadAuditBoardCredentials = (
         20,
         120
       )
-      auditBoardCreds.text(
-        auditBoardCreds.splitTextToSize(
-          `${window.location.origin}/auditboard/${board.passphrase}`,
-          180
-        ),
-        20,
-        140
+      const urlText: string[] = auditBoardCreds.splitTextToSize(
+        `${window.location.origin}/auditboard/${board.passphrase}`,
+        180
       )
-      auditBoardCreds.setTextColor(0, 0, 238)
-      auditBoardCreds.textWithLink(
-        'Or you can click here to go there directly.',
-        20,
-        160,
-        { url: `${window.location.origin}/auditboard/${board.passphrase}` }
+      const urlHeight = urlText.reduce(
+        (a: number, t: string) => auditBoardCreds.getTextDimensions(t).h + a,
+        0
       )
-      auditBoardCreds.setTextColor(0, 0, 0)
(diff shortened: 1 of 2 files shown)


client/src/components/MultiJurisdictionAudit/RoundManagement/generateSheets.ts
```diff
@@ -126,6 +126,14 @@ export const downloadAuditBoardCredentials = (
...
         140
       )
+      auditBoardCreds.setTextColor(0, 0, 238)
+      auditBoardCreds.textWithLink(
+        'Or you can click here to go there directly.',
+        20,
+        160,
+        { url: `${window.location.origin}/auditboard/${board.passphrase}` }
+      )
+      auditBoardCreds.setTextColor(0, 0, 0)
```
jonahkagan, line 136: my understanding of this task was to change the existing url text to be a hyperlink, not to add a separate link

MorganLove: @jonahkagan Yep, the hope was that that would work, but unfortunately it does not. You cannot create a multi-line hyperlink using jsPDF. The only way would be to create a clickable zone over the top of the link, but since the shape of the link is variable, that would be involved. Adding a clickable link made the most sense to me and is the easiest solution. @mcchilders do you think this is adequate?

jonahkagan: Ok this is kind of ridiculous but I just went and checked the source of textWithLink and it actually does what you describe - it just makes a clickable zone over the top of the link (https://artskydj.github.io/jsPDF/docs/modules_annotations.js.html#line241). so it might actually be totally reasonable to do that. you could make the width `max(width of the text, page width)` or honestly just use `page width` to make it simpler

MorganLove: @jonahkagan The documentation for jsPDF is honestly a pain sometimes, haha. Good job digging further! I was still having issues because for some reason just using the dimensions of the text itself wasn't adequate, with the pdf browser still having an edge around it that was giving the wrong link. Also, having the height be dynamic with the unknown string length possibly wrapping made it interesting. Let me know what you think of my solution.

jonahkagan (approved):
looks good to me. might be worth adding a comment to explain a bit about what's going on, since it sounded a bit tricky

PR #835, opened 2020-10-28
Adding padding below ja screens

### LINKED ISSUE #823: Add padding to bottom of JA screens
Now that we've added more things to the JA screens, the buttons sometimes bump up against the very bottom of the screen (like on the file upload screen). Let's add some padding under the buttons so that doesn't happen.

MorganLove:
**Description**

Closes #823 

**Testing**

Updated snapshots

**Progress**

Done

@jonahkagan is there a reason we don't just have bottom padding on the `Wrapper` component by default?

### DIFF
--- a/client/src/__snapshots__/App.test.tsx.snap
+++ b/client/src/__snapshots__/App.test.tsx.snap
@@ -6,7 +6,7 @@ exports[`App / renders aa logged in properly 1`] = `
     class="Toastify"
   />
   <div
-    class="sc-cIShpX dhyUlm"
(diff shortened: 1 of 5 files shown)


jonahkagan:
I think if you wanna try adding it to `Wrapper` that could work too. Just need to check the other screens that use it

MorganLove:
> I think if you wanna try adding it to `Wrapper` that could work too. Just need to check the other screens that use it

@jonahkagan Seems like doing that works fine, switched to that method.

jonahkagan (approved):

PR #837, opened 2020-10-28
Minerva changes

umbernhard:
**Description**
Reconfigures the data passed to the math to be friendlier to Minerva. Introduces a new DB query to get results by round, not just cumulatively, and updates BRAVO and applicable files. 

**Testing**

N/A

**Progress**
Ready

### DIFF
--- a/server/api/rounds.py
+++ b/server/api/rounds.py
@@ -112,12 +112,20 @@ def count_audited_votes(election: Election, round: Round):
             db_session.add(result)
 
 
+# Sum the audit results for each contest choice from all rounds so far
+def cumulative_contest_results(contest: Contest) -> Dict[str, int]:
+    results_by_choice: Dict[str, int] = defaultdict(int)
+    for result in contest.results:
+        results_by_choice[result.contest_choice_id] += result.result
+    return results_by_choice
+
+
 # Get round-by-round audit results
 def contest_results_by_round(contest: Contest) -> Dict[str, Dict[str, int]]:
-    results_by_round: Dict[str, Dict[str, int]] = defaultdict(lambda: defaultdict(int))
+    results_by_choice: Dict[str, Dict[str, int]] = defaultdict(lambda: defaultdict(int))
     for result in contest.results:
-        results_by_round[result.round_id][result.contest_choice_id] = result.result
-    return results_by_round
+        results_by_choice[result.round_id][result.contest_choice_id] = result.result
+    return results_by_choice
 
 
 # { batch_key: { contest_id: { choice_id: votes }}}
--- a/server/api/rounds.py
+++ b/server/api/rounds.py
@@ -119,6 +119,12 @@ def cumulative_contest_results(contest: Contest) -> Dict[str, int]:
         results_by_choice[result.contest_choice_id] += result.result
     return results_by_choice
 
+# Get round-by-round audit results
+def contest_results_by_round(contest: Contest) -> Dict[str, Dict[str, int]]:
+    results_by_choice: Dict[str, Dict[str, int]] = defaultdict(lambda: defaultdict(int))
+    for result in contest.results:
+        results_by_choice[result.round_id][result.contest_choice_id] = result.result
+    return results_by_choice
 
 # { batch_key: { contest_id: { choice_id: votes }}}
 BatchTallies = Dict[Tuple[str, str], Dict[str, Dict[str, int]]]
@@ -298,8 +304,8 @@ def calculate_risk_measurements(election: Election, round: Round):
             p_values, is_complete = ballot_polling.compute_risk(
                 election.risk_limit,
                 sampler_contest.from_db_contest(contest),
-                cumulative_contest_results(contest),
-                "BRAVO",
+                contest_results_by_round(contest),
+                BallotPollingType.BRAVO,
             )
             p_value = max(p_values.values())
         elif election.audit_type == AuditType.BATCH_COMPARISON:
--- a/server/api/rounds.py
+++ b/server/api/rounds.py
@@ -112,20 +112,12 @@ def count_audited_votes(election: Election, round: Round):
             db_session.add(result)
 
 
-# Sum the audit results for each contest choice from all rounds so far
-def cumulative_contest_results(contest: Contest) -> Dict[str, int]:
-    results_by_choice: Dict[str, int] = defaultdict(int)
-    for result in contest.results:
-        results_by_choice[result.contest_choice_id] += result.result
-    return results_by_choice
-
-
 # Get round-by-round audit results
 def contest_results_by_round(contest: Contest) -> Dict[str, Dict[str, int]]:
-    results_by_choice: Dict[str, Dict[str, int]] = defaultdict(lambda: defaultdict(int))
+    results_by_round: Dict[str, Dict[str, int]] = defaultdict(lambda: defaultdict(int))
     for result in contest.results:
-        results_by_choice[result.round_id][result.contest_choice_id] = result.result
-    return results_by_choice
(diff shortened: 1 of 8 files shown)


server/audit_math/ballot_polling.py
```diff
@@ -16,20 +16,40 @@ class SampleSizeOption(TypedDict):
 
 
 def get_sample_size(
-    risk_limit: int, contest: Contest, sample_results: Dict[str, int], mathtype: str
+    risk_limit: int,
+    contest: Contest,
+    sample_results: Dict[str, Dict[str, int]],
+    mathtype: str,
```
jonahkagan, line 22: should the type of `mathtype` be `BallotPollingType`? also a more conventional variable name would be `math_type`

server/api/sample_sizes.py
```diff
@@ -26,16 +26,16 @@ def sample_size_options(
     def sample_sizes_for_contest(contest: Contest):
         assert election.risk_limit is not None
         if election.audit_type == AuditType.BALLOT_POLLING:
-            cumulative_results = (
-                {choice.id: 0 for choice in contest.choices}
+            sample_results = (
+                {"round0": {choice.id: 0 for choice in contest.choices}}
```
jonahkagan, line 30: this seems kinda funky to me. could the math just take `None` if we haven't started the first round yet?

server/api/rounds.py
```diff
@@ -298,8 +306,8 @@ def calculate_risk_measurements(election: Election, round: Round):
             p_values, is_complete = ballot_polling.compute_risk(
                 election.risk_limit,
                 sampler_contest.from_db_contest(contest),
-                cumulative_contest_results(contest),
```
jonahkagan, line 301: can you remove `cumulative_contest_results` now, or is it still being used elsewhere?

server/api/rounds.py
```diff
@@ -120,6 +120,14 @@ def cumulative_contest_results(contest: Contest) -> Dict[str, int]:
     return results_by_choice
 
 
+# Get round-by-round audit results
+def contest_results_by_round(contest: Contest) -> Dict[str, Dict[str, int]]:
+    results_by_choice: Dict[str, Dict[str, int]] = defaultdict(lambda: defaultdict(int))
+    for result in contest.results:
+        results_by_choice[result.round_id][result.contest_choice_id] = result.result
+    return results_by_choice
```
jonahkagan, line 128: nit: i think `results_by_round` makes more sense than `results_by_choice` given the name of the function

server/api/sample_sizes.py
jonahkagan: or an empty dict?

server/audit_math/ballot_polling.py
umbernhard: ah, for sure.

server/api/sample_sizes.py
umbernhard: Hmm, probably. I'll mess with it and see what I can do.

server/api/rounds.py
umbernhard: I will check.

umbernhard: Nope, I removed it.

server/api/sample_sizes.py
umbernhard: Yep, the math takes and empty dict just fine.

jonahkagan: awesome! do you think we could do the same for the batch results just below?

also, i just noticed that the ballot comparison case doesn't take into account `round_one` at all, which makes me think there's a bug

umbernhard: Ballot comparison does, it just does it by setting the observed errors all to zero, since we haven't started yet. I think that's correct.

I'll look at batch and get back to you.

umbernhard: Looks like yes, and we can even remove the conditional logic. Pushing soon.

umbernhard: Done

jonahkagan: here's what i meant about ballot comparison: https://github.com/votingworks/arlo/pull/838

jonahkagan: you still need the conditional logic (read the comment at the top of the function to understand why). if there's no test that catches that we should add one

jonahkagan: it's probably my least favorite thing i've done in this codebase, cuz it's pretty confusing. if you can think of a better way to deal with it i'm open

two ideas i've had:
- store the sample size in the db once we start the first round, then return the stored sample size from the endpoint
- have a required `round_number` argument instead of an optional `round_one` flag

umbernhard: I see. This feels like a good question for @mcchilders and/or a future PR. For now, I'm going to revert to the preexisting behavior (fixed for this PR).

server/tests/audit_math/test_bravo.py
```diff
@@ -479,35 +479,37 @@ def test_compute_risk_empty(contests):
...
+    "test4": {"round0": {"cand1": 0,}},
+    "test5": {"round0": {"cand1": 0, "cand2": 0,}},
+    "test6": {"round0": {"cand1": 0, "cand2": 0, "cand3": 0}},
+    "test7": {"round0": {"cand1": 0, "cand2": 0, "cand3": 0}},
+    "test8": {"round0": {"cand1": 0, "cand2": 0, "cand3": 0}},
+    "test9": {"round0": {"cand1": 0, "cand2": 0, "cand3": 0}},
+    "test10": {"round0": {"cand1": 0, "cand2": 0, "cand3": 0}},
+    "test11": {"round0": {"cand1": 0, "cand2": 0}},
+    "test12": {"round0": {"cand1": 0, "cand2": 0, "cand3": 0}},
+    "test_small_third_candidate": {"round0": {"cand1": 0, "cand2": 0, "cand3": 0}},
```
jonahkagan, line 494: should these all just be `None` now?

server/api/rounds.py
jonahkagan: i think the change for this got left behind somewhere

server/api/sample_sizes.py
```diff
@@ -45,20 +42,12 @@ def sample_sizes_for_contest(contest: Contest):
...
-                    }
-                    for batch_key, batch_results in sample_results.items()
-                }
+            cumulative_batch_results = rounds.cumulative_batch_results(election)
             sample_size = macro.get_sample_sizes(
                 election.risk_limit,
                 sampler_contest.from_db_contest(contest),
                 rounds.batch_tallies(election),
-                sample_results,
+                cumulative_batch_results,
```
jonahkagan, line 50: we need to do something similar to what you did for ballot polling here. and we need a test case to catch this. im happy to work on the test case - i have an idea of what to do

server/api/rounds.py
umbernhard: Fixed

server/tests/audit_math/test_bravo.py
umbernhard: Ah, yep. It doesn't really matter, but I will fix it.

server/api/sample_sizes.py
umbernhard: I'll work on changing MACRO. Since this PR only pertains to ballot polling, do you mind if we make that a separate PR

jonahkagan: great, separate PR sounds good. but then we'll prob want to avoid introducing buggy behavior in this PR, so maybe just undo the changes to this bit?

umbernhard: Done

jonahkagan (approved):

Issue #741, opened 2020-09-16
Count phantoms as zombies programmatically


mcchilders:
Make sure that any ballot that is:
- Not found (audit board clicks "ballot not found" button)

Counts as a worst-case-scenario "zombie" ballot - e.g. 2 vote overstatement

PR #842, opened 2020-10-29
Count not found ballots as two-vote overstatement

jonahkagan:
Task: #741 

If an audit board can't find a ballot, the research says we should
conservatively count that as a two-vote overstatement.

We pass `None` as the `sample_cvr` value for this case, and then
interpret that in `compute_discrepancies`.

Note: the typechecker is not smart enough to deal with this change to
the type, so there are also some other changes to accommodate that.

### DIFF
--- a/server/audit_math/supersimple.py
+++ b/server/audit_math/supersimple.py
@@ -168,7 +168,13 @@ def compute_discrepancies(
                     if e:
                         # we found a discrepancy!
                         found = True
-                    e_weighted = Decimal(e) / Decimal(V_wl)
+
+                    if V_wl == 0:
+                        # In this case the error is undefined
+                        e_weighted = Decimal("inf")
+                    else:
+                        e_weighted = Decimal(e) / Decimal(V_wl)
+
                     if e_weighted > e_r:
                         e_r = e_weighted
                         e_int = e
--- a/server/audit_math/supersimple.py
+++ b/server/audit_math/supersimple.py
@@ -16,14 +16,16 @@ u1: Decimal = Decimal(0.001)
 o2: Decimal = Decimal(0.0001)
 u2: Decimal = Decimal(0.0001)
 
-# { ballot_id: { contest_id: { choice_id: 0 | 1 }}}
+
+# CVR: { contest_id: { choice_id: 0 | 1 }}
+# CVRS: { ballot_id: CVR }
 CVR = Dict[str, Dict[str, int]]
-CVRS = Dict[str, CVR]
+CVRS = Dict[str, Optional[CVR]]
 
 
 class SampleCVR(TypedDict):
     times_sampled: int
-    cvr: CVR
+    cvr: Optional[CVR]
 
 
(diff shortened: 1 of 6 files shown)


server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison.py
```diff
@@ -109,39 +109,36 @@
 \r
 ######## ROUNDS ########\r
 Round Number,Contest Name,Targeted?,Sample Size,Risk Limit Met?,P-Value,Start Time,End Time,Audited Votes\r
-1,Contest 1,Targeted,22,No,0.1495325187,DATETIME,DATETIME,Choice 1-1: 10; Choice 1-2: 9\r
-1,Contest 2,Opportunistic,,No,0.1124219565,DATETIME,DATETIME,Choice 2-1: 15; Choice 2-2: 9; Choice 2-3: 8\r
-2,Contest 1,Targeted,41,Yes,0.0013218788,DATETIME,DATETIME,Choice 1-1: 24; Choice 1-2: 17\r
-2,Contest 2,Opportunistic,,Yes,0.0113963098,DATETIME,DATETIME,Choice 2-1: 21; Choice 2-2: 10; Choice 2-3: 13\r
+1,Contest 1,Targeted,22,No,3.9787903602,DATETIME,DATETIME,Choice 1-1: 10; Choice 1-2: 8\r
+1,Contest 2,Opportunistic,,No,2.9913452992,DATETIME,DATETIME,Choice 2-1: 14; Choice 2-2: 8; Choice 2-3: 8\r
+2,Contest 1,Targeted,24,No,4.2908708982,DATETIME,DATETIME,Choice 1-1: 12; Choice 1-2: 7\r
```
jonahkagan, line 114: @umbernhard this seems weird to me - why would the sample size for round 2 go down when the total discrepancies for round 1 went up? I generally try to have this test meet the risk limit in round 2, but it's not happening because the sample size for round 2 is so small that there's no way to balance out the discrepancies from round 1.

umbernhard: Hrm. I'm not sure, I'm gonna look into it tomorrow.

server/audit_math/supersimple.py
```diff
@@ -126,48 +133,53 @@ def compute_discrepancies(
...
-                if e_weighted > e_r:
-                    e_r = e_weighted
-                    e_int = e
+        # Special case: if ballot can't be found by audit board, count it as a
+        # two-vote overstatement
+        if ballot_sample_cvr is None:
+            e_int = 2
+            e_weighted = Decimal(e_int) / Decimal(
+                contest.diluted_margin * contest.ballots
+            )
```
jonahkagan, line 143: i made this up - please check

umbernhard: Yep, looks good

umbernhard (approved):

PR #844, opened 2020-10-30
Added frontend for math selection

umbernhard:
**Description**

Adds frontend radio button for selecting what kind of math to use in ballot polling.

**Testing**

None yet, hoping @MorganLove or @jonahkagan  will have suggestions here. 

**Progress**
Functional, but failing tests. I think I need some help from @MorganLove here.


MorganLove:
@umbernhard Great job! It just needed snapshots to be updated so I did that real quick for you and rebased. :slightly_smiling_face: 

To update snapshots you just need to run `yarn test:update` or if you want to inspect them first, run `yarn test` and look through the diffs, then hit `u` and it'll update them.

MorganLove (dismissed):
You just need to add the following line after 228:

`ballotPollingType,`

And that will send the value back to the server. Other than that, looks good to go!

client/src/components/HomeScreen.tsx
```diff
@@ -285,6 +298,26 @@ const CreateAudit: React.FC = () => {
...
+                          )
+                        }
+                        selectedValue={values.ballotPollingType}
+                      >
+                        <Radio value="BRAVO">BRAVO</Radio>
+                        <Radio value="MINERVA">Minerva (Not recommended)</Radio>
+                      </RadioGroup>
+                    </label>
+                  </BallotPollingWrapper>
+                ) : null}
```
jonahkagan, line 320: i think we should disable this feature until Minerva is implemented

client/src/components/HomeScreen.tsx
```diff
@@ -235,6 +247,7 @@ const CreateAudit: React.FC = () => {
         organizationId: meta!.organizations[0].id,
         auditName: '',
         auditType: 'BALLOT_POLLING',
+        ballotPollingType: 'BRAVO',
```
jonahkagan, line 250: maybe add a test that selecting different ballot polling types sends the right request to the backend

client/src/components/HomeScreen.tsx
umbernhard: Agreed. I think we're also contemplating implementing sequential BRAVO, so we might also have that as an option too. Pending discussions.

client/src/components/HomeScreen.tsx
```diff
@@ -291,12 +306,43 @@ const CreateAudit: React.FC = () => {
...
+                    }
+                    case 'BALLOT_COMPARISON': {
+                      setFieldValue('auditMathType', 'SUPERSIMPLE')
+                      break
+                    }
+                    default: {
+                      setFieldValue('auditMathType', 'MACRO')
+                      break
+                    }
+                  }
```
jonahkagan, line 324: Two things:
1. I'm a little sketched out by calling `setFieldValue('auditType', e.currentTarget.value)` and then immediately trying to access `values.auditType` in the next line. Maybe it works, but it seems suspicious/not clear to me, since I don't know how `setFieldValue` works. Also, I would expect that `setFieldValue` triggers a re-render each time it's called. I think it might be better to use Formik's `setValues` and set both auditType and auditMathType in the same call, so that only one re-render is triggered.
2. I generally like to make all cases explicit instead of using default (in case another case gets added later). one way to do that is with an object.

So overall I might try this:

```ts
const auditType = e.currentTarget.value
const auditMathType = {
  BALLOT_POLLING: "BRAVO",
  BALLOT_COMPARISON: "SUPERSIMPLE",
  BATCH_COMPARISON: "MACRO",
}[auditType]
setValues({ ...values, auditType, auditMathType })
```

jonahkagan (approved):
This looks good! I left one comment but it's not blocking - feel free to take it or leave it

PR #847, opened 2020-10-30
Minerva integration

umbernhard:
**Description**

Integrates frontend changes into back end. 

**Testing**
Adds test for new configuration variable `ballotPollingType`.

**Progress**

Good to go pending frontend.


server/api/election_settings.py
```diff
@@ -46,6 +47,7 @@ def serialize_election_settings(election: Election) -> JSONDict:
         "riskLimit": election.risk_limit,
         "state": election.state,
         "auditType": election.audit_type,
+        "ballotPollingType": election.ballot_polling_type,
```
jonahkagan, line 50: i think this field should only be present for ballot polling audits

server/api/election_settings.py
```diff
@@ -31,6 +31,7 @@
         # update the auditType/auditName in this endpoint - they get set on
         # audit creation only.
         "auditType": {"type": "string"},
+        "ballotPollingType": {"type": "string"},
```
jonahkagan, line 34: update comment

server/api/elections.py
```diff
@@ -50,6 +56,7 @@ def create_election():
         id=str(uuid.uuid4()),
         audit_name=election["auditName"],
         audit_type=election["auditType"],
+        ballot_polling_type=election["ballotPollingType"],
```
jonahkagan, line 59: if `ballotPollingType` is not a required field, then if it's not present in the request body, this will fail. you can use `election.get("ballotPollingType")` to defend against that

server/api/reports.py
```diff
@@ -187,10 +187,22 @@ def contest_rows(election: Election):
 def audit_settings_rows(election: Election):
     return [
         heading("AUDIT SETTINGS"),
-        ["Audit Name", "Audit Type", "Risk Limit", "Random Seed", "Online Data Entry?"],
+        [
+            "Audit Name",
+            "Audit Type",
+            "Ballot Polling Type",
```
jonahkagan, line 193: only include this column for ballot polling audits

server/api/rounds.py
```diff
@@ -295,11 +295,12 @@ def calculate_risk_measurements(election: Election, round: Round):
         contest = round_contest.contest
 
         if election.audit_type == AuditType.BALLOT_POLLING:
+            assert election.ballot_polling_type is not None
```
jonahkagan, line 298: why do you need this assertion here but not in `sample_sizes.py` below?

server/models.py
```diff
@@ -43,6 +43,7 @@ class Election(BaseModel):
     # audit_name must be unique within each Organization
     audit_name = Column(String(200), nullable=False)
     audit_type = Column(Enum(AuditType), nullable=False)
+    ballot_polling_type = Column(Enum(BallotPollingType), nullable=True)
```
jonahkagan, line 46: leave out `nullable=True` since that's the default. that way we only ever see the `nullable` option when it's false

server/tests/test_errors.py
```diff
@@ -140,6 +140,7 @@ def test_bgcompute_batch_tallies_errors(
 ):
     election = Election.query.get(election_id)
     election.audit_type = AuditType.BATCH_COMPARISON
+    election.ballot_polling_type = BallotPollingType.BRAVO
```
jonahkagan, line 143: seems like `election.ballot_polling_type` shouldn't be set in a batch comparison audit

server/tests/helpers.py
```diff
@@ -132,6 +133,7 @@ def create_election(
         {
             "auditName": audit_name or f"Test Audit {datetime.utcnow()}",
             "auditType": audit_type,
+            "ballotPollingType": ballot_polling_type,
```
jonahkagan, line 136: in order to make consistent test data, ballot polling type shouldn't be set if audit type isnt ballot polling

server/api/elections.py
```diff
@@ -16,6 +16,12 @@
             "type": "string",
             "enum": [audit_type.value for audit_type in AuditType],
         },
+        "ballotPollingType": {
+            "type": "string",
+            "enum": [
+                ballot_polling_type.value for ballot_polling_type in BallotPollingType
+            ],
+        },
```
jonahkagan, line 24: we should also add some validation in `validate_election` that `ballotPollingType` is always set for ballot polling audits and never set for other audit types

server/api/election_settings.py
umbernhard: Ooh, good nudge. I anticipate that we will eventually be supporting a wide variety of math types, not just for ballot polling. E.g., if we get SHANGRLA implemented it will apply to all three methods. So I could make this more general (audit_math_type or something). Thoughts?

server/api/rounds.py
umbernhard: The type checker was having issues with it (without this it sees it as `Optional` but it expects not `Optional`

jonahkagan: yeah that's what i figured. i'm just surprised it didn't have the same issue in sample_sizes

server/migrations/versions/8bc5c2037187_adding_audit_math_type.py
```diff
@@ -0,0 +1,42 @@
...
+        """
+        UPDATE election
+        SET audit_math_type = 'BRAVO'
+        """
+    )
+    op.alter_column("election", "audit_math_type", nullable=False)
+    # ### end Alembic commands ###
+
+
+def downgrade():
```
jonahkagan, line 39: add `# pragma: no cover` so that this function doesnt count against test coverage

server/tests/test_errors.py
```diff
@@ -140,6 +140,7 @@ def test_bgcompute_batch_tallies_errors(
 ):
     election = Election.query.get(election_id)
     election.audit_type = AuditType.BATCH_COMPARISON
+    election.audit_math_type = AuditMathType.BRAVO
```
jonahkagan, line 143: this should be `MACRO` i think

jonahkagan (approved):
two small things, but otherwise looks great!

PR #848, opened 2020-10-30
Swapping out sample manifest file for BC

### LINKED ISSUE #822: Add a new Ballot Manifest sample file that reflects Ballot Comp constraints
[SampleManifestBC.csv](https://zube.io/files/votingworks/dd3b457816bc4342d425e6ff427730b8-samplemanifestbc.csv)
See attached - for Ballot Comparison audits *only,* we should link to this as the sample file rather than the existing sample file

MorganLove:
**Description**

Closes #822 

**Testing**

- None necessary

**Progress**

- Done

### DIFF
--- a/client/public/sample_manifest_BC.csv
+++ b/client/public/sample_manifest_BC.csv
@@ -1,0 +1,13 @@
+﻿Container,Tabulator,Batch Name,Number of Ballots
+Box 1,1,1,3
+Box 1,2,1,2
+Box 1,2,2,5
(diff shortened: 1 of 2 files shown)


jonahkagan (approved):
any reason you don't want to add a test for this?

MorganLove:
@jonahkagan Because the change will be covered by the snapshots, but these tests are currently skipped. So when we get caught back up with testing it will be reflected in those.

Issue #845, opened 2020-10-30
Fix overvotes in candidate vote tallies (ballot comp only)


mcchilders:
We need to check to see whether the votes, marked on a particular ballot and contest on that ballot,  equal more than the votes allowed (an overvote) in that contest. If they do, we need to make sure those get tallied as no vote/0 in the candidate vote tallies. 

e.g. more CVR cells on a particular row are marked “1” in a particular contest than the # allowed in the “Vote for X” param for that same contest, the votes should go to 0.

(Note that this only matters for the automatic vote tallying portion of ballot comp - comparing the CVRs to the votes entered doesn’t require this, because the audit board will enter all the votes they see (even if it is an overvote) so those will match with no issues.

PR #852, opened 2020-11-02
Skip overvotes when counting CVR vote totals

jonahkagan:
Task: #845 

When the CVR contains an overvote, we don't want to count that ballot in
the number of votes for each candidate that we parse from the CVR
(because those ballots wouldn't have counted in the official tallies for
the election).


server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison_manifests.py
```diff
@@ -10,111 +10,14 @@
 snapshots[
     "test_ballot_comparison_container_manifest 1"
 ] = """Container,Tabulator,Batch Name,Ballot Number,Imprinted ID,Ticket Numbers,Already Audited,Audit Board
-CONTAINER1,TABULATOR1,BATCH1,1,1-1-1,0.243550726331576894,N,Audit Board #1
```
umbernhard, line 13: Couple of questions: Why are we deleting most of this file? Also, why do manifests contain ticket numbers?

umbernhard (approved):

server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison_manifests.py
jonahkagan: sorry, meant to leave a comment and explain this change but then forgot! this test case was actually a tie before, so i changed it to be not a tie (see https://github.com/votingworks/arlo/pull/852/files/013ca05bcb7287382fb21c3542f5f32cf0fc9cfc#diff-0de3e56ae4db4922e32954dc0501c689b12bfeb42da260106f8e8ef323616cd0L128). so that radically changed the sample size.

i'm pretty sure this snapshot is actually of the ballot retrieval list, not the manifest

umbernhard: Ah I see. Cool.

PR #854, opened 2020-11-03
Adds round sizes to the audit_math.sampler_contest object

umbernhard:
**Description**

This PR changes the sampler_contest object to include the number of samples drawn per round, so that it may be consumed by Minerva. 

**Testing**
N/A

**Progress**

Ready


jonahkagan:
could this data just be another argument to the minerva interface? i'm not sure i understand why it makes sense to add it to the contest object - it doesn't seem to really have much to do with the contest

umbernhard:
@jonahkagan this was the most elegant place I could think to put it. If we change the Minerva interface, we have to change the ballot polling interface too. This way, if we ever need this data for other audit types (I suspect we might in the future), it's there.

PR #856, opened 2020-11-03
passing round sizes to math module

umbernhard:
**Description**

Passes round sizes to minerva.py.

**Testing**
N/A

**Progress**

Ready.

### DIFF
--- a/server/api/rounds.py
+++ b/server/api/rounds.py
@@ -198,6 +198,12 @@ def cumulative_batch_results(election: Election) -> BatchTallies:
     }
 
 
+def round_sizes(election: Election) -> Dict[int, int]:
+    return {
+        rnd.round_num: len(list(rnd.sampled_ballot_draws)) for rnd in election.rounds
+    }
+
+
 def cvrs_for_contest(contest: Contest) -> supersimple.CVRS:
     choice_name_to_id = {choice.name: choice.id for choice in contest.choices}
 
@@ -305,6 +311,7 @@ def calculate_risk_measurements(election: Election, round: Round):
                 election.risk_limit,
                 sampler_contest.from_db_contest(contest),
                 contest_results_by_round(contest),
+                round_sizes(election),
                 BallotPollingType.BRAVO,
             )
             p_value = max(p_values.values())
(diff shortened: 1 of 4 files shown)


server/api/rounds.py
```diff
@@ -198,6 +198,12 @@ def cumulative_batch_results(election: Election) -> BatchTallies:
     }
 
 
+def round_sizes(election: Election) -> Dict[int, int]:
+    return {
+        rnd.round_num: len(list(rnd.sampled_ballot_draws)) for rnd in election.rounds
```
nealmcb, line 203: If the round_num values are always just counting up, would it make it easier to just return a simple List[int]? Then we don't have to argue whether the first round is 0 or 1 ;)

server/audit_math/ballot_polling.py
```diff
@@ -19,6 +19,7 @@ def get_sample_size(
     risk_limit: int,
     contest: Contest,
     sample_results: Optional[Dict[str, Dict[str, int]]],
+    round_sizes: Dict[int, int],
```
nealmcb, line 22: It seems more intuitive to me to pass the round sizes before the sample results (mirroring the sequence of events in the audit procedure) but of course it matters little.

nealmcb (approved):
Thanks for the quick work!

server/api/rounds.py
jonahkagan: I like it being explicit here, for actually a similar reason - there is no "round 0".

jonahkagan (approved):

server/api/rounds.py
jonahkagan: Also If you wanted to do this in one database query, you could. slightly more efficient, but not a big deal either way

```py
return dict(
  Round.query.filter_by(election_id=election.id)
  .join(SampledBallotDraw).group_by(Round.id)
  .values(Round.round_num, func.count(SampledBallotDraw.ticket_number))
)
```

umbernhard: Yeah, I'll use the db query, just couldn't quite figure out how to make it work.

PR #859, opened 2020-11-04
Ballot auditing options updates

### LINKED ISSUE #843: Hide the "Audit Board Can't Agree" button on data entry screen (all methods)
Given the debate about how these ballots should properly be counted if they are labeled as "can't agree," we're going to hide this button, and instead train locals to follow their existing adjudication guidance on what they should do when an adjudication board can't agree (usually, ask the election supervisor to tie break etc.) This preserves the existing/familiar workflow while also pushing the audit board to do the same process that would happen during original counting of the ballots, which is probably the right answer, anyway.

### LINKED ISSUE #846: Add new button for "not on ballot"
On the ballot data entry screen, we currently have a button that says "Blank vote/Not on ballot." We need to take that label back to its original form (E.g. just "Blank vote") and add a second button for "Not on ballot." Should have the same behaviors as "blank vote" - if you click this button, it clears any other votes, etc.

- Change the label back to "Blank vote"
- Add a new button for "Not on ballot" (somewhere at the bottom, wherever is fine)
- Hooking it up so that it actually records "Not on ballot" when you click it.

MorganLove:
**Description**

Closes #843 
Closes #846 

Is the front end for #853 

**Testing**

- Updated interactions with the changed buttons
- Updated snapshots

**Progress**

- Done!

### DIFF
--- a/client/src/components/DataEntry/BallotAudit.tsx
+++ b/client/src/components/DataEntry/BallotAudit.tsx
@@ -35,11 +35,11 @@ const BallotAudit: React.FC<IProps> = ({
           If the voter did not vote in the contest, select &quot;Blank vote/Not
           on Ballot.&quot;
         </p>
-        <p>
+        {/* <p>
           If the audit board cannot agree, select &quot;Audit board can&apos;t
           agree.&quot; You may add a comment for additional information about
           the disagreement.
-        </p>
+        </p> */}
         <Formik
           initialValues={{ interpretations }}
           enableReinitialize
--- a/client/src/components/DataEntry/BallotAudit.tsx
+++ b/client/src/components/DataEntry/BallotAudit.tsx
@@ -143,12 +143,12 @@ const BallotAuditContest = ({
           label={c.name}
         />
       ))}
-      <BlockCheckbox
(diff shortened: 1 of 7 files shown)


client/src/components/DataEntry/BallotAudit.tsx
```diff
@@ -32,14 +32,14 @@ const BallotAudit: React.FC<IProps> = ({
           marked on the paper ballot.
         </p>
         <p>
-          If the voter did not vote in the contest, select &quot;Blank vote/Not
-          on Ballot.&quot;
+          If the voter did not vote in the contest, select &quot;Blank
+          vote.&quot;
```
jonahkagan, line 36: Should we add instructions for the "Not on ballot" case? Maybe:

If this contest is not shown on this ballot, select "Not on ballot"

jonahkagan (approved):

client/src/components/DataEntry/BallotAudit.tsx
MorganLove: I'll do that, and if @mcchilders has a better suggestion we can do that in another PR.

PR #860, opened 2020-11-04
Opportunistic contests should not be required

### LINKED ISSUE #836: Should be able to leave opportunistic contests blank
When I get to the Opportunistic Contests screen, if I click "Save & Next", I see a bunch of validation errors for all the empty form fields. I would expect to be able to click "Save & Next" without entering anything into the form.

MorganLove:
**Description**

Closes #836 

- On Opportunistic Contests doesn't submit or trigger validation of the form if the fields aren't touched or match the existing values
- Don't do this on Targeted Contests, and keep current behavior

**Testing**

- Testing the utilities created for deep equals and empty object testing
- Adding tests for above behavior

**Progress**

- Done

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/AASetup/Contests/ContestForm.tsx
+++ b/client/src/components/MultiJurisdictionAudit/AASetup/Contests/ContestForm.tsx
@@ -1,6 +1,7 @@
 /* eslint-disable jsx-a11y/label-has-associated-control */
 import React from 'react'
 import { useParams } from 'react-router-dom'
+import equal from 'fast-deep-equal'
 import { Formik, FormikProps, Field, FieldArray } from 'formik'
 import { Spinner } from '@blueprintjs/core'
 import FormWrapper from '../../../Atoms/Form/FormWrapper'
@@ -25,7 +26,7 @@ import { IContest, ICandidate, IAuditSettings } from '../../../../types'
 import DropdownCheckboxList from './DropdownCheckboxList'
 import Card from '../../../Atoms/SpacedCard'
 import { testNumber } from '../../../utilities'
-import { isObjectEmpty, isObjectEqual } from '../../../../utils/objects'
+import isObjectEmpty from '../../../../utils/objects'
 
 interface IProps {
   isTargeted: boolean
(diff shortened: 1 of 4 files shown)


client/src/components/MultiJurisdictionAudit/AASetup/Contests/ContestForm.tsx
```diff
@@ -77,26 +78,43 @@ const ContestForm: React.FC<IProps> = ({
...
+    contests: filteredContests.length ? filteredContests : contestValues,
+  }
+
+  const isOpportunisticFormClean = (
+    touched: {},
+    values: { contests: IContest[] }
+  ) => {
+    return (
+      !isTargeted &&
+      isObjectEmpty(touched) &&
```
jonahkagan, line 91: Does this mean that if I click into a field and then click out of it without entering anything, it still won't let me move to the next screen? Could we be more permissive and let them skip this step as long as the actual data still equals the initial default empty values?

MorganLove: @jonahkagan It does mean that, yeah. If we changed it to an `||` we could make it like you're saying. I agree with you that that would be better.

MorganLove:
Also realized that we already have `fast-deep-equal` installed so the additional util isn't necessary. Oops, haha.

client/src/components/MultiJurisdictionAudit/AASetup/Contests/index.test.tsx
```diff
@@ -258,6 +258,66 @@ describe('Audit Setup > Contests', () => {
...
+        nextStage={nextStage}
+        prevStage={prevStage}
+      />
+    )
+
+    await findByText('Target Contests')
+    fireEvent.click(getByText('Save & Next'), { bubbles: true })
+    await waitFor(() => {
+      expect(nextStage.activate).toHaveBeenCalledTimes(0)
+    })
```
jonahkagan, line 294: will this check work? since you're checking to see if `nextStage.activate` has been called 0 times, is it possible that this check will pass before the button handler has time to get called? if so, then you might want to check something else - e.g. that the button is disabled or that an error message pops up. if you think this check will work, then i would remove the `await waitFor`, since that implies that there's some asynchronous action being waited for

jonahkagan (approved):

client/src/components/MultiJurisdictionAudit/AASetup/Contests/index.test.tsx
MorganLove: @jonahkagan Oh interesting, good catch. :thinking: It should bring up the error messages from the fields and I can use that.

Issue #850, opened 2020-11-02
Expire inactive user sessions


jonahkagan:
When a user is inactive for a while, we should automatically log them out (i.e. expire the session).

Implementation: set permanent session lifetime to 1 hour (https://flask.palletsprojects.com/en/1.1.x/config/#PERMANENT_SESSION_LIFETIME) and make sure the session refreshes on every request.

PR #864, opened 2020-11-05
Expire sessions

jonahkagan:
Task: #850 

Configures Flask to expire session cookies after 2 hours of inactivity. Also updates
the frontend to display a message when this happens (note: we just
display the message when the frontend gets a 401, which could also
happen from trying to access a route while not logged in for some other
reason, but this is the most common case so I figured why not).

### DIFF
--- a/client/src/App.test.tsx
+++ b/client/src/App.test.tsx
@@ -19,11 +19,7 @@ jest.unmock('react-toastify')
 const apiMocks = {
   failedAuth: {
     url: '/api/me',
-    response: {},
-    error: {
-      status: 401,
-      statusText: 'UNAUTHORIZED',
-    },
+    response: null,
   },
   abAuth: {
     url: '/api/me',
--- a/client/src/components/HomeScreen.tsx
+++ b/client/src/components/HomeScreen.tsx
@@ -5,8 +5,9 @@ import {
   RadioGroup,
   Radio,
   HTMLSelect,
+  Callout,
 } from '@blueprintjs/core'
-import { useHistory } from 'react-router-dom'
+import { useHistory, useLocation } from 'react-router-dom'
 import styled from 'styled-components'
 import { Formik, FormikProps, Field } from 'formik'
 import { useAuthDataContext } from './UserContext'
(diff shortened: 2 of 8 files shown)


client/src/App.test.tsx
```diff
@@ -19,11 +19,7 @@ jest.unmock('react-toastify')
 const apiMocks = {
   failedAuth: {
     url: '/api/me',
-    response: {},
-    error: {
-      status: 401,
-      statusText: 'UNAUTHORIZED',
-    },
+    response: null,
```
benadida, line 22: I'm a little bit irked by a null response vs an empty object... but I think I can live with it.

client/src/components/HomeScreen.tsx
```diff
@@ -72,9 +73,20 @@ const LoginWrapper = styled.div`
...
       <img height="50px" src="/arlo.png" alt="Arlo, by VotingWorks" />
+      {query.get('logged-out') && (
+        <Callout
+          icon="lock"
+          intent="warning"
+          style={{ margin: '20px 0 20px 0' }}
+        >
+          You have been logged out due to inactivity.
+        </Callout>
+      )}
```
benadida, line 89: suggest doing a 

`location.replace('/')`

so that clicking the back button later doesn't make that notification show back up, which could be confusing.

client/src/App.test.tsx
jonahkagan, line 22: this test was out of date

client/src/components/HomeScreen.tsx
benadida: oh also that would work only if you change `?logged_out=1` to `#loggedout` as a fragment, so that `location.replace` doesn't actually do a page reload.

server/tests/test_auth.py
```diff
@@ -305,6 +308,25 @@ def test_auth_me_not_logged_in(client: FlaskClient):
     assert json.loads(rv.data) is None
 
 
+def test_session_expiration(client: FlaskClient, aa_email: str):
+    original_session_lifetime = app.permanent_session_lifetime
+    assert original_session_lifetime > timedelta(minutes=1)
+    assert app.config["SESSION_REFRESH_EACH_REQUEST"] is True
```
benadida, line 314: where is this set? Does this mean tests fail if you don't have the right env var?

benadida (approved):
I'm good to go with this, my suggestions could be applied as a followup PR (especially given timing.)

client/src/App.test.tsx
jonahkagan: @benadida didn't you make this change? https://github.com/votingworks/arlo/pull/795

server/tests/test_auth.py
jonahkagan: this is a default flask config variable. i just added it here to make sure it didn't get changed, since we rely on its behavior. i can add an explanatory comment

client/src/App.test.tsx
benadida: Lol, I did. I did. I remember worrying about it and then saying screw it. And apparently I forgot about the "screw it" part.

Never mind.

server/tests/test_auth.py
benadida: oh got it. Yeah a comment would be great for the future.

client/src/components/HomeScreen.tsx
jonahkagan: updated f71da731

Issue #751, opened 2020-09-23
AA clicking on an in-progress audit should default to Audit Progress screen


jonahkagan:
Currently it goes to setup screen

PR #866, opened 2020-11-05
Default to audit progress screen once audit is launched

jonahkagan:
Task: #751 

Previously, we tried to use `isSetupComplete`, but this only tells us
if the audit is _ready_ to be launched, not whether it's actually
launched yet.

Also fix some lint errors.


MorganLove (approved):
LGTM!

PR #867, opened 2020-11-05
useJurisdictions return null while still loading

jonahkagan:
This is the pattern used by all our other data fetching hooks.
Previously, there was no way to check if the jurisdictions list was
empty or still loading, which meant that components were being rendered
that depended on the jurisdictions list before it was loaded.


MorganLove (approved):
Makes sense to me! Looks good. :)

PR #868, opened 2020-11-06
Implement Minerva ballot-polling audits for significantly greater efficiency

### LINKED ISSUE #803: Enable ballot-polling audits via the more efficient Minerva stopping rule, from the Athena class
As described at [The Athena Class of Risk\-Limiting Ballot Polling Audits](https://arxiv.org/abs/2008.02315), the Minerva stopping rule is specifically designed for the kinds of round-by-round audits that are common in risk-limiting election audits, since they leverage all the information available at the end of the round.

Adding support for Minerva can reduce the workload significantly for election officials compared to selection-ordered BRAVO. If a stopping probability of 90% is requested, the average reduction in the size of the first round is over 25% for close contests. The reduction is over 50% if the ballots are entered in offline mode, when contrasted with the corresponding end-of-round BRAVO stopping rule.

nealmcb:
**Description**

Implement Minerva ballot-polling audits for significantly greater efficiency.

Fixes #803

Adds more logging to the server, configurable $ARLO_LOGLEVEL, and allows use of fstrings in logging messages without complaints by pylint.

This PR contains a few other changes for the documentation as an aid to testers like me.

**Testing**

Note that the [Athena](https://github.com/filipzz/athena) module has to be installed from its github repo, as configured in Pipfile.

This code works fine so far with manual testing.

Doctests are used in the audit_math code to both illustrate the APIs and test functionality,
which can also be run via `python -m server.audit_math.minerva`

The external athena module which implements the Minerva algorithm has its own tests.

**Progress**

Round size calculation should now always be under 10 seconds or so.

The only slow parts now, for huge audits, are risk calculations. which for hundreds of thousands of ballots in a 2nd round can take over a minute on the first round for a 350000 ballot sample, and 90 seconds for 500,000, but is more like 10 seconds for 100,000. They can also be relatively slower in the second round for a given round size.
Web server timeouts, e.g. on Heroku, should take this into account.

On the horizon, we see a factor of 10x improvement coming for calculate_risk though!

Note that I suggest that the change in BRAVO for #884 (dcc801a726dae7da39fb5da85effc707ad812c40) should be moved to a point higher level in the app, e.g. in `sample_sizes.py`. Even a ballot-level comparison audit in a small jurisdiction can yield a sample size that is so large that a full hand count is the better choice.

Should be ready to merge now.

- [x] Fix test failures from test_minerva.py, 
- [x] 100% Test coverage
- [x] Incorporate latest athena package to resolve some rare numerical issues
- [x] Document performance characteristics

- [x] Add `$ARLO_MINERVA_MULTIPLE` environmental variable to allow tuning of future round sizes to handle various contingencies and dependencies on the initial value.
 Based on reviews of the paper and proofs to date, to be conservative in guaranteeing that the risk limit is achieved, for the time being we are not allowing future round sizes to be conditioned on the sample results to date. So the stopping probability for the first round can be selected as normal, but each round size after that is simply a multiple of the previous round size, using a provisional "MINERVA_MULTIPLE" of 1.5, which corresponds well to a 90% stopping probability for the next round when the first round uses the 90% option.  But for flexibility in a given audit situation, we suggest this step to allow that multiple to be changed. Though note that any deterministic round schedule would be ok.

- [x] Eliminate shim.py and move remaining functionality to minerva.py, to reduce clutter and clarity

Propose deferring to other PRs:
- [ ] make magic `"0.9"` stopping probability key for second and future rounds a defined constant for clarity.


.pylintrc
```diff
@@ -163,6 +163,7 @@ disable=print-statement,
         duplicate-code,
         broad-except,
         no-else-raise,
+        logging-fstring-interpolation,
```
umbernhard, line 166: we'd like this to happen in another PR. We're also in the process of setting up logging, so this all may change.

README.md
```diff
@@ -109,11 +109,17 @@ Rather than manually config the environment, you can also run the setup script d
 
 ### Creating Organizations and Administrators
 
-Organizations are, for example, the State of
```
umbernhard, line 112: See comment above. Unless the changes pertain specifically to Minerva, it'd be great if you could move them to a separate PR.

client/src/components/HomeScreen.tsx
```diff
@@ -208,9 +208,7 @@ const CreateAuditWrapper = styled.div`
   padding: 30px;
 `
 
-// TODO: remove "display: none;" when we implement Minerva
 const BallotPollingWrapper = styled.div`
-  display: none;
```
umbernhard, line 213: I think for now we're going to do this a little differently (@benadida is working on a configuration change). So this will probably get a little more complex.

docs/auth0.md
```diff
@@ -12,13 +12,13 @@ Auth0 is used for authentication. Key things to keep in mind:
 - we use two separate Auth0 tenants, one for audit administrators, one
   for jurisdiction administrators, each with its own single
   application, so we can use completely different login screens for
-  both, specifically 2FA for audit administrators and passwordless for
-  jurisdiction administrators.
+  both, specifically 2FA for audit administrators, and both 2FA
```
umbernhard, line 15: Same comment as for the README.

runtime.txt
```diff
@@ -0,0 +1 @@
+python-3.7.8
```
umbernhard, line 1: Is this file important? It's pegging a version of python that we don't use anywhere else. If you don't need it, could you please remove?

server/audit_math/bravo.py
```diff
@@ -122,6 +123,8 @@ def get_test_statistics(
                     Decimal((1 - winners[winner]["swl"][cand]) / 0.5) ** votes
                 )
 
+    logging.debug(f"bravo test_stats: T={T}")
```
umbernhard, line 126: See comment above about logging.

server/audit_math/minerva.py
```diff
@@ -1,11 +1,111 @@
...
 
+from athena.audit import Audit  # type: ignore
 from .sampler_contest import Contest
+from .shim import minerva_sample_sizes  # type: ignore
+
+# FIXME: make this an environmental variable
+MINERVA_MULTIPLE = 1.5
+
+
+def compute_cumulative_sample(sample_results):
```
umbernhard, line 26: I'm wondering if I should pull this up to the ballot_polling module so we don't have to repeat the code...

server/audit_math/minerva.py
```diff
@@ -1,11 +1,111 @@
...
+    Computes a cumulative sample given a round-by-round sample
+    """
+    cumulative_sample = defaultdict(int)
+    for rd in sample_results:
+        for cand in sample_results[rd]:
+            cumulative_sample[cand] += sample_results[rd][cand]
+    return cumulative_sample
+
+
+def make_arlo_contest(tally, num_winners=1, votes_allowed=1):
```
umbernhard, line 37: Is this function necessary? Is the constructor for `Contest` not sufficient?

server/audit_math/minerva.py
```diff
@@ -1,11 +1,111 @@
...
 """
-# TODO: remove the following once we are using the arguments
-# pylint: disable=unused-argument
-from typing import Dict, Tuple, Optional
+from decimal import Decimal
+from collections import defaultdict
+import logging
+from typing import List, Dict, Tuple, Optional
 
+from athena.audit import Audit  # type: ignore
```
umbernhard, line 18: For the sake of clarity, could you import this as `AthenaAudit` or something?

server/audit_math/shim.py
```diff
@@ -0,0 +1,130 @@
...
+        risk_limit      - the risk-limit for this audit
+        p_w             - the fraction of vote share for the winner
+        p_r             - the fraction of vote share for the loser / runner-up
+    """
+
+    # calculate the undiluted "two-way" share of votes for the winner
+    p_wr = p_w + p_r
+    p_w2 = p_w / p_wr
+
+    contest_ballots = 100000
```
umbernhard, line 30: I assume this code will go away once it gets integrated into minerva.py? Since we have the vote counts there, I don't see why you would need to keep fudging it like this

server/config.py
```diff
@@ -1,4 +1,5 @@
 import os
+import logging
```
umbernhard, line 2: See point above about logging.

umbernhard (changes requested):
We're getting there. I know you have some changes already planned, hopefully these aren't too bad.

.pylintrc
nealmcb: I have found the logging calls are very helpful for preserving a record of the audit and troubleshooting. f-strings, especially with the new `=` debugging sugar, are great for logging.  But lint complains unless this option is turned on.
So rather than removing or changing all the logging calls, I think it best to keep this one in to accommodate them.
See discussion at https://github.com/PyCQA/pylint/issues/2354#issuecomment-414526879 which doesn't even note the additional benefits provided by the f-string debugging syntax.

README.md
nealmcb: Yeah - it makes perfect sense to separate this out.

client/src/components/HomeScreen.tsx
nealmcb: Got it. Should I take it out then?

runtime.txt
nealmcb: Oops - thanks!

server/audit_math/minerva.py
nealmcb: This is actually being bypassed right now, and I actually may not need it at all in Minerva, when I redo this code.

server/audit_math/minerva.py
nealmcb: There are various circumstances (in doctests, in python notebooks, etc) where I find it handy to have a function that uses defaults and a bit of math to reduce the boilerplate necessary.
I guess another option would be to add something like this to the Contest class.

server/audit_math/minerva.py
nealmcb: Good idea.

server/audit_math/minerva.py
nealmcb: I pulled it out.

server/audit_math/minerva.py
umbernhard: Okay, that's fine then.

.pylintrc
umbernhard: I'm just worried that this (and the other logging calls) will interfere with our own logging deployment that's currently in the works. Could you at least comment out your logging code so we can be sure that won't happen?

client/src/components/HomeScreen.tsx
umbernhard: I think it's probably fine to leave in.

server/audit_math/shim.py
nealmcb: Yes - that will go away, but since we aren't actually using the code at this point except for the first round with no samples, I may just leave this for a future PR, if that's ok.

umbernhard:
@nealmcb btw I'm working on this over on #874 so I can rebase and setup a deploy.

.pylintrc
nealmcb: If you're worried about the logging config chunk in `config.py`, I can comment that out, and just uncomment locally for my own testing.But leaving the normal calls to log things in place, and the config that allows fstrings, wouldn't seem to affect your updates and deployment, and dealing with all of them during rebases would add burdens.
Unless there's something novel about the way you're deploying it, I don't see any conflicts. The minerva.py logging calls happen rarely, and actually the config changes I made help mask the more verbose logging in the athena module itself.

umbernhard:
@nealmcb if you can edit your Pipfile to be like this one, we can get this deployed so we can test it. 

https://github.com/votingworks/arlo/pull/875/files#diff-230078d672f10d17463a8a6265cad825b790885898256a3365be90685caac58dR43

nealmcb:
> @nealmcb if you can edit your Pipfile to be like this one, we can get this deployed so we can test it.
> 
> https://github.com/votingworks/arlo/pull/875/files#diff-230078d672f10d17463a8a6265cad825b790885898256a3365be90685caac58dR43

Done :)
And curious how the different syntax helps....

emergentview:
Done and passing (all but the client build, for whatever reason)

umbernhard:
@nealmcb I think the issue with the client tests is that your snapshots need to be regenerated. You can do that by running `make test-client`, running all the tests with `a` at the prompt, and then updating the snapshots with `u` once you get the prompt again. 

Also the branch is out of date again. I think we're getting really close here.

nealmcb:
Thanks.  I pushed a rebase.

I tried the tests, and they all succeeded, but I didn't see anything about updating - just watching.
ss-test-client-2020-11-10T13_47_57

umbernhard:
Well, it’s passing, so ¯\_(ツ)_/¯

server/audit_math/minerva.py
```diff
@@ -14,14 +100,193 @@ def get_sample_size(
...
+            "type": None,
+            "size": 4,
+            "prob": 1.0,
+        }
+
+        return samples
+
+    # If we haven't returned yet, actually do the math for the first round
+    for quant in quants:
+        size = minerva_sample_sizes(alpha, p_w, p_l, 0, 0, quant)
```
umbernhard, line 189: @nealmcb It occurs to me that maybe this isn't the right way to do this. Looking at the library example [here](https://github.com/filipzz/athena/blob/master/code/athena.ipynb) it seems like we could pass a list of quantiles to the Athena library directly, which might dramatically speedup sample size computation, since it won't have to recompute the whole distribution three to four times. Thoughts?

server/audit_math/shim.py
```diff
@@ -0,0 +1,130 @@
...
+
+    risk_limit = float(alpha)
+    p_w = float(p_w)  # type: ignore
+    p_r = float(p_r)  # type: ignore
+
+    # calculate the undiluted "two-way" share of votes for the winner
+    p_wr = p_w + p_r
+    p_w2 = p_w / p_wr
+
+    audit = make_election(risk_limit, p_w, p_r)  # type: ignore
```
umbernhard, line 102: So it looks like this is the major performance hit we're taking. Because we're calling this function 4 times per sample_size generation, we're rebuilding the distribution which is slow. So we ought to pass in a list of quantiles rather than calling this function once per quantile.

umbernhard (changes requested):

server/audit_math/shim.py
emergentview: I've pushed a much more straightforward version of the interface code, which uses a single AthenaAudit instance, rather than borrowing from the old BRAVO approach, and eliminates shim.py
But it turns out that there is very little speedup for the first round, which is the only time we're currently calling the code.

The big speedup comes from the concurrent change in athena 0.7.9 to use a very good approximation to the round size for margins under 1.5%.  Above that, on my machine, the total time is about 5 seconds, so Arlo / Heroku should not really notice much slowdown any more.

The slow parts now are risk calculations. which for hundreds of thousands of ballots in a 2nd round can take over a minute on the first round for a 350000 ballot sample, and 90 seconds for 500,000, but is more like 10 seconds for 100,000.  They are slower in the second round.

It looks like a factor of 10x improvement is on the way for calculate_risk though!

nealmcb: I've pushed a much more straightforward version of the interface code, which uses a single AthenaAudit instance, rather than borrowing from the old BRAVO approach, and eliminates shim.py
But it turns out that there is very little speedup for the first round, which is the only time we're currently calling the code.

The big speedup comes from the concurrent change in athena 0.7.9 to use a very good approximation to the round size for margins under 1.5%.  Above that, on my machine, the total time is about 5 seconds, so Arlo / Heroku should not really notice much slowdown any more.

The slow parts now are risk calculations. which for hundreds of thousands of ballots in a 2nd round can take over a minute on the first round for a 350000 ballot sample, and 90 seconds for 500,000, but is more like 10 seconds for 100,000.  They are slower in the second round.

It looks like a factor of 10x improvement is on the way for calculate_risk though!

server/audit_math/shim.py
nealmcb: Done

umbernhard:
Okay @nealmcb, in reviewing, I created a test election with a targeted contest with 100,000 votes, where one candidate got 60,000 and the other got 40,000. I also added an opportunistic contest with 5500, 4000, and 500 votes for the candidates. I got a sample size, and entered data for the first round as follows: 100 to 100 for the targeted contest, and 80, 20, 10 for the opportunistic. When I tried to advance to the next round, I got an error `'asn'`. Any idea what's going on here?

umbernhard (approved):
Looks good, pending a rebase onto main.

nealmcb:
Good - that makes sense. All looks good now. Thanks for all your help, Matt!

umbernhard:
@nealmcb looks like you’re out of date again, could you rebase one more time for me? I’ll get it merged in I promise

PR #873, opened 2020-11-09
Multiwinner fix

### LINKED ISSUE #871: BRAVO sample size fail on multi-winner contest: KeyError: '0.9'
Multi-winner contests fail in ballot-polling audits

To reproduce:

BRAVO 2-winner contest (400, 400, 200)

It sort of works for the first round, but you only get an ASN round size choice.

Then it fails with this at the beginning of the second round:

```Python
  File "/srv/s/electionaudits/arlo/arlo/server/api/rounds.py", line 673, in <dictcomp> 
    contest_id: options[sample_size_key]["size"] 
KeyError: '0.9' 
```

because the sample size only indicates the ASN, not the 90% stopping probability choice that is hardcoded into rounds.py

This is due to this code in bravo.get_sample_size:

```Python
    # For multi-winner, do nothing
    if contest.num_winners != 1:
        return {"asn": {"type": "ASN", "size": asn, "prob": None}}
```

it thus always only presents the ASN as an option.

The p-values look right for the results of the first round that I tried.

The Minerva code for the same case seems to work.

I suggest at the same time, making the required key for future rounds ('0.9' in this case) at least a defined constant referenced anywhere that needs it, to help coders avoid this sort of thing in other cases.

umbernhard:
**Description**
Adds a default return to sample size option selection in subsequent rounds, fixes #871.

**Testing**

Adds three new tests to check that round progression in multiwinner contests works. 

**Progress**

Ready.


server/api/rounds.py
```diff
@@ -515,6 +515,7 @@ def draw_sample_for_contest(contest: Contest, sample_size: int) -> List[BallotDr
     samples = [
         draw_sample_for_contest(contest, sample_sizes[contest.id])
         for contest in contests
+        if contest.is_targeted
```
jonahkagan, line 518: this filter is already applied to the input for this function. see L465

server/api/rounds.py
```diff
@@ -670,7 +671,13 @@ def create_round(election: Election):
...
         }[AuditType(election.audit_type)]
         sample_sizes = {
-            contest_id: options[sample_size_key]["size"]
+            contest_id: (
+                (sample_size_options[contest_id]).get(
+                    sample_size_key, sample_size_options[contest_id]["asn"]
+                )["size"]
+                if election.audit_type == AuditType.BALLOT_POLLING
+                else sample_size_options[contest_id][sample_size_key]["size"]
+            )
```
jonahkagan, line 680: I think this might be a little cleaner to read like this:

```python
sample_size_key = {
  AuditType.BALLOT_POLLING: "0.9" if "0.9" in sample_size_options[contest_id] else "asn",
  ...
}
```
and then the rest can stay the same as it was before

server/tests/api/test_rounds.py
```diff
@@ -83,6 +83,67 @@ def test_rounds_create_one(
     assert sorted(sampled_jurisdictions) == sorted(jurisdiction_ids[:2])
 
 
+def test_multiwinner_rounds_create_one(
```
jonahkagan, line 86: the pattern i've been using for special concerns like this one has been to make a new test file (e.g. `test_multiple_targeted_contests.py`. that way you can group all the tests for that special concern together and override the existing test fixtures (so you don't have to create a bunch of new fixture names). i'm hoping it will keep the tests easier to navigate and understand. what do you think about moving these new tests and fixtures to one file together?

server/tests/api/test_rounds.py
```diff
@@ -137,6 +198,60 @@ def test_rounds_create_two(
...
+    rv = client.get(
+        f"/api/election/{election_id}/jurisdiction/{jurisdiction_ids[0]}/round"
+    )
+    rounds = json.loads(rv.data)
+    compare_json(rounds, expected_rounds)
+
+    ballot_draws = SampledBallotDraw.query.filter_by(
+        round_id=rounds["rounds"][1]["id"]
+    ).all()
+    # Check that we automatically select the 90% prob sample size
```
jonahkagan, line 246: i thought the point here was to make sure we were falling back to the `asn` - is this comment just out of date?

in general, if the only thing we need to test is that the `asn` sample size gets picked for round 2 in the multiwinner case, i think this test case alone should suffice.

do you think there are other important things to test about multiwinner that made you want to add more test cases? (i'm not opposed to more tests, just want to understand what they are testing)

jonahkagan: i see that in multi-winner we are also expecting a different response for the first round sample sizes than in single-winner, which makes me think we should also have a test that confirms the sample size options that are returned in the first round

server/api/rounds.py
umbernhard: Will fix.

server/api/rounds.py
umbernhard: See I knew there was a fancy syntax way to do it, I just couldn't figure it out.

server/tests/api/test_rounds.py
umbernhard: Sure, that's reasonable.

jonahkagan (approved):

PR #878, opened 2020-11-10
Enable Minerva math in dev/staging but not production

jonahkagan:
Rather than implement a feature flag, we simply disable the Minerva math
selection in production. Since we don't really need a lot of
configurability, this is a quick solution for today.

Replaces #877


benadida (approved):
This shortcut feels weird to me because it is focused on our deployment, vs the other solution which might enable someone else doing a production deploy to use the new math.

I can live with it, but I think the other approach was better and not much more expensive.

jonahkagan:
@benadida oh that's a good point. my motivation was just avoiding complications, but i wasn't thinking about other ppl deploying it

Issue #876, opened 2020-11-09
Auth0 errors should not 500


jonahkagan:
https://sentry.io/organizations/votingworks/issues/2018789935/?referrer=slack

I don’t think it should cause a 500 and server crash, but rather should be returned to the user as a user error. Idea for a fix is to add a Flask error handler callback for OAuthErrors. Here’s the error log in Auth0: https://manage.auth0.com/dashboard/us/arlo-aa/logs/90020201109173331673000256466111263420581919314463948882

PR #879, opened 2020-11-10
Show Auth0 login errors to users

jonahkagan:
Task: #876 

If something goes wrong in the OAuth flow, Auth0 sends an error message
to the callback. We catch those errors and redirect the user to the
login screen, displaying the error.

Two changes to make this possible:
- Standardize the way we show pass messages to the login screen
- Change the auth tests to use the `patch` method of the mocking module,
which automatically clears the mocks out once the test is done
(previously, they were carrying over between tests)

### DIFF
--- a/client/src/components/utilities.ts
+++ b/client/src/components/utilities.ts
@@ -21,7 +21,11 @@ export const api = async <T>(
       // If we get a 401, it most likely means the session expired, so we
       // redirect to the login screen with a flag to show a message.
       if (response.status === 401) {
-        window.location.replace('/#logged-out')
+        window.location.replace(
+          encodeURI(
+            '/?error=unauthorized&message=You have been logged out due to inactivity.'
+          )
+        )
         return null
       }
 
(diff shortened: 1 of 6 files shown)


client/src/components/utilities.ts
```diff
@@ -21,7 +21,11 @@ export const api = async <T>(
       // If we get a 401, it most likely means the session expired, so we
       // redirect to the login screen with a flag to show a message.
       if (response.status === 401) {
-        window.location.replace('/#logged-out')
+        window.location.replace(
```
benadida, line 24: I should have been clearer with the guidance I meant to give on this. `window.location.replace` will replace the current history entry with the new one. For popup error messages, ideally you do a `replace` *after* you show the message to remove the error message from the history.. but if you're using a query parameter that's going to reload the page anyways.

So maybe we just live with the error message in the history, and do a `window.location = ` here cause `replace` is almost certainly not what we want.

benadida (approved):

PR #884, opened 2020-11-11
Support sampling all ballots

jonahkagan:
In large elections, if the targeted contest has a close margin, the
sample size may be so large that it would be more efficient to audit all
the ballots.

This change supports that case by:
- returning just one sample size option for all ballots
- not actually drawing a sample (unnecessary, plus spares the db)
- setting the audit to offline data entry
- tweaking some logic that relied on sampled ballots being in the db


server/api/rounds.py
```diff
@@ -399,34 +399,64 @@ def is_round_complete(election: Election, round: Round) -> bool:
...
-            .group_by(Jurisdiction.id, Contest.id)
-            .having(func.count(JurisdictionResult.result) == 0)
-            .count()
-        )
         return num_jurisdictions_without_results == 0
 
 
+def sampled_all_ballots(round: Round, election: Election):
+    return election.audit_type == AuditType.BALLOT_POLLING and any(
+        round_contest.sample_size >= round_contest.contest.total_ballots_cast
```
nealmcb, line 452: This code seems to set the threshold for "more efficient to audit all the ballots" at a sample size equal to `total_ballots_cast`. I think it is more likely to be a tenth of that or less, but the threshold varies by jurisdiction and circumstances. You might want to multiply by a fraction here which can be configured, e.g. via an environmental variable.

client/src/components/MultiJurisdictionAudit/Progress/JurisdictionDetail.tsx
```diff
@@ -138,24 +139,43 @@ const RoundStatusSection = ({
...
+    const jurisdictionStatus =
+      jurisdiction.currentRoundStatus && jurisdiction.currentRoundStatus.status
+
+    if (round.sampledAllBallots) {
+      if (jurisdictionStatus === JurisdictionRoundStatus.COMPLETE)
+        return <p>Data entry complete</p>
+      if (auditBoards.length === 0)
+        return <p>Waiting for jurisdiction to set up audit boards</p>
+      return (
+        <p>Auditing all {jurisdiction.ballotManifest.numBallots} ballots</p>
```
benadida, line 153: I'm seeing `jurisdiction.ballotManifest.numBallots` here, and `jurisdiction.numBallots` elsewhere. Is that right? Are these kept in sync?

server/api/reports.py
```diff
@@ -12,7 +12,11 @@
 from ..util.isoformat import isoformat
 from ..util.group_by import group_by
 from ..audit_math import supersimple, sampler_contest
-from ..api.rounds import cvrs_for_contest, sampled_ballot_interpretations_to_cvrs
+from ..api.rounds import (
```
benadida, line 15: curious, what do the parens do here in the `import` statement?

server/api/rounds.py
benadida: Agreed with the goal, but this is not where that happens. The decision to offer "all in" happens in `audit_math/bravo.py` based on a 25% threshold (hard-wired for now, we can configure it later), at which point the sample size selected is all the ballots, and then this code here triggers.

server/api/rounds.py
```diff
@@ -448,23 +478,34 @@ def draw_sample(election: Election, round: Round, sample_sizes: Dict[str, int]):
     )
 
     # Create RoundContest objects to include the contests in this round
-    for contest in contests_that_havent_met_risk_limit:
-        round_contest = RoundContest(
-            round_id=round.id,
-            contest_id=contest.id,
+    round.round_contests = [
```
benadida, line 481: Double checking that we don't need to `db_session.add(round)` after this?

server/api/rounds.py
```diff
@@ -448,23 +478,34 @@ def draw_sample(election: Election, round: Round, sample_sizes: Dict[str, int]):
...
     ]
 
+    # Special case: if we are sampling all ballots, we don't need to actually
+    # draw a sample. Instead, we force an offline audit.
+    if sampled_all_ballots(round, election):
+        if len(contests_to_sample) > 1:
+            raise Conflict(
+                "Cannot sample all ballots when there are multiple targeted contests."
+            )
+        election.online = False
```
benadida, line 506: no `db_session.add(election)` ?

Is this code supposed to happen on every round or just the first? I assume it can only happen on round 1, given that we're looking at everything but... just checking.

server/api/offline_results.py
```diff
@@ -64,12 +64,20 @@ def validate_offline_results(
...
             ballot_draws_by_contest.get(contest.id, 0)
             if contest.is_targeted
             else ballots_sampled
         )
+
+        # Special case: if we sampled all ballots, then the max results allowed
+        # should be based on the ballot manifest, since we don't have any
+        # sampled ballots in the db
+        if sampled_all_ballots(round, election):
+            num_ballots = jurisdiction.manifest_num_ballots
```
benadida, line 79: nit: overwriting the variable just computed, maybe cleaner to have an if/else statement here for this case and the case just prior.

server/auth/routes.py
```diff
@@ -93,7 +93,12 @@ def auth_me():
                 for org in user.organizations
             ],
             jurisdictions=[
-                {"id": j.id, "name": j.name, "election": serialize_election(j.election)}
+                {
+                    "id": j.id,
+                    "name": j.name,
+                    "election": serialize_election(j.election),
+                    "numBallots": j.manifest_num_ballots,
```
benadida, line 100: ahah, I see, this is where you're copying the `manifest.numBallots` to `jurisdiction.numBallots`, but still want to make sure you're confident in using both, as per my earlier comment.

benadida (approved):
a handful of questions to double check / confirm behavior, once checked, good to go.

client/src/components/MultiJurisdictionAudit/Progress/JurisdictionDetail.tsx
jonahkagan: - `jurisdiction.numBallots` is part of the JA /api/me response
- `jurisdiction.ballotManifest.numBallots` is part of the AA /jurisdiction response

So different endpoints for different user types. They come from the same db field, yes.

jonahkagan: If you think it makes more sense to return the whole `ballotManifest` object to the JA as well, I'm open to that. Adding just `numBallots` was the quickest/smallest change, but I didn't think it through deeply

server/api/reports.py
jonahkagan: i think it's just an autoformat thing

server/api/rounds.py
jonahkagan: doesn't seem like it? if the RoundContests didn't get created successfully, everything else downstream would break, so i think it's working

server/api/rounds.py
jonahkagan: Similar to above, I think when you've loaded a SQLAlchemy object, it's loaded into the session, so you don't need to re-add it to the session when you modify it. I think you only need to add new objects you're creating to the session.

I agree that there can only be one round in this case. But it wouldn't hurt for this particular line to run every round.

client/src/components/MultiJurisdictionAudit/RoundManagement/OfflineBatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,229 @@
...
+      enableReinitialize
+      onSubmit={async values => {
+        // TODO validation to prevent partially empty rows
+        // Omit empty rows
+        const filteredResults = values.results.filter(
+          result =>
+            result.batchName !== '' &&
+            Object.values(result.choiceResults).every(
+              value => (value as string | number) !== ''
+            )
```
MorganLove, line 46: @jonahkagan If you're wanting to omit empty rows, wouldn't it be better to test `.some(value => (value as string | number) !== '')`? That way you're just checking that at least one field has a value in it? Though if what you're wanting is to make sure they put in at least a '0' I think this would work, just that's not what's described in the comment, so I wanted to make sure.

client/src/components/MultiJurisdictionAudit/RoundManagement/OfflineBatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,229 @@
...
+                  ))}
+                </tr>
+              </thead>
+              <FieldArray
+                name="results"
+                render={arrayHelpers => (
+                  <tbody>
+                    {values.results.map((_, r) => (
+                      // eslint-disable-next-line react/no-array-index-key
+                      <tr key={`batchName-${r}`}>
```
MorganLove, line 79: @jonahkagan Why not use the `batchName` field on the result object here?

client/src/components/MultiJurisdictionAudit/RoundManagement/OfflineBatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,229 @@
...
+//         >
+//           {alreadySubmittedResults
+//             ? `Already Submitted Data for Round ${round.roundNum}`
+//             : `Submit Data for Round ${round.roundNum}`}
+//         </BottomButton>
+//       </form>
+//     )}
+//   </Formik>
+// )
+// }
```
MorganLove, line 227: @jonahkagan Remember to remove.

client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
```diff
@@ -20,13 +20,15 @@ import BatchRoundDataEntry from './BatchRoundDataEntry'
...
+import OfflineBatchRoundDataEntry from './OfflineBatchRoundDataEntry'
 
 const PaddedWrapper = styled(Wrapper)`
   flex-direction: column;
   align-items: flex-start;
-  width: 510px;
   padding: 30px 0;
 `
+// TODO
+// width: 510px;
```
MorganLove, line 31: @jonahkagan Confused about what's going on here.

client/src/components/MultiJurisdictionAudit/RoundManagement/OfflineBatchRoundDataEntry.tsx
MorganLove: @jonahkagan I recommend you use something like this:

```
                    {values.results.map((result, r) => (
                      <tr key={`batchName-${result.batchName + r}`}>
```
This means you won't get the error about identical batchnames when you create multiple empty rows before filling them in, and it also retains unique keys when the order changes, even when people use the same batch name for multiple batches (which they shouldn't but it needs to work even if they do).

client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
jonahkagan: the new form needs the full width. this width constraint was here to make the other views rendered in this wrapper more narrow so they would look better. we'll need to come back and style those views specifically at some point

benadida (approved):

PR #885, opened 2020-11-11
Make JA upload success more obvious

MorganLove:
**Description**

- Colors the callout for the status box green when all the needed files are successfully uploaded.

**Testing**

- I'm looking for the test that needs updating for this, since it didn't show any failing after the change.

**Progress**

- It works, just investigating tests.

### DIFF
--- a/client/src/components/Atoms/Form/_helpers.tsx
+++ b/client/src/components/Atoms/Form/_helpers.tsx
@@ -10,7 +10,10 @@ export const generateOptions = (count: number): JSX.Element[] => {
 }
 
 export const ErrorLabel = styled.p`
-  color: #ff0000;
+  color: #a82a2a;
+`
+export const SuccessLabel = styled.p`
+  color: #0a6640;
 `
 
 export default generateOptions
--- a/client/src/components/MultiJurisdictionAudit/CSVForm/index.tsx
+++ b/client/src/components/MultiJurisdictionAudit/CSVForm/index.tsx
@@ -10,7 +10,7 @@ import { ErrorLabel, SuccessLabel } from '../../Atoms/Form/_helpers'
 import FormSection, {
   FormSectionDescription,
 } from '../../Atoms/Form/FormSection'
-import { IFileInfo } from '../useJurisdictions'
+import { FileProcessingStatus, IFileInfo } from '../useJurisdictions'
 
 export const Select = styled(HTMLSelect)`
   margin-top: 5px;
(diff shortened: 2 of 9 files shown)


jonahkagan:
I think it's a little confusing still because the first place my eyes go is the "The audit has not started" in big bold green, which doesn't sound like a success message.

Screen Shot 2020-11-12 at 11 43 43 AM

Another pattern that might be worth exploring is just putting a message somewhere near the upload buttons when it uploads successfully, e.g.
Screen Shot 2020-11-12 at 11 48 19 AM

MorganLove:
@jonahkagan Yeah, the message sends mixed messages methinks. Maybe say "Ballot manifest successfully uploaded" below it?

We could actually change the CSVForm component to give a success indicator similarly to how it has the error indicator. That would put a green success message (like in your example) on every CSV file upload, which wouldn't be a bad thing imo.

jonahkagan:
@MorganLove yes i like the idea of adding it to CSVForm. then we get this upgrade for free everywhere we use that component

client/src/components/MultiJurisdictionAudit/CSVForm/index.tsx
```diff
@@ -121,6 +121,12 @@ const CSVFileForm = ({
                   {processing && processing.error && (
                     <ErrorLabel>{processing.error}</ErrorLabel>
                   )}
+                  {processing && processing.completedAt && (
+                    <SuccessLabel>
+                      Upload successfully completed at{' '}
+                      {new Date(`${processing.completedAt}Z`).toLocaleString()}!
+                    </SuccessLabel>
+                  )}
```
jonahkagan, line 129: this shows up in the error case as well:
Screen Shot 2020-11-17 at 9 44 31 AM

client/src/components/MultiJurisdictionAudit/StatusBox/index.tsx
```diff
@@ -40,9 +41,10 @@ const StatusBox: React.FC<IStatusBoxProps> = ({
   buttonLabel,
   onButtonClick,
   children,
+  success,
 }: IStatusBoxProps) => {
   return (
-    <Wrapper>
+    <Wrapper intent={success ? 'success' : 'none'} icon={null}>
```
jonahkagan, line 47: i think you should remove the status bar changes and just go with the upload success message, since it's still confusing in the way we previously discussed

MorganLove: Fair enough. :slightly_smiling_face:

MorganLove: I did leave the 'successfully' addition in the header, though it's not green anymore. I think that helps clarify things a bit more without being confusing.

client/src/components/Atoms/Form/_helpers.tsx
```diff
@@ -10,7 +10,10 @@ export const generateOptions = (count: number): JSX.Element[] => {
 }
 
 export const ErrorLabel = styled.p`
-  color: #ff0000;
+  color: #a82a2a;
+`
+export const SuccessLabel = styled.p`
+  color: #0a6640;
```
jonahkagan, line 16: Oh I meant to say on the previous review - i realized recently you can use the `Colors` object from Blueprint. So this could be `Colors.GREEN2` or something (not sure exactly which color to use). doesn't need to be changed necessarily but nice to know about i think

jonahkagan (approved):

PR #889, opened 2020-11-12
fix the deletion query in bulk upload of jurisdiction admins

benadida:
needs a regression test, adding it shortly.

### DIFF
--- a/server/util/jurisdiction_bulk_update.py
+++ b/server/util/jurisdiction_bulk_update.py
@@ -43,7 +43,11 @@ def bulk_update_jurisdictions(
     with session.begin_nested():
         # Clear existing admins.
         session.query(JurisdictionAdministration).filter(
-            Jurisdiction.election == election
+            JurisdictionAdministration.jurisdiction_id.in_(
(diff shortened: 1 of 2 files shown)


server/util/jurisdiction_bulk_update.py
```diff
@@ -43,7 +43,11 @@ def bulk_update_jurisdictions(
     with session.begin_nested():
         # Clear existing admins.
         session.query(JurisdictionAdministration).filter(
-            Jurisdiction.election == election
+            JurisdictionAdministration.jurisdiction_id.in_(
+                Jurisdiction.query.filter(Jurisdiction.election == election).values(
```
jonahkagan, line 47: maybe just `Jurisdiction.query.filter_by(election_id=election.id).values(Jurisdiction.id).subquery()` - that's our usual convention for filtering (so it's sure to work) and maybe the subquery makes it more efficient?

jonahkagan (approved):
what's the test situation for this? assuming you tested manually, but i definitely would like to add an automated test at some point

PR #890, opened 2020-11-12
Changelog for offline batch results

jonahkagan:
Record the before/after of each change to offline batch results, as well
as the user who changed it.


server/models.py
```diff
@@ -611,6 +611,21 @@ class OfflineBatchResult(BaseModel):
...
+class OfflineBatchResultChangelog(BaseModel):
+    user_id = Column(
+        String(200), ForeignKey("user.id", ondelete="cascade"), nullable=False,
+    )
+    jurisdiction_id = Column(
+        String(200), ForeignKey("jurisdiction.id", ondelete="cascade"), nullable=False,
+    )
+
+    before = Column(JSON)
+    after = Column(JSON)
```
benadida, line 624: should these be not null?

benadida (approved):

server/models.py
jonahkagan: 🤦

PR #891, opened 2020-11-12
Offline batch results form improvements

jonahkagan:
- Disable finalizing when form is unsaved
- Validate that total results don't exceed number of ballots in manifest (also defends against integer overflow errors)
- Don't allow saving results after finalizing
- Test and improve basic request validation

### DIFF
--- a/server/tests/test_sample_all_ballots.py
+++ b/server/tests/test_sample_all_ballots.py
@@ -457,19 +457,6 @@ def test_offline_batch_results_validation(
             [
                 {
                     "batchName": "Batch 1",
-                    "batchType": "bad type",
-                    "choiceResults": {
-                        choice["id"]: choice["numVotes"] / 4
-                        for choice in contest["choices"]
-                    },
-                }
-            ],
-            "'bad type' is not one of ['Absentee By Mail', 'Advance', 'Election Day', 'Provisional', 'Other']",
-        ),
-        (
-            [
-                {
-                    "batchName": "Batch 1",
                     "batchType": "Provisional",
                     "choiceResults": {
                         choice["id"]: choice["numVotes"] / 4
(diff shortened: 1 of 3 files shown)


server/tests/test_sample_all_ballots.py
```diff
@@ -382,3 +380,186 @@ def test_all_ballots_audit(
...
+                {
+                    "batchName": "Batch 1",
+                    "batchType": "bad type",
+                    "choiceResults": {
+                        choice["id"]: choice["numVotes"] / 4
+                        for choice in contest["choices"]
+                    },
+                }
+            ],
+            "'bad type' is not one of ['Absentee By Mail', 'Advance', 'Election Day', 'Provisional', 'Other']",
```
umbernhard, line 467: Looks like this one got duplicated?

server/tests/test_sample_all_ballots.py
```diff
@@ -382,3 +380,186 @@ def test_all_ballots_audit(
...
+                {
+                    "batchName": "Batch 1",
+                    "batchType": "Provisional",
+                    "choiceResults": {
+                        choice["id"]: choice["numVotes"] / 4
+                        for choice in contest["choices"][:1]
+                    },
+                }
+            ],
+            "Invalid choice ids for batch Batch 1",
```
umbernhard, line 480: Do we want a more informative error here? "All choices must have total (even if it is 0)"?

server/tests/test_sample_all_ballots.py
```diff
@@ -382,3 +380,186 @@ def test_all_ballots_audit(
...
+                {
+                    "batchName": "Batch 1",
+                    "batchType": "Provisional",
+                    "choiceResults": {
+                        choice["id"]: choice["numVotes"]
+                        for choice in contest["choices"]
+                    },
+                }
+            ],
+            "Total results (2000000) cannot be greater than the number of ballots reported in the ballot manifest (1000000)",
```
umbernhard, line 503: I don't think we want to enforce this. If the manifest undercounted ballots, the audit should be able to deal with that.

umbernhard (changes requested):

server/tests/test_sample_all_ballots.py
jonahkagan: users should never see this - the frontend doesnt allow it. it's just defense

MorganLove:
LGTM!

MorganLove (approved):
Approving the front end changes.

server/tests/test_sample_all_ballots.py
jonahkagan: fixed

umbernhard (approved):

PR #892, opened 2020-11-12
Fix JA status box when all ballots sampled

jonahkagan:
Basically decided to just bail for now since we don't have great status
info available, nor a good audit report (it's currently empty).


MorganLove (approved):
Makes sense.

PR #894, opened 2020-11-12
Offline batch results sticky headers

jonahkagan:
ezgif com-gif-maker


MorganLove (approved):

MorganLove:
@jonahkagan looks good! Just remember to do an interactive rebase to leave off the first two commits that were included in #893

PR #896, opened 2020-11-12
add whitelist for Cloudflare servers

danwallach:
**Description**

This PR contains a list of the IP address ranges used by Cloudflare and the necessary pre-request hook for Flask to reject requests coming from anywhere other than Cloudflare.

**Testing**

We're going to need to try this out on the staging server. I've dorked around with adding and removing `127.0.0.0/8` and other such things to the whitelist for local testing, and it seems to work.

### DIFF
--- a/server/web_whitelist.py
+++ b/server/web_whitelist.py
@@ -7,12 +7,10 @@ from ipaddress import (
     ip_address,
 )
 from typing import List, Union
-
 from flask import request, abort
-
-# list of Cloudflare IPs: https://www.cloudflare.com/ips/
 from .app import app
 
+# list of Cloudflare IPs: https://www.cloudflare.com/ips/
 cloudflare_ip_list = [
     # V4 address blocks
     "173.245.48.0/20",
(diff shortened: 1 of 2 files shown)


server/web_whitelist.py
```diff
@@ -0,0 +1,57 @@
...
+    IPv4Address,
+    IPv6Address,
+    ip_address,
+)
+from typing import List, Union
+
+from flask import request, abort
+
+# list of Cloudflare IPs: https://www.cloudflare.com/ips/
+from server.app import app
```
danwallach, line 14: Is this the right way to reference the "real" Flask app?

server/web_whitelist.py
```diff
@@ -0,0 +1,57 @@
...
+            return True
+    return False
+
+
+# https://stackoverflow.com/questions/24222220/block-an-ip-address-from-accessing-my-flask-app-on-heroku
+@app.before_request
+def block_method():
+    ip = request.environ.get("REMOTE_ADDR")
+    if not addr_is_whitelisted(ip_address(ip)):
+        abort(403)  # "Forbidden"
```
umbernhard, line 57: Is there any way to add a more specific message about why we're serving a 403? Speaking from an internet measurement perspective, it might be nice to do so if we can (helps with this kind of stuff: https://ensa.fi/papers/403forbidden_imc18.pdf)

umbernhard (changes requested):
It looks like we need to explicitly allow the testing env IPs in addition to CF. I'm not sure if it's just a matter of adding localhost or if it's more involved than that, but that's why the tests are failing.

jonahkagan:
Closing to clean up long-standing open PRs

PR #897, opened 2020-11-13
Strips newline characters from CVRs and standardized contests

umbernhard:
**Description**

Strips newline characters out of CVRs and standardized contest files.

**Testing**

Adds testing of files with newlines in them. 

**Progress**

Tests pass?

### DIFF
--- a/server/api/cvrs.py
+++ b/server/api/cvrs.py
@@ -68,7 +68,9 @@ def process_cvr_file(session: Session, jurisdiction: Jurisdiction, file: File):
 
         # Parse out all the initial metadata
         _election_name = next(cvrs)[0]
-        contest_row = next(cvrs)
+        contest_row = [
+            contest.replace("\r", "").replace("\n", "  ") for contest in next(cvrs)
+        ]
         first_contest_column = next(
             c for c, value in enumerate(contest_row) if value != ""
         )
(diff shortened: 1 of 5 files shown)


server/api/cvrs.py
```diff
@@ -68,7 +68,9 @@ def process():
 
         # Parse out all the initial metadata
         _election_name = next(cvrs)[0]
-        contest_row = next(cvrs)
+        contest_row = [
+            contest.replace("\r", "").replace("\n", "  ") for contest in next(cvrs)
```
jonahkagan, line 72: if you wanted to do it in one function call, you could do `re.sub(r"\r*\n+", " ", contest)`

jonahkagan: actually, a better option might be `" ".join(contest.splitlines())` because splitlines is very smart about handling all types of newlines

jonahkagan (commented):
def add tests!

server/api/cvrs.py
umbernhard: oooh that's a great suggestion

server/api/cvrs.py
```diff
@@ -68,7 +68,7 @@ def process():
 
         # Parse out all the initial metadata
         _election_name = next(cvrs)[0]
-        contest_row = next(cvrs)
+        contest_row = [" ".join(contest.splitlines()) for contest in next(cvrs)]
```
jonahkagan, line 71: the only suggestion i have is to pull this logic out into a utility function - maybe `util.remove_newlines` or something like that. that way it would be more clear what's happening here. but it's not blocking

jonahkagan (approved):

PR #899, opened 2020-11-13
Offline batch results: assorted fixes

jonahkagan:
- max length of 200 characters for batch names
- wrap words in cells
- use a timezone for finalized_at
- disable finalizing when no results recorded


MorganLove (approved):
Approving front end

PR #902, opened 2020-11-14
Add per-batch-results API for parallel data entry

benadida:
backend ready, frontend needed.

### DIFF
--- a/server/api/offline_batch_results.py
+++ b/server/api/offline_batch_results.py
@@ -186,6 +186,89 @@ def record_offline_batch_results(
 
 
 @api.route(
+    "/election/<election_id>/jurisdiction/<jurisdiction_id>/round/<round_id>/results/batch/",
+    methods=["POST"],
+)
+@restrict_access([UserType.JURISDICTION_ADMIN])
+def record_offline_batch_results_add(
+    election: Election,  # pylint: disable=unused-argument
+    jurisdiction: Jurisdiction,
+    round: Round,
+):
+    batch_result = request.get_json()
+    validate_offline_batch_results(election, jurisdiction, round, [batch_result])
+
+    for contest_choice_id, result in batch_result["choiceResults"].items():
+        db_session.add(
+            OfflineBatchResult(
+                jurisdiction_id=jurisdiction.id,
+                batch_name=batch_result["batchName"],
+                batch_type=batch_result["batchType"],
+                contest_choice_id=contest_choice_id,
+                result=result,
+            )
+        )
+
+    db_session.commit()
+
+    return jsonify(status="ok")
+
+
+@api.route(
+    "/election/<election_id>/jurisdiction/<jurisdiction_id>/round/<round_id>/results/batch/<batch_name>",
+    methods=["PUT"],
+)
+@restrict_access([UserType.JURISDICTION_ADMIN])
+def record_offline_batch_results_update(
+    election: Election,  # pylint: disable=unused-argument
+    jurisdiction: Jurisdiction,
+    round: Round,
+    batch_name: str,
+):
+    batch_result = request.get_json()
+    validate_offline_batch_results(election, jurisdiction, round, [batch_result])
+
+    for contest_choice_id, result in batch_result["choiceResults"].items():
+        new_batch_result = OfflineBatchResult.query.filter_by(
+            jurisdiction_id=jurisdiction.id,
+            batch_name=batch_name,
+            contest_choice_id=contest_choice_id,
+        ).one_or_none()
+
+        if not new_batch_result:
+            raise Conflict("this batch has been deleted")
+
+        new_batch_result.batch_name = batch_result["batchName"]
+        new_batch_result.batch_type = batch_result["batchType"]
+        new_batch_result.result = result
+        db_session.add(new_batch_result)
+
+    db_session.commit()
+
+    return jsonify(status="ok")
+
+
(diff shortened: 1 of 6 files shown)


server/api/offline_batch_results.py
```diff
@@ -185,6 +185,89 @@ def record_offline_batch_results(
...
+@restrict_access([UserType.JURISDICTION_ADMIN])
+def record_offline_batch_results_add(
+    election: Election,  # pylint: disable=unused-argument
+    jurisdiction: Jurisdiction,
+    round: Round,
+):
+    offline_batch_results = request.get_json()
+    validate_offline_batch_results(election, jurisdiction, round, offline_batch_results)
+
+    for batch_results in offline_batch_results:
```
jonahkagan, line 201: shouldn't this endpoint just take one batch at a time?

server/api/offline_batch_results.py
```diff
@@ -185,6 +185,89 @@ def record_offline_batch_results(
     return jsonify(status="ok")
 
 
+@api.route(
+    "/election/<election_id>/jurisdiction/<jurisdiction_id>/round/<round_id>/results/batch/",
```
jonahkagan, line 189: im gonna let this trailing slash slide this time ;)

server/api/offline_batch_results.py
```diff
@@ -185,6 +185,89 @@ def record_offline_batch_results(
...
+    election: Election,  # pylint: disable=unused-argument
+    jurisdiction: Jurisdiction,
+    round: Round,
+    batch_name: str,
+):
+    offline_batch_results = request.get_json()
+    validate_offline_batch_results(election, jurisdiction, round, offline_batch_results)
+
+    # there should be only one
+    batch_results = offline_batch_results[0]
```
jonahkagan, line 233: maybe just take a single object instead of an array then?

server/api/offline_batch_results.py
```diff
@@ -185,6 +185,89 @@ def record_offline_batch_results(
...
+                )
+            )
+
+    db_session.commit()
+
+    return jsonify(status="ok")
+
+
+@api.route(
+    "/election/<election_id>/jurisdiction/<jurisdiction_id>/round/<round_id>/results/batch/<batch_name>",
```
jonahkagan, line 219: what if we just give each batch a uuid so editing batch names gets easier?

server/api/offline_batch_results.py
```diff
@@ -185,6 +185,89 @@ def record_offline_batch_results(
...
+@restrict_access([UserType.JURISDICTION_ADMIN])
+def record_offline_batch_results_delete(
+    election: Election,  # pylint: disable=unused-argument
+    jurisdiction: Jurisdiction,
+    round: Round,  # pylint: disable=unused-argument
+    batch_name: str,
+):
+    OfflineBatchResult.query.filter_by(
+        jurisdiction_id=jurisdiction.id, batch_name=batch_name
+    ).delete()
```
jonahkagan, line 264: maybe do a find first to make sure it's actually there, error if not

use case: somebody tries to delete a batch that's already been deleted by somebody else

server/api/offline_batch_results.py
```diff
@@ -185,6 +185,89 @@ def record_offline_batch_results(
...
+    "/election/<election_id>/jurisdiction/<jurisdiction_id>/round/<round_id>/results/batch/<batch_name>",
+    methods=["PUT"],
+)
+@restrict_access([UserType.JURISDICTION_ADMIN])
+def record_offline_batch_results_update(
+    election: Election,  # pylint: disable=unused-argument
+    jurisdiction: Jurisdiction,
+    round: Round,
+    batch_name: str,
+):
```
jonahkagan, line 228: do we need to protect against two simultaneous edits to the same batch?

server/api/offline_batch_results.py
```diff
@@ -185,6 +185,89 @@ def record_offline_batch_results(
     return jsonify(status="ok")
 
 
+@api.route(
```
jonahkagan, line 188: idea for forcing users to update: keep the existing endpoint above but make it just return an error message that says "Arlo has been updated. Please refresh your browser for the latest version."

benadida: I expect some counties will need to parallelize 4-5x, constantly entering data. We could literally see livelock. So no, I think we have to allow for truly parallel addition.

server/api/offline_batch_results.py
benadida: lol can't get it past you.

server/api/offline_batch_results.py
benadida: it should, but I was trying to keep it simple so I could reuse your validation. I'll look at simplifying.

server/api/offline_batch_results.py
benadida: maybe as a future improvement but... I *really* don't want to modify the data model right now.

server/api/offline_batch_results.py
benadida: No, if two desks update the same batch at the same time, I'm comfortable with last-write wins.

server/api/offline_batch_results.py
benadida: seems okay for the second user to just succeed without an error though?

jonahkagan (approved):

PR #903, opened 2020-11-14
Add UTC timezone to all timestamps

jonahkagan:
Because we store timestamps with no timezone info in the db, they were getting read from the db without timezone info, and therefore getting misinterpreted as local time by the frontend (instead of UTC time).

Here we make a few changes to fix this without modifying the database schema or changing timezones already in the database:
- Change everywhere we create datetimes from `datetime.utcnow` (which doesn't have timezone info) to use `datetime.now(timezone.utc)` (which does)
- Require that every datetime written to the db has timezone info
- When reading timezones from the db, set the timezone info to UTC

The latter two changes are accomplished with a custom sqlalchemy column type.

Lastly, we remove the workaround on the frontend that was appending `Z` to the timestamps to interpret them as UTC.


benadida (approved):
looks good, but let's wait to merge until after the current big audit, just in case.

PR #905, opened 2020-11-14
Fix: can't edit batch name to existing batch name

jonahkagan:
Also add batch name to edit modal title


benadida (approved):
wrong value for batch name in edit batch, otherwise looks good.

benadida (approved):

PR #908, opened 2020-11-16
Fix display of finalized timestamp

jonahkagan:
We store UTC timestamps in the db, but they don't have timezone info attached. This changes the frontend to interpret the timestamps with the UTC timezone.

Note that there's an additional bug: even though we create our timestamps with the UTC timezone, the way they get stored in the database actually depends on the local timezone of the servers. It appears that Heroku has UTC as the local timezone, whereas my machine has PT:

On Heroku:
```
>>> datetime.now().isoformat()
'2020-11-16T22:10:24.927692'
```

On my machine:
```
>>> datetime.now().isoformat()
'2020-11-16T14:10:24.927692'
```

We should probably resolve this at the database level eventually.


benadida:
might be worth doing a 

`SET TIME ZONE 'UTC';`

in our Postgres connection setup.

benadida (approved):

PR #909, opened 2020-11-16
Don't end round automatically after last jurisdiction finalizes

jonahkagan:
Add a script to do it instead.


benadida (approved):
nice, great touch having the actual script :)

PR #911, opened 2020-11-17
Fix offline batch results copy/paste

jonahkagan:
The empty table header above the edit buttons wasn't getting copied, so
the header row was one shorter than the body rows, which meant the
headers were off by one when pasting the results into a spreadsheet.

So we just exclude the whole column from copy/paste.

Also adds a "Copy to clipboard" button

ezgif com-gif-maker (3)

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -86,6 +86,7 @@
     "@types/styled-components": "^5.1.2",
     "@types/uuidv4": "^2.0.0",
     "@types/yup": "^0.26.22",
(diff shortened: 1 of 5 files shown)


benadida (approved):
nice. While we're making a change... why don't we add a button that says "copy to clipboard" so it's an easy click and it's discoverable?

jonahkagan:
@benadida added!

benadida (approved):

PR #912, opened 2020-11-18
Visual PDF snapshots

jonahkagan:
Use pdf-visual-diff to create image snapshots for label/placeholder
pdfs.


client/src/components/MultiJurisdictionAudit/RoundManagement/generateSheets.test.tsx
```diff
@@ -33,64 +33,54 @@ describe('generateSheets', () => {
...
+      const pdf = downloadLabels(
         1,
         dummyBallots.ballots,
         mockJurisdiction.name,
         mockJurisdiction.election.auditName
       )
-      const deterministicPDF = pdf
-        .replace(/CreationDate \([^)]+\)/g, '') // remove the timestamp
-        .replace(/ID \[[^\]]+\]/g, '') // remove the unique id
-      expect(deterministicPDF).toMatchSnapshot()
```
MorganLove, line 46: @jonahkagan Is it no longer including the id and creation timestamp when creating the pdf?

jonahkagan: they aren't visible in the pdf, so we don't need to remove them anymore. they're probably still in the file metadata, but we dont test that

MorganLove: Aaaah, that's awesome!! Only testing what actually matters is so awesome. :D

MorganLove (approved):

PR #913, opened 2020-11-18
Add container/tabulator to labels and placeholders

jonahkagan:
For ballot comparison audits


client/src/components/MultiJurisdictionAudit/RoundManagement/generateSheets.ts
```diff
@@ -2,42 +2,55 @@ import jsPDF from 'jspdf'
 import { IAuditBoard } from '../useAuditBoards'
 import { IBallot } from './useBallots'
 
+// Label constants in points
```
jonahkagan, line 5: accidentally measured these in points instead of millimeters and didnt feel like going back to remeasure

MorganLove (approved):

PR #917, opened 2020-11-19
Run server tests in parallel

jonahkagan:


umbernhard (approved):
:raised_hands: :raised_hands: :raised_hands:

PR #923, opened 2020-11-23
Improve launching audit UX

jonahkagan:
- Fix a bug on the audit progress screen where it would crash right
after launching the audit since the jurisdiction data hadn't been
refreshed yet (so currentRoundStatus wasn't populated yet)
- Add a spinner to the launch button until the request finishes
- Change the "Close without launching" button to just say "Cancel" (more
conventional language) and disable it while launching


MorganLove (approved):
LGTM, just need to update the test to reflect the wording change to "cancel"

PR #924, opened 2020-11-23
Sampler manifest from CVR

jonahkagan:
In ballot comparison audits, we should only allow the sampler to choose
from ballots that had a result recorded in the CVR for the targeted
contest. So we load the CvrBallots for all the targeted contests
jurisdictions, and then parse out whether or not each ballot had a
result recorded for each contest when building the sampler manifest.


server/api/rounds.py
```diff
@@ -528,22 +564,47 @@ def draw_sample_for_contest(contest: Contest, sample_size: int) -> List[BallotDr
...
-        # (deterministic real world ids) instead of the jurisdiction and batch
-        # ids (non-deterministic uuids that we generate for each audit).
-        manifest = {
-            (jurisdiction.name, batch.tabulator, batch.name): list(
-                range(1, batch.num_ballots + 1)
-            )
-            for jurisdiction in contest.jurisdictions
-            for batch in jurisdiction.batches
-        }
+        if election.audit_type == AuditType.BALLOT_POLLING:
```
umbernhard, line 567: Do we have to worry about batch comparison too?

server/api/rounds.py
```diff
@@ -518,6 +519,41 @@ def sample_ballots(
...
+    # inputs. So the sampler expects the same input for the same real world
+    # data. Thus, we use the jurisdiction name and batch keys
+    # (deterministic real world ids) instead of the jurisdiction and batch
+    # ids (non-deterministic uuids that we generate for each audit).
+    batch_id_to_key = {
+        batch.id: (jurisdiction.name, batch.tabulator, batch.name)
+        for jurisdiction in participating_jurisdictions
+        for batch in jurisdiction.batches
+    }
+    batch_key_to_id = {
```
umbernhard, line 536: Is there not a builtin that does this for you? Seems a little clunky, but I'm not sure.

umbernhard (approved):
Looks good so far!

server/api/rounds.py
jonahkagan: not that i could find

server/api/rounds.py
jonahkagan: it's handled in a separate `sample_batches` function

server/api/rounds.py
```diff
@@ -518,6 +519,41 @@ def sample_ballots(
     contests: List[Contest],
     sample_sizes: Dict[str, int],
 ):
+    participating_jurisdictions = {
+        jurisdiction for contest in contests for jurisdiction in contest.jurisdictions
```
benadida, line 523: this line is freaking me out, but I think I get it.

benadida (approved):

PR #926, opened 2020-11-23
Using ID instead of index for contest selection

### LINKED ISSUE #925: Contest selection table does not maintain proper selection when filtered
If you select a contest and then filter so that the indexes change, a different contest will show as selected because it's at the index of the original selection.

MorganLove:
**Description**

- Using the index broke when filtering the contests because the indexes changed.

Closes #925 

**Testing**

- We don't have any tests for the contest selection page on ballot comparison audits. That needs to be corrected, but we also need to get this fix out, too.

**Progress**

- Tested manually and it works. Could use more manual confirmation.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -63,7 +63,7 @@
     "coverageThreshold": {
       "global": {
         "statements": 83,
-        "branches": 77,
+        "branches": 76,
         "functions": 81,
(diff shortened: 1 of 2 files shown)


jonahkagan:
Closed by #955

PR #936, opened 2020-11-25
Fix sample size options minerva

umbernhard:
**Description**
Fixes escalation for Minerva, where there is no `asn` value but there is a `0.9` value. 

**Testing**

Adds new tests for the rounds api to check that Minerva works as expected. 

**Progress**

Ready.


server/tests/ballot_polling/conftest.py
```diff
@@ -0,0 +1,17 @@
...
+@pytest.fixture
+def election_id(client: FlaskClient, org_id: str, request):
+    set_logged_in_user(client, UserType.AUDIT_ADMIN, DEFAULT_AA_EMAIL)
+    return create_election(
+        client,
+        audit_name=f"Test Audit {request.node.name}",
+        audit_type=AuditType.BALLOT_POLLING,
+        audit_math_type=AuditMathType.MINERVA,
+        organization_id=org_id,
+    )
```
jonahkagan, line 17: is there a reason to have a conftest.py here? since you only have one test file (and it's specific to minerva) i think it would be simpler to just move that fixture there

server/tests/ballot_polling/test_minerva_ballot_polling.py
```diff
@@ -0,0 +1,335 @@
...
+            ],
+            "totalBallotsCast": 1000,
+            "numWinners": 1,
+            "votesAllowed": 1,
+            "jurisdictionIds": jurisdiction_ids,
+        },
+    ]
+    rv = put_json(client, f"/api/election/{election_id}/contest", contests)
+    assert_ok(rv)
+    return [str(c["id"]) for c in contests]
```
jonahkagan, line 41: does these two fixtures need to be overridden? seems like it would be fine to just use the default settings and contests

server/tests/ballot_polling/test_minerva_ballot_polling.py
```diff
@@ -0,0 +1,335 @@
...
+    snapshot.assert_match(
+        {
+            f"{result.contest.name} - {result.contest_choice.name}": result.result
+            for result in RoundContestResult.query.filter_by(round_id=round_2_id).all()
+        }
+    )
+
+    set_logged_in_user(client, UserType.AUDIT_ADMIN, DEFAULT_AA_EMAIL)
+    rv = client.get(f"/api/election/{election_id}/report")
+    assert_match_report(rv.data, snapshot)
```
jonahkagan, line 335: these tests seem fine, but i would recommend trimming them down to only test the things that are actually different about minerva audits. e.g. i don't think we need to test whether the round is over or not after only one jurisdiction records results, since that logic doesn't have anything to do with minerva. same with the round contest results.

conversely, i do think we should have a test for the sample size options, since that does rely on minerva math.

benadida (approved):
I don't know enough about Minerva to tell if the tests are truly complete, so good to get @jonahkagan's review, too. On structure and fix, I'm good with it.

server/tests/ballot_polling/conftest.py
umbernhard: Okay, can do. I was trying to preserve as much of the naming as possible so I could be lazy. I was trying to follow the examples of ballot/batch comparison, but I can push it up.

server/tests/ballot_polling/test_minerva_ballot_polling.py
umbernhard: I will check. If I recall I was getting some errors from Minerva when there were two contests.

server/tests/ballot_polling/test_minerva_ballot_polling.py
umbernhard: Okay, I can work on fixing that.

umbernhard:
@jonahkagan Filip fixed the problem that was causing my tests to fail, so this is ready for re-review. I had to upgrade the version of the Athena library we're using, and that affected the text in an error message that I had to account for.

server/tests/ballot_polling/test_minerva_ballot_polling.py
```diff
@@ -0,0 +1,358 @@
...
+    manifests,  # pylint: disable=unused-argument
+    snapshot,
+):
+    set_logged_in_user(client, UserType.AUDIT_ADMIN, DEFAULT_AA_EMAIL)
+    rv = client.get(f"/api/election/{election_id}/sample-sizes")
+    assert rv.status_code == 200
+
+    sample_size_options = json.loads(rv.data)["sampleSizes"][contest_ids[0]]
+    assert len(sample_size_options) == 3
+    print(sample_size_options)
```
jonahkagan, line 47: remove

jonahkagan (approved):

PR #938, opened 2020-11-25
Test fixes

MorganLove:
**Description**

- re-enabling tests

**Testing**

- Homescreen and timers

**Progress**

- ready to go


client/src/components/utilities.ts
```diff
@@ -50,7 +50,7 @@ export const poll = (
   condition: () => Promise<boolean>,
   callback: () => void,
   errback: (arg0: Error) => void,
-  timeout: number = 120000,
+  timeout: number = 120000, // should we lengthen this since we tend to poll every minute now?
```
jonahkagan, line 53: not sure i understand this comment - did you want to address it?

jonahkagan (approved):

client/src/components/utilities.ts
MorganLove: @jonahkagan Oh yes, I put that in there to remind me to ask about it, then forgot about it, lol. So, on reflection, we're actually not even using `poll` where I was thinking of when I made the comment, since we moved to `useInterval` there. So we can disregard this. :grin:

PR #939, opened 2020-11-30
Expire session based on inactivity or lifetime

jonahkagan:
Since Flask's permanent session feature only supports one timeout, we build our own expiration scheme on top of the session.

The session has two fields, _created_at (set on login) and _last_request_at (set on request, only if authorization is successful). We check these fields and corresponding timeouts every request when authorizing.

There's a small wrinkle for superadmins: since they can log in as other users, we ensure _created_at doesn't get set when they log in as other users, since that would allow someone with a superadmin cookie to keep the cookie alive by logging in as other users repeatedly.

We also update the test helpers for auth to use the actual session manipulation code rather than try to replicate it, since it's getting more complex.

### DIFF
--- a/server/auth/lib.py
+++ b/server/auth/lib.py
@@ -32,17 +32,6 @@ _CREATED_AT = "_created_at"
 _LAST_REQUEST_AT = "_last_request_at"
 
 
-def is_session_expired(session):
-    return (
-        datetime.now(timezone.utc)
-        > datetime.fromisoformat(session[_CREATED_AT]) + config.SESSION_LIFETIME
-    ) or (
-        datetime.now(timezone.utc)
-        > datetime.fromisoformat(session[_LAST_REQUEST_AT])
-        + config.SESSION_INACTIVITY_TIMEOUT
-    )
-
-
 def set_loggedin_user(
     session, user_type: UserType, user_key: str, from_superadmin: bool = False
 ):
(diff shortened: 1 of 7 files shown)


server/auth/lib.py
```diff
@@ -26,21 +28,47 @@ class UserType(str, enum.Enum):
...
+
+
+def set_loggedin_user(
+    session, user_type: UserType, user_key: str, from_superadmin: bool = False
+):
     session[_USER] = {"type": user_type, "key": user_key}
+    # We don't want to set the created time when a superadmin logs in as
+    # another user, since it was already set when the superadmin logged in.
+    if not from_superadmin:
+        session[_CREATED_AT] = datetime.now(timezone.utc).isoformat()
```
benadida, line 53: good call!

server/auth/lib.py
```diff
@@ -26,21 +28,47 @@ class UserType(str, enum.Enum):
...
+    session[_LAST_REQUEST_AT] = datetime.now(timezone.utc).isoformat()
 
 
-def get_loggedin_user() -> Union[Tuple[UserType, str], Tuple[None, None]]:
+def get_loggedin_user(session) -> Union[Tuple[UserType, str], Tuple[None, None]]:
     user = session.get(_USER, None)
-    return (user["type"], user["key"]) if user else (None, None)
+    if not user or is_session_expired(session):
+        return (None, None)
```
benadida, line 60: if session is expired, I would suggest actively expiring the whole session so the cookie gets cleared / renewed.

server/auth/lib.py
```diff
@@ -183,8 +215,11 @@ def restrict_access_superadmin(route: Callable):
 
     @functools.wraps(route)
     def wrapper(*args, **kwargs):
-        if not is_superadmin():
+        if not is_superadmin(session):
             raise Forbidden(description="requires superadmin privileges")
+
+        update_last_request_timestamp(session)
```
benadida, line 221: is this happening twice, once in restrict-access and once here?

benadida (approved):

server/auth/lib.py
jonahkagan: these two decorators cover different sets of routes

server/auth/lib.py
jonahkagan: 65019224

benadida (approved):

PR #943, opened 2020-12-03
Move state input from Participants to Settings screen

jonahkagan:
There are some issues going on with the two file uploads in the Participants screen in ballot comparison audits. This is the first step to simplify that form so that we can fix the issues there by taking out this extra field that never really belonged there anyway.

Changes:
- Move the state input
- Remove the localization scheme for states (there's no need for localization - we're probably only going to use Arlo in the US in English for now)
- Restyle the settings screen to be more consistent
- Add an integration test for the settings screen
- Add a percent sign to the risk limit display values


client/src/components/MultiJurisdictionAudit/useSetupMenuItems/_mocks/index.ts
```diff
@@ -49,7 +43,7 @@ export const auditSettings: {
     randomSeed: null,
     riskLimit: null,
     auditType: 'BATCH_COMPARISON',
-    auditMathType: 'BRAVO',
+    auditMathType: 'MACRO',
```
MorganLove, line 46: @jonahkagan Just curious, what's the change here?

MorganLove (approved):
@jonahkagan LGTM! I was initially confused by leaving the Formik instance with the `state` initialValue in the Participants view even if it isn't submitted, but then I realized you're fixing that in the other PR, so that's fine.

This way of doing things makes the types look a lot cleaner, yey!

client/src/components/MultiJurisdictionAudit/useSetupMenuItems/_mocks/index.ts
jonahkagan: just making the data more realistic. BRAVO can only be used with BALLOT_POLLING, and MACRO only with BATCH_COMPARISON

PR #945, opened 2020-12-03
Sequential uploading of jurisdictions/standardized contests files

jonahkagan:
There were two issues happening with the previous uploading scheme:
- After selecting both a jurisdictions and standardized contests file for upload and clicking "Save & Next," you would sometimes get an error saying you have to upload the jurisdictions file first and then you'd have to click "Save & Next" again to upload the contests file.
- If either file upload failed during background processing, the error message wouldn't get displayed

Here we use our existing CSV upload hook and component to make the AA upload these two files in sequence and display any error messages in processing. This is how we do multiple file uploads for JAs and it works great, so it seems like a safe way to go.

ezgif com-gif-maker (4)


MorganLove (approved):
@jonahkagan Nice work! This is much cleaner and easier to use.

PR #952, opened 2020-12-07
Don't validate CSV upload on blur

jonahkagan:
"You must upload a file" was getting displayed every time the file
dialog opened, since it blurred the input. We just turn that off and let
that message pop up if they try to click the Upload File button before
selecting a file.

Also added some small styling tweaks and a test for CVR upload.


client/src/components/MultiJurisdictionAudit/CSVForm/index.tsx
```diff
@@ -107,7 +106,6 @@ const CSVFileForm = ({
                     }}
                     hasSelection={!!values.csv}
                     text={values.csv ? values.csv.name : 'Select a CSV...'}
-                    onBlur={handleBlur}
```
MorganLove, line 110: @jonahkagan Can we use the `validateOnBlur` setting on the Formik instance instead? [It's documented here.](https://formik.org/docs/api/formik#validateonblur-boolean) I think that the method you use here has the same effect, but I'm not entirely sure, and it's not the documented way to do it so it might cause issues later on. Also, since in most form uses the `handleBlur` function is passed in to the form element invisibly, having it _not_ here doesn't document that it _isn't_ there. Having it explicitly set to not validate on blur in the main Formik options will clearly document that.

MorganLove (changes requested):

client/src/components/MultiJurisdictionAudit/CSVForm/index.tsx
jonahkagan: sure, changed here: 58d7c752

MorganLove (approved):

Issue #954, opened 2020-12-07
Add support for num winners >1 in ballot comparison


mcchilders:
Right now, the number of winners in any ballot comparison contest is hard-coded to "1" and we don't support multi-winner contests. This is because the CVR files don't contain the info we need to derive this value, so we need to build a way for audit admins to set the value for any given targeted contest in order to support it. To do so, we'll need:
- a place in the ballot comp data model to store number of winners, a la ballot polling/batch
- a UI that allows the AA to set that value for any targeted/opportunistic contest they want to audit 

A couple notes:
- We don't want to make AAs set this value for every contest (there could be hundreds) but just for the ones they want to audit (either targeted/opportunistic)
- We can make an intelligent guess that the number of *winners* allowed is likely to be the same as the number of *votes* allowed, which is in the CVR, but because of where this falls in the setup flow it will be easier to default to "1" and use our existing target/opportunistic contest interface.

![NumWiners.png](https://zube.io/files/votingworks/206abfd1ce59326fba4262709895b73e-numwiners.png)

PR #955, opened 2020-12-08
Add num winners field to contest select for ballot comparison

jonahkagan:
Task: #954 (also fixes #926)

- Updates the /contest endpoint to accept numWinners
- Reworks the ContestSelect component to key contests by name to avoid issues with filtering by index
- Adds a "Winners" column to the ContestSelect table
- Adds filtering contests by jurisdiction name (an outstanding TODO in the code)

ezgif com-gif-maker (5)


MorganLove (approved):
@jonahkagan This looks great!

PR #959, opened 2020-12-10
Audit board jurisdiction name in /api/me response

jonahkagan:
Updated: originally just wanted to remove this field from the frontend, but instead decided to add it to /api/me so that the frontend can continue to use it.

### DIFF
--- a/server/auth/routes.py
+++ b/server/auth/routes.py
@@ -108,6 +108,7 @@ def auth_me():
             type=user_type,
             id=audit_board.id,
             jurisdictionId=audit_board.jurisdiction_id,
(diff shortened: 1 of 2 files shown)


MorganLove:
@jonahkagan The mockups have this as part of the design, and if I remember right, it did at one point exist. Did it get lost in the middle somewhere? Should we add it back in instead of removing references to it? @mcchilders what are your thoughts?

mcchilders:
Yeah I'd actually love to have the jurisdiction name back, here!

MorganLove (approved):

PR #961, opened 2020-12-10
Change UserContext data format

jonahkagan:
- Use `null` to represent data still being loaded (our standard pattern)
- Put the `/me` response in a `{ user }` object so that we can easily
extend it with superadmin user info (in a PR coming soon)
- Remove `isAuthenticated`, since it's redundant with `user`, which is
`null` when not authenticated (plus, the meaning of this field will be
less clear once we also have superadmin user info).
- Move the relevant types into the UserContext module

### DIFF
--- a/client/src/App.test.tsx
+++ b/client/src/App.test.tsx
@@ -24,7 +24,10 @@ const apiMocks = {
   },
   abAuth: {
     url: '/api/me',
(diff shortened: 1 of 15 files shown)


MorganLove (approved):
Much cleaner. :)

Issue #958, opened 2020-12-09
Superadmin 2.0


jonahkagan:
- Move to React
- Track which superadmin user is logged in and display it in Arlo when impersonating an AA/JA
- New features:
  - Create/update/delete org
  - Create/update/delete audit admin
  - Add jurisdiction admin after audit starts
  - Delete audit?

PR #963, opened 2020-12-14
Support Tools header

jonahkagan:
Task: #958 

- Store superadmin email in session
- Add superadminUser to UserContext
- Redesign the AA header to include the logged in user's email (and add icons to the buttons)
- Add a Support Tools header above the AA header to show when a superadmin is logged in

Screen Shot 2020-12-14 at 10 46 25 AM


benadida (approved):
backed only: approved!

MorganLove (approved):

PR #966, opened 2020-12-17
show download buttons even without audit boards

benadida:


jonahkagan (approved):
From reading the code, I think the design of the page you modified might look a little weird because you put the buttons above the heading. I'd like to see a screenshot. But everything looks functionally correct.

Ideally there would also be some frontend tests for this change.

benadida:
closing this as it seems we're not going this route.

PR #972, opened 2021-01-04
Delete AuditBoards without cascading

jonahkagan:
Since SampledBallots/Batches are assigned to AuditBoards through sampling, they shouldn't be deleted when an AuditBoard is deleted. In other words, the AuditBoard isn't really the parent of SampledBallot/Batch in our data model, SampledBallot's true parent is Batch, and Batch's true parent is Jurisdiction.

So we change those foreign keys to have `on delete set null` instead of `on delete cascade`.


benadida (approved):
Make sure that migration actually works in staging, but otherwise great!

PR #973, opened 2021-01-04
Handle invalid audit board passphrases

jonahkagan:
Currently, using an invalid audit board passphrase leads to a database exception. This has happened a few times in production: https://sentry.io/organizations/votingworks/issues/2028541746/.

Instead, we redirect to the login screen with an error message.


jonahkagan:
@benadida looks like you ignored this issue in Sentry, which meant we didn't get notified as it continued to happen.

benadida (approved):

PR #976, opened 2021-01-05
Adds test for multi-winner ballot comparison

umbernhard:
**Description**
Adds a test case for mutlwinner ballot comparison to confirm it does the right thing. 

**Testing**

Adds a new test election to `test_supersimple.py` with two winners. 

**Progress**

Ready.


jonahkagan:
Does this test the discrepancy computation? That seems like a part of the math that might be sensitive to a change in the number of winners.

umbernhard:
Yep, it does. See `test_find_no_discrepancies` and `test_find_one_discrepancy`.

server/tests/audit_math/test_supersimple.py
```diff
@@ -112,10 +121,70 @@ def test_find_one_discrepancy(contests, cvrs):
...
+
+def test_two_vote_overstatement_discrepancies(contests, cvrs):
+    sample_cvr = {
+        0: {
+            "times_sampled": 1,
+            "cvr": {
+                "Contest A": {
+                    "winner": 0,
+                    "loser": 1,
+                },  # One of the reported loser ballots was actually a winner ballot
```
jonahkagan, line 170: is this comment incorrect? just noticed it because it's the same as the comment in the previous test case but the actual data is different

jonahkagan (approved):

server/tests/audit_math/test_supersimple.py
umbernhard: Ah, good catch. Will fix.

PR #977, opened 2021-01-06
Support Tools interface

jonahkagan:
Adds a new Support Tools interface (formerly known as the superadmin interface) built in React.

Features:
- Home screen with list of organizations
- Org screen with list of audits, list of audit admins (with login buttons), and form to create new audit admin (including integrating with Auth0)
- Audit screen with list of jurisdictions and list of jurisdiction admins (with login buttons)

I took this opportunity to test out two new libraries to help with data loading (react-query) and forms (react-hook-form).

https://user-images.githubusercontent.com/530106/103838547-31e3ac00-5042-11eb-9fb3-41588a7d9e26.mov

### DIFF
--- a/server/api/support.py
+++ b/server/api/support.py
@@ -34,21 +34,23 @@ def auth0_get_token() -> str:
     return str(response["access_token"])
 
 
-def auth0_create_audit_admin(email: str):
+def auth0_create_audit_admin(email: str) -> str:
     token = auth0_get_token()
     auth0 = Auth0(AUTH0_DOMAIN, token)
     try:
-        auth0.users.create(
+        user = auth0.users.create(
             dict(
                 email=email,
                 password=secrets.token_urlsafe(),
                 connection="Username-Password-Authentication",
             )
         )
+        return str(user["user_id"])
     except Auth0Error as error:
         # If user already exists in Auth0, no problem!
         if error.status_code == 409:
-            return
+            users = auth0.users_by_email.search_users_by_email(email.lower())
+            return str(users[0]["user_id"])
         raise error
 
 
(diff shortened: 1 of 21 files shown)


server/api/support.py
```diff
@@ -0,0 +1,171 @@
...
+    token = auth0_get_token()
+    auth0 = Auth0(AUTH0_DOMAIN, token)
+    try:
+        auth0.users.create(
+            dict(
+                email=email,
+                password=secrets.token_urlsafe(),
+                connection="Username-Password-Authentication",
+            )
+        )
```
benadida, line 47: should we store an `external_user_id`, or is the user keyed by `email` in the Auth0 DB?

benadida (approved):
one question about ensuring that we have a proper reference to the auth0 user row in our database, otherwise good!

server/api/support.py
jonahkagan: added here: 972b1d4

server/api/support.py
```diff
@@ -0,0 +1,174 @@
...
+        return str(user["user_id"])
+    except Auth0Error as error:
+        # If user already exists in Auth0, no problem!
+        if error.status_code == 409:
+            users = auth0.users_by_email.search_users_by_email(email.lower())
+            return str(users[0]["user_id"])
```
benadida, line 53: What if the user is not in our DB? Unlikely given the sequence of calls, but probably something to check.

benadida (approved):
one small error check to add, otherwise good.

server/api/support.py
benadida: read the first line wrong. Ignore.

PR #981, opened 2021-01-07
Fixing double submission in DataEntry

MorganLove:
**Description**

- Following a similar pattern to what Formik does for the other forms

**Testing**

- TBD

**Progress**

- Manually tested, need to add a regression test for this


MorganLove:
Closed in favor of #989

PR #983, opened 2021-01-07
Create organization from support tools

jonahkagan:
- New endpoint to create an organization as a superadmin
- Unique constraint on Organization.name
- Form on support tools home screen to create org

https://user-images.githubusercontent.com/530106/103957798-59994980-5100-11eb-9bb0-3c8a8360eb6f.mov


server/api/support.py
```diff
@@ -64,6 +64,29 @@ def list_organizations():
...
+@api.route("/support/organizations", methods=["POST"])
+@restrict_access_superadmin
+def create_organization():
+    organization = request.get_json()
+    validate(organization, ORGANIZATION_SCHEMA)
+
+    if Organization.query.filter_by(name=organization["name"]).one_or_none():
+        raise Conflict("Organization already exists")
+
+    db_session.add(Organization(id=str(uuid.uuid4()), name=organization["name"]))
```
benadida, line 84: when there is a DB constraint on uniqueness, I prefer to just `try` and `catch` the constraint exception, because technically doing the if then `add` doesn't fully protect you from the race condition. Not strictly necessary, I can live with this, just highlighting a different, slightly more resilient approach.

server/tests/helpers.py
```diff
@@ -78,9 +78,11 @@ def create_user(email=DEFAULT_AA_EMAIL) -> User:
 
 
 def create_org_and_admin(
-    org_name: str = "Test Org", user_email: str = DEFAULT_AA_EMAIL
+    org_name: str = None, user_email: str = DEFAULT_AA_EMAIL
```
benadida, line 81: nit: are we sure we want the API contract for this call to be that if you explicitly pass in a `None`, then it gets replaced with `Test Org` ? Is that sufficiently clear?

benadida (approved):
One suggestion and question, neither blocking merge.

server/api/support.py
jonahkagan: great point. this has come up before, and i actually initially implemented my first find_or_create in Arlo that way, but I had to do a string match on the constraint name in some exception metadata field, and it seemed so hacky that i stopped doing it

i would love to have a resilient, generalized find_or_create function in Arlo that actually checked for the db exception. but it probably doesn't matter for this particular PR

server/tests/helpers.py
jonahkagan: the way this test helper usually gets used is with no arguments passed in because the test case doesn't care about the org name or user email. in the rare test cases that do care, they can configure them using these optional arguments. not totally sure if that answers your question though.

PR #989, opened 2021-01-11
Adding Formik to BallotReview

### LINKED ISSUE #969: Protect forms from double-click errors (audit board interface)
Let's make sure the forms that the audit board uses to enter data don't allow double data entry via double-clicks

MorganLove:
**Description**

Closes #969 

- Stops double submission errors on DataEntry ballot audit review

**Testing**

- Added regression test that successfully fails without this fix and passes with it.

**Progress**

- Done!


jonahkagan (approved):
nice! 👍 for the regression test

PR #993, opened 2021-01-11
adding imprinted id to placeholder if present

MorganLove:
**Testing**

- added test case

**Progress**

- done!


client/src/components/MultiJurisdictionAudit/RoundManagement/generateSheets.ts
```diff
@@ -79,12 +79,14 @@ export const downloadPlaceholders = (
     ballots.forEach(ballot => {
       if (pageCount > 0) placeholders.addPage('letter')
 
+      const imprintedId = ballot.imprintedId ? [ballot.imprintedId] : []
```
jonahkagan, line 82: do we want it to say `Imprinted ID:` at the beginning like we do with all the other fields?

client/src/components/MultiJurisdictionAudit/RoundManagement/generateSheets.ts
```diff
@@ -79,12 +79,14 @@ export const downloadPlaceholders = (
...
       const lines = [
         ballot.auditBoard!.name,
         ballot.batch.container && `Container: ${ballot.batch.container}`,
         ballot.batch.tabulator && `Tabulator: ${ballot.batch.tabulator}`,
         `Batch: ${ballot.batch.name}`,
         `Ballot Number: ${ballot.position}`,
+        ...imprintedId,
```
jonahkagan, line 89: for consistency, i would use the pattern established above on lines 85 and 86 to add this item to the list (falsey values are filtered out on L91)

client/src/components/MultiJurisdictionAudit/RoundManagement/generateSheets.ts
MorganLove: @jonahkagan The length of the `imprintedId` in the test data at least was right at the edge already, and adding even 'ID' before the line made it run off.

client/src/components/MultiJurisdictionAudit/RoundManagement/generateSheets.ts
MorganLove: Oh right, somehow I missed that. :stuck_out_tongue_closed_eyes:

client/src/components/MultiJurisdictionAudit/RoundManagement/generateSheets.ts
jonahkagan: yeah my guess was that the test data was using an intentionally long batch name to see what happens with long batch names. i dont think it's a common case. i could be wrong though. in any case, I think it should be consistent with the other fields in its labeling, so if we want to come up with a solution for long lines i would vote we do that separately

MorganLove: @jonahkagan the test data just concatenates the batch name and ballot position, which gets long. If that's what an imprintedId is (which would make having it on a separate line pretty pointless, since all the info is already there), then we'll have to solve the long lines issue. If it's something else, we shouldn't need to worry.

jonahkagan: Sometimes it's the concatenation of tabulator-batch name-ballot position, sometimes it's not (every voting system has their own format for imprinted ids). In practice the imprinted ids I've seen haven't been super long - maybe 10 characters max - but I agree it could be an issue.

jonahkagan (approved):

client/src/components/MultiJurisdictionAudit/RoundManagement/generateSheets.ts
MorganLove: Ah, then we should likely implement a solution for long lines soon. I'll make an issue.

PR #997, opened 2021-01-12
Fix test fixture that resets db

jonahkagan:
It was incorrectly catching the case where the tests weren't being run
in parallel.


umbernhard (approved):
Was this what was causing the minerva fails?

jonahkagan:
Ohhh interesting question! I'm not sure. Could be worth putting up a PR that reinstates those tests and running them on CI a few times through to check

jonahkagan:
Though thinking about it more I don't think it's likely to be the cause

PR #998, opened 2021-01-12
Script to clear audit boards for a jurisdiction

jonahkagan:
Tested all cases manually.

### DIFF
--- a/scripts/clear-audit-boards.py
+++ b/scripts/clear-audit-boards.py
@@ -15,6 +15,8 @@ if __name__ == "__main__":
         print("Jurisdiction not found")
         sys.exit(1)
 
+    print(f"Org: {jurisdiction.election.organization.name}")
+    print(f"Audit: {jurisdiction.election.audit_name}")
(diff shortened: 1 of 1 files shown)


scripts/clear-audit-boards.py
```diff
@@ -0,0 +1,41 @@
...
+        print("Usage: python -m scripts.clear-audit-boards <jurisdiction_id>")
+        sys.exit(1)
+
+    jurisdiction_id = sys.argv[1]
+    jurisdiction = Jurisdiction.query.get(jurisdiction_id)
+    if not jurisdiction:
+        print("Jurisdiction not found")
+        sys.exit(1)
+
+    print(f"Jurisdiction: {jurisdiction.name}")
```
benadida, line 18: suggest printing the election name & org name, too.

benadida (approved):

scripts/clear-audit-boards.py
jonahkagan: good call, added

PR #1003, opened 2021-01-14
Cypress tests

MorganLove:
**Description**

- End to end integration tests

**Testing**

- Yup.

**Progress**

- WIP


client/cypress/end-to-end/ballot-comparison-audit-board.spec.js
```diff
@@ -0,0 +1,121 @@
...
+  const id = 0
+
+  before(() => {
+    id = uuid()
+    cy.visit('/')
+    cy.loginAuditAdmin('audit-admin-cypress@example.com')
+    cy.get('input[name=auditName]').type(`TestAudit${id}`)
+    cy.get('input[value="BALLOT_COMPARISON"]').check({ force: true })
+    cy.findByText("Create Audit").click()
+    cy.wait(1000)
```
jonahkagan, line 14: I wouldn't expect you to need to call `wait`, since Cypress has built in retries/timeouts. I think if the default timeout (4 seconds) is too short for some of these checks, it would be better to increase the timeout for those checks rather than add a `wait`, since `wait` makes the test always wait that long even if it's not necessary, whereas a timeout will let the check finish as soon as possible. 

More on that here: https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Timeouts

Another option that I just found out about for this specific case would be to intercept the request for the new route and wait on that: https://docs.cypress.io/guides/guides/network-requests.html#Waiting

client/cypress/end-to-end/ballot-comparison-audit-board.spec.js
```diff
@@ -0,0 +1,121 @@
...
+    cy.findAllByText('Upload File').click()
+    cy.findAllByText(/Upload successfully completed/).should('have.length', 2)
+    cy.wait(2000)
+    cy.findByText('Next').click()
+    cy.findAllByText('Target Contests').should('have.length', 2)
+    cy.get('input[type="checkbox"]').first().check({ force: true })
+    cy.findByText('Save & Next').click()
+    cy.findAllByText('Opportunistic Contests').should('have.length', 2)
+    cy.findByText('Save & Next').click()
+    cy.get('#state').select('AL')
```
jonahkagan, line 44: When possible, I think it would be better to use the testing library style of queries so that we are testing by what the user sees rather than our internal dom structure or css. For this one, I would use `cy.findByRole('select', {name: 'State'})` or if that didn't work `cy.findByLabelText('State')`.

When I set up Cypress, I installed a testing library plugin so that all the same queries we use in our unit tests are available.

package.json
```diff
@@ -22,6 +22,9 @@
       "sort-package-json"
     ]
   },
+  "dependencies": {
+    "pdf-parse": "^1.1.1"
+  },
```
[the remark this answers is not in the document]
MorganLove: Haha, oops! I must have done `yarn add` when I was in the wrong directory and didn't notice. Thanks!

client/cypress/support/commands.js
```diff
@@ -26,6 +26,7 @@
 
 import url from 'url'
 import qs from 'querystring'
+import 'cypress-file-upload'
```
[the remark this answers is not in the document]
MorganLove: Yeah, it just makes it so you don't have to import it in every test suite that uses `.attachFile()`. But that kind of does document where that comes from, so it's useful to do that anyway, so I moved it.

client/cypress/end-to-end/ballot-comparison-audit-board.spec.js
MorganLove: Yes, I'm trying to minimize their usage. There are a few places where things fail without them that still perplex me, though, and I'm working on figuring them out. I just went through and cleaned up a bunch of extra ones though.

Thanks for pointing out the network request interceptions! That's a really cool feature, and I'll have to take a look and see if they can help resolve some of these.

client/cypress/end-to-end/ballot-comparison-audit-board.spec.js
```diff
@@ -0,0 +1,121 @@
...
+    cy.wait(1000)
+    cy.contains('Number of Audit Boards')
+    cy.findByText('Save & Next').click()
+    cy.findByText('Download Audit Board Credentials').click()
+    cy.logout()
+    cy.wait(2000)
+  })
+
+  it('Audit Board - Login with Credentials', () => {
+    cy.task('getPdfContent', `cypress/fixtures/PDFs/Audit Board Credentials\ -\ Death Star\ -\ TestAudit${id}.pdf`).then((content) => {
```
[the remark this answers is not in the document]
MorganLove: Yes, PDF parsing wasn't possible with Cypress, and so I used a node command to read & process, and as far as I could see, the only way to do that was via a task instead of a Cypress custom command.

client/cypress/end-to-end/ballot-comparison-audit-board.spec.js
```diff
@@ -0,0 +1,287 @@
...
+      cy.findByText('Next').click()
+      cy.contains(/Ballot Cards to Audit/)
+      cy.get('table tbody tr').each(($el, index, list) => {
+          if(index == 0) {
+              cy.findByText('Start Auditing').click()
+          }
+          cy.get('input[type="checkbox"]').first().click({force: true})
+          cy.findByText('Review').click() 
+          cy.findByText('Submit & Next Ballot').click() 
+      })
```
jonahkagan, line 120: This particular bit confused me - it took a while for me to figure out why you were looping over the rows of the table, since you're actually not on that screen anymore once you click "Start Auditing".

As a user, I think I would just click "Start Auditing" and then audit all of the ballots until it took me back to the main screen with the table of ballots. A way to implement that would be to define a function `auditBallot` that audits one currently shown ballot, clicks "Submit & Next Ballot", and then checks to see whether it's looking at another ballot or if it's back on the main screen. If it's looking at another ballot, it can call `auditBallot` again recursively. Otherwise it can return and the test script can continue. (I based this idea on this: https://github.com/cypress-io/cypress/issues/8100#issuecomment-664919436)

Or if you want to stick with this, maybe you could add some comments explaining what's going on

client/cypress/end-to-end/ballot-comparison-audit-creation.spec.js
```diff
@@ -0,0 +1,370 @@
+import 'cypress-file-upload'
+
+before(() => cy.exec('./cypress/seed-test-db.sh'))
+
+describe('Ballot Comparison Audit Creation', () => {
```
jonahkagan, line 5: It seems like the audit board test suite also has to do the entire audit creation and setup, which covers the same functionality as this test suite. I would recommend combining them into one test suite that goes through the entire audit flow. It will be faster to run and less test code to maintain.

jonahkagan: It's also a bit confusing to me that there is state maintained between the different test cases within a test suite. That's generally not a recommended pattern because the test cases can't be run separately, and it seems like Cypress docs also advise against it: https://docs.cypress.io/guides/references/best-practices.html#Having-tests-rely-on-the-state-of-previous-tests

client/cypress/end-to-end/ballot-comparison-audit-creation.spec.js
```diff
@@ -0,0 +1,370 @@
...
+        fileContent: fileContent.toString(),
+        fileName: 'sample_standardized_contests_contest_name_col_error.csv',
+        mimeType: 'csv'
+      })
+    })
+    cy.findAllByText('Upload File').click()
+    cy.contains("Missing required column: Contest Name.")      
+  })
+
+  it('Standardized Contests invalid CSV - Jurisdiction column error', () => {
```
jonahkagan, line 127: I don't think we need to test all the different file error cases in Cypress. Those are already covered by unit tests. One error case per file type should suffice to convince us that the integration between the frontend and backend is working correctly.

client/cypress/end-to-end/ballot-comparison-audit-creation.spec.js
MorganLove: Yeah, I think I'm going to restructure things down to just a couple suites with happy paths and get those merged, then figure out the most efficient way to cover the necessary error conditions without any bloat.

Could you elaborate on what you mean by maintained state? Do you mean the id? I haven't noticed any tests that break when run independently and I've run almost all if not all with `.only` during testing.

client/cypress/end-to-end/ballot-comparison-audit-board.spec.js
MorganLove: Yeah, I see what you're saying. I'll see if I can find a more straightforward and simple way to do it, and refactor to that or add comments.

client/cypress/end-to-end/ballot-comparison-audit-creation.spec.js
jonahkagan: Sounds good :)

By maintain state, I meant state in the database. But I think I just misread the tests because there's a `beforeEach` happening to set up the audit before each `it` test case. My bad.

I guess the root of that point though is that if the `beforeEach` has a lot of steps, running it before each `it` will make the tests take a lot longer, so making one long `it` for each audit you want to run might save a lot of time.

MorganLove:
@jonahkagan I cut this PR down to just two happy paths so we can get it merged (I kept the other test cases in a separate branch locally, and I'll be refining them and adding them as separate PRs). I also refactored it to use getters based more off user interaction.

I did try modifying the audit board ballot polling as you suggested, but in order to use conditions in cypress it ended up adding more obfuscations than simplifying things. I also realized that we want to test that it goes through exactly the number of ballots that are present, and so I ended up going with the original method, just with a clarifying comment.

client/cypress/end-to-end/ballot-polling-happy-path.spec.js
```diff
@@ -0,0 +1,178 @@
...
+    cy.logout()
+    cy.loginAuditAdmin('audit-admin-cypress@example.com')
+    cy.findByText(`TestAudit${id}`).click()
+    cy.findByText('Review & Launch').click()
+    cy.findAllByText('Review & Launch').should('have.length', 2)
+    cy.findByText('Launch Audit').click()
+    cy.findAllByText('Launch Audit').spread((firstButton, secondButton) => {
+      secondButton.click()
+    })
+    cy.get('tbody').children('tr').its('length').should('be.gt', 0)
```
jonahkagan, line 77: any reason you didn't enter jurisdiction results and finish this audit?

client/cypress/end-to-end/login-page.spec.js
```diff
@@ -0,0 +1,19 @@
+before(() => cy.exec('./cypress/seed-test-db.sh'))
+
+describe('Audit creation, filling in standard ballot comparison values', () => {
```
jonahkagan, line 3: this test description seems wrong

client/cypress/fixtures/CSVs/contest/sample_standardized_contests.csv
```diff
@@ -0,0 +1,4 @@
+Contest Name,Jurisdictions
```
jonahkagan, line 1: remove CSV files you aren't using?

jonahkagan (approved):

client/cypress/end-to-end/ballot-polling-happy-path.spec.js
MorganLove: I've been focusing on the ballot comparison route and just copying over what's applicable to the ballot polling side for the most part. Hadn't gotten to testing the offline audit completion. Should be easy to add in a separate PR though.

Issue #986, opened 2021-01-11
Improve CVR upload error messages


jonahkagan:
Need specific errors for:
- when batch names don't match manifest
- when contest names don't match standardized contest list

Should follow the patterning we've used elsewhere for error messages. 
Let's look and see if showing a possible similar field that they might have meant is helpful, may be? Maybe not. Would be helpful to see all the errors at once ... maybe show the first one and let them expand to see others if wanted? "64 additional errors... click to expand" etc

jonahkagan:
First PR up that handles batch names that don't match manifests and a few other errors we've seen: https://github.com/votingworks/arlo/pull/1006

Outstanding work:
- Contest names that don't match standardized contests is a bit tricker, since we only catch that when computing the sample size later.
- We've seen a race condition if you upload a large CVR and then upload a new CVR while the first uploaded CVR is still parsing.

PR #1006, opened 2021-01-19
Better CVR parsing error messages

jonahkagan:
Task: #986 

Adds user-facing error messages for three issues we've seen so far:
- Empty CVR file
- Invalid contest name format (missing the "(Vote For=1)" suffix)
- Tabulator/batch name that doesn't match manifest

### DIFF
--- a/server/api/cvrs.py
+++ b/server/api/cvrs.py
@@ -5,6 +5,8 @@ import csv
 import typing
 from collections import defaultdict
 import re
+import difflib
+import ast
 from datetime import datetime
 from sqlalchemy.orm.session import Session
 from flask import request, jsonify, Request
@@ -87,6 +89,11 @@ def process_cvr_file(session: Session, jurisdiction: Jurisdiction, file: File):
         contest_votes_allowed = []
         for contest_header in contest_headers:
             match = re.match(r"^(.+) \(Vote For=(\d+)\)$", contest_header)
+            if not match:
+                raise UserError(
+                    f"Invalid contest name: {contest_header}."
+                    + " Contest names should have this format: Contest Name (Vote For=1)."
+                )
             contest_names.append(match[1])
             contest_votes_allowed.append(int(match[2]))
 
@@ -121,7 +128,7 @@ def process_cvr_file(session: Session, jurisdiction: Jurisdiction, file: File):
 
             for row in cvrs:
                 [
-                    _cvr_number,
+                    cvr_number,
                     tabulator_number,
                     batch_id,
                     record_id,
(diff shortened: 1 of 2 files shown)


server/api/cvrs.py
```diff
@@ -83,6 +89,11 @@ def process():
         contest_votes_allowed = []
         for contest_header in contest_headers:
             match = re.match(r"^(.+) \(Vote For=(\d+)\)$", contest_header)
+            if not match:
+                raise UserError(
+                    f"Invalid contest name: {contest_header}."
+                    + " Contest names should have this format: Contest Name (Vote For=1)."
```
umbernhard, line 95: can we also say something like "And must match standardized contest file exactly", maybe including a "Expected <blah> from standardized contests, got <blah> from the CVR"?

server/api/cvrs.py
```diff
@@ -117,15 +128,45 @@ def process():
...
+                            (
+                                " The closest match we found in the ballot manifest was:"
+                                f" {closest_match[0]}, {closest_match[1]}."
+                            )
+                            if closest_match
+                            else ""
+                        )
+                        + " Please check your CVR file and ballot manifest thoroughly"
+                        " to make sure these values match - there may be a similar"
+                        " inconsistency in other rows in the CVR file."
```
umbernhard, line 167: Do we have a check for duplicate rows? Would be nice to call that out explicitly too.

umbernhard: Also total rows

umbernhard (changes requested):
Looks good. I'd like to see the changes I suggested made, but if you or @mcchilders think that's too much I'll change to an Approve.

server/api/cvrs.py
jonahkagan: Hmm, we're not currently guaranteed to have the standardized contests file uploaded when the CVRs are uploaded, so we can't check that here.

server/api/cvrs.py
jonahkagan: I don't think it's likely to happen since it's a file generated by their voting system. I'm taking the approach of only building error messages for the errors we've actually seen, since there are tons of ways this file _could_ be wrong. I'm hoping that we can trust the generated files to have the correct format, and discover in practice which ways our assumptions about the format are incorrect.

server/api/cvrs.py
umbernhard: I thought we were? What are we matching against if not?

jonahkagan: We match against the contests that the AA selects from the standardized contests file, but we don't do that until later when generating sample size options. It's kind of awkward and something I'd like to discuss, but whatever we decide I'd like to handle it in a separate PR.

umbernhard: Oh I see, just a generic regex. I feel like we should compare with the standardized contests file, but I defer to @mcchilders

umbernhard (approved):

PR #1013, opened 2021-01-20
JA audit board status improvements

jonahkagan:
- Fix the status box to only show audit boards complete once they're
signed off
- Add tags/intent to the progress bars for individual audit boards in
order to differentiate which boards have signed off

Screen Shot 2021-01-20 at 12 49 13 PM

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/StatusBox/index.tsx
+++ b/client/src/components/MultiJurisdictionAudit/StatusBox/index.tsx
@@ -309,9 +309,8 @@ export const JurisdictionAdminStatusBox = ({
       )
 
     const numCompleted = auditBoards.filter(
-      ({ currentRoundStatus }) =>
-        currentRoundStatus.numAuditedBallots ===
-        currentRoundStatus.numSampledBallots
+      ({ currentRoundStatus, signedOffAt }) =>
+        currentRoundStatus.numSampledBallots === 0 || signedOffAt
     ).length
     const details = [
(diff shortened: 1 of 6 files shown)


client/src/components/MultiJurisdictionAudit/StatusBox/index.tsx
```diff
@@ -309,9 +309,8 @@ export const JurisdictionAdminStatusBox = ({
       )
 
     const numCompleted = auditBoards.filter(
-      ({ currentRoundStatus }) =>
-        currentRoundStatus.numAuditedBallots ===
-        currentRoundStatus.numSampledBallots
+      ({ currentRoundStatus, signedOffAt }) =>
+        currentRoundStatus.numSampledBallots === 0 || signedOffAt
```
benadida, line 313: is this the right logic? `numSampledBallots === 0`? Or should it be `numAuditedBallots === numSampledBallots` ?

benadida (approved):
IMPORTANT: check one logic line, but otherwise lgtm

client/src/components/MultiJurisdictionAudit/StatusBox/index.tsx
jonahkagan: if an audit board doesn't have any ballots assigned, the audit logic considers it "complete" because there's no work to be done. but it won't have `signedOffAt` populated. we account for this already everywhere but here

PR #1014, opened 2021-01-21
Support Tools: Reopen audit board

jonahkagan:
- Reorganize support tools to have a separate screen for each jurisdiction with a list of audit boards and JAs
- Add a button to reopen an audit board after it has signed off

https://user-images.githubusercontent.com/530106/105403986-485f3b00-5bde-11eb-8b2a-1e9b7d050d2f.mov

### DIFF
--- a/client/src/components/Header.tsx
+++ b/client/src/components/Header.tsx
@@ -61,6 +61,7 @@ const UserMenu = styled.div`
   .bp3-button-text {
     overflow: hidden;
     text-overflow: ellipsis;
+    white-space: nowrap;
   }
   .bp3-menu {
     width: 200px;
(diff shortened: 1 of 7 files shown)


client/src/components/Header.tsx
```diff
@@ -61,6 +61,7 @@ const UserMenu = styled.div`
   .bp3-button-text {
     overflow: hidden;
     text-overflow: ellipsis;
+    white-space: nowrap;
```
jonahkagan, line 64: this fixes an unrelated bug where long user emails with a hyphen in them were wrapping onto two lines

MorganLove (approved):
Looks great! I had fun playing with this. So easy to navigate and work with.

PR #1019, opened 2021-01-25
Draw sample in background

jonahkagan:


jonahkagan:
@benadida would love an initial review before I go further with this approach. Key files are server/worker/tasks.py and server/models.py

server/worker/tasks.py
```diff
@@ -0,0 +1,112 @@
...
+
+    # Claim this task by updating the `started_at` timestamp in such a way that
+    # it must not have been set before.
+    result = db_session.execute(
+        update(BackgroundTask.__table__)  # pylint: disable=no-member
+        .where(BackgroundTask.id == task.id)
+        .where(BackgroundTask.started_at.is_(None))
+        .values(started_at=datetime.now(timezone.utc))
+    )
+    if result.rowcount == 0:
```
benadida, line 71: Unless this is its own transaction in this statement that commits right away, this will just block other threads during the whole process of running the task until the transaction commits. Is that what `db_session.execute` does?

If you're planning for multiple workers, we may also want to plan for cleanup on backend job crash that would otherwise leave tasks uncompleted.

server/worker/tasks.py
```diff
@@ -0,0 +1,112 @@
...
+        )
+        logger.info(f"TASK_START {task_metadata}")
+
+        with db_session.begin_nested():  # pylint: disable=no-member
+            if task.task_name == TaskName.DRAW_SAMPLE:
+                round = Round.query.filter_by(
+                    id=dict(task.payload)["round_id"],
+                    election_id=dict(task.payload)["election_id"],
+                ).one()
+                rounds.draw_sample(round)
```
benadida, line 87: should this be more of a registered handler that we dispatch to, rather than a set of `if/then` statements?

server/models.py
```diff
@@ -743,6 +746,21 @@ class File(BaseModel):
...
 
 
+class TaskName(str, enum.Enum):
+    DRAW_SAMPLE = "DRAW_SAMPLE"
+
+
+class BackgroundTask(BaseModel):
+    id = Column(String(200), primary_key=True)
+    task_name = Column(Enum(TaskName), nullable=False)
+    # All tasks must have election_id in the payload
```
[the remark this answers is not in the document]
jonahkagan: I wanted to require the election_id so that it would always be logged and hopefully make debugging easier, not for any correctness concern, so that's why it seemed kind of unnecessary in the data model. I tried putting a type constraint on the payload but the type system wasn't making it easy so I decided to just make it an informal constraint. But if you think it merits a formal field I'm open to being convinced

server/models.py
```diff
@@ -446,6 +446,9 @@ class Round(BaseModel):
     round_num = Column(Integer, nullable=False)
     ended_at = Column(UTCDateTime)
 
+    draw_sample_task_id = Column(String(200), ForeignKey("background_task.id"))
```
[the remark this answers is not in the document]
jonahkagan: good call, just forgot. thanks!

server/worker/tasks.py
jonahkagan: definitely would make for a slightly cleaner API. i'll try it out

server/worker/tasks.py
jonahkagan: I actually just copied this over from the code that @eventualbuddha wrote for processing files, so I didn't take the time to think deeply about it.

I did a little research:
- db_session.execute is just a way to run a query without using the ORM layer, it doesn't do anything else special
- This statement executes in the current transaction, not in its own transaction
- Since the default transaction isolation level in Postgres is "read committed", I think that if there were multiple threads trying to execute this statement, you are correct that they would get blocked until the first thread to execute this statement commits the transaction (according to https://www.postgresql.org/docs/9.5/transaction-iso.html)

I'm not totally sure what you mean about planning for cleanup beyond what we already do (which is to rollback the nested transaction and then commit the outer transaction, recording the error).

That being said, maybe we don't need any of this since we're running in a single-worker context?

benadida: right, if we continue to assume single-worker context (and if so we should make the explicitly clear in the code), we can simplify this.

benadida (approved):

jonahkagan:
@benadida added one new update. can you give it a look? https://github.com/votingworks/arlo/pull/1019/commits/510d205bf753b4d7c7e2bbd10ddd5c1a4ee8c76a

benadida (approved):
Assuming this is all done as a transaction, which I imagine it is, then I'm good!

PR #1022, opened 2021-01-28
Draw sample in background - frontend

jonahkagan:
There are two places that we create a new round - the audit setup review screen (for round 1) and the status box (for subsequent rounds). In order to support both of these places creating rounds and waiting for the sample drawing to complete in the background, we check if the sample is being drawn before rendering any of the the audit view, and show a spinner if the background task is still in progress.

Technical changes:
- Consolidated round creation in the useRoundsAuditAdmin hook (startNewRounds function), since we need to reload the rounds once the new round is created
- Since startNewRounds requires sample sizes as numbers, simplified the Review screen to use a numeric input


client/src/components/MultiJurisdictionAudit/useSetupMenuItems/_mocks/index.ts
```diff
@@ -123,12 +123,7 @@ export const auditSettings: {
 }
 
 export const roundMocks: {
-  [key in
-    | 'empty'
-    | 'singleIncomplete'
-    | 'twoIncomplete'
-    | 'singleComplete'
-    | 'needAnother']: IRound[]
+  [key: string]: IRound[]
```
MorganLove, line 126: I find the explicit defining of each state to be useful in IDE autocompletion, personally.

client/src/components/testUtilities.tsx
```diff
@@ -128,9 +128,10 @@ export const withMockFetch = async (
     }
 
     if (expectedRequest) {
+      const requestIndex = requests.length - requestsLeft.length - 1
       // eslint-disable-next-line no-console
       console.error(
-        'Expected fetch request:\n',
+        `Expected fetch request (${requestIndex}):\n`,
```
MorganLove, line 134: Nice!

MorganLove (approved):
Looks good! One suggestion that's not blocking, that's all.

client/src/components/MultiJurisdictionAudit/useSetupMenuItems/_mocks/index.ts
jonahkagan: Ohhh cool I was wondering why they were there. Thanks for the tip. I'll put them back. I wish the IDEs were smart enough to figure it out without the type annotation...

PR #1029, opened 2021-02-08
Adding backend validation for sample sizes

### LINKED ISSUE #740: Backend: validate custom sample sizes
ballot polling: custom sample size should be no greater than the total ballots cast (value provided by AA in audit setup)
batch comparison: custom sample size should be no greater the total number of batches across all jurisdictions (e.g. batches across all manifests)
ballot comparison: custom sample size should be no greater than the total ballots (ballots across all CVRs)
Note that, in any of these scenarios, the value *can* be equal to these values ... just not greater than these values.

MorganLove:
**Description**

Closes #740 

**Testing**

- None yet

**Progress**

- WIP

Initial run at this, let me know if I'm off base. :slightly_smiling_face:

### DIFF
--- a/server/api/rounds.py
+++ b/server/api/rounds.py
@@ -709,6 +709,55 @@ CREATE_ROUND_REQUEST_SCHEMA = {
 def validate_round(round: dict, election: Election):
     validate(round, CREATE_ROUND_REQUEST_SCHEMA)
 
+    total_batches = 0
+    if election.audit_type == AuditType.BATCH_COMPARISON:
+        contest_sample_size_arr = list(round["sampleSizes"].values())
+        contest_sample_size = contest_sample_size_arr[0]
+        for single_contest in election.contests:
+            if single_contest.is_targeted:
+                for jurisdiction in single_contest.jurisdictions:
+                    total_batches += (
+                        jurisdiction.manifest_num_batches
+                        if jurisdiction.manifest_num_batches
+                        else 0
+                    )
+                if contest_sample_size > total_batches:
+                    raise Conflict(
+                        f"Sample size must be less than or equal to: {total_batches} (the total number of batches in the targeted contest)"
+                    )
+
+    count = 0
+    if election.audit_type == AuditType.BALLOT_POLLING:
+        for single_contest in election.contests:
+            if single_contest.is_targeted:
+                contest_sample_size_arr = list(round["sampleSizes"].values())
+                contest_sample_size = contest_sample_size_arr[count]
+                if contest_sample_size > single_contest.total_ballots_cast:
+                    raise Conflict(
+                        f"Sample size must be less than or equal to: {single_contest.total_ballots_cast} (the total number of ballots in the targeted contest)"
+                    )
+                count += 1
+
+    total_ballots = 0
+    if election.audit_type == AuditType.BALLOT_COMPARISON:
+        contest_sample_size_arr = list(round["sampleSizes"].values())
+        sample_size_count = len(contest_sample_size_arr) - 2
+        for single_contest in election.contests:
+            contest_sample_size = contest_sample_size_arr[sample_size_count - 1]
+            total_ballots = 0
+            if single_contest.is_targeted:
+                for jurisdiction in single_contest.jurisdictions:
+                    total_ballots += (
+                        jurisdiction.manifest_num_batches
+                        if jurisdiction.manifest_num_batches
+                        else 0
+                    )
+                if contest_sample_size > total_ballots:
+                    raise Conflict(
+                        f"Sample size must be less than or equal to: {total_ballots} (the total number of ballots in the targeted contest)"
+                    )
+                sample_size_count -= 1
+
     current_round = get_current_round(election)
     if current_round and not current_round.draw_sample_task.completed_at:
         raise Conflict("Arlo is already currently drawing the sample for this round.")
--- a/server/api/rounds.py
+++ b/server/api/rounds.py
@@ -709,55 +709,6 @@ CREATE_ROUND_REQUEST_SCHEMA = {
 def validate_round(round: dict, election: Election):
     validate(round, CREATE_ROUND_REQUEST_SCHEMA)
 
-    total_batches = 0
-    if election.audit_type == AuditType.BATCH_COMPARISON:
-        contest_sample_size_arr = list(round["sampleSizes"].values())
-        contest_sample_size = contest_sample_size_arr[0]
-        for single_contest in election.contests:
-            if single_contest.is_targeted:
-                for jurisdiction in single_contest.jurisdictions:
-                    total_batches += (
-                        jurisdiction.manifest_num_batches
-                        if jurisdiction.manifest_num_batches
-                        else 0
-                    )
-                if contest_sample_size > total_batches:
-                    raise Conflict(
-                        f"Sample size must be less than or equal to: {total_batches} (the total number of batches in the targeted contest)"
-                    )
-
-    count = 0
-    if election.audit_type == AuditType.BALLOT_POLLING:
-        for single_contest in election.contests:
-            if single_contest.is_targeted:
-                contest_sample_size_arr = list(round["sampleSizes"].values())
-                contest_sample_size = contest_sample_size_arr[count]
-                if contest_sample_size > single_contest.total_ballots_cast:
-                    raise Conflict(
-                        f"Sample size must be less than or equal to: {single_contest.total_ballots_cast} (the total number of ballots in the targeted contest)"
-                    )
-                count += 1
-
-    total_ballots = 0
-    if election.audit_type == AuditType.BALLOT_COMPARISON:
-        contest_sample_size_arr = list(round["sampleSizes"].values())
-        sample_size_count = len(contest_sample_size_arr) - 2
-        for single_contest in election.contests:
-            contest_sample_size = contest_sample_size_arr[sample_size_count - 1]
-            total_ballots = 0
-            if single_contest.is_targeted:
-                for jurisdiction in single_contest.jurisdictions:
-                    total_ballots += (
-                        jurisdiction.manifest_num_batches
-                        if jurisdiction.manifest_num_batches
-                        else 0
-                    )
-                if contest_sample_size > total_ballots:
-                    raise Conflict(
-                        f"Sample size must be less than or equal to: {total_ballots} (the total number of ballots in the targeted contest)"
-                    )
-                sample_size_count -= 1
-
     current_round = get_current_round(election)
     if current_round and not current_round.draw_sample_task.completed_at:
         raise Conflict("Arlo is already currently drawing the sample for this round.")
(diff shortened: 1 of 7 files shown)


server/api/rounds.py
```diff
@@ -709,6 +709,55 @@ def sample_batches(
...
+                    total_ballots += (
+                        jurisdiction.manifest_num_batches
+                        if jurisdiction.manifest_num_batches
+                        else 0
+                    )
+                if contest_sample_size > total_ballots:
+                    raise Conflict(
+                        f"Sample size must be less than or equal to: {total_ballots} (the total number of ballots in the targeted contest)"
+                    )
+                sample_size_count -= 1
```
jonahkagan, line 759: for ballot comparison, we can actually use the exact same strategy as ballot polling. `contest.total_ballots_cast` will be populated based on the number of ballots in the CVRs by the time the round gets created.

jonahkagan (changes requested):
excited to have you contributing on backend! there's obviously a large codebase to grok, so i'm happy to answer questions as you dive in

server/api/rounds.py
```diff
@@ -709,6 +709,55 @@ def sample_batches(
 def validate_round(round: dict, election: Election):
     validate(round, CREATE_ROUND_REQUEST_SCHEMA)
 
+    total_batches = 0
```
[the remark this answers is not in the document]
MorganLove: @jonahkagan Using `targeted_contest_ids` is simplifying some things, but now in trying to access the contest object itself is being odd. I tried 

```for contest_id in targeted_contest_ids:
            single_contest = election.contests[election.contests.index(contest_id)]```

But I got the error `error: Value of type "Iterable[Contest]" is not indexable` so that evidently doesn't work. What would you suggest?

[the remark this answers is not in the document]
Comment (jonahkagan) [reply server/api/rounds.py:712]: ah, the problem here might be that `election.contests` is not actually a list, it's an iterator (a lazy list returning items one by one, in this case, from the database). so to fix that error, i think doing `list(election.contests)` will fix it

that being said, it might be easier to change `targeted_contest_ids` to `targeted_contests` and make it a list of contest objects instead of ids, that way you don't need to go searching for the contest objects again

```python
targeted_contests = [contest for contest in election.contests if contest.is_targeted]
if set(round["sampleSizes"].keys()) != {c.id for c in targeted_contests}:
  ... etc ...
```

server/api/rounds.py
```diff
@@ -724,10 +724,49 @@ def validate_round(round: dict, election: Election):
...
 
-        targeted_contest_ids = {c.id for c in election.contests if c.is_targeted}
-        if set(round["sampleSizes"].keys()) != targeted_contest_ids:
+        targeted_contests = [
+            contest for contest in election.contests if contest.is_targeted
+        ]
+        if set(round["sampleSizes"].keys()) != {c.id for c in targeted_contests}:
             raise BadRequest("Sample sizes provided do not match targeted contest ids")
 
+    if election.audit_type == AuditType.BATCH_COMPARISON:
```
jonahkagan, line 733: move these over one indentation - they should only apply for round 1

server/api/rounds.py
```diff
@@ -724,10 +724,49 @@ def validate_round(round: dict, election: Election):
...
+
+    count = 0
+    if (
+        election.audit_type == AuditType.BALLOT_POLLING
+        or election.audit_type == AuditType.BALLOT_COMPARISON
+    ):
+        for single_contest in targeted_contests:
+            contest_sample_size_arr = list(round["sampleSizes"].values())
+            contest_sample_size = contest_sample_size_arr[count]
+            if contest_sample_size > single_contest.total_ballots_cast:
```
jonahkagan, line 752: use the same pattern from L739: `round["sampleSizes"][single_contest.id]`

MorganLove:
@jonahkagan locally I have one server test failing: `server/tests/audit_math/test_minerva.py::test_get_sample_size`. The diff is expecting the `size` field to have `4` but it's getting `1000`. I was going to ask you what your advice was on that, but on CI it's giving a completely different set of errors that I'm not making heads or tails of. Thoughts?

server/api/rounds.py
```diff
@@ -724,10 +724,36 @@ def validate_round(round: dict, election: Election):
...
+            or election.audit_type == AuditType.BALLOT_COMPARISON
+        ):
+            for single_contest in targeted_contests:
+                if (
+                    round["sampleSizes"][single_contest.id]
+                    > single_contest.total_ballots_cast
+                ):
+                    raise Conflict(
+                        f"Sample size must be less than or equal to: {single_contest.total_ballots_cast} (the total number of ballots in the targeted contest)"
+                    )
```
jonahkagan, line 755: might be helpful to include the contest name in the error message so the user knows which contest they need to correct the sample size for

server/tests/ballot_comparison/test_ballot_comparison.py
```diff
@@ -469,3 +483,49 @@ def test_ballot_comparison_cvr_metadata(
 
     ballot_missing_contest = next(b for b in ballots if b["imprintedId"] == "2-2-4")
     assert ballot_missing_contest["contestsOnBallot"] == [contests[0]["id"]]
+
+
+def test_custom_sample_size_validation(
```
jonahkagan, line 488: test case function names need to be unique - i think that's why you're getting the error in CI

server/tests/ballot_comparison/test_ballot_comparison.py
```diff
@@ -161,6 +161,13 @@ def test_ballot_comparison_two_rounds(
     sample_size = sample_size_options[target_contest_id][0]
     snapshot.assert_match(sample_size)
 
+    contest = Contest.query.get(target_contest_id)
+    assert contest.total_ballots_cast is None
+    assert contest.votes_allowed is None
+    assert contest.choices == []
+
+    set_contest_metadata_from_cvrs(contest)
+
```
jonahkagan, line 170: why did you add this block?

jonahkagan: ah, i think i understand. now that we're validating the sample size against the contest metadata, it needs to be set before running that validation. this means there's actually an implicit assumption in the create round endpoint that the contest metadata has already been set. i think we should make that explicit so there's no hidden failure case.

My recommendation would be to move the block of code that sets the metadata up from later in the create rounds endpoint to before the validation. This is the block:

```python
    # For ballot comparison audits, we need to lock in the contest metadata we
    # parse from the CVRs when we launch the audit.
    if (
        election.audit_type == AuditType.BALLOT_COMPARISON
        and json_round["roundNum"] == 1
    ):
        for contest in election.contests:
            set_contest_metadata_from_cvrs(contest)
```

Will probably have to be tweaked to `json_round.get("roundNum")` since it will come before we actually validate the JSON. A bit weird, but I think it should work fine.

Another option would be to split the validation into two functions - one for the round generally, and one for the sample sizes, and then call the sample size validation function after setting the contest metadata.

server/tests/batch_comparison/test_batch_comparison.py
```diff
@@ -167,7 +167,7 @@ def test_batch_comparison_round_1(
     assert jurisdictions[1]["currentRoundStatus"] is None
 
     # Use an artificially large sample size in order to have enough samples to work with
-    sample_size = 20
+    sample_size = 10
```
jonahkagan, line 170: i guess this failed the validation? i think we needed that many samples for the rest of the test to be meaningful (e.g. where we check below how the batches got divvied up between the audit boards - with a small sample that check doesn't really check anything)

maybe instead of lowering the sample size, we could change the manifest so the validation passes?

server/tests/ballot_comparison/test_ballot_comparison.py
jonahkagan: not sure if that's the cause of the minerva error - maybe fix this and then let me know if you're still seeing that error too

server/tests/batch_comparison/test_batch_comparison.py
MorganLove: Ah, gotcha. Good point. What tweak would you recommend on the manifest so the math is happy?

jonahkagan: I'm not totally sure, but I would try out increasing the number of ballots until the sample size validation passes. I would recommend removing the `manifests` fixture from this test case's arguments and just copying the code in that fixture into this test case, then tweaking it.

server/tests/ballot_comparison/test_ballot_comparison.py
MorganLove: @jonahkagan Okay, I went for the second option, let me know what you think!

server/tests/ballot_comparison/test_ballot_comparison.py
MorganLove: @jonahkagan Fixed this, but still getting the minerva error.

server/api/rounds.py
jonahkagan: the error message for batch comparison also needs this change

server/tests/ballot_comparison/test_ballot_comparison.py
jonahkagan: can you remove this block from all the tests now?

server/api/rounds.py
```diff
@@ -720,13 +720,42 @@ def validate_round(round: dict, election: Election):
...
 
-    if round["roundNum"] == 1:
-        if "sampleSizes" not in round:
-            raise BadRequest("Sample sizes are required for round 1")
 
-        targeted_contest_ids = {c.id for c in election.contests if c.is_targeted}
-        if set(round["sampleSizes"].keys()) != targeted_contest_ids:
-            raise BadRequest("Sample sizes provided do not match targeted contest ids")
+def validate_custom_sample_size(round: dict, election: Election):
+    validate(round, CREATE_ROUND_REQUEST_SCHEMA)
```
jonahkagan, line 725: dont need to run this validation again

server/api/rounds.py
```diff
@@ -720,13 +720,42 @@ def validate_round(round: dict, election: Election):
...
         raise Conflict("The current round is not complete")
 
-    if round["roundNum"] == 1:
-        if "sampleSizes" not in round:
-            raise BadRequest("Sample sizes are required for round 1")
 
-        targeted_contest_ids = {c.id for c in election.contests if c.is_targeted}
-        if set(round["sampleSizes"].keys()) != targeted_contest_ids:
-            raise BadRequest("Sample sizes provided do not match targeted contest ids")
+def validate_custom_sample_size(round: dict, election: Election):
```
jonahkagan, line 724: `validate_sample_sizes` would be more accurate, since it's not only custom, and there are potentially multiple sample sizes

server/tests/batch_comparison/test_batch_comparison.py
```diff
@@ -285,6 +338,7 @@ def test_batch_comparison_round_2(
     assert rv.status_code == 200
     batches = json.loads(rv.data)["batches"]
 
+    print(batches)
```
jonahkagan, line 341: remove

server/tests/audit_math/test_minerva.py
```diff
@@ -20,6 +20,7 @@ def contests():
 # Note also doctests in minerva.py module.
 
 
+# @pytest.mark.skip(reason="Failing currently with 4 instead of 1000")
```
jonahkagan, line 23: remove

jonahkagan (approved):

PR #1030, opened 2021-02-08
Optimize assigning ballots to audit boards

jonahkagan:
Task: #1027 

Previously, we loaded all the sampled ballots into memory at once, which
was memory-intensive and led to high water marking.

With this new approach, we count the sampled ballots per batch in the
database, and use bulk update operations to assign the audit boards, so
we never need to load the ballots into memory.


benadida (approved):
Nice!

PR #1032, opened 2021-02-09
Count ballots instead of loading all ballots

jonahkagan:
Task: #1031

When we load JA screens, we show a count of the ballots assigned, as well as (sometimes) buttons to download placeholders/labels. Previously, we would load all the ballots from the backend to implement both of these features. This made page loads slow, when we really only need to load the ballots if a JA clicks to download placeholders/labels.

This PR makes a few changes:
- adds a `count` query parameter to the `/ballots` endpoint to return just a count instead of all the ballots
- only loads ballots on demand when a button is clicked (this can be improved in the future with pagination and/or caching)
- adds a new AsyncButton component that disables the button while the ballot loading/pdf generation/downloading is happening, since this can take a while


MorganLove (approved):
This makes me happy. :slightly_smiling_face:

server/api/ballots.py
```diff
@@ -199,6 +199,17 @@ def list_ballots_for_jurisdiction(
...
+        count = (
+            SampledBallot.query.join(Batch)
+            .filter_by(jurisdiction_id=jurisdiction.id)
+            .join(SampledBallotDraw)
+            .filter_by(round_id=round.id)
+            .distinct(SampledBallot.id)
+            .count()
+        )
+        return jsonify({"count": count})
+
```
benadida, line 212: I'm a little bit irked by returning different data entirely based on a query parameter, as opposed to adding a `/ballots-count` endpoint. I can live with it, just not quite what I would expect from that API signature.

benadida (approved):

server/api/ballots.py
jonahkagan: I think the two common ways I've seen to do this are:
1. Use a response envelope and paginate the request, so that you get back something like: `{ ballots: [...], page: 1, count: 1000 }` (this is what I was emulating, though without the pagination, I didn't know what to do with the `ballots` field, so I just omitted it, figuring we could move to this pattern if/when we paginate)
2. Putting pagination and count in headers

I've never really seen a REST pattern for getting a count without having pagination, have you?

benadida: ah, I see, this is a prelude to pagination... OK, I'm more comfortable with that then!

PR #1041, opened 2021-02-11
bgcompute: commit before sleeping

jonahkagan:
Migrations would often hang when the worker was running. The `pg_stat_activity` table would show an "idle in transaction" query coming from bgcompute. Committing the transaction before sleeping seems to solve it!

### DIFF
--- a/server/worker/bgcompute.py
+++ b/server/worker/bgcompute.py
@@ -185,6 +185,10 @@ def bgcompute_forever():
     while True:
         bgcompute()
         run_new_tasks()
+        # Before sleeping, we need to commit the curren transaction, otherwise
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -189,4 +189,4 @@ workflows:
       - preflight
       - build-and-test-server
       - build-and-test-client
-      - cypress
+      # - cypress
```
benadida, line 192: is this a temporary thing that you want to remove before merging?

benadida (approved):
excellent!

.circleci/config.yml
jonahkagan: just a commit error. will remove

PR #1044, opened 2021-02-12
Remove contest round status

jonahkagan:
This was originally built to feed an "Audit Progress by Contest" screen, which we never built and don't plan to build anytime soon. The contests endpoint is one of our slowest/most-memory intensive endpoints because of this extra, unused data. So we remove it for now.

Also removes the navigation sidebar on the Audit Progress screen since there's only one option: Jurisdictions.


MorganLove (approved):
LGTM!

PR #1045, opened 2021-02-13
upgrading cypress

MorganLove:
A test PR to figure out what's going on with the versioning of cypress and CI


MorganLove:
Successfully figured out what was up.

PR #1046, opened 2021-02-16
Fix progress screen flash when drawing sample

jonahkagan:
Previously, the progress screen would render briefly after launching the audit before the spinner would get rendered. This happened because the old round data (an empty array) hadn't been updated from the backend yet. To fix this, we use the `startNextRound` function from the same hook that loaded that data, so that it will invalidate and reload the data automatically before resolving the `startNextRound` promise.


MorganLove (approved):
let's go!

PR #1057, opened 2021-02-17
Inject Flask env vars into index.html

jonahkagan:
Previously, we were trying to use create-react-app's custom env vars feature to set env vars readable by the React app, but those env vars are compiled into the app at build time. Heroku doesn't supply env vars at build time, so they weren't actually getting read.

Instead, we use Flask's templating to inject the env vars from the server into the index.html page, where we load them in JS to make them available to the React app.


benadida (approved):
good catch!

Issue #1047, opened 2021-02-17
Create new "hybrid" audit method for SUITE


mcchilders:
Create a new audit type for the hybrid audit; add to the options for creating a new audit as "Hybrid (SUITE - ballot comparison & ballot polling)"

PR #1059, opened 2021-02-17
Hybrid audit type

jonahkagan:
Task: #1047


client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
```diff
@@ -164,6 +164,8 @@ const Review: React.FC<IProps> = ({
                     BALLOT_POLLING: 'Ballot Polling',
                     BATCH_COMPARISON: 'Batch Comparison',
                     BALLOT_COMPARISON: 'Ballot Comparison',
+                    HYBRID:
+                      'Hybrid (SUITE - Ballot Comparison & Ballot Polling)',
```
benadida, line 168: Are we labeling this SUITE even though it is only one math type of hybrid audit?

benadida (approved):

client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
mcchilders: SUITE describes the stats that are used, not what types of audits are being combined in the hybrid, but you'll notice it's labeled with all three. (While it's theoretically possible to extend SUITE to other types of hybrid audits beyond ballot comp + ballot polling, it's never been implemented, so this is as robust an implementation as anyone has done.)

PR #1060, opened 2021-02-18
Adding frontend validation for custom sample sizes

MorganLove:
**Description**

- Matching the back to the front

**Testing**

- Missing unit tests written

**Progress**

- Done

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
+++ b/client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
@@ -140,35 +140,32 @@ const Review: React.FC<IProps> = ({
   ).length
 
   const getBatchTotal = () =>
-    contests
-      ? participatingJurisdictions.reduce(
-          (a, { ballotManifest: { numBatches } }) =>
-            numBatches !== null ? a + numBatches : a,
-          0
-        )
-      : 0
+    participatingJurisdictions.reduce(
+      (a, { ballotManifest: { numBatches } }) =>
+        numBatches !== null ? a + numBatches : a,
+      0
+    )
 
   const getBallotTotal = (jurisdictionIds: string[]) =>
-    contests
-      ? jurisdictions
-          .filter(jurisdiction =>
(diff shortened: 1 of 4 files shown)


client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
```diff
@@ -139,6 +139,58 @@ const Review: React.FC<IProps> = ({
...
+      return testNumber(
+        Number(getBatchTotal()),
+        `Must be less than or equal to: ${Number(
+          getBatchTotal()
+        )} (the total number of batches in this targeted contest)`
+      )
+    }
+    if (auditType === 'BALLOT_COMPARISON') {
+      return testNumber(
+        Number(getBallotTotal(jurisdictionIds)),
```
jonahkagan, line 184: could you just use `totalBallotsCast` for this case as well?

client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
```diff
@@ -139,6 +139,58 @@ const Review: React.FC<IProps> = ({
     isFileProcessed(j.ballotManifest)
   ).length
 
+  const getBatchTotal = () =>
+    contests
```
jonahkagan, line 143: why do you need to check `contests`? isn't it guaranteed to be defined when this function is called?

client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
MorganLove: @jonahkagan It comes up `null` here when on Ballot Comparison audits.

jonahkagan: oh I see why that is. it gets set when computing the sample sizes, but since the api call for contest probably happens before the api call for sample sizes, it doesn't show up. side effects like that are tricky. i guess it's okay to keep what you have here, though it's not the same logic as the backend validation

jonahkagan (approved):

PR #1062, opened 2021-02-19
Store which sample size option was selected

jonahkagan:
When launching the audit, we used to just store the sample size number that was selected. This meant that there was no way to display which sample size option corresponding with that number had been chosen (e.g. was it "asn" or was it "custom"?)

This changes the create round endpoint to take the entire sample size option and store it in the RoundContest object. It also makes the /sample-sizes endpoint return which sample sizes were selected for the first round so we can show them to the user.

This also opens up the ability for more complex sample size option formats, which we'll need for hybrid audits.


MorganLove:
Looks fine on a visual review. I'll poke at it functionally in the am. :)

server/migrations/versions/07859b6b370b_json_sample_sizes.py
```diff
@@ -0,0 +1,32 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.alter_column(
+        "round_contest",
+        "sample_size",
+        type_=sa.JSON,
+        postgresql_using="json_build_object('key', 'custom', 'size', sample_size::int, 'prob', null)",
```
benadida, line 25: I'm not familiar with this syntax, you've checked this will do the right thing for old sample sizes that were ints? I think it will, just making sure you're sure :)

benadida (approved):

server/migrations/versions/07859b6b370b_json_sample_sizes.py
jonahkagan: yeah i tried it out. i was actually surprised that it worked on the first try!

PR #1066, opened 2021-02-24
Check for ballots not getting assigned to audit boards

jonahkagan:
Manually tested the logging works

### DIFF
--- a/server/api/audit_boards.py
+++ b/server/api/audit_boards.py
@@ -2,9 +2,9 @@ import uuid
 import itertools
 from datetime import datetime
 from typing import List, Dict
-from flask import jsonify, request
+from flask import jsonify, request, current_app
 from xkcdpass import xkcd_password as xp
-from werkzeug.exceptions import Conflict, BadRequest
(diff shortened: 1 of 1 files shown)


server/api/audit_boards.py
```diff
@@ -122,6 +122,34 @@ def assign_sampled_ballots(
...
+                dict(
+                    jurisdiction_id=jurisdiction.id,
+                    round_id=round.id,
+                    num_sampled_ballots=num_sampled_ballots,
+                    num_assigned_ballots=num_assigned_ballots,
+                    buckets=balanced_buckets,
+                )
+            )
+        )
+        raise InternalServerError("Error assigning ballots to audit boards")
```
benadida, line 151: will this fail for the user in the front-end? Probably okay if it does, just checking.

benadida (approved):

server/api/audit_boards.py
jonahkagan: yes it will - i did that intentionally so they would know to try again

PR #1067, opened 2021-02-24
Waiting for sample sizes before launching

MorganLove:
**Description**

- Disabling audit launch until sample sizes are loaded

**Testing**

- Waiting for the audit launch to be not disabled before clicking in cypress

**Progress**

- It should work, but the cypress tests aren't behaving yet

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
+++ b/client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
@@ -424,7 +424,7 @@ const Review: React.FC<IProps> = ({
               <FormButton
                 intent="primary"
                 disabled={
-                  sampleSizeOptions === null ||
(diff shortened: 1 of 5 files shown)


client/src/components/MultiJurisdictionAudit/AASetup/Review/index.tsx
```diff
@@ -424,6 +424,7 @@ const Review: React.FC<IProps> = ({
               <FormButton
                 intent="primary"
                 disabled={
+                  sampleSizeOptions === null ||
```
jonahkagan, line 427: should be `(shouldShowSampleSizes && sampleSizeOptions === null)`

MorganLove: I wondered about that, haha.

jonahkagan (approved):

PR #1068, opened 2021-02-25
adding batch comparison failure case tests

MorganLove:
**Testing**

- Same pattern as for ballot comparison failure case tests, just for batch comparison

**Progress**

- There's a wait at the beginning because there's a rerender that happens as the component is loading which sometimes lands just as it is clicking the Upload File button, and the button gets detached from the DOM, making it fail. I'm puzzled on how to make this stable without the wait.

- Also, I'm thinking that we won't need to do a separate failure cases suite for ballot polling, since its functionality is a subset of the other two, and it'd be a duplication.


client/cypress/end-to-end/batch-comparison-failure-cases.spec.js
```diff
@@ -0,0 +1,153 @@
...
+    cy.fixture('CSVs/candidate-total-batch/sample_candidate_totals_by_batch_col_error.csv').then(fileContent => {
+      cy.get('input[type="file"]').last().attachFile({
+        fileContent: fileContent.toString(),
+        fileName: 'sample_candidate_totals_by_batch_col_error.csv',
+        mimeType: 'csv'
+      })
+    })
+    cy.findAllByText('Upload File').last().click()
+    cy.contains(/Missing required column: Palpatine/)
+
```
jonahkagan, line 110: I think the test could probably end here, since there are no further error cases to check

jonahkagan (commented):
I agree it's probably fine to not do another test case for ballot polling errors

client/cypress/end-to-end/batch-comparison-failure-cases.spec.js
MorganLove: Makes sense. I've updated it, but decided to cut it off after the replace file process since that's technically part of the process of dealing with bad files, and the integration there isn't able to be effectively covered with the unit tests.

jonahkagan (approved):

PR #1070, opened 2021-03-01
Redesign contests section of Review screen

jonahkagan:
In preparation for hybrid audits, show contest metadata on the Review screen. Also adds in a scrollable table component to display the contest universe, which was previously taking up a lot of real estate.

Screen Shot 2021-03-01 at 3 48 00 PM
Screen Shot 2021-03-01 at 3 51 29 PM


client/src/components/Atoms/Table.tsx
```diff
@@ -135,4 +135,55 @@ export const sortByRank = <T extends object>(rank: (data: T) => number) =>
...
+  width: 100%;
+
+  thead {
+    flex: 0 0 auto;
+    box-shadow: inset 0 -1px 0 0 rgb(16 22 26 / 15%); /* Copied from Blueprint */
+    width: 100%;
+  }
+
+  thead tr {
+    /* Remove approx width of scrollbar so headers line up with columns */
```
MorganLove, line 158: Out of curiosity, does this approach work across different OSs and browsers?

jonahkagan: Tried a couple browsers and it was subtly different. I found a new solution where you add a hidden scrollbar to the header row instead: f198cbc9

MorganLove: Huh, cool!

MorganLove (approved):

PR #1074, opened 2021-03-02
Fixing cypress

MorganLove:
Pinning the chrome version so it changing versions doesn't randomly break things

This is specifically what broke this time: https://github.com/cypress-io/cypress/issues/15300

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,7 +93,7 @@ commands:
       - run:
           name: Install Chrome
           command: |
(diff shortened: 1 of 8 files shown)


MorganLove:
I'm running out of ideas for what to look at here.

MorganLove:
I think this is the problem: https://github.com/cypress-io/cypress/issues/15300

jonahkagan (approved):

Issue #1049, opened 2021-02-17
Calculate & store vote totals per audit segment (CVR vs. no-CVR)


mcchilders:
The SUITE math requires not just that the vote totals for each candidate in each target contest be known, but also that they be known within each audit segment (CVR & no-CVR). The CVR segment totals can be derived from the CVRs themselves and then added up, as we do normally with ballot comparison, but the no-CVR segment must be calculated by taking the overall totals (as reported by the AA on the Target/Opportunistic contest screens) and subtracting the vote totals from the CVR segment. 

Once calculated, these values should also be displayed on the audit review and launch screen so that the AA can check them.

PR #1075, opened 2021-03-03
Hybrid audit: Count CVR/non-CVR vote totals

jonahkagan:
Task: #1049 

Compute the CVR/non-CVR vote totals and display them on the Review screen.


server/api/cvrs.py
```diff
@@ -66,6 +67,53 @@ def set_contest_metadata_from_cvrs(contest: Contest):
...
+        return None
+
+    cvr_choice_votes = {choice.id: 0 for choice in contest.choices}
+    for jurisdiction in contest.jurisdictions:
+        cvr_contests_metadata = typing.cast(
+            JSONDict, jurisdiction.cvr_contests_metadata
+        )
+        contest_metadata = cvr_contests_metadata[contest.name]
+        for choice_name, choice_metadata in contest_metadata["choices"].items():
+            choice = next(c for c in contest.choices if c.name == choice_name)
```
benadida, line 105: is `contest.choices` going to do a DB query every time? If so should we pull those once outside of the two `for` loops?

benadida (approved):
one question on perf

server/api/cvrs.py
jonahkagan: Nope, once you load a relationship (e.g. `contest.choices`), it's just a Python object.

PR #1076, opened 2021-03-03
Hybrid audit: sample size integration

jonahkagan:
Integrate with the sample size math for hybrid audits. There is one big design decision to think through here: We need to separate the audit results for CVR ballots and non-CVR ballots.

For CVR ballots, the math needs a CVR-like dictionary with the audit result for each sampled ballot. We already compute this for ballot comparison, so we simply add a filter to make sure its only looking at CVR ballots.

For non-CVR ballots, it's a bit trickier. For ballot polling, we compute the vote counts for each contest choice at the end of each round and store them in the RoundContestResult object. Here, I chose to add a filter during that process to only count non-CVR ballots, so the resulting counts only apply to that segment.

The downside to this is that we won't be storing a cumulative total across both segments. But that's not a big deal, since the math never actually requires the cumulative total. We may want to show it in the audit report, but it's easy to compute by adding together the totals from the two segments.

I do think this makes things a bit confusing, since certain database objects have different meanings in different audit types. I'm curious about exploring ideas to pay back some of this "design debt" that we have incurred by adding the audit methods one-by-one. One direction that might be fruitful is to explore where database objects are being reused in different ways across audit types and consider separating them out into separate tables specific to an audit type.

TODO:
- Testing
- Front-end changes to show the two segments of the sample size


server/api/sample_sizes.py
```diff
@@ -99,7 +106,83 @@ def sample_sizes_for_contest(contest: Contest):
...
+                -1,  # TODO what is sample_size supposed to be
+            )
+
+            options = suite.get_sample_size(
+                election.risk_limit,
+                # TODO is this the right contest?
+                sampler_contest.from_db_contest(contest),
+                non_cvr_stratum,
+                cvr_stratum,
+            )
```
jonahkagan, line 174: @umbernhard I took a crack at using the suite interface from your PR. I've got a few questions marked TODO where I wasn't sure. Can you help me figure them out?

server/api/sample_sizes.py
```diff
@@ -99,7 +106,83 @@ def sample_sizes_for_contest(contest: Contest):
...
+            # In hybrid audits, we only store round contest results for non-CVR
+            # ballots
+            non_cvr_sample_results = (
+                None if round_one else rounds.contest_results_by_round(contest)
+            )
+            non_cvr_stratum = suite.BallotPollingStratum(
+                non_cvr_contest,
+                non_cvr_sample_results,
+                -1,  # TODO what is sample_size supposed to be
```
umbernhard, line 138: This is the number of ballots sampled so far.

server/api/sample_sizes.py
```diff
@@ -99,7 +106,83 @@ def sample_sizes_for_contest(contest: Contest):
...
+                        for choice_id, vote_count in vote_counts.items()
+                    },
+                },
+            )
+            cvr_reported_results = rounds.cvrs_for_contest(contest)
+            # The CVR sample results are filtered to only CVR ballots
+            cvr_sample_results = rounds.sampled_ballot_interpretations_to_cvrs(contest)
+            # TODO how to use get_misstatements to create the stratum?
+            cvr_misstatements = suite.get_misstatements(
+                cvr_contest, cvr_reported_results, cvr_sample_results, "???", "???"
```
umbernhard, line 159: We need to iterate over every winner-loser pair and get the misstatements. So my data type is probably not correct here for elections with more than two candidates. Misstatements should be 

```
{
    (winner1,loser1): {"o1": 0, "o2": 0, "u1": 0, "u2": 0},
    (winner1,loser2): {...},
    ...,
}
```

server/api/sample_sizes.py
```diff
@@ -99,7 +106,83 @@ def sample_sizes_for_contest(contest: Contest):
...
+            )
+            cvr_stratum = suite.BallotComparisonStratum(
+                cvr_contest,
+                cvr_reported_results,
+                cvr_misstatements,
+                -1,  # TODO what is sample_size supposed to be
```
umbernhard, line 165: same as the other strata, the number of ballots sampled so far

server/audit_math/suite_stub.py
```diff
@@ -0,0 +1,81 @@
...
+    return {}
+
+
+def get_sample_size(
+    risk_limit: int,
+    contest: Contest,
+    non_cvr_stratum: BallotPollingStratum,
+    cvr_stratum: BallotComparisonStratum,
+):
+    return {
```
umbernhard, line 76: I don't think we'll be able to return stopping probabilities like this. So the output of this function will just be a tuple of `(cvr_sample_size, no_cvr_sample_size)`

server/api/sample_sizes.py
```diff
@@ -1,14 +1,72 @@
...
+    misstatements: suite.MISSTATEMENTS = {}
+    for winner, loser in product(contest.winners, contest.losers):
+        misstatement_counts = Counter(
+            misstatement(
+                winner, loser, audited_result["cvr"], reported_results.get(ballot)
+            )
+            for ballot, audited_result in audited_results.items()
+        )
+        misstatements[(winner, loser)] = {
+            "o1": misstatement_counts[1],
```
umbernhard, line 63: whoa this is a really cool way to do this

server/api/sample_sizes.py
```diff
@@ -99,7 +157,77 @@ def sample_sizes_for_contest(contest: Contest):
...
+                {} if round_one else rounds.contest_results_by_round(contest)
+            )
+            non_cvr_stratum = suite.BallotPollingStratum(
+                total_ballots.non_cvr,
+                non_cvr_vote_counts,
+                non_cvr_sample_results,
+                non_cvr_previous_samples,
+            )
+
+            cvr_reported_results = rounds.cvrs_for_contest(contest)
```
umbernhard, line 203: it occurs to me that this isn't actually used by the stratum object now that we have misstatements, so I guess I should take it out.

umbernhard (approved):
Looks good to me, the only change might be removing the cvr results from the ballot comparison stratum.

server/api/sample_sizes.py
jonahkagan: yeah, discovered this when we did ballot comparison!

benadida (approved):

PR #1078, opened 2021-03-04
Suite

umbernhard:
**Description**
Implements SUITE functionality for ballot polling and comparison audits. 

**Testing**
Introduces tests to verify tweaks made to ballot polling and comparison, as well as new math with SUITE. 

**Progress**

Tests need some more love, and the code probably ought to be restyled for consistency/clarity.


jonahkagan (commented):
didn't review the implementations of the math functions yet

server/audit_math/suite.py
```diff
@@ -0,0 +1,650 @@
...
+
+
+from decimal import Decimal
+from ..models import AuditMathType
+from .sampler_contest import Contest, CVR, CVRS, SAMPLE_CVRS, SampleCVR
+from .supersimple import Discrepancy
+from . import bravo
+
+
+GAMMA = 1.03905  # This GAMMA is used in Stark's tool, AGI, and CORLA
```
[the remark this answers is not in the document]
umbernhard: I'm gonna switch it to Decimals, just hadn't gotten there yet.

server/audit_math/suite.py
```diff
@@ -0,0 +1,650 @@
...
+        pvalues.append(maximize_fisher_combined_pvalue(alpha, contest, strata, winner,loser))
+
+    max_p = max(pvalues)
+
+    return max_p, max_p <= alpha
+
+
+
+
+def compute_discrepancies(
```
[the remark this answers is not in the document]
umbernhard: yeah, I'm unsure how to resolve this. The difference is that this implementation only computes discrepancies for the two candidates we care about, while the one in `supersimple` gets the maximum discrepancy count per ballot.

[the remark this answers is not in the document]
jonahkagan: ahhh i didn't read closely. now i see the difficulty. i think you could solve this by making a function:

```py
def compute_discrepancy(winner_reported: CVR, loser_reported: CVR, winner_audited: CVR, loser_audited: CVR) -> Discrepany
```

then you can use it like so for `supersimple`:

```py
for ballot in ballots:
  ballot_discrepancy = max(
    compute_discrepancy(
      reported[ballot][winner],
      reported[ballot][loser],
      audited[ballot][winner],
      audited[ballot][loser],
    )
    for winner, loser in product(winner, losers)
  )
```

and in `suite`:
```pyt
for ballot in ballots:
  ballot_discrepancy = compute_discrepancy(
      reported[ballot][winner],
      reported[ballot][loser],
      audited[ballot][winner],
      audited[ballot][loser],
  )

File: server/audit_math/suite.py [hunk 3]
```diff
@@ -0,0 +1,660 @@
...
+                        ...
+                    }
+            sample_results : the vote totals for this stratum in the sample so far. E.g.,
+                    {
+                        "winner1": 10,
+                        "winner2": 7,
+                        "loser1": 5,
+                        "loser2": 2,
+                        ...
+                    }
```
jonahkagan, line 71: i think sample_results is a dict mapping round id to the vote totals example you have here

server/audit_math/suite.py
```diff
@@ -0,0 +1,660 @@
...
+            contest: contest information for this stratum, including candidate
+                     vote totals and stratum size
+            results: The CVRs for this stratum.
+            misstatements: All of the misstatements observed so far, by winner-loser pair. E.g.,
+                    {
+                        (winner1, loser1): {"o1": 0, "o2": 0, "u1": 0, "u2": 0},
+                        (winner1, loser2): ...,
+                        ...,
+                        (winner2, loser1): ...
+                    }
```
jonahkagan, line 210: where do you get winner-loser pairs from if you are creating a BallotComparisonStratum?

server/audit_math/suite.py
```diff
@@ -0,0 +1,691 @@
...
+        )
+        pvalue = np.exp(log_pvalue)
+        return float(np.min([pvalue, 1.0]))  # cast for the typechecker
+
+
+def get_misstatements(contest, reported_cvr, sample, winner, loser):
+    o1, o2, u1, u2 = 0, 0, 0, 0
+    for ballot in reported_cvr:
+        if ballot not in sample:
+            continue
```
jonahkagan, line 260: this seems a bit different from the logic in supersimple, where we iterate over the sample results, not the reported results, and also have some special casing for a few different edge cases

jonahkagan (commented):
I think things are looking a lot simpler and more straightforward! I just did a very high level review of the interface - didn't dig into the math

server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison.py
```diff
@@ -64,7 +64,7 @@
 \r
 ######## ROUNDS ########\r
 Round Number,Contest Name,Targeted?,Sample Size,Risk Limit Met?,P-Value,Start Time,End Time,Audited Votes\r
-1,Contest 1,Targeted,40,No,179125340.4921900034,DATETIME,DATETIME,Choice 1-1: 17; Choice 1-2: 12\r
+1,Contest 1,Targeted,40,No,179125340.4921896756,DATETIME,DATETIME,Choice 1-1: 17; Choice 1-2: 12\r
```
umbernhard, line 67: I'm not sure what's going on here. It could be just that we're at the end of our precision so there's some jittering.

server/audit_math/suite.py
umbernhard: Oh you're right, will fix.

server/audit_math/suite.py
umbernhard: (winner, loser) come from the overall contest results. They aren't the winners/losers of the stratum, but of the overall contest.

server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison.py
jonahkagan: yeah this is unsettling since you didn't change any of the code...

server/audit_math/suite.py
umbernhard: My thinking about the difference is that in a total ballot comparison audit, there may be ballots that don't have CVRs and aren't included in the results, which may get sampled. Here, it seems like those ballots would get dumped into the ballot polling stratum and so it's safe to assume that there aren't any ballots that don't have CVRs here. I suppose I could just iterate over the sample, I think this was left over from some of my thinking. 

Actually this function is never used, I might just take it out altogether. It should probably move to the api code that initializes these objects.

server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison.py
umbernhard: well now it's failing on CI. Ghosts?

umbernhard: I wonder if it's a quirk of the processor architecture on my new machine. I.e. if little- vs. big-endian machines cause small precision differences in the Decimal module.

jonahkagan (approved):

Issue #1077, opened 2021-03-04
Sample from manifest in ballot comparison/hybrid audits


jonahkagan:
Previously, in ballot comparison audits, we constructed the ballot universe to sample from based on the ballots that were present in the CVR and had a result recorded for the target contest. The motivation for this was to avoid sampling a bunch of ballots that didn't even have the contest we wanted to audit, wasting a bunch of the auditors' time.

The issue with this approach is that it exposes the audit to a particular attack: if an attacker were able to remove a row from the CVR, they could change the tabulation results of the election. We would never be able to sample that ballot, since it wasn't in the CVR.

Instead, we need to draw the sample based on the _declared_ ballots in the ballot manifest, which is generated independently from the tabulator data. Then, if a ballot we sample isn't present in the CVR, we should count that as a two-vote overstatement - the worst-case scenario.

PR #1080, opened 2021-03-08
Sample from manifest in ballot comparison audits

jonahkagan:
Task: #1077

- Draw sample based on manifest, not CVR
- Don't hide contests from audit board that aren't on the CVR

Also updates ballot comparison tests to have expected discrepancy values written into the test (previously we just manually verified them by looking at the audit report snapshot).


server/api/ballots.py
```diff
@@ -172,20 +170,11 @@ def serialize_ballot(ballot: SampledBallot) -> JSONDict:
 
 
 def serialize_ballot_comparison_ballot(
-    ballot: SampledBallot, cvr_ballot: CvrBallot, jurisdiction: Jurisdiction
```
benadida, line 175: I thought we wanted to maybe keep the jurisdiction around to filter by jurisdiction... but maybe that's for later?

server/audit_math/supersimple.py
```diff
@@ -134,9 +133,9 @@ def compute_discrepancies(
 
         found = False
 
-        # Special case: if ballot can't be found by audit board, count it as a
-        # two-vote overstatement
-        if ballot_sample_cvr is None:
+        # Special cases: if ballot wasn't in CVR or ballot can't be found by
+        # audit board, count it as a two-vote overstatement
+        if ballot_sample_cvr is None or ballot_cvr is None:
```
benadida, line 138: this is a math change, can we get a :+1: from @umbernhard ?

benadida (approved):
looks good, one more check on the math from @umbernhard would make me happy.

server/audit_math/supersimple.py
umbernhard: So this can only possibly happen if we're sampling from the manifest, not the CVR. That might be another discussion like the one we just had @jonahkagan. I feel like we should be doing that...

In any case, I don't think this matters for now, since if we're just sampling from the CVR then the `or` will never be true.

server/api/ballots.py
jonahkagan: I think what you're referring to is that we only want to show the audit board the contests that meet two conditions:
1. The contest should be selected for audit (either targeted or opportunistic)
2. The jurisdiction should be in the contest universe

We meet (1) because Arlo only stores the contests that are selected for audit. We meet (2) because the /contests endpoint for audit boards scopes the contests returned to only the ones with the jurisdiction in their contest universe.

PR #1083, opened 2021-03-09
Only let AA enter contest total ballots cast in ballot polling

jonahkagan:
In all other audit types, we should compute this from the manifests.

See #1081


client/cypress/end-to-end/batch-comparison-failure-cases.spec.js
```diff
@@ -51,7 +51,6 @@ describe('Batch Comparison', () => {
     cy.findByLabelText('Votes for Candidate/Choice 1').type('2700')
     cy.findByLabelText('Name of Candidate/Choice 2').type('Palpatine')
     cy.findByLabelText('Votes for Candidate/Choice 2').type('2620')
-    cy.findByLabelText('Total Ballots for Contest').type('5320')
```
benadida, line 54: should we be testing for absence of this field?

benadida (approved):

client/cypress/end-to-end/batch-comparison-failure-cases.spec.js
jonahkagan: covered that in the unit test i added

PR #1084, opened 2021-03-10
Script to parse XML CVRs

jonahkagan:


benadida (approved):
looks good to me, obviously some tests would help if possible, but I understand we don't have much source data.

PR #1085, opened 2021-03-11
Reprocess standardized contests on jurisdiction file change

jonahkagan:
Modifies process_file to use a nested transaction so that multiple files can be processed in the same transaction.

### DIFF
--- a/server/api/standardized_contests.py
+++ b/server/api/standardized_contests.py
@@ -27,50 +27,54 @@ STANDARDIZED_CONTEST_COLUMNS = [
 ]
 
 
+def set_standardized_contests(election: Election, file: File):
+    standardized_contests_csv = parse_csv(file.contents, STANDARDIZED_CONTEST_COLUMNS)
+
+    standardized_contests = []
+    for row in standardized_contests_csv:
+        if row[JURISDICTIONS].strip() == "all":
+            jurisdictions = election.jurisdictions
+        else:
+            jurisdiction_names = {
+                name.strip() for name in row[JURISDICTIONS].split(",")
+            }
+            jurisdictions = list(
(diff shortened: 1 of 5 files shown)


server/api/standardized_contests.py
```diff
@@ -30,7 +30,7 @@
 def process_standardized_contests_file(
     session: Session, election: Election, file: File
 ):
-    def process():
+    def process() -> None:
```
jonahkagan, line 33: realized this wasn't getting typechecked because it didn't have any type annotation. changes below are just for the typechecker

server/util/jurisdiction_bulk_update.py
```diff
@@ -30,6 +33,26 @@ def process():
...
+        logger.info(
+            f"START_REPROCESSING_STANDARDIZED_CONTESTS {dict(election_id=election.id)}"
+        )
+        # First, clear out the previously processed data.
+        election.standardized_contests = None
+        election.contests = []
+        election.standardized_contests_file.processing_started_at = None
+        election.standardized_contests_file.processing_completed_at = None
+        election.standardized_contests_file.processing_error = None
+        session.flush()  # Make sure process_file can read the changes we just made
```
benadida, line 48: I think what you're doing here is updating the DB without committing... and we're sure `process_file` will be within the same transaction, right? If so, sounds good.

benadida (approved):

server/util/jurisdiction_bulk_update.py
jonahkagan: Yes, the idea is to make sure those changes are readable by queries that happen in the same transaction in process_file. So instead of committing them (which would end the transaction and start a new one), we flush them, which moves them from the Sqlalchemy session to the database.

PR #1088, opened 2021-03-12
Fix ballot comparison round 2 sample sizes

jonahkagan:
We were giving the math the number of previously sampled ballots across all
contests, instead of just for the given contest.


server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison.py
```diff
@@ -20,6 +20,11 @@
 TABULATOR2,BATCH2,6,2-2-6,0.300053574780458718,N,Audit Board #1
 """
 
+snapshots["test_ballot_comparison_multiple_targeted_contests_sample_size 1"] = [
+    ({"key": "supersimple", "prob": None, "size": 30},),
+    ({"key": "supersimple", "prob": None, "size": 30},),
+]
```
jonahkagan, line 26: verified that these were different with the old code

benadida (approved):

Issue #1054, opened 2021-02-17
Sample ballots appropriately from each segment, combine into single retrieval list


mcchilders:
SUITE math will provide a certain number of ballots to be audited per target contest per segment (CVR and no-CVR), so we need to take our typical contest universes and further segment them by CVR/no-CVR to appropriately sample ballots, and then be able to combine these into a single retrieval list while also tracking which ballots are sampled for which.

PR #1093, opened 2021-03-15
Hybrid: draw sample

jonahkagan:
Task: #1054

Draw two samples - one for CVR ballots and one for non-CVR ballots.

### DIFF
--- a/server/api/rounds.py
+++ b/server/api/rounds.py
@@ -499,6 +499,8 @@ class SampleSize(TypedDict):
     size: int
     key: str
     prob: Optional[float]
+    sizeCvr: int  # Only in hybrid audits
+    sizeNonCvr: int  # Only in hybrid audits
 
 
 class BallotDraw(NamedTuple):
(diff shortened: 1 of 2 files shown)


umbernhard (approved):

server/api/rounds.py
```diff
@@ -565,23 +572,35 @@ def draw_sample_for_contest(
...
         # this one. Note that this corresponds to the number of SampledBallotDraws,
         # not SampledBallots.
-        num_previously_sampled = SampledBallotDraw.query.filter_by(
-            contest_id=contest.id
-        ).count()
+        num_previously_sampled = (
+            SampledBallotDraw.query.filter_by(contest_id=contest.id)
+            .join(SampledBallot)
+            .join(Batch)
+            .filter_by(has_cvrs=filter_has_cvrs)
```
benadida, line 579: just checking: you explicitly want this to select for `has_cvrs = null` in the case of ballot comparison and batch comparison?

server/api/rounds.py
```diff
@@ -565,23 +572,35 @@ def draw_sample_for_contest(
...
+            .count()
+        )
 
         # Create the pool of ballots to sample (aka manifest) by combining the
         # manifests from every jurisdiction in the contest's universe.
         manifest = {
             batch_id_to_key[batch.id]: list(range(1, batch.num_ballots + 1))
             for jurisdiction in contest.jurisdictions
             for batch in jurisdiction.batches
+            if batch.has_cvrs == filter_has_cvrs
```
benadida, line 589: seems consistent with above point, so so far so good :)

benadida (approved):
one question about filtering, which I'm pretty sure you got right, but worth confirming.

^^ may want to add a comment in the data model about `has_cvrs`, when is it null, specifically that it seems to be null in a ballot comparison audit even though, of course, there are CVRs.

server/api/rounds.py
jonahkagan: yes, will add a comment

PR #1095, opened 2021-03-16
Adding more tests

MorganLove:
More unit tests

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/index.test.tsx
+++ b/client/src/components/MultiJurisdictionAudit/index.test.tsx
@@ -12,6 +12,8 @@ import {
   prettifyRefreshStatus,
 } from './index'
 import {
+  jurisdictionFileMocks,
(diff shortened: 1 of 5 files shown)


client/src/components/MultiJurisdictionAudit/index.test.tsx
```diff
@@ -127,6 +133,185 @@ describe('AA setup flow', () => {
...
+          <Router>
+            <AuditAdminViewWithAuth />
+          </Router>
+        </AuthDataProvider>
+      )
+
+      // check file upload of jurisdiction
+      await waitFor(() => {
+        expect(queryAllByText(/Upload successfully completed/).length).toBe(1)
+      })
```
jonahkagan, line 295: simpler to use `await screen.findByText(...)` like you did on L311. i think there are a few other places where you use this waitfor/expect/query pattern that could be simplified

jonahkagan (approved):

PR #1097, opened 2021-03-17
Hybrid: calculate risk measurement

jonahkagan:


server/api/rounds.py
```diff
@@ -342,6 +353,68 @@ def sampled_ballot_interpretations_to_cvrs(
...
+    # For targeted contests, count the number of audited ballot draws so far
+    if contest.is_targeted:
+        num_previous_samples_dict = dict(
+            SampledBallotDraw.query.filter_by(contest_id=contest.id)
+            .join(SampledBallot)
+            .join(Batch)
+            .group_by(Batch.has_cvrs)
+            .values(Batch.has_cvrs, func.count(SampledBallotDraw.ticket_number))
+        )
+    # For opportunistic contests, count the number of audited ballots so far
```
umbernhard, line 378: this comment isn't very descriptive of the difference between targeted/opportunistic. Maybe the one for targeted should say "count audited ballots draws **for this contest**", and opportunistic should say "count **any** audited ballot", or something?

server/api/rounds.py
```diff
@@ -342,6 +353,68 @@ def sampled_ballot_interpretations_to_cvrs(
...
+        num_previous_samples_dict = dict(
+            SampledBallot.query.join(Batch)
+            .join(Jurisdiction)
+            .join(Jurisdiction.contests)
+            .filter_by(id=contest.id)
+            .group_by(Batch.has_cvrs)
+            .values(Batch.has_cvrs, func.count(SampledBallot.id))
+        )
+
+    non_cvr_previous_samples = num_previous_samples_dict.get(False, 0)
```
umbernhard, line 389: For later, it would be really nice to make True and False here aliased to a more descriptive name like `HAS_CVR` and `NO_CVR` or whatever.

server/tests/ballot_comparison/test_ballot_comparison.py
```diff
@@ -319,6 +319,116 @@ def test_contest_names_dont_match_cvr_contests(
...
+            target_choices = ([choice_1_1] if vote_choice_1_1 == "1" else []) + (
+                [choice_1_2] if vote_choice_1_2 == "1" else []
+            )
+            audit_ballot(
+                ballot,
+                target_contest_id,
+                (
+                    Interpretation.VOTE
+                    if vote_choice_1_1 != ""
+                    else Interpretation.CONTEST_NOT_ON_BALLOT
```
umbernhard, line 382: This feels weird to me. We're saying that a vote for the loser is the same as the contest not being on the ballot?

umbernhard (approved):

server/api/rounds.py
jonahkagan: hmm, the opportunistic case does actually filter on contest as well, it just does it via the contest universe instead of the SampledBallotDraw.contest_id field

server/api/rounds.py
jonahkagan: interesting, any idea how to do that?

server/tests/ballot_comparison/test_ballot_comparison.py
jonahkagan: this line is just interpreting test data for instructions on how to audit the ballots. it's copied from what we already had in the ballot comparison tests. it just says that if we put an empty string in the test data, audit that as "contest not on ballot"

server/api/rounds.py
jonahkagan: i can clarify that a bit

server/api/rounds.py
umbernhard: we could just define `HAS_CVR = True` and then pass that in.

server/tests/ballot_comparison/test_ballot_comparison.py
umbernhard: ah I gotcha

PR #1100, opened 2021-03-17
Hybrid: cleanup all places we check audit type

jonahkagan:
- JA status box
- Support tools audit type tag
- ballot retrieval list
- round 2 sample size selection


MorganLove (approved):
LGTM :)

PR #1101, opened 2021-03-17
Fixing hanging spinner

MorganLove:
**Description**

- There was a bug where after clicking on Download Audit Report, it opens a new tab to download, and after the file is downloaded, the loader still keeps spinning.

download-audit-report-button

**Testing**

- done

**Progress**

- ready for review

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
+++ b/client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
@@ -291,7 +291,7 @@ describe('StatusBox', () => {
         expect(window.open).toHaveBeenCalledTimes(1)
         expect(window.open).toBeCalledWith(`/api/election/1/report`)
       })
-      expect(downloadReportButton.classList.contains('bp3-loading')).toBe(false)
+      // haven't found a way to close the window in order to test that the spinner ceases to spin
     })
   })
 
@@ -481,9 +481,7 @@ describe('StatusBox', () => {
           '/api/election/1/jurisdiction/1/report'
         )
       })
-      await expect(downloadReportButton.classList.contains('bp3-loading')).toBe(
-        false
-      )
+      // haven't found a way to close the window in order to test that the spinner ceases to spin
     })
 
     cvrAuditTypes.forEach(auditType => {
--- a/client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
+++ b/client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
@@ -278,16 +278,20 @@ describe('StatusBox', () => {
           />
         </Router>
       )
-      fireEvent.click(
-        screen.getByRole('button', { name: 'Download Audit Report' }),
-        {
-          bubbles: true,
-        }
+      const downloadReportButton = screen.getByRole('button', {
+        name: 'Download Audit Report',
+      })
+      fireEvent.click(downloadReportButton, {
+        bubbles: true,
+      })
+      await expect(downloadReportButton.classList.contains('bp3-loading')).toBe(
+        true
       )
       await waitFor(() => {
         expect(window.open).toHaveBeenCalledTimes(1)
         expect(window.open).toBeCalledWith(`/api/election/1/report`)
       })
+      expect(downloadReportButton.classList.contains('bp3-loading')).toBe(false)
     })
   })
 
@@ -462,11 +466,14 @@ describe('StatusBox', () => {
           />
         </Router>
       )
-      fireEvent.click(
-        screen.getByRole('button', { name: 'Download Audit Report' }),
-        {
-          bubbles: true,
-        }
+      const downloadReportButton = screen.getByRole('button', {
+        name: 'Download Audit Report',
+      })
+      fireEvent.click(downloadReportButton, {
+        bubbles: true,
+      })
+      await expect(downloadReportButton.classList.contains('bp3-loading')).toBe(
+        true
       )
       await waitFor(() => {
         expect(window.open).toHaveBeenCalledTimes(1)
@@ -474,6 +481,9 @@ describe('StatusBox', () => {
           '/api/election/1/jurisdiction/1/report'
         )
       })
+      await expect(downloadReportButton.classList.contains('bp3-loading')).toBe(
+        false
+      )
     })
 
     cvrAuditTypes.forEach(auditType => {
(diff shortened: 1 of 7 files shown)


MorganLove:
@jonahkagan let me know what you think about this approach that I took (very different from the first attempt). I'm still working on figuring out the test complexities this introduces, but it works from my manual tests. It spins and prevents double clicks while it's fetching the reports, and then stops when it prompts the download.

MorganLove:
So I tried to figure out a way to close the download window in order to test that the spinner actually stops spinning, but wasn't able to. So I haven't been able to do a regression test for this.

client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
```diff
@@ -278,16 +278,20 @@ describe('StatusBox', () => {
...
+        bubbles: true,
+      })
+      await expect(downloadReportButton.classList.contains('bp3-loading')).toBe(
+        true
       )
       await waitFor(() => {
         expect(window.open).toHaveBeenCalledTimes(1)
         expect(window.open).toBeCalledWith(`/api/election/1/report`)
       })
+      // haven't found a way to close the window in order to test that the spinner ceases to spin
```
jonahkagan, line 294: maybe just call `window.onbeforeunload` to simulate it

jonahkagan, line 294: maybe just call `window.onbeforeunload` to simulate it

client/src/components/MultiJurisdictionAudit/StatusBox/index.tsx
```diff
@@ -78,16 +79,30 @@ const StatusBox: React.FC<IStatusBoxProps> = ({
...
+const downloadAuditAdminReport = (electionId: string) =>
+  new Promise((resolve, reject) => {
+    apiDownload(`/election/${electionId}/report`)
+      .then(() => {
+        resolve('done')
+      })
+      .catch(err => {
+        reject(err)
+      })
+  })
```
jonahkagan, line 91: i dont think this change is necessary - since `apiDownload` returns a promise, then the code that was here previously is equivalent

client/src/components/MultiJurisdictionAudit/StatusBox/index.tsx
```diff
@@ -78,16 +79,30 @@ const StatusBox: React.FC<IStatusBoxProps> = ({
...
+) =>
+  new Promise((resolve, reject) => {
+    apiDownload(`/election/${electionId}/jurisdiction/${jurisdictionId}/report`)
+      .then(() => {
+        resolve('done')
+      })
+      .catch(err => {
+        reject(err)
+      })
+  })
```
jonahkagan, line 105: same here, can revert to what was here before

client/src/components/utilities.ts
```diff
@@ -44,7 +44,19 @@ export const api = async <T>(
...
+  new Promise((resolve, reject) => {
+    try {
+      const windowObj = window.open(`/api${endpoint}`)
+      if (windowObj != null) {
+        windowObj.onbeforeunload = () => {
+          resolve('done')
+        }
+      }
+    } catch (err) {
+      reject(err)
```
jonahkagan, line 57: maybe log the error to the console here so we don't need to catch it in the calling functions

client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
MorganLove: @jonahkagan I had tried that a few different ways, but the `window.open` opens a new window, and I wasn't able to figure out how to trigger `onbeforeunload` for that new window.

jonahkagan: since here we're mocking `window.open`, maybe that mock can return a mock `Window` object representing the new window

client/src/components/utilities.ts
```diff
@@ -44,7 +44,20 @@ export const api = async <T>(
...
+    try {
+      const windowObj = window.open(`/api${endpoint}`)
+      if (windowObj != null) {
+        windowObj.onbeforeunload = () => {
+          resolve('done')
+        }
+      }
+    } catch (err) {
+      // eslint-disable-next-line no-console
+      console.error(err)
```
jonahkagan, line 58: i think you should still reject the promise, otherwise wouldn't it just hang indefinitely and cause the spinner to keep showing?

client/src/components/MultiJurisdictionAudit/StatusBox/index.tsx
```diff
@@ -64,6 +64,7 @@ const StatusBox: React.FC<IStatusBoxProps> = ({
                 <Button
                   intent="success"
                   loading={isSubmitting}
+                  disabled={isSubmitting}
```
jonahkagan, line 67: `loading` automatically disables the button I believe, so you don't need to add this

client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
MorganLove: Tried that too, but it was returning undefined. :stuck_out_tongue_closed_eyes:

jonahkagan: this works for me:

```ts
    it('downloads audit report', async () => {
      const mockDownloadWindow: { onbeforeunload?: () => void } = {}
      window.open = jest.fn().mockReturnValue(mockDownloadWindow)
      render(
        <Router>
          <AuditAdminStatusBox
            rounds={roundMocks.singleComplete}
            startNextRound={jest.fn()}
            jurisdictions={jurisdictionMocks.allComplete}
            contests={contestMocks.filledTargeted.contests}
            auditSettings={auditSettings.all}
          />
        </Router>
      )
      const downloadReportButton = screen.getByRole('button', {
        name: 'Download Audit Report',
      })
      userEvent.click(downloadReportButton)
      expect(downloadReportButton).toBeDisabled()
      await waitFor(() => {
        expect(window.open).toHaveBeenCalledTimes(1)
        expect(window.open).toBeCalledWith(`/api/election/1/report`)
      })
      mockDownloadWindow.onbeforeunload!()
      await waitFor(() => {
        expect(downloadReportButton).toBeEnabled()
      })
    })
```

jonahkagan (approved):

PR #1113, opened 2021-03-19
Suite tests

umbernhard:
**Description**

Adds tests to get SUITE up to 100% test coverage. Also removes unreachable code. 

**Testing**

Added testing for all the corner cases. 

**Progress**

Ready to go.

### DIFF
--- a/server/audit_math/suite.py
+++ b/server/audit_math/suite.py
@@ -102,9 +102,11 @@ class BallotPollingStratum:
         """
 
         n = self.sample_size
-
-        if n == 0:
+        if n == 0 or reported_margin == 0:
             return 1.0
+
+        if self.sample_size == self.num_ballots:
+            return 0.0
 
         sample = bravo.compute_cumulative_sample(self.sample)
         n_w = sample[winner]
@@ -248,9 +250,11 @@ class BallotComparisonStratum:
             pvalue - the pvalue for the hypothesis given the null_lambda
         """
 
-        if self.sample_size == 0:
-            print("No sample size?")
+        if self.sample_size == 0 or reported_margin == 0:
             return 1.0
+
+        if self.sample_size == self.num_ballots:
+            return 0.0
 
         o1, o2, u1, u2 = (
             self.misstatements[(winner, loser)]["o1"],
@@ -316,6 +320,10 @@ def maximize_fisher_combined_pvalue(
 
     V = cvr_winner_votes - cvr_loser_votes + bp_winner_votes - bp_loser_votes
 
+    # The election is tied
+    if V == 0:
+        return 1.0
+
     lambda_lower = (
         np.amax(
             [
@@ -344,6 +352,7 @@ def maximize_fisher_combined_pvalue(
     Wn = bp_sample_winner_votes
     Ln = bp_sample_loser_votes
     Un = bp_stratum.sample_size - bp_sample_winner_votes - bp_sample_loser_votes
+    print(Wn, Ln, Un)
     assert Wn >= 0, f"{Wn, Ln, Un}"
     assert Ln >= 0, f"{Wn, Ln, Un}"
     assert Un >= 0, f"{Wn, Ln, Un}"
@@ -459,7 +468,6 @@ def try_n(
         A Fisher's combined maximized p-value for the given sample size, contest,
         winner-loser pair, and strata.
     """
-
     n1_original = cvr_stratum.sample_size
     n2_original = bp_stratum.sample_size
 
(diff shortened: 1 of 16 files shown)


server/audit_math/suite.py
```diff
@@ -102,9 +102,11 @@ def compute_pvalue(
         """
 
         n = self.sample_size
+        if n == 0 or reported_margin == 0:
+            return 1.0
 
-        if n == 0:
-            return 1
+        if self.sample_size == self.num_ballots:
+            return 0.0
```
jonahkagan, line 109: maybe just use `self.sample_size` everywhere and remove `n` for readability?

server/audit_math/suite.py
```diff
@@ -133,17 +135,14 @@ def compute_pvalue(
...
+        upper_n_w_limit = (self.num_ballots - n_u + null_margin) / 2.0
         lower_n_w_limit = np.max([n_w, n_l + null_margin])
 
         # For extremely small or large null_margins, the limits do not
         # make sense with the sample values.
         if upper_n_w_limit < n_w or (upper_n_w_limit - null_margin) < n_l:
             raise ValueError("Null is impossible, given the sample")
 
-        if lower_n_w_limit > upper_n_w_limit:
-            lower_n_w_limit, upper_n_w_limit = upper_n_w_limit, lower_n_w_limit
```
jonahkagan, line 145: how are you sure this is unreachable?

server/audit_math/suite.py
```diff
@@ -153,13 +152,6 @@ def compute_pvalue(
             )
         )
 
-        # Sometimes the upper_n_w_limit is too extreme, causing illegal 0s.
-        # Check and change the limit when that occurs.
-        if np.isinf(null_logLR(upper_n_w_limit)) or np.isinf(
-            LR_derivative(upper_n_w_limit)
-        ):
-            upper_n_w_limit -= 1
```
jonahkagan, line 161: same q: how do you know it's unreachable?

server/audit_math/suite.py
```diff
@@ -172,10 +164,11 @@ def compute_pvalue(
             nuisance_param = sp.optimize.brentq(
                 LR_derivative, lower_n_w_limit, upper_n_w_limit
             )
+        # print(nuisance_param)
```
jonahkagan, line 167: remove

server/audit_math/suite.py
```diff
@@ -534,24 +532,27 @@ def try_n(
         # Otherwise use the sample we've seen so far
         hyp_sample["hyp_round"] = {
             winner: min(
-                math.ceil(
-                    ((n2 - hyp_sample_size) * cumulative_sample[winner])
-                    / hyp_sample_size
+                int(
```
jonahkagan, line 535: why `int` vs `math.ceil`?

jonahkagan (approved):
i'd still like to go through and review all the math, which i haven't done yet, so this is prob fine to merge. if you want to answer my questions now, feel free, or we can talk about them when i review the math

server/audit_math/suite.py
umbernhard: Ah, yeah I mean to do that. Will clean up.

server/audit_math/suite.py
umbernhard: Yes. Taken with the statement above it, it is mathematically impossible for one of these ifs to be true when the other isn't. Therefore, if we pass the first if, this one can never happen. I worked out the math, I can take a picture of my notes if you would like!

server/audit_math/suite.py
umbernhard: upper_n_w_limit is guaranteed to be a finite number. The only way for it to become infinite is if we do something like divide it by really, really small numbers. If you look in these two functions, that can't ever happen. Again, I worked this out on paper (and with a coding example.

server/audit_math/suite.py
umbernhard: we want our sample to always be conservative, so we want to round votes for the winner down, not up. This was actually an error I had made in my original implementation. I'm using int because we have to do the cast anyways, and `int(math.floor(`is redundant.

jonahkagan: looks like `int` and `math.floor` are slightly different, fyi. dunno if it matters here though: https://stackoverflow.com/questions/31036098/what-is-the-difference-between-int-and-floor-in-python-3

umbernhard: hmm, tbh truncation probably saves us more headaches. So I think it's fine.

PR #1114, opened 2021-03-19
Added tests, removed unreachable code to get BRAVO to 100% coverage

umbernhard:
**Description**

Adds testing and removed unreachable logic to get BRAVO tests to 100% coverage. 

**Testing**

Tweaked some tests to cover corner cases. 

**Progress**

Ready to go.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -6,7 +6,7 @@ omit =
     scripts/*
 
 [report]
-fail_under = 97
+fail_under = 98
 precision = 2
 skip_covered = true
 show_missing = true
--- a/server/audit_math/suite.py
+++ b/server/audit_math/suite.py
@@ -101,15 +101,16 @@ class BallotPollingStratum:
             pvalue: the pvalue from testing the hypothesis that null margin is not the acual margin
         """
 
-        n = self.sample_size
+        if self.sample_size == 0 or reported_margin == 0:
+            return 1.0
 
-        if n == 0:
-            return 1
+        if self.sample_size == self.num_ballots:
+            return 0.0
 
         sample = bravo.compute_cumulative_sample(self.sample)
         n_w = sample[winner]
         n_l = sample[loser]
-        n_u = n - n_w - n_l
+        n_u = self.sample_size - n_w - n_l
 
         v_w = self.vote_totals[winner]
         v_l = self.vote_totals[loser]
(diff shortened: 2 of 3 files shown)


benadida (approved):
looks ok to me, but recommend a @jonahkagan review on this one, too.

server/audit_math/bravo.py
```diff
@@ -342,6 +344,12 @@ def get_sample_size(
 
     samples: Dict = {}
 
+    if round_sizes:
+        num_sampled = sum([round_sizes.get(rnd, 0) for rnd in round_sizes])
+        # If we've already sampled all the ballots, we should never be here
```
jonahkagan, line 349: if we should never be here, shouldn't we just raise an exception?

server/audit_math/bravo.py
```diff
@@ -374,7 +382,7 @@ def get_sample_size(
             best_loser = loser
 
     # If we're in a single-candidate race, set sample to 0
-    if margin["losers"] == 0:
+    if not margin["losers"]:
```
jonahkagan, line 385: why this change?

server/audit_math/bravo.py
```diff
@@ -432,6 +423,7 @@ def get_sample_size(
     if (
         num_ballots > large_election_threshold
         and samples["0.9"]["size"] >= all_ballots_threshold
+        or p_w == p_l
```
jonahkagan, line 426: i think this might error in case that `p_w == p_l` because `samples` won't have key `"0.9"` set.

server/tests/audit_math/test_bravo.py
```diff
@@ -598,3 +672,53 @@ def test_tied_contest():
         "all-ballots": {"type": "all-ballots", "size": 4992420, "prob": None},
     },
 }
+
+
+round2_sample_sizes = {
+    "test1": {
+        "asn": {"type": "ASN", "size": -12, "prob": 0.0},
```
jonahkagan, line 679: what's with all these negative sample sizes?

server/audit_math/bravo.py
umbernhard: I can do that!

server/audit_math/bravo.py
umbernhard: `margin["losers"]` contains an object, it's not a number.

server/audit_math/bravo.py
umbernhard: Hmm, it doesn't in the tests, but that's an interesting point. I did this to get around the linter warning that I was returning too many times.

server/tests/audit_math/test_bravo.py
umbernhard: It's a quirk of the math. I guess we could be rounding them to zero, but these are cases that should never really be seen by Arlo....

PR #1117, opened 2021-03-19
Gets ballot_polling.py to 100% coverage. 

umbernhard:
**Description**
Adds a test to make sure our ballot polling math type logic works. Gets ballot_polling.py to 100% coverage. 

**Testing**
Adds a test to make sure our ballot polling math type logic works. 

**Progress**

Ready.

### DIFF
--- a/server/tests/audit_math/test_ballot_polling.py
+++ b/server/tests/audit_math/test_ballot_polling.py
@@ -41,7 +41,6 @@ def test_get_sample_sizes(contests):
 
 def test_compute_risk(contests):
     for contest in contests:
-        print(contest)
         sample = round1_sample_results[contest]
 
         bravo_test_stat, bravo_decision = bravo.compute_risk(
--- a/server/tests/audit_math/test_ballot_polling.py
+++ b/server/tests/audit_math/test_ballot_polling.py
@@ -40,23 +40,39 @@ def test_get_sample_sizes(contests):
 
 
 def test_compute_risk(contests):
+    print("testing!")
     for contest in contests:
+        print(contest)
         sample = round1_sample_results[contest]
 
         bravo_test_stat, bravo_decision = bravo.compute_risk(
             RISK_LIMIT, contests[contest], sample
         )
 
-        assert bravo_test_stat, bravo_decision == ballot_polling.compute_risk(
+        computed_stat, computed_decision = ballot_polling.compute_risk(
             RISK_LIMIT, contests[contest], sample, AuditMathType.BRAVO, {1: 119}
         )
+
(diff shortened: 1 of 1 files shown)


server/tests/audit_math/test_ballot_polling.py
```diff
@@ -0,0 +1,91 @@
...
+        minerva_sample_size = minerva.get_sample_size(
+            RISK_LIMIT, contests[contest], None, None
+        )
+        assert minerva_sample_size == ballot_polling.get_sample_size(
+            RISK_LIMIT, contests[contest], None, AuditMathType.MINERVA, None
+        )
+
+
+def test_compute_risk(contests):
+    print("testing!")
```
benadida, line 43:
```suggestion
```

server/tests/audit_math/test_ballot_polling.py
```diff
@@ -0,0 +1,91 @@
...
+    for contest in contests:
+        print(contest)
```
benadida, line 45:
```suggestion
```

server/tests/audit_math/test_ballot_polling.py
```diff
@@ -0,0 +1,91 @@
...
+        sample = round1_sample_results[contest]
+
+        bravo_test_stat, bravo_decision = bravo.compute_risk(
+            RISK_LIMIT, contests[contest], sample
+        )
+
+        computed_stat, computed_decision = ballot_polling.compute_risk(
+            RISK_LIMIT, contests[contest], sample, AuditMathType.BRAVO, {1: 119}
+        )
```
benadida, line 54: I would do this kind of test differently, just ensuring that `ballot_polling.compute_risk(...)` calls the proper `bravo.compute_risk`, as this test here could potentially be pretty inefficient... but I can live with it, just a comment.

benadida (approved):

jonahkagan:
I would have expected these two `if` statements to be covered by our integration tests (tests for bravo and minerva audits respectively). Are they not? Unit testing small wrappers like this doesn't seem that worthwhile to me, regardless of whether you do it this way or mock it the way @benadida suggested.

umbernhard:
@jonahkagan the first one is, but the second one isn't. Apparently we don't have a unit test for Minerva where we compute the risk measurement? I could dump this PR if you think I should, but we'd need to add a test somewhere else.

jonahkagan:
I think we temporarily skipped the minerva integration tests because they were causing concurrency issues on CI. maybe we can try bringing them back and see if the issues come back?

umbernhard:
I can look at that, sure. In the mean time I'm gonna close this PR for now.

PR #1118, opened 2021-03-19
Full coverage on sampler.py

umbernhard:
**Description**

Adds a test to get full coverage of sampler.py

**Testing**
Adds a new test with a contest that isn't in every batch. 

**Progress**

Ready.


server/tests/audit_math/snapshots/snap_test_sampler.py
```diff
@@ -20,6 +33,8 @@
     ("0.230651143", "pct 7", 2),
 ]
 
+snapshots["test_draw_macro_sample_wrong_contest 1"] = []
```
jonahkagan, line 36: where does this come from? i can't find a test case that corresponds

server/tests/audit_math/test_sampler.py
```diff
@@ -134,6 +134,31 @@ def test_macro_recount_sample(close_macro_batches, close_macro_contest, snapshot
...
+    }
+
+    other_contest = Contest(name, info_dict)
+
+    for batch in macro_batches:
+        pct = int(batch.split(" ")[-1])
+        if pct < 10:
+            macro_batches[batch]["test2"] = {"cand1": 40, "cand2": 10, "ballots": 50}
+
+    # This should give us zeros for error
```
jonahkagan, line 155: i don't quite understand this comment. can you explain a bit more?

also not totally sure what this test is for - how does it test multiple contests?

umbernhard: I will add a more explanatory comment. Basically there's a branch in macro that happens if the max possible error is 0. One way to make that happen is to compute the error for a batch that doesn't have the contest on its ballots. Here I'm adding a second contest to only some of the batches to make sure that happens.

server/tests/audit_math/snapshots/snap_test_sampler.py
umbernhard: hmm, that's weird. Lemme regen the snapshots for this branch.

jonahkagan (approved):

PR #1119, opened 2021-03-19
Adds new tests to macro to get 100% coverage

umbernhard:
**Description**
Adds new tests to macro to get 100% coverage

**Testing**

New tests for corner cases in MACRO like recounts and worst-case scenarios. 

**Progress**
Ready

### DIFF
--- a/server/tests/audit_math/test_macro.py
+++ b/server/tests/audit_math/test_macro.py
@@ -138,6 +138,48 @@ def test_max_error(contests, batches):
             )
 
 
+def test_get_sizes_extra_contests(contests, batches):
+    name = "test2"
+
+    info_dict = {
+        "cand1": 400,
+        "cand2": 100,
+        "ballots": 500,
+        "numWinners": 1,
+        "votesAllowed": 1,
+    }
+
+    other_contest = Contest(name, info_dict)
+    contests[name] = other_contest
+
+    for batch in batches:
+        if "AV" in batch:
+            continue
+        pct = int(batch.split(" ")[-1])
+        if pct < 10:
+            batches[batch]["test2"] = {"cand1": 40, "cand2": 10, "ballots": 50}
+
+    expected_first_round = {
+        "Contest A": 29,
+        "Contest B": 15,
+        "Contest C": 10,
+        name: 3,
+    }
+
+    # This should give us zeros for error
+    sample = {}
+    for contest in contests:
+        computed = macro.get_sample_sizes(
+            RISK_LIMIT, contests[contest], batches, sample
+        )
+
+        assert (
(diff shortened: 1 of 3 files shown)


server/tests/audit_math/test_macro.py
```diff
@@ -138,6 +138,48 @@ def test_max_error(contests, batches):
             )
 
 
+def test_get_sizes_extra_contests(contests, batches):
```
jonahkagan, line 141: not sure i understand what this test is testing - can you explain?

server/tests/audit_math/test_macro.py
```diff
@@ -208,6 +250,83 @@ def test_get_sample_sizes(contests, batches):
...
 
 
+def test_full_recount(contests, batches):
+    # Do a full recount:
+    sample = batches
+    expected_recount_size = {
+        "Contest A": 1,
+        "Contest B": 1,
+        "Contest C": 1,
+    }
```
jonahkagan, line 260: why would a full recount only result in a sample size of one batch? i would expect the sample size to be the total number of batches

umbernhard: This is default behavior for the sake of being conservative. In this test we're passing in our batches as the sample, so we've already done a full recount but we're asking for more ballots. I could change it to throw an error like I have done elsewhere.

server/tests/audit_math/test_macro.py
umbernhard: this is similar to what's going on in the `sampler` tests. There's a branch we need to test where the collected possible error is zero, so we achieve that by adding a contest that is only in some batches. If we compute the error for that contests in a batch that doesn't contain it, we get zero.

server/tests/audit_math/test_macro.py
jonahkagan: ah i get it. didn't see that we had already done the full recount. yes, i like the consistency of erroring here

server/tests/audit_math/test_macro.py
```diff
@@ -259,15 +259,17 @@ def test_full_recount(contests, batches):
...
+        with pytest.raises(ValueError, match=r"All ballots have already been counted!"):
+            computed = macro.get_sample_sizes(
+                RISK_LIMIT, contests[contest], batches, sample
+            )
+
+            assert (
+                expected_recount_size[contest] == computed
+            ), "Third round sample expected {}, got {}".format(
+                expected_recount_size[contest], computed
+            )
```
jonahkagan, line 272: i think this line is unreachable now, right?

jonahkagan (approved):

PR #1120, opened 2021-03-20
Adding tests

MorganLove:
**Testing**

- Filling in the gaps

### DIFF
--- a/client/cypress/end-to-end/ballot-comparison-failure-cases.spec.js
+++ b/client/cypress/end-to-end/ballot-comparison-failure-cases.spec.js
@@ -217,6 +217,13 @@ describe('Ballot Comparison Failure Cases', () => {
     cy.loginAuditAdmin(auditAdmin)
     cy.findByText(`TestAudit${id}`).click()
     cy.findByText('Review & Launch').click()
+  
+    // custom size validation test
+    cy.findByText('Enter your own sample size (not recommended)').click()
+    cy.findByRole('spinbutton').type('30').blur()
+    cy.findByText('Must be less than or equal to: 15 (the total number of ballots in the contest)')
+
+    cy.get('input[type="radio"]').first().click({force: true})
     cy.findByRole('button', { name: 'Launch Audit' })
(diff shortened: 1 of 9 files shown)


MorganLove:
These tests are being oddly flaky.

client/src/components/DataEntry/SignOff.tsx
```diff
@@ -37,7 +37,7 @@ const SignOff = ({ auditBoard, submitSignoff }: IProps) => {
                 label={`Audit Board Member: ${member.name}`}
               >
                 <LabelText htmlFor={`[${i}]`}>Full Name</LabelText>
-                <NameField name={`[${i}]`} />
+                <NameField name={`[${i}]`} id={`[${i}]`} />
```
jonahkagan, line 40: why this change?

client/cypress/end-to-end/ballot-comparison-failure-cases.spec.js
```diff
@@ -217,6 +217,13 @@ describe('Ballot Comparison Failure Cases', () => {
     cy.loginAuditAdmin(auditAdmin)
     cy.findByText(`TestAudit${id}`).click()
     cy.findByText('Review & Launch').click()
+  
+    // custom size validation test
+    cy.findByText('Enter your own sample size (not recommended)').click()
+    cy.findByRole('spinbutton').type('30').blur()
+    cy.findByText('Must be less than or equal to: 15 (the total number of ballots in the contest)')
```
jonahkagan, line 224: this is frontend validation, right? if so, i think this would be better tested in a unit test, not an end-to-end test, since it doesn't involve communication between front and backend

client/src/components/DataEntry/SignOff.tsx
MorganLove: @jonahkagan Needed to find the input field of 'Full Name' by label, and the reference of the label wasn’t returning anything because the id of the input was not available.

jonahkagan (approved):
yay for more tests!

PR #1121, opened 2021-03-22
Adding audit name to status bar

### LINKED ISSUE #1116: Add audit name to status bar in AA and JA workflows
For both the AA and JA workflows, we'd like to show the name of the audit you're currently working on at the top left of the status bar (see screenshot).

MorganLove:
**Description**

Closes #1116 

**Testing**

- Test that it shows up in the status bars
- Updated snapshots

**Progress**

- Ready to go!

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
+++ b/client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
@@ -286,6 +286,25 @@ describe('StatusBox', () => {
     })
   })
 
+  it('shows audit name', () => {
+    render(
+      <Router>
+        <AuditAdminStatusBox
+          rounds={[]}
+          startNextRound={jest.fn()}
+          jurisdictions={[]}
+          contests={[]}
+          auditSettings={auditSettings.blank!}
+        />
+      </Router>
+    )
+    screen.getByRole('heading', {
+      name: 'Test Audit',
+    })
+    screen.getByText('Audit setup is not complete.')
+    screen.getByText('The audit has not started.')
+  })
+
   describe('JurisdictionAdminStatusBox', () => {
     it('renders unuploaded ballot manifest initial state', () => {
       render(
@@ -297,6 +316,7 @@ describe('StatusBox', () => {
             batchTallies={{ file: null, processing: null }}
             cvrs={{ file: null, processing: null }}
             auditType="BALLOT_POLLING"
+            auditName="Test Audit"
           />
         </Router>
       )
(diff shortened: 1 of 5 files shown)


client/src/components/MultiJurisdictionAudit/StatusBox/index.tsx
```diff
@@ -42,6 +44,7 @@ const StatusBox: React.FC<IStatusBoxProps> = ({
     <Wrapper icon={null}>
       <Inner>
         <div className="text">
+          <H3>{auditName}</H3>
```
jonahkagan, line 47: It looks a bit weird to me to have both the audit name and the headline be styled as headers. I'm not totally sure what a better option would be, but they seem to visually compete.

If we do really want them to both be headers, I think a little more spacing in between them would be helpful.

client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
```diff
@@ -286,6 +286,25 @@ describe('StatusBox', () => {
     })
   })
 
+  it('shows audit name', () => {
```
jonahkagan, line 289: instead of adding a new test case for the audit name, i think it would be simpler to just add an assertion checking for the audit name to one of the existing test cases

client/src/components/MultiJurisdictionAudit/StatusBox/index.tsx
MorganLove: @jonahkagan Yeah, I agree. I tried putting in a `<hr />` between them and it definitely creates a clear distinction. Let me know what you think. @mcchilders if you could weigh in here with design ideas that'd be awesome too.

jonahkagan: The hr definitely creates a distinction, but I worry that it takes up a lot of vertical screen real estate. I have been playing around a bit but haven't come up with much except for adding a little bit more spacing between the two headers.

MorganLove: Good point. I've reduced the spacing around the hr, and I think it looks better. I like having the clear distinction better than just spacing between the headers myself.

jonahkagan: Hmm, but the hr extends all the way to the other end and interferes with the button when there's a button:
Screen Shot 2021-03-24 at 11 25 51 AM

MorganLove: Oh darnit, forgot about the button. :thinking: Okay, I made it shorter as well, and I think that should do it.

mcchilders: Sorry, responded by email but it apparently didn't put it in here! I don't really like the horizontal line, but otherwise it looks fine to me.

jonahkagan (approved):

PR #1122, opened 2021-03-22
Minerva to 100% coverage. 

umbernhard:
**Description**

Adds a new test to Minerva to get 100% coverage. Also tweaks code to eliminate unnecessary function calls. 

**Testing**

Tests the round size function. 

**Progress**

Ready.


server/audit_math/minerva.py
```diff
@@ -117,7 +110,7 @@ def get_sample_size(
     >>> get_sample_size(20, c3, None, [])
     {'0.7': {'type': None, 'size': 87, 'prob': 0.7}, '0.8': {'type': None, 'size': 110, 'prob': 0.8}, '0.9': {'type': None, 'size': 156, 'prob': 0.9}}
     >>> get_sample_size(10, c3, make_sample_results(c3, [[55, 40, 3]]), {1: 100})
-    {'0.9': {'type': None, 'size': 150, 'prob': 0.9}}
+    {'0.9': {'type': None, 'size': 225, 'prob': 0.9}}
```
jonahkagan, line 113: why did this change?

jonahkagan (approved):

server/audit_math/minerva.py
umbernhard: I took out a function to compute the next round sample size and put it inline (it was only one line so I didn't see a reason not to). when I did that, I realized that it was computing zero-indexed rounds, and for consistency I changed it to be 1-indexed. That means that this number increased by a factor of 1.5 (the multiple we're using for round size escalation).

jonahkagan: Does that mean it was incorrect before? I wouldn't expect a refactor to change the output for the same input

umbernhard: I would argue that it does, but ultimately it's not that important: since we're just estimating sample sizes so long as we aren't overshooting or undershooting by a ton, it's not that important. I think it's more important to have consistency in how we number rounds.

PR #1127, opened 2021-03-22
New  tests to get BRAVO up to 100%

umbernhard:
Description

Adds testing and removed unreachable logic to get BRAVO tests to 100% coverage.

Testing

Tweaked some tests to cover corner cases.

Progress

Ready to go.


server/audit_math/bravo.py
```diff
@@ -406,30 +401,26 @@ def get_sample_size(
...
+            size = bravo_sample_sizes(alpha, p_w, p_l, sample_w, sample_l, quant)
+            samples[str(quant)] = {"type": None, "size": size, "prob": quant}
 
     # If the computed sample size is a good chunk of the ballots, recommend
     # auditing all ballots, since this is actually less work than auditing a
     # large proportion (for large elections).
     large_election_threshold = 100000
     all_ballots_threshold = num_ballots * 0.25
-    if (
+    if p_w == p_l or (
```
jonahkagan, line 423: maybe update the comment to include ties?

server/audit_math/suite.py
```diff
@@ -163,6 +164,7 @@ def compute_pvalue(
             nuisance_param = sp.optimize.brentq(
                 LR_derivative, lower_n_w_limit, upper_n_w_limit
             )
+        # print(nuisance_param)
```
jonahkagan, line 167: remove

server/audit_math/bravo.py
```diff
@@ -342,6 +344,12 @@ def get_sample_size(
 
     samples: Dict = {}
 
+    if round_sizes:
+        num_sampled = sum([round_sizes.get(rnd, 0) for rnd in round_sizes])
```
jonahkagan, line 348: small nit: pretty sure it's impossible for `round_sizes.get(rnd, 0)` to hit the default case and give 0, since you're iterating over `round_sizes`. You could probably just say `sum(round_sizes.values())`. not a big deal though

server/audit_math/bravo.py
```diff
@@ -406,23 +411,19 @@ def get_sample_size(
 
         return samples
 
-    # If we haven't seen anything yet, initialize sample_w and sample_l
-    if not cumulative_sample:
-        sample_w = 0
-        sample_l = 0
-    else:
+    if p_w != p_l:
```
jonahkagan, line 414: this `if` is no longer necessary i believe

server/tests/audit_math/test_bravo.py
```diff
@@ -210,11 +232,28 @@ def test_bravo_sample_sizes_round1_incomplete():
 
 def test_get_sample_size(contests):
     for contest in contests:
+        print(contest)
```
jonahkagan: remove

server/tests/audit_math/test_bravo.py
```diff
@@ -563,14 +639,19 @@ def test_tied_contest():
...
     },
-    "test3": {"asn": {"type": "ASN", "size": 1, "prob": 1.0},},
-    "test4": {"asn": {"type": "ASN", "size": 1, "prob": 1.0},},
-    "test5": {
-        "asn": {"type": "ASN", "size": 1000, "prob": 1},
-        "0.7": {"type": None, "size": 1000, "prob": 0.7},
-        "0.8": {"type": None, "size": 1000, "prob": 0.8},
-        "0.9": {"type": None, "size": 1000, "prob": 0.9},
+    "test3": {
+        "asn": {"type": "ASN", "size": -1.0, "prob": -1.0},
```
jonahkagan, line 649: why did these all change to size -1.0? seems weird that it's negative and that it's a float

server/tests/audit_math/test_bravo.py
```diff
@@ -598,3 +679,53 @@ def test_tied_contest():
         "all-ballots": {"type": "all-ballots", "size": 4992420, "prob": None},
     },
 }
+
+
+round2_sample_sizes = {
```
jonahkagan, line 690: it kind of seems like all these round 2 sample sizes that are negative shouldn't be allowed. maybe it's not easy to disallow that though?

at least an explanatory comment would be helpful

server/tests/audit_math/test_bravo.py
umbernhard: you're right about the float. The -1 is because these are single-candidate races and invalid. I could throw an exception instead.

jonahkagan: Exceptions seem clearer to me

server/tests/audit_math/test_bravo.py
umbernhard: I fixed it so that no negative sample sizes will get returned, and cases where it would have happened result in errors being thrown.

jonahkagan (approved):
nice work!

PR #1129, opened 2021-03-23
Reprocess CVR file when ballot manifest reuploaded

jonahkagan:
Task: #1124 

Since the CVR file depends on the batches created from the manifest, if the manifest changes, we need to reprocess the CVR file.

### DIFF
--- a/server/api/ballot_manifest.py
+++ b/server/api/ballot_manifest.py
@@ -1,4 +1,5 @@
 import uuid
+import logging
 from datetime import datetime
 from sqlalchemy.orm.session import Session
 from flask import request, jsonify, Request
@@ -16,6 +17,9 @@ from ..util.process_file import (
 from ..util.csv_download import csv_response
 from ..util.csv_parse import decode_csv_file, parse_csv, CSVValueType, CSVColumnType
 from ..audit_math.suite import HybridPair
+from .cvrs import process_cvr_file
+
+logger = logging.getLogger("arlo")
 
 CONTAINER = "Container"
 TABULATOR = "Tabulator"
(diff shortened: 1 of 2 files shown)


server/api/ballot_manifest.py
```diff
@@ -115,6 +119,36 @@ def process():
...
+                Batch.query.filter_by(jurisdiction_id=jurisdiction.id)
+                .with_entities(Batch.id)
+                .subquery()
+            )
+        ).delete(synchronize_session=False)
+        jurisdiction.cvr_contests_metadata = None
+        jurisdiction.cvr_file.processing_started_at = None
+        jurisdiction.cvr_file.processing_completed_at = None
+        jurisdiction.cvr_file.processing_error = None
+        # Because process_cvr_file uses a COPY command outside of the session
```
benadida, line 140: oh booooooo. Are we sure we can't get COPY to run in the current transaction? Is it a sqlalchemy limitation?

(I understand this isn't a new issue, so we don't have to fix it now, but still makes me cry.)

server/api/ballot_manifest.py
```diff
@@ -115,6 +119,36 @@ def process():
 
     process_file(session, file, process)
 
+    # If CVR file already uploaded, try reprocessing it, since it depends on
+    # batch names from the manifest
+    if jurisdiction.cvr_file:
```
benadida, line 124: should we always reprocess the CVR file, or only if the batch set has changed? Or maybe we also need to if the numbers have changed?

I suppose it won't kill us to reprocess, so I can live with this solution if it's simpler.

benadida (approved):

server/api/ballot_manifest.py
jonahkagan: I know it's sad, but I really do think we're in ok shape here. If the CVR reprocessing fails, the file will get picked up and tried again by bgcompute. In fact, we don't even _need_ to call process_cvr_file here explicitly, though I think it's clearer to do so.

I can look into trying to run COPY in the current transaction, though I couldn't get it to work the first time around.

server/api/ballot_manifest.py
jonahkagan: Yeah we could be smarter about it, but I don't see a huge advantage. Right now I'm using the dumb approach for all the file reprocessing.

PR #1131, opened 2021-03-23
Reprocess batch tallies file when ballot manifest reuploaded

jonahkagan:
Task: #1130 

Since the batch tallies file depends on the batches created from the manifest, if the manifest changes, we need to reprocess the batch tallies file.

### DIFF
--- a/server/tests/batch_comparison/snapshots/snap_test_batch_comparison.py
+++ b/server/tests/batch_comparison/snapshots/snap_test_batch_comparison.py
@@ -8,15 +8,15 @@ from snapshottest import Snapshot
 snapshots = Snapshot()
 
 snapshots["test_batch_comparison_round_1 1"] = {
-    "numSamples": 14,
+    "numSamples": 10,
     "numSamplesAudited": 0,
-    "numUnique": 5,
+    "numUnique": 4,
     "numUniqueAudited": 0,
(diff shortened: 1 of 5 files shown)


server/tests/batch_comparison/snapshots/snap_test_batch_comparison.py
```diff
@@ -8,15 +8,15 @@
 snapshots = Snapshot()
 
 snapshots["test_batch_comparison_round_1 1"] = {
-    "numSamples": 14,
+    "numSamples": 10,
```
benadida, line 11: why did these change?

benadida (approved):

server/tests/batch_comparison/snapshots/snap_test_batch_comparison.py
jonahkagan: due to the changes i had to make in the test data in test_batch_comparison.py below.

the test in question was uploading the manifests and cvrs via standard fixtures and then updating the ballot manifest to something different than the test fixtures. since the new ballot manifest didn't match the cvrs, this code change caused the test to break.

the simplest fix was to remove the part of that test that changed the manifest. i confirmed that the modified manifest wasn't necessary. that lowered the sample size for that test, which changed these snapshots.

PR #1133, opened 2021-03-23
suite review

jonahkagan:
delete suite.py in order to review line by line


server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-        np.amax(
-            [
-                cvr_winner_votes - cvr_loser_votes - cvr_stratum.num_ballots,
-                V - (bp_winner_votes - bp_loser_votes + bp_stratum.num_ballots),
-            ]
-        )
-        / V
-    )
-    lambda_upper = (
-        np.amax(
```
jonahkagan, line 324: this should be `np.amin`

server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-            try:
-                pvalue1 = np.min(
-                    [
-                        1,
-                        cvr_stratum.compute_pvalue(
-                            reported_margin, winner, loser, test_lambda
-                        ),
-                    ]
-                )
-            except ValueError:
```
jonahkagan, line 375: when would we see a ValueError and why return 0 in that case?

server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-    Wn = bp_sample_winner_votes
-    Ln = bp_sample_loser_votes
-    Un = bp_stratum.sample_size - bp_sample_winner_votes - bp_sample_loser_votes
-    assert Wn >= 0, f"{Wn, Ln, Un}"
-    assert Ln >= 0, f"{Wn, Ln, Un}"
-    assert Un >= 0, f"{Wn, Ln, Un}"
-
-    T2 = (
-        lambda delta: 2
-        * bp_stratum.sample_size
```
jonahkagan, line 348: this should be `cvr_stratum.sample_size`

server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-        * np.log(1 + reported_margin * delta / (2 * bp_stratum.num_ballots * GAMMA))
```
jonahkagan, line 349: `bp_stratum.num_ballots` should be `cvr_stratum.num_ballots`

server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-    """
-    maximized_pvalue = 0.0
-    # find range of possible lambda
-    cvr_winner_votes = cvr_stratum.vote_totals[winner]
-    bp_winner_votes = bp_stratum.vote_totals[winner]
-
-    cvr_loser_votes = cvr_stratum.vote_totals[loser]
-    bp_loser_votes = bp_stratum.vote_totals[loser]
-
-    V = cvr_winner_votes - cvr_loser_votes + bp_winner_votes - bp_loser_votes
```
jonahkagan, line 308: Is `V` equivalent to `reported_margin` below?  I believe it is.

If yes, then I think you can remove `vote_totals` from the ballot comparison stratum entirely, since they aren't used anywhere else.

server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-        else:
-            nuisance_param = sp.optimize.brentq(
-                LR_derivative, lower_n_w_limit, upper_n_w_limit
-            )
-        logLR = alt_logLR - null_logLR(nuisance_param)
-        LR = float(np.exp(logLR))  # This value is always a float, but np.exp
-        # can return a vector. casting for the typechecker.
-        # Note if this value overflows, the p-value becomes 0.
-
-        return 1.0 / LR if 1.0 / LR < 1 else 1.0
```
jonahkagan, line 171: nit: why not `min(1.0/LR, 1.0)` like they have in the original code?

server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-                        ),
-                    ]
-                )
-            except ValueError as e:
-                # If the sprt throws an error, set its pvalue to 0.
-                # This is per the Stark code
-                if "Alternative" in str(e):
-                    pvalue2 = 1.0
-                else:
-                    pvalue2 = 0
```
jonahkagan, line 393: I couldn't find this logic in the original code. Where did you see it? Also curious why the two different cases here based on the error message

server/audit_math/suite.py
umbernhard: that returns `Optional` which causes mypy to freak out

server/audit_math/suite.py
umbernhard: I think that's correct.

server/audit_math/suite.py
umbernhard: it feels like there's a broader discussion of why this is a max/min at all. It seems to me like the first value is the right one per the paper. The other value is useful if there are a lot of no-vote ballots in the CVR stratum I guess? In which case max seems right to me?

server/audit_math/suite.py
umbernhard: Good catch. Idk why they called it T2, doesn't this correspond to T1 in the paper?

server/audit_math/suite.py
umbernhard: If lambda is overly large, it can cause the p-value calculation to try to take a log of a negative value. In that case, we're saying that the sample is so good that the alternative (i.e. the reported margin) is impossible. I guess that means we should be setting the p-value to 1 instead. Good catch.

server/audit_math/suite.py
umbernhard: Here's the corresponding code for the null error: https://github.com/pbstark/CORLA18/blob/master/code/sprt.py#L94

umbernhard: For the alternative, it's roughly here: https://github.com/pbstark/CORLA18/blob/master/code/sprt.py#L60

Following the same logic as above, if the alternative isn't compatible with the sample, something is really wacky, so return a p-value of 1.

server/audit_math/suite.py
jonahkagan: I just tried it and mypy didn't complain. Maybe the issue was using `np.amin` vs just the builtin `min`?

server/audit_math/suite.py
jonahkagan: Hmm, in section 2.2 of the paper, it says:
> Some values of λ can be ruled out a priori, because (for instance) ωwℓ,s ≤ Vwℓ,s + Ns, where Ns is the number of ballots cast in stratum s, and thus 1 − (Vwℓ2 + N2) / Vwℓ ≤ λ ≤ (Vwℓ,1 + N1) / Vwℓ

Due to the "for instance," I read that as a non-exhaustive exploration of bounds - i.e. there could be other bounds as well.

In the comment for the function that the code here is based on, there's a derivation that gives the bounds implemented in the code. Seems reasonable to me to trust that.

My guess as to why they use `min` for the upper bound is that if both bounds are valid, whichever one is smaller is stricter and therefore more useful.

server/audit_math/suite.py
jonahkagan: I couldn't find T1 in the paper, except in the section that defined it as "Let Tj be the value of tp for the batch p selected in the jth draw," which seems different from this.

What I did find is that they refer to alternative test statistic functions (Liptak's or Tippet's) as T, so maybe T has something to do with that?

server/audit_math/suite.py
jonahkagan: Should that just be handled within the p-value computing function?

server/audit_math/suite.py
jonahkagan: Similar comment to above - should the p-value function itself make these decisions?

server/audit_math/suite.py
umbernhard: Oh maybe so. I can update it.

server/audit_math/suite.py
umbernhard: Fair point. I can fix that.

server/audit_math/suite.py
umbernhard: aha, that makes total sense. Will fix.

server/audit_math/suite.py
umbernhard: T1 is in Appendix A on page 12 (as an expected value ET1). I think it's T for test statistic, 1 for stratum 1 (ballot comparison in this case). I dunno though.

server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-        contest         - the overall contest information
-        bp_stratum: a stratum object containing the ballot polling stratum
-        cvr_stratum: a stratum object containing the ballot comparison stratum
-    Outputs:
-        sample_sizes    - A Tuple of (cvr_strata_size, no_cvr_strata_size).
-    """
-
-    alpha = float(risk_limit) / 100
-
-    for winner, loser in product(contest.winners, contest.losers):
```
jonahkagan, line 583: i'm not sure i understand how this approach works. it seems like you perform the search for the right sample sizes for each winner/loser pair and then return the sample sizes for whichever winner/loser pair was tested last.

i.e. i think could refactor this function to look like this:

```python
def find_sample_sizes(winner, loser):
  ... everything that happens in the body of the for loop ...

for winner, loser in product(winners, losers):
  sample_sizes = find_sample_size(winner, loser)
return sample_sizes
```

it seems like the iterating over winner/loser pairs should happen within the p-value calculation when trying a sample size. something more like this:

```python
def hypothetical_pvalue(candidate_sample_size):
  hypothetical_strata = ... set up strata ...
  return compute_risk(hypothetical_strata) # reuse function that already knows how to compute risk across winner/loser pairs

candidate_sample_size = num_sampled
candidate_pvalue = 1
while candidate_pvalue > risk_limit:
  candidate_sample_size = ... some next sample size to try ...
  candidate_pvalue = hypothetical_pvalue(candidate_sample_size)
  
return candidate_sample_size
```

server/audit_math/suite.py
jonahkagan: Yeah that's where I got the quote I cited from - just a few lines before that. Little t is for taint there.

server/audit_math/suite.py
umbernhard: Ope, that's a bug. I think what you're proposing is the right approach.

server/audit_math/suite.py
umbernhard: Ah, no we do need cvr_winner/loser votes below.

server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-        )
-
-        ballots_to_sample = num_sampled
-
-        expected_pvalue = 1.0
-
-        # this allows us to exactly match CORLA18's estimate_n and estimate_escalation_n
-        coefficient = 1.1
-        if bp_stratum.sample_size == 0 and cvr_stratum.sample_size == 0:
-            coefficient = 2.0
```
jonahkagan, line 598: do you think the use of different coefficients was intentional? or just a typo or something?

server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-
-    for winner, loser in product(contest.winners, contest.losers):
-        n_ratio = cvr_stratum.num_ballots / (
-            cvr_stratum.num_ballots + bp_stratum.num_ballots
-        )
-        num_sampled = max(
-            MIN_SAMPLE_SIZE, cvr_stratum.sample_size + bp_stratum.sample_size
-        )
-
-        ballots_to_sample = num_sampled
```
jonahkagan, line 591: any reason to have `num_sampled` and then rename it `ballots_to_sample`?

server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-
-        expected_pvalue = 1.0
-
-        # this allows us to exactly match CORLA18's estimate_n and estimate_escalation_n
-        coefficient = 1.1
-        if bp_stratum.sample_size == 0 and cvr_stratum.sample_size == 0:
-            coefficient = 2.0
-
-        # step 1: linear search, increasing n by a factor of 1.1 each time
```
jonahkagan, line 600: update comment here and below to replace 1.1 with "coefficient"

jonahkagan, line 600: this comment and the one below are slightly inaccurate since its not always 1.1

server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-        low_n = ballots_to_sample / coefficient
-        high_n = ballots_to_sample
-        mid_pvalue = 1.0
-        # TODO: do we need this tolerance?
-        risk_limit_tol = 0.8
-        # while (mid_pvalue > alpha) or (expected_pvalue is np.nan):
-        while (
-            (mid_pvalue > alpha)
-            or (mid_pvalue < risk_limit_tol * alpha)
-            or (expected_pvalue is np.nan)
```
jonahkagan, line 631: is this supposed to be `expected_pvalue` or `mid_pvalue`?

server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-                bp_stratum,
-                cvr_stratum,
-                n_ratio,
-            )
-
-        # step 2: bisection between n/1.1 and n
-        low_n = ballots_to_sample / coefficient
-        high_n = ballots_to_sample
-        mid_pvalue = 1.0
-        # TODO: do we need this tolerance?
```
jonahkagan, line 625: i think i understand the intention behind it - to provide a lower bound for the expected p value so that we don't end up with a bigger sample than necessary to meet the risk limit

server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-        o1_rate = cvr_stratum.misstatements[(winner, loser)]["o1"] / n1_original
-        o2_rate = cvr_stratum.misstatements[(winner, loser)]["o2"] / n1_original
-        u1_rate = cvr_stratum.misstatements[(winner, loser)]["u1"] / n1_original
-        u2_rate = cvr_stratum.misstatements[(winner, loser)]["u2"] / n1_original
-
-    n1 = math.ceil(n_ratio * n)
-    n2 = int(n - n1)
-
-    if (n1 < n1_original) or (n2 < n2_original):
-        return 1.0
```
jonahkagan, line 472: i didn't see this case in the original code. curious why you needed to add it

server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-        + cvr_stratum.misstatements[(winner, loser)]["o2"]
-    )
-    u1 = (
-        math.floor(u1_rate * (n - n1_original))
-        + cvr_stratum.misstatements[(winner, loser)]["u1"]
-    )
-    u2 = (
-        math.floor(u2_rate * (n - n1_original))
-        + cvr_stratum.misstatements[(winner, loser)]["u2"]
-    )
```
jonahkagan, line 489: nit: could pull `cvr_stratum.misstatements[(winner, loser)]` out into a variable for readability

server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-    cumulative_sample = bravo.compute_cumulative_sample(bp_stratum.sample)
-
-    # Add fake ballots to the hypothetical sample:
-    if prev_sample_size == 0:
-        # If no ballots have been sampled, assume the sample is roughly
-        # the margin
-        hyp_sample["hyp_round"] = {
-            winner: min(
-                int(n2 * (bp_stratum.vote_totals[winner] / bp_stratum.num_ballots)),
-                bp_stratum.vote_totals[winner],
```
jonahkagan, line 520: why the `min` with `bp_stratum.vote_totals[winner]` - seemed like the original code just used the value on L519

server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-            ),
-            loser: min(
-                math.ceil(
-                    (n2 - prev_sample_size)
-                    * (cumulative_sample[loser] / prev_sample_size)
-                ),
-                bp_stratum.vote_totals[loser],
-            ),
-        }
```
jonahkagan, line 546: my reading of the original code is that they create this expected new sample results and add it to the observed sample so far: https://github.com/pbstark/CORLA18/blob/master/code/suite_tools.py#L639-L641

so here i think that would mean taking these values and adding `cumulate_sample[winner or loser]` to them, right?

jonahkagan: oh wait, nevermind, this is just for the hypothetical new round. ignore

server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-
-    hyp_sample_size = n2
-    hyp_sample_size = min(bp_stratum.num_ballots, hyp_sample_size)
```
jonahkagan, line 549: why this `min`? couldn't find in the orig code

server/audit_math/suite.py
umbernhard: I think it's intentional. By limiting it for subsequent rounds you do your best to get lower sample sizes?

server/audit_math/suite.py
umbernhard: probably not!

server/audit_math/suite.py
```diff
@@ -1,718 +0,0 @@
...
-
-        # step 1: linear search, increasing n by a factor of 1.1 each time
-        while (expected_pvalue > alpha) or (expected_pvalue is np.nan):
-            ballots_to_sample = int(coefficient * ballots_to_sample)
-            if ballots_to_sample > contest.ballots:
-                cvr_ballots_to_sample = math.ceil(n_ratio * contest.ballots)
-                bp_ballots_to_sample = int(contest.ballots - cvr_ballots_to_sample)
-                return HybridPair(
-                    cvr=cvr_ballots_to_sample, non_cvr=bp_ballots_to_sample
-                )
```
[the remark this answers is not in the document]
umbernhard: At that point it's probably more efficient to do a full recount rather than count like 75% of your ballots. Also it's hard to do the computations with all the ballots, as it leads to zeros in denominators and stuff, which I suspect is really why this case is here.

server/audit_math/suite.py
umbernhard: `expected_pvalue`: https://github.com/pbstark/CORLA18/blob/master/code/suite_tools.py#L421

Since the first loop breaks if it's `nan`, it might be `nan` in the first iteration and cause the other two checks to fail.

server/audit_math/suite.py
umbernhard: It's here: https://github.com/pbstark/CORLA18/blob/master/code/suite_tools.py#L616

Basically if our new sample sizes (which should be cumulative and include previous rounds) are lower than the ones we started with, something has gone terrible wrong.  Typically it happens during the linear search, so this just makes sure we get back to where we're supposed to be.

server/audit_math/suite.py
umbernhard: I suppose it probably doesn't matter, but I didn't want to be sampling more ballots than we had, since I don't know what that means.

server/audit_math/suite.py
umbernhard: again, doesn't make sense to sample more ballots than we have. Helps avoid some degenerate cases.

server/audit_math/suite.py
jonahkagan: i think the first loop actually _continues_ if `expected_pvalue` is `nan`

when is a pvalue `nan`? that might help us make sense of this

server/audit_math/suite.py
jonahkagan: yeah, that does seem like the tradeoff. i wonder if there's a way to simulate with/without and get more info about it's usefulness

server/audit_math/suite.py
umbernhard: Oh shoot you're right. Looking at the code again, I'm not sure we can actually encounter a `nan`? Like it should only happen if a p-value is negative, which can't actually happen. So we can probably take these checks out.

Issue #1107, opened 2021-03-18
Bug - Retrieval list sortation thinks "Batch 11" comes before "Batch 2"


mcchilders:
See retrieval list attached - sorting looks great, container, then batch, then record ... except that for some reason it thinks Batch 11 comes before Batch 2 ...

jonahkagan:
It sorted by tabulator before batch name, that's why.

jonahkagan:
Also it sorts "lexicographically", which means it treats digits like letters, not like numbers. Since 1 comes before 2, 11 also comes before 2 (in the same way that "aa" comes before "b").

PR #1134, opened 2021-03-24
"Human sort" ballots

jonahkagan:
Task: #1107 

Container, tabulator, and batch names (as well as audit board names) often contain text and numbers. Regular alphanumeric sorting orders numbers in a way that humans wouldn't - e.g. 10 comes before 2.

To order these in a human-friendly way, we implement a database-level human_sort function.


benadida (approved):
Nice.

I have a small concern that sorting by the function could make the query slow, since there's no way to index it, but... it's probably good enough. Let's keep an eye on slow queries at some point.

PR #1135, opened 2021-03-24
Suite revamp

umbernhard:
**Description**

Implementing changes discussed in #1133 

**Testing**

Adds new tests to catch some previously uncaught problems. 

**Progress**

Ready?


server/audit_math/suite.py
```diff
@@ -631,13 +629,12 @@ def get_sample_size(
...
-
-        max_bp_sample_size = max(max_bp_sample_size, bp_ballots_to_sample)
-        max_cvr_sample_size = max(max_cvr_sample_size, cvr_ballots_to_sample)
+        cvr_ballots_to_sample = int(math.ceil(n_ratio * high_n))
+        bp_ballots_to_sample = int(math.ceil(high_n - cvr_ballots_to_sample))
+        sample_sizes.append((cvr_ballots_to_sample, bp_ballots_to_sample))
 
-    return HybridPair(cvr=max_cvr_sample_size, non_cvr=max_bp_sample_size)
+    sample_size = sorted(sample_sizes, key=sum, reverse=True)[0]
+    return HybridPair(cvr=sample_size[0], non_cvr=sample_size[1])
```
jonahkagan, line 637: crazy idea: since `n_ratio` is constant throughout the search, just append `high_n` to the list of sample sizes, let `sample_size = max(sample_sizes)` here, and then do the splitting by stratum last. less identical to the orig code but maybe simpler? (i know it's bikeshedding at this point...)

server/audit_math/suite.py
```diff
@@ -631,20 +619,22 @@ def get_sample_size(
             or (expected_pvalue is np.nan)
         ):
             mid_n = int(np.floor((low_n + high_n) / 2))  # cast for typechecker
-            if mid_n in [low_n, high_n]:
-                break
             mid_pvalue = try_n(
                 mid_n, alpha, contest, winner, loser, bp_stratum, cvr_stratum, n_ratio,
             )
+            if mid_n in [low_n, high_n]:
+                break
```
jonahkagan, line 626: why this change?

server/audit_math/suite.py
umbernhard: you're not wrong. I think for comparability to the CORLA code we should keep it like this.

server/audit_math/suite.py
umbernhard: I was debugging and needed to see the value before it broke. I put it back!

server/audit_math/suite.py
jonahkagan: cool, sounds good. in that case how would you feel about replacing `sorted(...)` with `max` for clarity? (`max` also takes a `key` function)

umbernhard: I tried that but the type checker complained

jonahkagan: ugh so annoying! i got this to work: `sample_size = max(sample_sizes, key=lambda pair: sum(pair))`. take it or leave it!

umbernhard: lol now the linter is complaining that it's an unnecessary lambda :joy:

umbernhard: I'm just gonna stick with `sorted`

jonahkagan (approved):

Issue #1115, opened 2021-03-19
Add "delete" function & confirmation modal for old audits


mcchilders:
From the View/Create New Audit screen, the AA should be able to delete any audit in their organization. Let's add a "delete" icon to the end of the buttons for each audit that, when clicked, triggers a confirmation modal where the AA can either cancel or confirm deletion of the audit. The modal should also warn the user that the action cannot be undone.

"Are you sure you want to delete <Audit Name?> This action cannot be undone. Cancel/Delete"

![Screen Shot 2021-03-19 at 2.52.04 PM.png](https://zube.io/files/votingworks/fd29f8feae587549e96f719844caa06a-screen-shot-2021-03-19-at-2-52-04-pm.png)

MorganLove:
@Jonah-Kagan Do we have a delete endpoint for these? I was just looking in elections.py and only saw a POST endpoint. Or is there a different way we should be deleting these?

jonahkagan:
We don't have an endpoint to delete audits yet

PR #1137, opened 2021-03-25
Delete audit button

jonahkagan:
Task: #1115 

- Endpoint to delete an audit
- Delete audit button on Home Screen

For now, we do a soft delete - we set a flag `Election.deleted_at` in the database, but don't actually delete it. That way it's easier to restore if someone accidentally deletes an audit.

https://user-images.githubusercontent.com/530106/112557928-d3011980-8d8a-11eb-89e5-0c60d0fcdc83.mov


server/models.py
```diff
@@ -155,6 +155,10 @@ class Election(BaseModel):
     )
     standardized_contests = Column(JSON)
 
+    # When a user deletes an audit, we keep it in the database just in case
+    # they change their mind, but flag it so that we can restrict access
+    is_deleted = Column(Boolean)
```
benadida, line 160: suggestion: either make this not-null default false, or make it `deleted_at` with a nullable timestamp, which then gives you extra information about when the delete happened.

benadida (approved):
one suggestion.

server/models.py
jonahkagan: good call. I originally made it not-null default false, but then got lazy when working on the migration. I actually like the timestamp better - it matches the pattern we've been using - can't believe I didn't think of that!

benadida (approved):

PR #1143, opened 2021-03-30
Hybrid: enable round 2 test

jonahkagan:


server/tests/hybrid/snapshots/snap_test_hybrid.py
```diff
@@ -73,6 +73,76 @@
...
+Jurisdiction Name,Audit Board Name,Member 1 Name,Member 1 Affiliation,Member 2 Name,Member 2 Affiliation\r
+J1,Audit Board #1,,,,\r
+J2,Audit Board #1,,,,\r
+\r
+######## ROUNDS ########\r
+Round Number,Contest Name,Targeted?,Sample Size,Risk Limit Met?,P-Value,Start Time,End Time,Audited Votes\r
+1,Contest 1,Targeted,22,No,1.0,DATETIME,DATETIME,Choice 1-1: 11; Choice 1-2: 3\r
+1,Contest 2,Opportunistic,,No,1.0,DATETIME,DATETIME,Choice 2-1: 5; Choice 2-2: 1; Choice 2-3: 1\r
+2,Contest 1,Targeted,49,Yes,0,DATETIME,DATETIME,Choice 1-1: 4; Choice 1-2: 2\r
+2,Contest 2,Opportunistic,,Yes,0,DATETIME,DATETIME,Choice 2-1: 0; Choice 2-2: 0; Choice 2-3: 0\r
```
jonahkagan, line 101: @umbernhard kind of surprised I got a p-value of 0 in this integration test. Any idea what case might be triggering this?

jonahkagan: maybe all ballots got sampled? there are 50 ballots total. the sample sizes add up to more than 50. at the same time, ut there are less than 50 unique ballots in the table that got sampled according to the table below

jonahkagan: When I count the ballots below that are from the non-CVR stratum (all the ones from Tabulator 3), I count 20, which would be all the ballots in the stratum, which would explain the p-value of 0 according to what you said above? It does seem like that shouldn't necessarily yield a p-value of 0.

umbernhard: Ah I missed the tabulator field. If we've counted all the ballots for the stratum, the p-value should be zero for that stratum. Even if I remove the branch that sets the fisher p-value to 0, doing the computation still results in an overall p-value of zero. I think this is a corner case that I'm not really sure how to handle. We could force the math to just return the other p-value if one of the p-values is zero.

umbernhard (approved):

PR #1144, opened 2021-03-31
Suite fixes

umbernhard:
**Description**

Adds handling for a corner case where one stratum gets recounted but the other doesn't. 

**Testing**

Adds new test for this corner case.

**Progress**
Ready.


benadida (approved):
Good to land, let's have @jonahkagan do a post-merge review for anything I missed.

PR #1146, opened 2021-04-02
Fix sample size return in API

umbernhard:
**Description**

We were indexing an object like a Tuple, which I think has undefined behavior. Fix it. 

**Testing**

N/A

**Progress**

Still testing.


jonahkagan:
@umbernhard I think we fixed this bug in a different PR, but maybe you still want to add the test case here?

PR #1147, opened 2021-04-03
Adding display for JAs without audits and tests

MorganLove:
**Description**

- When an audit is deleted a JA might get stranded without an audit to list on their dash. Before, they only saw a blank screen, this lets them know there aren't any audits to show.

**Testing**

- Added a few more tests for covering the audit deletion function

**Progress**

- Ready for review

### DIFF
--- a/client/cypress/end-to-end/ballot-comparison-happy-path.spec.js
+++ b/client/cypress/end-to-end/ballot-comparison-happy-path.spec.js
@@ -18,6 +18,19 @@ describe('Ballot Comparison', () => {
     cy.findByText('Create Audit').click()
     cy.viewport(1000, 2000)
     cy.contains('Audit Setup')
+
+    // test deletion of audit
+    cy.findByText("View Audits").click()
+    cy.findByRole('button', { name: `TestAudit${id}` }).siblings('button').click()
+    cy.findByRole('button', { name: 'Delete' }).click()
+    cy.findByText(`TestAudit${id}`).should('not.exist')
+
+    // create new audit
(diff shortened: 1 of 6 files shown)


client/cypress/end-to-end/ballot-comparison-happy-path.spec.js
```diff
@@ -18,6 +18,19 @@ describe('Ballot Comparison', () => {
     cy.findByText('Create Audit').click()
     cy.viewport(1000, 2000)
     cy.contains('Audit Setup')
+
+    // test deletion of audit
+    cy.findByText("View Audits").click()
+    cy.findByRole('button', { name: `TestAudit${id}` }).siblings('button').click()
+    cy.findByRole('button', { name: 'Delete' }).click()
+    cy.findByText(`TestAudit${id}`).should('not.exist')
```
jonahkagan, line 26: i already added a test for audit deletion to cypress

jonahkagan (approved):

client/src/components/HomeScreen.tsx
```diff
@@ -210,29 +210,37 @@ const ListAuditsJurisdictionAdmin = ({
...
-            >
-              {name}
-            </LinkButton>
-          ))}
-        </div>
-      ))}
+      {Object.entries(jurisdictionsByAudit).length === 0 ? (
+        <Callout intent="warning">
+          You don&apos;t have any available audits at the moment
+        </Callout>
```
jonahkagan, line 216: does this new case need a test?

PR #1148, opened 2021-04-05
Adding support for saving cy screencaps as artifacts

MorganLove:
**Description**

- So we can see more about what is failing when it happens on CI

(Based on this: https://circleci.com/docs/2.0/artifacts/?section=pipelines)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,7 +104,7 @@ commands:
           name: Saving screenshots & videos from Cypress
           command: |
             mkdir /tmp/artifacts;
-            cp -r client/cypress/screenshots client/cypress/videos /tmp/artifacts;
+            cp -r ./client/cypress/screenshots ./client/cypress/videos /tmp/artifacts;
 
       - store_artifacts:
           path: /tmp/artifacts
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -98,6 +98,17 @@ commands:
...
+  save-screencaps:
+    steps:
+      - run:
+          name: Saving screenshots & videos from Cypress
+          command: |
+            mkdir /tmp/artifacts;
+            cp -r ./client/cypress/screenshots ./client/cypress/videos /tmp/artifacts;
+
+      - store_artifacts:
+          path: /tmp/artifacts
```
jonahkagan, line 110: any reason the files need to be copied to `/tmp/artifacts`? could this just be:

```
- store_artifacts:
    path: ./client/cypress/screenshots
- store_artifacts:
    path: ./client/cypress/videos
```

if so, i would probably just put this directly into the steps of the cypress job below

MorganLove: I was wanting to name it, so we'd see the step in the process, but that's probably not necessary.

PR #1151, opened 2021-04-05
Now throwing an error if we need to do a full recout before we do it

umbernhard:
**Description**

`suite.get_sample_size` now errors if a recount of either strata is expected. 

**Testing**

Modified test cases accordingly. 

**Progress**

Ready.

### DIFF
--- a/server/audit_math/suite.py
+++ b/server/audit_math/suite.py
@@ -658,6 +658,12 @@ def get_sample_size(
 
         sample_size = sorted(sample_sizes, key=sum, reverse=True)[0]
 
+    if (
+        sample_size[0] == cvr_stratum.num_ballots
+        or sample_size[1] == bp_stratum.num_ballots
+    ):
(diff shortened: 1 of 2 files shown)


server/audit_math/suite.py
```diff
@@ -658,6 +658,12 @@ def get_sample_size(
 
         sample_size = sorted(sample_sizes, key=sum, reverse=True)[0]
 
+    if (
+        sample_size[0] == cvr_stratum.num_ballots
+        or sample_size[1] == bp_stratum.num_ballots
+    ):
+        raise ValueError("One or both strata need to be recouted.")
```
jonahkagan, line 665: "recouted" should be "recounted"

jonahkagan (approved):
are you planning to change the p-value computation to handle this case differently at all?

umbernhard:
I figured I'd leave it, but I can remove it if you think I should.

PR #1153, opened 2021-04-06
Background task for sample size options

jonahkagan:
Task: #1150 

Backend changes
- Creates a new background task for computing sample size options for the first round
- Modifies the /sample-sizes endpoint to trigger the background task, returning task status and results (once the task completes). The results are cached and available for 5 seconds
- For later round sample sizes, modifies the draw_sample background task to compute the sample size options and pick one (this was previously done in the create round endpoint)

Frontend changes
- Changes the useSampleSizes hook to return the entire response (including task status and error) and poll for status updates until the task completes
- Updates the review screen to consume this new data format. Notably, we can now easily display any validation errors when computing the sample sizes persistently in the page (previously we toasted them)
- Refactors the review screen component to be a bit easier to read by colocating all of the sample size stuff near the rendering of the sample sizes form


server/api/sample_sizes.py
```diff
@@ -186,18 +202,72 @@ def sample_sizes_for_contest(contest: Contest):
...
+        task["status"]
+        not in [ProcessingStatus.READY_TO_PROCESS, ProcessingStatus.PROCESSING]
+    ):
+        election.sample_size_options = None
+        election.sample_size_options_task = create_background_task(
+            first_round_sample_size_options, dict(election_id=election.id)
+        )
+        db_session.commit()
+
+    # In tests, the background task will complete immediately, so we return
```
benadida, line 266: this makes me wonder if there's anything we can do, in tests, to make background tasks not happen immediately, to queue them up on some event loop... I know no such thing, but you know what i mean.

for later though.

server/models.py
```diff
@@ -196,6 +196,14 @@ class Election(BaseModel):
     )
     standardized_contests = Column(JSON)
 
+    # During audit setup, we compute sample size options for the first
+    # round of the audit in a background task and store the results here.
+    sample_size_options_task_id = Column(
+        String(200), ForeignKey("background_task.id", ondelete="cascade")
```
benadida, line 202: cascade? Shouldn't it be set to null?

benadida (approved):
one important data model cascade question. Otherwise looks good.

I am slightly irked by the background job being kicked off even if there are pre-condition errors, but I can see the point of a consistent API for all errors / results.

server/api/sample_sizes.py
jonahkagan: there's a config variable for whether or not to run the tests immediately. i designed it this way. i explicitly disable that behavior in one of the tests i added that actually checks all the various states of the background task. but for most tests, it's simpler to just have it complete immediately.

server/models.py
jonahkagan: whoops, just copied it from elsewhere without thinking it through. fixed here: 1746f087. I think there may be other places this is misconfigured as well

Issue #1152, opened 2021-04-06
Bug: Fix CVR ballot imprint lookup for CVRs with holes


mcchilders:
Record IDs in CVR batches are ordered, but not always sequential - meaning you might jump from ID 2 to ID 6 in a batch. When we're looking up the CVR record for a particular ballot from the manifest (e.g. Tab 1, Batch 12, ballot 16), we need to actually find the 16th ballot in the ordered list of ballots in Batch 12, which might be Record ID 16 ... or might not. This may require indexing ordered ballots within batches so that we fill any holes where the record IDs are not sequential.

PR #1154, opened 2021-04-06
Compute ballot position for CVR ballots based on record id

jonahkagan:
Task: #1152 

The record ids in for each batch in the CVR might not always be a continuous sequence - there might be missing numbers. So if we assume that the record id is equivalent to the ballot's physical "position" in the batch, we might sample ballots that we can't find in the CVR.

This PR separates the concept of a record id and a ballot position. After loading the CvrBallots into the db, we order all the record ids for each batch and assign them a ballot position based on the index in that ordering.


benadida (approved):
nifty!

PR #1156, opened 2021-04-06
Fix CvrBallot constraints

jonahkagan:
The primary key constraint for batch_id+ballot_position was causing errors when trying to assign ballot_positions. Instead, we change the primary key to batch_id+record_id and create a separate unique index for batch_id+ballot_position.


server/models.py
```diff
@@ -786,15 +786,18 @@ class CvrBallot(Base):
     record_id = Column(Integer, nullable=False)
     # ballot_position is the counting index of the ballot among all ballots
     # from the batch in the CVR - we use this to match with sampled ballots from the manifest
-    ballot_position = Column(Integer)
+    ballot_position = Column(Integer, nullable=False)
```
jonahkagan, line 789: accidentally left this out from the previous PR

benadida: grumble, sorry I missed this.

benadida (approved):

PR #1160, opened 2021-04-08
Make background sample sizes backwards compatible

jonahkagan:
Existing audits that were launched before we made a background task for samples sizes won't have a background task. We now expect that after the audit launches, we can load the previously computed sample size options from the cached value from the background task.

This PR changes the sample sizes endpoint to start a background task after the audit launches if there isn't one already in order to backfill this data for existing audits.

### DIFF
--- a/server/api/sample_sizes.py
+++ b/server/api/sample_sizes.py
@@ -246,54 +246,42 @@ def serialize_sample_size_options(sample_size_options):
 @api.route("/election/<election_id>/sample-sizes", methods=["GET"])
 @restrict_access([UserType.AUDIT_ADMIN])
 def get_sample_sizes(election: Election):
-    task = serialize_background_task(election.sample_size_options_task)
-
-    # If we've already started the first round, return which sample size was
-    # selected for each contest so we can show the user
-    if len(list(election.rounds)) > 0:
-        selected_sample_sizes = dict(
-            RoundContest.query.join(Round)
-            .filter_by(election_id=election.id, round_num=1)
-            .values(RoundContest.contest_id, RoundContest.sample_size)
-        )
-        return jsonify(
-            sampleSizes=serialize_sample_size_options(election.sample_size_options),
-            selected=selected_sample_sizes,
-            task=task,
-        )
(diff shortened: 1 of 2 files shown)


server/api/sample_sizes.py
```diff
@@ -246,54 +246,42 @@ def serialize_sample_size_options(sample_size_options):
...
-    # Otherwise, start a background task to compute sample size options (as
-    # long as there isn't already one in progress).
-    if task is None or (
-        task["status"]
-        not in [ProcessingStatus.READY_TO_PROCESS, ProcessingStatus.PROCESSING]
-    ):
+        # Don't need to recompute after the audit launches
+        and len(list(election.rounds)) == 0
+    )
+    if not election.sample_size_options_task or existing_options_expired:
```
benadida, line 264: isn't the first condition of this `if` already checked in the `existing_options_expired` conjunction?

benadida: maybe change to `existing_options_sad` ;)

benadida (approved):

server/api/sample_sizes.py
jonahkagan: it's subtle but it's not the same. `election.sample_size_options_task` is only checked in the definition of `existing_options_expired` so that i could pull it out into a separate variable. if you substitute in it looks like this:
```python
    if not election.sample_size_options_task or (
        election.sample_size_options_task
        and election.sample_size_options_task.completed_at
        and (
            datetime.now(timezone.utc) - election.sample_size_options_task.completed_at
            > timedelta(seconds=5)
        )
        # Don't need to recompute after the audit launches
        and len(list(election.rounds)) == 0
    ):
```
Checking `election.sample_size_options_task` after the `or` is not necessary due to short-circuiting here. But removing the first clause `not election.sample_size_options_task` would totally change the behavior

Hopefully that makes sense

benadida: Oh ... duh, got it. Not that subtle, just not enough caffeine on this end.

PR #1166, opened 2021-04-13
Show error from drawing sample

jonahkagan:


client/src/components/MultiJurisdictionAudit/useRoundsAuditAdmin.ts
```diff
@@ -51,6 +51,9 @@ const postRound = async (
 export const isDrawSampleComplete = (rounds: IRound[]) =>
   rounds[rounds.length - 1].drawSampleTask.completedAt !== null
 
+export const drawSampleError = (rounds: IRound[]) =>
+  rounds[rounds.length - 1].drawSampleTask.error
```
benadida, line 55: I wonder if we might just more aggressively say that if there's *any* error on any of the tasks, rather than just the last one.

benadida (approved):

client/src/components/MultiJurisdictionAudit/useRoundsAuditAdmin.ts
jonahkagan: it's not possible to have an error on an earlier round than the most recent round, because that earlier round wouldn't have a sample drawn, meaning the audit wouldn't have been able to proceed

PR #1168, opened 2021-04-15
Making ballot audit status more clear

### LINKED ISSUE #1162: Show ballots in "not_audited" status on Audit Board ballot list
Sometimes audit boards who skip/not_found a ballot need to come back to it and enter data about that ballot, but it can be hard to remember to do that if there's not some visual cue onscreen. Since we mark these ballots "not_audited" when they are skipped, we should show the Audit Board some sort of indication of which ballots those were in the list of ballots - make sure we show a color or icon in the status column that makes them look at it.

MorganLove:
**Description**

Closes #1162 

screenshot

**Testing**

- Updated snapshots

**Progress**

@mcchilders is this what you're thinking?

### DIFF
--- a/client/src/components/DataEntry/BoardTable.tsx
+++ b/client/src/components/DataEntry/BoardTable.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import styled from 'styled-components'
-import { H1 } from '@blueprintjs/core'
+import { H1, H5 } from '@blueprintjs/core'
 import { Link } from 'react-router-dom'
 import { Column } from 'react-table'
 import { Table } from '../Atoms/Table'
@@ -12,7 +12,8 @@ import StatusTag from '../Atoms/StatusTag'
 
 const RightWrapper = styled.div`
   display: flex;
-  justify-content: flex-end;
+  align-items: flex-start;
+  justify-content: space-between;
   margin: 20px 0;
   .bp3-button {
     margin-left: 10px;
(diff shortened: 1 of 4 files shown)


MorganLove:
@mcchilders added a summary at the top, but the phrasing might be confusing for people. What do you think?

Screenshot 2

mcchilders:
@MorganLove this is perfect!

jonahkagan (commented):
A few thoughts based on the screenshot:
- Love the addition of colors!
- "Ballot Audit status" has inconsistent capitalization. also maybe not even a necessary header - it seems pretty obvious that those are counts without the header
- Any reason we're only showing the count of not audited/not found? Why not also show the count of audited?
- "not found" is very rare - maybe doesn't make sense to show a summary count of that unless it happens
- the vertical alignment of the buttons on the right with the "ballot audit status" section looks weird to me

MorganLove:
@mcchilders @jonahkagan here's an updated screenshot:

Screenshot 3

client/src/components/DataEntry/BoardTable.tsx
```diff
@@ -8,10 +8,12 @@ import { BallotStatus } from '../../types'
 import LinkButton from '../Atoms/LinkButton'
 import { IBallot } from '../MultiJurisdictionAudit/RoundManagement/useBallots'
 import { IAuditBoard } from '../UserContext'
+import StatusTag from '../Atoms/StatusTag'
 
 const RightWrapper = styled.div`
```
jonahkagan, line 13: `RightWrapper` doesn't really make sense as a name for this anymore

client/src/components/DataEntry/BoardTable.tsx
```diff
@@ -8,10 +8,12 @@ import { BallotStatus } from '../../types'
...
   display: flex;
-  justify-content: flex-end;
+  align-items: flex-start;
```
jonahkagan, line 15: i think `center` or `baseline` would look better, so the text lines up within the tags/buttons

client/src/components/DataEntry/BoardTable.tsx
```diff
@@ -86,6 +107,18 @@ const BoardTable: React.FC<IProps> = ({ boardName, ballots, url }: IProps) => {
     b => b.status === BallotStatus.NOT_AUDITED
   )
 
+  const getTotalAudited = ballots.filter(
```
jonahkagan, line 110: nit: maybe just `totalAudited` since this is just an int, not a function to "get" something

jonahkagan (approved):
A few more questions but overall looks good!

PR #1170, opened 2021-04-21
Batch data entry fixes

MorganLove:
**Description**

- Console warning issue: initial values to be set so that it stops complaining about fields changing between controlled and uncontrolled. For choice results, it was just assigning blank array and there was no values initially. So added code to fetch the choices and assigned a blank value to it. This then required changing the data type of choice results value to `number | string` so that `''` can be assigned. (We could assign 0 as well, but it will then display 0 by default.)

- Dialogs were opening still after results finalized: once the results are finalized, made changes so that when buttons are clicked, the popup doesn’t open.

**Testing**

- Added a cypress test for handling offline batch data entry paths

### DIFF
--- a/client/cypress/end-to-end/offline-batch-data-entry-tests.spec.js
+++ b/client/cypress/end-to-end/offline-batch-data-entry-tests.spec.js
@@ -1,0 +1,296 @@
+import 'cypress-file-upload'
+
+before(() => cy.exec('./cypress/seed-test-db.sh'))
+
+describe('Offline Batch Data Entry', () => {
+  const auditAdmin = 'audit-admin-cypress@example.com'
+  const jurisdictionAdmin = 'wtarkin@empire.gov'
+  const uuid = () => Cypress._.random(0, 1e6)
+  let id = 0
+
+  beforeEach(() => {
+    id = uuid()
+    cy.visit('/')
+    cy.loginAuditAdmin(auditAdmin)
+    cy.get('input[name=auditName]').type(`TestAudit${id}`)
+    cy.get('input[value="BALLOT_POLLING"]').check({ force: true })
+    cy.get('input[value="BRAVO"]').check({ force: true })
+    cy.findByText('Create Audit').click()
+    cy.viewport(1000, 2000)
+    cy.contains('Audit Setup')
+  })
+
+  it('happy path', () => {
+    cy.fixture('CSVs/jurisdiction/sample_jurisdiction_filesheet.csv').then(
+      fileContent => {
+        cy.get('input[type="file"]')
+          .first()
+          .attachFile({
+            fileContent: fileContent.toString(),
+            fileName: 'sample_jurisdiction_filesheet.csv',
+            mimeType: 'csv',
+          })
+      }
+    )
+    cy.findAllByText('Upload File').spread((firstButton, secondButton) => {
+      firstButton.click()
+    })
+    cy.contains('Upload successfully completed')
+
+    cy.wait(100) // gets stuck in an infinite loop without a 100ms wait here
+    cy.findByText('Next').click()
+    cy.get('input[name="contests[0].name"]').type('Contest')
+    cy.get('input[name="contests[0].choices[0].name"]').type('A')
+    cy.get('input[name="contests[0].choices[0].numVotes"]').type('300')
+    cy.get('input[name="contests[0].choices[1].name"]').type('B')
+    cy.get('input[name="contests[0].choices[1].numVotes"]').type('100')
+    cy.get('input[name="contests[0].totalBallotsCast"]').type('400')
+    cy.findByText('Select Jurisdictions').click()
+    cy.findByLabelText('Death Star').check({ force: true })
+    cy.findByText('Save & Next').click()
+    cy.findAllByText('Opportunistic Contests').should('have.length', 2)
+    cy.findByText('Save & Next').click()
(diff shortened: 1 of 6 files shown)


client/src/components/MultiJurisdictionAudit/RoundManagement/OfflineBatchRoundDataEntry.tsx
```diff
@@ -292,7 +301,10 @@ const OfflineBatchRoundDataEntry = ({ round }: IProps) => {
                     document.getElementById('results-table')!.outerHTML
                   }
                 />
-                <Button onClick={() => setIsConfirmOpen(true)}>
+                <Button
+                  onClick={() => setIsConfirmOpen(true)}
+                  disabled={!!finalizedAt}
```
jonahkagan, line 306: nice catch with these. it's a shame disabling the `fieldset` doesn't disable the buttons

client/src/components/MultiJurisdictionAudit/RoundManagement/OfflineBatchRoundDataEntry.tsx
```diff
@@ -292,7 +301,10 @@ const OfflineBatchRoundDataEntry = ({ round }: IProps) => {
...
+                >
```
jonahkagan, line 307: nice catch with these. it's a shame disabling the fieldset doesn't disable the buttons

client/src/components/MultiJurisdictionAudit/RoundManagement/OfflineBatchRoundDataEntry.tsx
```diff
@@ -128,12 +128,18 @@ const OfflineBatchRoundDataEntry = ({ round }: IProps) => {
...
 
   const total = (choiceId: string) =>
-    sum(results.map(batch => batch.choiceResults[choiceId]))
+    sum(results.map(batch => Number(batch.choiceResults[choiceId])))
+
+  // set default choiceResults to blank for controlled input
+  const defaultChoiceArr: { [key: string]: string } = {}
+  contest.choices.forEach(choice => {
+    defaultChoiceArr[choice.id] = ''
+  })
```
jonahkagan, line 137: dang, this is annoying. i did a little research and i think you're right that passing an empty string is the correct solution (https://github.com/facebook/react/issues/7779)

here's a way to do it that has a smaller footprint:

```
diff --git a/client/src/components/MultiJurisdictionAudit/RoundManagement/OfflineBatchRoundDataEntry.tsx b/client/src/components/MultiJurisdictionAudit/RoundManagement/OfflineBatchRoundDataEntry.tsx
index 6861c5a2..2c16a76e 100644
--- a/client/src/components/MultiJurisdictionAudit/RoundManagement/OfflineBatchRoundDataEntry.tsx
+++ b/client/src/components/MultiJurisdictionAudit/RoundManagement/OfflineBatchRoundDataEntry.tsx
@@ -83,6 +83,7 @@ const InputWithValidation = ({ field, form, ...props }: FieldProps) => {
         className={`bp3-input bp3-fill ${error ? 'bp3-intent-danger' : ''}`}
         {...field}
         {...props}
+        value={field.value || ''}
       />
     </div>
   )
```

client/cypress/end-to-end/offline-batch-data-entry-tests.spec.js
```diff
@@ -0,0 +1,296 @@
...
+      .siblings('button')
+      .click()
+    cy.contains('Number of Audit Boards')
+    cy.findByText('Save & Next').click()
+    cy.contains('No batches added. Add your first batch below.')
+
+    // check validation message when no fields are filled up
+    cy.findByRole('button', { name: /Add batch/ }).click()
+    cy.findByRole('button', {name: 'Save Batch'}).click()
+    cy.contains('Please fill in the empty fields above before saving this batch.')
```
jonahkagan, line 258: i'm really glad you added an end to end test for this flow - that's definitely something we were missing. at the same time, the cypress tests are already the longest running part of CI. so ideally we would only test the bare minimum in the cypress test and pull the rest out into unit tests. for example, this validation message is entirely frontend, so it could be tested with a unit test

client/src/components/MultiJurisdictionAudit/RoundManagement/useOfflineBatchResults.ts
```diff
@@ -11,7 +11,7 @@ export interface IOfflineBatchResult {
     | 'Other'
     | ''
   choiceResults: {
-    [choiceId: string]: number
+    [choiceId: string]: number | string
```
jonahkagan, line 14: is the `string` still necessary here? i wouldn't expect it to be

client/cypress/end-to-end/offline-batch-data-entry-tests.spec.js
```diff
@@ -0,0 +1,196 @@
...
+    cy.findByRole('button', {name: 'Save Batch'}).click()
+    cy.get('.Toastify')
+    .find('div')
+    .find('div')
+    .contains('Batch names must be unique')
+    .invoke('text')
+    .then(text =>
+      expect(text).to.equal('Batch names must be unique')
+    )
+    cy.get('.Toastify')
```
jonahkagan, line 137: maybe update to use the new `findAndCloseToast` command

jonahkagan (approved):
unit tests look great! a few small questions but otherwise looks good to go

Issue #1007, opened 2021-01-19
Handle misnamed contests in CVR for Ballot Comp & hybrid methods


mcchilders:
When you are auditing a contest that crosses jurisdictions and relies on CVRs, you need each jurisdiction to name the contest in exactly the same way in their voting system election definition (which controls the naming in the CVR files), otherwise Arlo has no way to correctly roll-up vote totals/ballot universes across jurisdictions. For example, if one county calls the presidential race "President of the United States" and a different county uses, "United States President," Arlo would treat those as two different races.

States generally specify the correct naming conventions for contests that cross jurisdiction boundaries, but we know from experience in Colorado that not every local jurisdiction follows the rules. So, as a safety mechanism, we need to:
- check when the AA hits the review screen whether any contest names in uploaded CVRs do not match one of the contest names on the standardized contest list
-- if all the names appear in the standardized contest list, you're good to go
-- if one or more names do not match, display the current name along with the jurisdiction name, and allow the AA to choose the correct name from a dropdown containing all the contest names that are associated with that jurisdiction from the standardized contest list. Note that each jurisdiction + contest name combination that doesn't match should have its own fix in the interface (no bulk edits).
-- there should also be a "no change" option, which is what the interface should default to. In practice you sometimes have dozens of errors, and in those cases AAs have often just wanted to edit the ones that "matter" (e.g. the names of the contests you are actually auditing, so the "no change" option essentially allows them to skip fixing names where it doesn't matter.

Screenshots of potential interface attached - could appear as a standard section on the review screen whenever corrections are able to be made, or we could show a warning that says corrections are needed and give the AA the option to open the interface in a modal when they are ready to fix it.

![Screen Shot 2021-03-30 at 2.03.21 PM.png](https://zube.io/files/votingworks/942fde209e6d8257187d18fada9867d0-screen-shot-2021-03-30-at-2-03-21-pm.png)

PR #1173, opened 2021-04-21
Standardize contest names

jonahkagan:
Task: #1007 

https://user-images.githubusercontent.com/530106/115788215-2d849a00-a378-11eb-80ac-8a88ef99efce.mov


MorganLove:
@jonahkagan Looks awesome! A couple issues I ran into:

1) After the election is launched, the option to change the contest names still appears. I think this should be disabled because in between submitting the ballots, I changed the contest name and while signing off it showed an internal server error.

2) Added 3 jurisdictions. For 1st jurisdiction, used the ballot comparison file from the fixtures.

For 2nd & 3rd jurisdiction, used a CVR with different contest names - Contest 12 & Contest 21.

On the Launch page, from the dialog to choose correct names, I then chose the 1st option for just the first jurisdiction and saved.

Opened again, and chose another option for the second jurisdiction, and an internal error was shown. And after that, it didn't update the contest names and spinner kept loading on submission.

jonahkagan:
Thanks @MorganLove. Addressed the first issue here: ab376aa9

jonahkagan:
And fixed the second issue here: 753544a

MorganLove (approved):
Nice!

benadida (approved):

PR #1174, opened 2021-04-22
Cypress: close toasts

jonahkagan:
Tests were failing because the (expected) toasts for error message were covering up the user menu that we need to click on for logout.


MorganLove (approved):
Looks awesome!

PR #1175, opened 2021-04-27
Refactoring cy tests to save time

MorganLove:
No need to cover a happy path separately when it's included in the failure case path as well.


jonahkagan (approved):
makes sense to me! good thinking

PR #1180, opened 2021-04-30
Redirecting JA to audit screen if only one audit

### LINKED ISSUE #1011: Bypass jurisdiction list if JA only has access to a single jurisdiction

MorganLove:
**Description**

Closes #1011 

**Testing**

- Added new unit test to verify this works
- Updated snapshots
- Updated cypress flows

**Progress**

Should be good to go if it works

### DIFF
--- a/client/src/components/HomeScreen.tsx
+++ b/client/src/components/HomeScreen.tsx
@@ -10,7 +10,7 @@ import {
   Button,
   Intent,
 } from '@blueprintjs/core'
-import { useHistory, useLocation } from 'react-router-dom'
+import { useHistory, useLocation, Redirect } from 'react-router-dom'
 import styled from 'styled-components'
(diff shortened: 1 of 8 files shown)


client/src/components/HomeScreen.tsx
```diff
@@ -51,7 +51,13 @@ const HomeScreen: React.FC = () => {
           </Inner>
         </Wrapper>
       )
-    case 'jurisdiction_admin':
+    case 'jurisdiction_admin': {
+      if (user.jurisdictions.length === 1) {
+        const electionId = user.jurisdictions[0].election.id
+        const userId = user.jurisdictions[0].id
```
jonahkagan, line 57: technically, this is the jurisdiction id, not the user id

jonahkagan (approved):

PR #1183, opened 2021-05-03
Reworks sampling for MACRO.

umbernhard:
**Description**

Swapping out consistent sampling for something that doesn't eat all of our memory

**Testing**

snapshots were updated. 

**Progress**

Need to resolve typing issues around ticket numbers no longer being here.


server/audit_math/sampler.py
```diff
@@ -110,55 +110,22 @@ def draw_ppeb_sample(
...
-
-    return sample
+    weighted_errors = [
+        macro.compute_max_error(batch_results[batch], contest) / U
+        for batch in batch_results
+    ]
+
+    return generator.choice(
+        list(batch_results.keys()), sample_size + num_sampled, p=weighted_errors
+    )
```
jonahkagan, line 131: do we need to cut out the first `num_sampled` from this result?

server/audit_math/sampler.py
```diff
@@ -110,55 +110,22 @@ def draw_ppeb_sample(
...
-    for item in faux_sample:
-        sample.append((item[0], item[1][0], item[2]))
-
-    return sample
+    weighted_errors = [
+        macro.compute_max_error(batch_results[batch], contest) / U
+        for batch in batch_results
+    ]
+
+    return generator.choice(
```
jonahkagan, line 129: maybe specify `replace=True` just to be sure - it's not totally clear that it's the default

umbernhard: Good point!

server/audit_math/sampler.py
umbernhard: Yep, you're right.

jonahkagan (approved):

PR #1185, opened 2021-05-04
Confirm tests

MorganLove:
**Testing**

- Adding unit tests for the Confirm component

**Progress**

- Should be almost done, looking for a way to avoid using `any`, and I'm not sure if the snapshots are working right.

### DIFF
--- a/client/src/components/Atoms/Confirm.test.tsx
+++ b/client/src/components/Atoms/Confirm.test.tsx
@@ -1,0 +1,94 @@
+import React from 'react'
+import { render, screen, within, waitFor } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { Intent } from '@blueprintjs/core'
+import { useConfirm, Confirm, IConfirmOptions } from './Confirm'
+
+const confirmMock: IConfirmOptions = {
+  title: 'Confirm Component',
+  description: (
+    <div>
+      <p>
+        <strong>This action cannot be undone.</strong>
+      </p>
+    </div>
+  ),
+  yesButtonLabel: 'Yes',
+  yesButtonIntent: Intent.PRIMARY,
+  onYesClick: () =>
+    new Promise<void>(resolve => {
+      resolve()
+    }),
+}
+
+jest.mock('react', () => ({
+  ...jest.requireActual('react'),
+  useState: jest.fn(),
+}))
+
+afterEach(() => {
+  jest.clearAllMocks()
+})
+
+describe('Confirm component', () => {
+  it('opens confirm component', async () => {
+    const setOptions = jest.fn()
+
+    // passing default values for state to open the component
+    // as state is mocked
+    const useStateMock: any = () => [confirmMock, setOptions]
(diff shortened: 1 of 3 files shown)


client/src/components/Atoms/Confirm.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+jest.mock('react', () => ({
+  ...jest.requireActual('react'),
+  useState: jest.fn(),
+}))
+
+afterEach(() => {
+  jest.clearAllMocks()
+})
+
+describe('Confirm component', () => {
```
jonahkagan, line 33: I think it might be simpler to test by creating a "consumer" component that mimics the way the `Confirm` component and `useConfirm` hooks interact in their real usage. Here's an example of what that could look like:

```ts
describe('Confirm', () => {
  it('shows a confirmation dialog', async () => {
    const onYesClickMock = jest.fn().mockResolvedValue(true)

    const ConfirmConsumer = () => {
      const { confirm, confirmProps } = useConfirm()
      return (
        <>
          <Button
            onClick={() =>
              confirm({
                title: 'Test Title',
                description: 'Test description',
                onYesClick: onYesClickMock,
              })
            }
          >
            Open Confirm
          </Button>
          <Confirm {...confirmProps} />
        </>
      )
    }
    render(<ConfirmConsumer />)

    userEvent.click(screen.getByRole('button', { name: 'Open Confirm' }))
    const dialog = screen
      .getByRole('heading', { name: 'Test Title' })
      .closest('.bp3-dialog') as HTMLElement
    userEvent.click(within(dialog).getByRole('button', { name: 'Ok' }))
    expect(onYesClickMock).toHaveBeenCalled()
    await waitFor(() => expect(dialog).not.toBeInTheDocument())
    // ... more interactions and assertions ...
  })
})
```

client/src/components/Atoms/Confirm.tsx
```diff
@@ -15,7 +15,7 @@ export const useConfirm = () => {
   // On close, we set options to null.
   const [options, setOptions] = useState<IConfirmOptions | null>(null)
 
-  const confirm = (newOptions: IConfirmOptions) => {
+  const confirm = (newOptions: IConfirmOptions | null) => {
```
jonahkagan, line 18: why this change?

client/src/components/Atoms/Confirm.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    userEvent.click(within(dialog).getByRole('button', { name: 'Yes' }))
+
+    await waitFor(() => {
+      expect(
+        screen.queryByText(
+          'This action cannot be undone. Do you want to continue?'
+        )
+      ).not.toBeInTheDocument()
+    })
+  })
```
jonahkagan, line 93: might be good to check that it calls `onYesClickMock` when you click Yes

client/src/components/Atoms/Confirm.tsx
MorganLove: Oops, that was left there from some earlier testing.

jonahkagan (approved):

PR #1201, opened 2021-05-17
Layout update

### LINKED ISSUE #1190: New VotingWorks logo
Update Arlo with the new VotingWorks logo

### LINKED ISSUE #1193: New design: Audit Board Ballot List
Implement @beausmith's new design for the Ballot List in the Audit Board data entry workflow.

### LINKED ISSUE #1194: New design: Audit Board Data Entry screen
Implement the new design for the Audit Board data entry screen.

### LINKED ISSUE #1195: New Design: Audit Board Data Confirmation Modal
Remove the Audit Board Data Entry Review Screen and replace with a confirmation modal, as shown here.

MorganLove:
**Description**

Closes #1193 

Closes #1195

Closes #1190

Closes #1194

**Testing**

- Updated cypress
- Updated unit tests and snapshots

**Progress**

All the tickets for the redesign are all in this one PR now for unified review.


jonahkagan:
I noticed that when I filled out the audit board member names and then click submit, they don't load in the header until I refresh the page.

jonahkagan:
Now that we put the audit board name in the middle of the header, I think having it in the right corner is redundant. I think we talked about just replacing that menu with a simple "Log out" button or link

Screen Shot 2021-06-17 at 5 32 40 PM

jonahkagan:
I find the colored text on the gray background kind of hard to read, which happens both in the top bar and in the status column. Not sure how to fix - maybe using Tags with colored background and white or black text would be more readable? That's what we used in the audit progress table. @beausmith any thoughts?

Screen Shot 2021-06-17 at 5 36 54 PM

jonahkagan:
This text has slightly weird capitalization. Maybe just lowercase "submit"?

Screen Shot 2021-06-17 at 5 37 58 PM

jonahkagan:
I realized that there's not a ton of visual feedback when you click "Ballot Not Found" - just the ballot info changes. It might be worth popping up the confirmation modal for that too. I almost clicked it twice when trying it because I wasn't sure it worked.

jonahkagan:
Maybe disable the "Submit Selections" button until the user has selected something. Otherwise it just pops up a blank confirmation modal.

jonahkagan:
The header looks a bit overlappy in the iPhone X view. Are we wanting to support phone usage?

Screen Shot 2021-06-17 at 5 51 28 PM

client/src/components/DataEntry/Ballot.tsx
```diff
@@ -77,12 +143,49 @@ const Ballot: React.FC<IProps> = ({
...
+    }
+  }, [
+    auditing,
+    confirm,
+    interpretations,
+    contests,
+    submitBallot,
+    ballot,
+    nextBallot,
+  ])
```
jonahkagan, line 187: You should be able to just call `confirm` directly when the user clicks the "Submit Selections" button instead of using this intermediate `auditing` state. Any particular reason for the `auditing` state?

jonahkagan (commented):
At a first glance, this looks really really good! Love the restyle

MorganLove:
> Now that we put the audit board name in the middle of the header, I think having it in the right corner is redundant. I think we talked about just replacing that menu with a simple "Log out" button or link
> 
> Screen Shot 2021-06-17 at 5 32 40 PM

I think the title there is helpful for disambiguating who is logged in when using the support user feature, perhaps. Or should we just exclude it in this instance, or do something different for indicating who is logged in?

MorganLove:
> The header looks a bit overlappy in the iPhone X view. Are we wanting to support phone usage?
> 
> Screen Shot 2021-06-17 at 5 51 28 PM

There are some phone issues that have cropped up in other places too, I think. Should we resolve all of them in this PR, or a separate one, do you think?

jonahkagan:
> > The header looks a bit overlappy in the iPhone X view. Are we wanting to support phone usage?
> > Screen Shot 2021-06-17 at 5 51 28 PM
> 
> There are some phone issues that have cropped up in other places too, I think. Should we resolve all of them in this PR, or a separate one, do you think?

I think that particular phone issue was introduced in this PR (by the addition of the audit board and member names in the middle of the header), so that seems like a good reason to resolve it in this PR. I would suggest as a rule of thumb: if code changes in this PR introduce an issue, resolve it here. If you just notice an existing issue, resolve it in a separate PR.

client/src/components/DataEntry/Ballot.tsx
```diff
@@ -73,14 +140,98 @@ const Ballot: React.FC<IProps> = ({
...
+    newInterpretations: IBallotInterpretation[]
+  ) => {
+    setInterpretations(newInterpretations)
+    initiateConfirm('general')
+  }
 
   const submitNotFound = async () => {
-    await submitBallot(ballot.id, BallotStatus.NOT_FOUND, [])
-    nextBallot()
+    initiateConfirm('not_found')
```
jonahkagan, line 234: i think this whole section will be simpler if you have two separate functions for showing the confirm dialog instead of one function that conditions on general/not_found.

```ts
const onNotFoundClick = () =>
  confirm({
     ... all the confirm options for not found ...
     onYesClick: async () => {
      submitBallot(ballot.id, BallotStatus.NOT_FOUND, [])
      nextBallot()
     }
  })

const onSubmitSelectionsClick = async () =>
  confirm({
    ... all the confirm options for when interpretations were selected ...
    onYesClick: async () => {
      await submitBallot(
        ballot.id,
        BallotStatus.AUDITED,
        interpretations.filter(
          ({ interpretation }) => interpretation !== null
        )
      )
      nextBallot()
    }
  })

File: client/src/components/DataEntry/Ballot.tsx [hunk 3]
```diff
@@ -73,14 +140,98 @@ const Ballot: React.FC<IProps> = ({
...
+  const initiateConfirm = (selectionType: 'not_found' | 'general') =>
+    confirm({
+      title: 'Confirm the Ballot Selections',
+      description: (
+        <>
+          {contests.map((contest, i) => (
+            <div key={contest.id}>
+              <p>{contest.name}</p>
+              {selectionType === 'not_found' ? (
+                <h3>Ballot Not Found</h3>
```
jonahkagan, line 201: do we want to show "Ballot Not Found" for each contest or just once overall in the review modal? i think just showing it once would make more sense

client/src/components/DataEntry/Ballot.tsx
```diff
@@ -73,14 +140,98 @@ const Ballot: React.FC<IProps> = ({
...
+        return <h3>Blank vote</h3>
+      case Interpretation.CONTEST_NOT_ON_BALLOT:
+        return <h3>Not on Ballot</h3>
+      // case Interpretation.CANT_AGREE:
+      // case for Interpretation.CANT_AGREE in case we decide to put it back in again
+      // return (
+      //   <LockedButton disabled large intent="primary">
+      //     Audit board can&apos;t agree
+      //   </LockedButton>
+      // )
```
jonahkagan, line 175: might be worth just removing this since the actual LockedButton part is now out of date. can leave a comment that we don't support CANT_AGREE

client/src/components/DataEntry/Ballot.tsx
MorganLove: I like that idea.

client/src/components/DataEntry/Ballot.tsx
```diff
@@ -73,14 +140,91 @@ const Ballot: React.FC<IProps> = ({
...
-  const [auditing, setAuditing] = useState(true)
   const [interpretations, setInterpretations] = useState<
     IBallotInterpretation[]
   >(contests.map(emptyInterpretation))
+  const { confirm, confirmProps } = useConfirm()
+
+  const interpretationsRef = useRef<IBallotInterpretation[]>(
+    contests.map(emptyInterpretation)
+  )
+  interpretationsRef.current = interpretations
```
jonahkagan, line 151: What's this ref for? Seems like an unnecessary duplicate of the interpretations state

client/src/components/DataEntry/Ballot.tsx
```diff
@@ -73,14 +140,91 @@ const Ballot: React.FC<IProps> = ({
...
+      default:
+        return null
+    }
+  }
+
+  const setInterpretationsFunc = (
+    newInterpretations: IBallotInterpretation[]
+  ) => {
+    setInterpretations(newInterpretations)
+  }
```
jonahkagan, line 179: seems like this function could be removed

client/src/components/DataEntry/Ballot.tsx
```diff
@@ -73,14 +140,91 @@ const Ballot: React.FC<IProps> = ({
...
-    nextBallot()
+    confirm({
+      title: 'Confirm the Ballot Selections',
+      description: (
+        <div>
+          <h3>Ballot Not Found</h3>
+        </div>
+      ),
+      onYesClick: async () => {
+        submitBallot(ballot.id, BallotStatus.NOT_FOUND, [])
```
jonahkagan, line 222: should we `await` this call to `submitBallot` to make sure it succeeds before moving to the next ballot?

client/src/components/DataEntry/Ballot.tsx
```diff
@@ -99,74 +243,93 @@ const Ballot: React.FC<IProps> = ({
...
+                  </NotFoundButton>
+                </div>
+              </BallotMainRow>
+              <FlushDivider />
+              <div>
+                <BallotAudit
+                  contests={contests}
+                  interpretations={interpretations}
+                  setInterpretations={setInterpretationsFunc}
+                  confirmSelections={() => confirmSelections()}
```
jonahkagan, line 301: unnecessary function wrapper

client/src/components/DataEntry/BoardTable.tsx
```diff
@@ -71,32 +157,65 @@ const BoardTable: React.FC<IProps> = ({ boardName, ballots, url }: IProps) => {
...
+            Re-Audit
+          </AuditBtn>
+        ) : (
+          <AuditBtn
+            to={`${url}/batch/${ballot.batch.id}/ballot/${ballot.position}`}
+            minimal
+            fill
+          >
+            <strong>Audit Ballot</strong>
+          </AuditBtn>
```
jonahkagan, line 193: since these buttons have the same props except the inner text, could move the conditional to inside the text of the button

client/src/components/UserContext.tsx
```diff
@@ -68,6 +69,7 @@ export interface ISupportUser {
 export interface IAuthData {
   user: IUser | null
   supportUser: ISupportUser | null
+  setAuthData: (arg0: IAuthData) => void
```
jonahkagan, line 72: Why is this added to the context? My understanding is that the context is supposed to load the auth data and then pass it down to children components. I don't think children components should also be loading the auth data and then setting it on the context - that violates the one-direction data flow principle of React.

jonahkagan (commented):
This PR is getting pretty big and it's making it kind of confusing to review. I'm wondering if it would be possible to break out any smaller chunks and get those reviewed and merged in first, or if it's all interdependent.

client/src/components/UserContext.tsx
MorganLove: @jonahkagan The issue I was running into was a bug on the DataEntry side of things for Audit Boards. The member names stopped showing in the header whenever the audit board forms were being submitted because it wasn't getting updated when the api was being called at that point. This just allows the DataEntry component to update the state from the top when it submits the member and the signoff forms.

MorganLove:
> This PR is getting pretty big and it's making it kind of confusing to review. I'm wondering if it would be possible to break out any smaller chunks and get those reviewed and merged in first, or if it's all interdependent.

@jonahkagan I'm looking to see if it's possible, but it seems pretty interwoven by now. There's some basic workflow changes, and those are interdependent with the style updates.

client/src/components/UserContext.tsx
jonahkagan: Gotcha. In that case, here are a few other suggestions:
- Hoist the auth state up a level and pass it into both the Header and the DataEntry components, adding a method to the auth context for the API request to set the audit board member names (I don't love this solution, since the auth data context is used for other user types as well)
- Make a separate header component for audit boards only and render within the DataEntry component. Manage the audit board state at the DataEntry component level. (This solution is more in line with the current approach. I'm pretty sure the reason we originally had to load the audit board within the DataEntry component instead of just using the auth context was so that we could reload it after setting the member names. Since the header is diverging in design for the audit boards anyway, this seems like it might make for simpler components)

MorganLove:
@jonahkagan all refactored! :slightly_smiling_face:

benadida (approved):

PR #1202, opened 2021-05-17
Add times_sampled to MACRO math

jonahkagan:
- Add a new argument `times_sampled` to `macro.compute_risk` to enable us to correctly take into account batches that are sampled multiple times 
- Change the sample results passed into the `macro` model to only count each batch's audit results once (instead of summing across all samples for the same batch like we were previously doing)
- Change `count_audited_votes` in the same way (the counts are only used in the audit report, not for any audit logic, but it's good for them to be consistent with the math)


server/api/rounds.py
```diff
@@ -212,6 +215,22 @@ def cumulative_batch_results(election: Election) -> BatchTallies:
...
 
 
+def batches_times_sampled(election: Election) -> Dict[Tuple[str, str], int]:
+    sampled_batch_draw_counts = (
+        SampledBatchDraw.query.join(Batch)
+        .join(Jurisdiction)
+        .filter_by(election_id=election.id)
+        .group_by(Jurisdiction.name, Batch.name)
+        .values(
+            Jurisdiction.name, Batch.name, func.count(SampledBatchDraw.ticket_number)
```
benadida, line 225: does having `SampledBatchDraw.ticket_number` in the `count` actually change anything?

benadida (approved):

server/api/rounds.py
jonahkagan: Seems to not make a difference. I'll remove

PR #1210, opened 2021-06-01
Batch comparison: fix full hand recount cases

jonahkagan:
In `macro.compute_risk`, we want to check that all batches have been sampled exactly once, not that the total number of samples matches the total number of batches.

To ensure that each batch gets sampled exactly once, we change the sampling code to draw each batch exactly once in the full recount case (to avoid sampling with replacement, which could draw some batches multiple times and some batches not at all).


umbernhard:
@jonahkagan can we add a test case to make sure this solves the problem?

umbernhard (approved):

umbernhard (approved):

PR #1216, opened 2021-06-02
Tweak JA ballot/batch count to avoid pluralization issues

jonahkagan:
The old language only made sense for more than one ballot/batch. This new language works for one or more ballots/batches.

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
+++ b/client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
@@ -83,14 +83,14 @@ const RoundManagement = ({
   }
 
   const ballotsOrBatches =
-    auditSettings.auditType === 'BATCH_COMPARISON' ? 'batches' : 'ballots'
+    auditSettings.auditType === 'BATCH_COMPARISON' ? 'Batches' : 'Ballots'
 
   if (numSamples === 0 && !round.sampledAllBallots) {
     return (
       <PaddedWrapper>
         <StrongP>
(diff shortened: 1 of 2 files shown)


benadida (approved):

client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
```diff
@@ -83,14 +83,14 @@ const RoundManagement = ({
   }
 
   const ballotsOrBatches =
-    auditSettings.auditType === 'BATCH_COMPARISON' ? 'batches' : 'ballots'
+    auditSettings.auditType === 'BATCH_COMPARISON' ? 'Batches' : 'Ballots'
```
MorganLove, line 86: @jonahkagan Shouldn't these be capitalized, since that looks more correct when used on their own, and since you're lowercasing them on line 93?

MorganLove (changes requested):

client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
jonahkagan: The first letter is capitalized so they can start the sentence on L106. I lowercase them on L93 so they can be used in the middle of a sentence

MorganLove (approved):

client/src/components/MultiJurisdictionAudit/RoundManagement/index.tsx
MorganLove: Oh gotcha, I was reading it wrong. Looks good!

PR #1217, opened 2021-06-03
Improve error on duplicate row in jurisdictions file

jonahkagan:
We were getting an internal database error. Instead, we use our CSV parsing validation to provide a useful error message.

### DIFF
--- a/server/util/jurisdiction_bulk_update.py
+++ b/server/util/jurisdiction_bulk_update.py
@@ -12,8 +12,8 @@ JURISDICTION_NAME = "Jurisdiction"
 ADMIN_EMAIL = "Admin Email"
 
 JURISDICTIONS_COLUMNS = [
-    CSVColumnType("Jurisdiction", CSVValueType.TEXT),
-    CSVColumnType("Admin Email", CSVValueType.EMAIL),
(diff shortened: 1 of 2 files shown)


server/util/jurisdiction_bulk_update.py
```diff
@@ -12,8 +12,8 @@
 ADMIN_EMAIL = "Admin Email"
 
 JURISDICTIONS_COLUMNS = [
-    CSVColumnType("Jurisdiction", CSVValueType.TEXT),
-    CSVColumnType("Admin Email", CSVValueType.EMAIL),
+    CSVColumnType("Jurisdiction", CSVValueType.TEXT, unique=True),
+    CSVColumnType("Admin Email", CSVValueType.EMAIL, unique=True),
```
benadida, line 16: does this preclude a jurisdiction from having two admins? Or an admin for being an admin for two jurisdictions?

jonahkagan: No, those cases are both still allowed. I know it's not the best interface to express a composite unique key (I though it was a bit confusing when I originally wrote it), but it's been working so it didn't seem worth changing.

benadida (approved):
might be worth a comment that it's a composite key, not just two unique keys, but otherwise good.

PR #1226, opened 2021-06-14
Map feature

### LINKED ISSUE #1209: Show a state-specific map on the AA progress screen
We'd like to augment the AA /progress screen with a map version of the same data that is shown in the table on the progress screen, using the same colors as the table to indicate which jurisdictions are in which state of the audit.

Basic CoAs:
- Data should be the same as what's shown on the progress table; so, prior to audit launch the data displayed in the map should indicate file upload status for each jurisdiction just like the table, and once launched both the table and the map should indicate progress with the audit in each jurisdiction.
- Colors/labels should match what's already used in the table in the table
- Which state to show in the map should key off the 'state' field the AA fills in when creating the audit
- Need a basic key included with the map explaining the colors
- Non-participating jurisdictions in a state can be white/clear fill (instead of grey, which indicates participating but no data yet)
- Minimally we need the following states: MI, RI, VA, GA, OH, CA, CO

Open questions:
- In most states, the jurisdictions are counties (RI/GA/OH/CA/CO) and we can easily map jurisdiction name -> a known list of county names. However, in some states elections are run by cities/towns (MI) - and still others are a mix of counties and cities/towns (VA). We can likely find a city/town map to deal with MI ... but how should we handle the non-county jurisdictions in a place like VA, where a county map still makes the most sense visually?

The general idea would be

MorganLove:
**Description**

Closes #1209 

**Testing**

- Updates existing unit tests and snapshots
- New tests written for map presence and some of the behavior

**Progress**

Ready for review


client/src/components/MultiJurisdictionAudit/_mocks.ts
```diff
@@ -209,6 +209,12 @@ export const jaApiCalls = {
     },
     response: { status: 'ok' },
   },
+  getMapData: {
+    url: 'https://cdn.jsdelivr.net/npm/us-atlas@3/counties-10m.json',
+    response: new Response(
+      'https://cdn.jsdelivr.net/npm/us-atlas@3/counties-10m.json'
+    ),
+  },
```
jonahkagan, line 217: why is `getMapData` part of `jaApiCalls`? isn't it only loaded by audit admins, not jurisdiction admins?

client/src/components/MultiJurisdictionAudit/Progress/ProgressMap.tsx
```diff
@@ -0,0 +1,367 @@
...
+          }
+          if (
+            currentRoundStatus.status === JurisdictionRoundStatus.NOT_STARTED ||
+            currentRoundStatus.status === null
+          ) {
+            return 'gray'
+          }
+        }
+      }
+      return 'default'
```
jonahkagan, line 207: since we want to ensure the progress table and the map always show the same status, it think it would be good to factor out the logic that determines the status. i'm imagining creating a function that takes a jurisdiction and returns the status that we want to show in the table/map

client/src/components/MultiJurisdictionAudit/Progress/ProgressMap.tsx
```diff
@@ -0,0 +1,367 @@
...
+}: IProps) => {
+  const width = 960
+  const height = 500
+  const d3Container = useRef(null)
+  const tooltipContainer = useRef(null)
+
+  const [usCounties, setUSCounties] = useState<Feature[] | undefined>(undefined)
+  const [usStates, setUSStates] = useState<Feature[] | undefined>(undefined)
+  const [usState, setUSState] = useState<Feature | undefined>(undefined)
+  const [jsonData, setJsonData] = useState<Topology | undefined>(undefined)
```
jonahkagan, line 136: i think these state hooks are redundant, since `usCounties`, `usStates`, and `usState` are derived from `jsonData`. i think it will simplify the component to just have one state hook for the topology data, and then derive all the other data you need as variables based on the value of the topology data

client/src/components/MultiJurisdictionAudit/Progress/ProgressMap.tsx
```diff
@@ -0,0 +1,367 @@
...
+  const [isMapLoaded, setIsMapLoaded] = useState<boolean>(false)
```
jonahkagan, line 137: what are we waiting for when we show the spinner? does it take time to load the topology data, or to execute the d3 code to set up the map? my guess is that it's almost all waiting for loading the topology data. if so, it might be virtually equivalent to stop showing the spinner as soon as the topology data is loaded, which means we could eliminate the `isMapLoaded` state and just check whether the topology data is loaded

client/src/components/MultiJurisdictionAudit/Progress/ProgressMap.tsx
```diff
@@ -0,0 +1,367 @@
...
+import { Colors, Spinner } from '@blueprintjs/core'
+import labelValueStates from '../AASetup/Settings/states'
+import { JurisdictionRoundStatus, IJurisdiction } from '../useJurisdictions'
+import { FileProcessingStatus, IFileInfo } from '../useCSV'
+
+interface IProps {
+  stateName?: string | null
+  jurisdictions: IJurisdiction[]
+  isRoundStarted: boolean
+  auditType:
```
jonahkagan, line 18: could use `AuditSettings['auditType']` so there's one source of truth

client/src/components/MultiJurisdictionAudit/Progress/ProgressMap.tsx
```diff
@@ -0,0 +1,367 @@
...
+  path {
+    fill: none;
+    stroke: none;
+    stroke-width: 0.5px;
+  }
+  #single-state {
+    stroke: ${Colors.BLACK};
+  }
+  .county {
+    cursor: pointer;
```
jonahkagan, line 49: having the pointer made me think i could click on the map. i think leaving the default might be better

client/src/components/MultiJurisdictionAudit/Progress/ProgressMap.tsx
```diff
@@ -0,0 +1,367 @@
...
+      <Tooltip id="tooltip" className="hide-tooltip" ref={tooltipContainer} />
+      {isMapLoaded ? (
+        <MapLabels>
+          {isRoundStarted ? (
+            <div>
+              <MapLabelsRow>
+                <MapLabelsBoxes className="success" /> Complete
+              </MapLabelsRow>
+              <MapLabelsRow>
+                <MapLabelsBoxes className="progress" /> In-progress
```
jonahkagan, line 328: Change to "In progress" for consistency with the other labels

client/src/components/MultiJurisdictionAudit/Progress/ProgressMap.tsx
```diff
@@ -0,0 +1,367 @@
...
+          ) : auditType === 'BALLOT_POLLING' ? (
+            <div>
+              <MapLabelsRow>
+                <MapLabelsBoxes className="success" /> Manifest uploaded
+              </MapLabelsRow>
+              <MapLabelsRow>
+                <MapLabelsBoxes className="danger" /> Manifest upload failed
+              </MapLabelsRow>
+              <MapLabelsRow>
+                <MapLabelsBoxes className="gray" /> No Manifest uploaded
```
jonahkagan, line 343: change to "No manifest uploaded"

client/src/components/MultiJurisdictionAudit/Progress/ProgressMap.tsx
```diff
@@ -0,0 +1,367 @@
...
+          {isRoundStarted ? (
+            <div>
+              <MapLabelsRow>
+                <MapLabelsBoxes className="success" /> Complete
+              </MapLabelsRow>
+              <MapLabelsRow>
+                <MapLabelsBoxes className="progress" /> In-progress
+              </MapLabelsRow>
+              <MapLabelsRow>
+                <MapLabelsBoxes className="gray" /> Not Started
```
jonahkagan, line 331: Change to "Not started"

jonahkagan (commented):
Very cool! The map seems to work perfectly. The main challenge I see is that the jurisdiction names very rarely match. I was poking through a few past audits and the most common issues I see are:
- the jurisdictions provided by the user have "County" in the name, while the topology data doesn't
- the jurisdictions provided by the user have different capitalization than the topology data

It might be worth attempting to normalize the jurisdiction names when matching.

client/src/components/MultiJurisdictionAudit/_mocks.ts
```diff
@@ -445,6 +451,12 @@ export const aaApiCalls = {
     url: '/api/election/1/standardized-contests/file',
     response,
   }),
+  getMapData: {
+    url: 'https://cdn.jsdelivr.net/npm/us-atlas@3/counties-10m.json',
+    response: new Response(
+      'https://cdn.jsdelivr.net/npm/us-atlas@3/counties-10m.json'
+    ),
```
jonahkagan, line 458: does this mean our tests rely on connecting to this CDN now? wondering if we can just put in some fake local data instead

client/src/components/MultiJurisdictionAudit/useJurisdictions.ts
```diff
@@ -22,6 +22,7 @@ export enum JurisdictionRoundStatus {
   NOT_STARTED = 'NOT_STARTED',
   IN_PROGRESS = 'IN_PROGRESS',
   COMPLETE = 'COMPLETE',
+  FAILED = 'FAILED',
```
jonahkagan, line 25: `FAILED` isn't a status that gets returned from the server, so i think it's a bit confusing to have it be part of this type. this type represents the jurisdiction's status with regard to the ballots it's been assigned to audit during the round.

two ideas for what `getJurisdictionStatus` should return:
- a union of `JurisdictionRoundStatus | FileProcessingStatus`
- a new enum `JurisdictionProgressStatus` that has states for file uploads (e.g. `UPLOADS_NOT_STARTED`, `UPLOADS_IN_PROGRESS`) as well as for auditing progress (`AUDITING_NOT_STARTED`, `AUDITING_IN_PROGRESS`)

client/src/components/MultiJurisdictionAudit/useSetupMenuItems/_mocks/index.ts
```diff
@@ -761,6 +761,88 @@ export const jurisdictionMocks: { [key: string]: IJurisdiction[] } = {
       currentRoundStatus: null,
     },
   ],
+  oneCompleteWithAlbamaJurisdictions: [
```
jonahkagan, line 764: typo, should be Alabama. and same typo below

client/src/components/MultiJurisdictionAudit/Progress/ProgressMap.tsx
```diff
@@ -0,0 +1,338 @@
...
+  }
+
+  const getJurisdictionStatusClass = useCallback(
+    (countyName: string) => {
+      const filteredJurisdiction = jurisdictions.find(jurisdiction => {
+        const jurisdictionNameLower = jurisdiction.name.toLowerCase()
+        const countyNameLower = countyName.toLowerCase()
+        return (
+          jurisdictionNameLower === countyNameLower ||
+          (jurisdictionNameLower.includes(' county') &&
```
jonahkagan, line 151: i think you can probably drop the `includes` since `replace` will just return the original string if it doesn't contain `' county'`

client/src/components/MultiJurisdictionAudit/Progress/ProgressMap.tsx
```diff
@@ -0,0 +1,353 @@
...
+  isRoundStarted,
+  auditType,
+}: IProps) => {
+  const width = 960
+  const height = 500
+  const d3Container = useRef(null)
+  const tooltipContainer = useRef(null)
+
+  const [jsonData, setJsonData] = useState<Topology | undefined>(undefined)
+  const [matchedJurisdictions, setMatchedJurisdictions] = useState<boolean>(
```
jonahkagan, line 136: i think this is derived from `jsonData` and `jurisdictions`, so it doesn't need  to be its own state variable (less state => less confusing state bugs 😄 )

client/src/components/MultiJurisdictionAudit/Progress/ProgressMap.tsx
```diff
@@ -0,0 +1,353 @@
...
+          .replace('county', '')
+          .trim()
+      )
+      const filteredCounties = Object.values(usCounties).filter(county =>
+        jurisdictionNames.includes(
+          county.properties && county.properties.name.toLowerCase()
+        )
+      )
+
+      if (filteredCounties.length / jurisdictions.length >= 0.5) {
```
jonahkagan, line 245: is there a reason for hiding the map within this `useEffect`? seems like it would be a bit simpler to just not render the map at all if we don't meet the threshold (similar to what you did with the labels below)

client/src/components/MultiJurisdictionAudit/Progress/ProgressMap.tsx
```diff
@@ -0,0 +1,353 @@
...
+  const getJurisdictionStatusClass = useCallback(
+    (countyName: string) => {
+      const filteredJurisdiction = jurisdictions.find(jurisdiction => {
+        const jurisdictionNameLower = jurisdiction.name.toLowerCase()
+        const countyNameLower = countyName.toLowerCase()
+        return (
+          jurisdictionNameLower === countyNameLower ||
+          jurisdictionNameLower.replace(' county', '') === countyNameLower
+        )
+      })
```
jonahkagan, line 156: I'm noticing that the logic that matches jurisdiction names to county names is duplicated here and below. Maybe a way to simplify would be to compute a mapping from county name to jurisdiction in the beginning, and then use that throughout the component.

```ts
const countyNameToJurisdiction = Object.fromEntries(
  countyNames.map(countyName =>
    [countyName, jurisdictions.find(jurisdiction =>
      jurisdiction.name.toLowerCase().replace(/\s+county(\s+|$)/i, '') === countyName.toLowerCase()
    )]
  )
)

// To get a jurisdiction by county name
countyNameToJurisdiction[countyName]

// To check how many jurisdictions matched a county
Object.values(countyNameToJurisdiction).filter(jurisdiction => jurisdiction !== null).length

File: client/src/components/MultiJurisdictionAudit/Progress/index.tsx [hunk 17]
```diff
@@ -97,50 +107,51 @@ const Progress: React.FC<IProps> = ({
...
-              {numComplete}/{files.length} files uploaded
-            </Status>
-          )
+        const jurisdictionStatus = getJurisdictionStatus(jurisdiction)
+        switch (jurisdictionStatus) {
+          case JurisdictionProgressStatus.UPLOADS_COMPLETE:
+            return (
+              <Status intent="success">
+                {(auditType === 'BALLOT_POLLING' && 'Manifest uploaded') ||
+                  `${numComplete}/${files.length} files uploaded`}
```
jonahkagan, line 124: could use the ternary operator here: `<condition> ? <value> : <else value>`

jonahkagan (approved):

Issue #1214, opened 2021-06-02
Add "totals by jurisdiction" QA file to AA audit setup review screen


mcchilders:
We already have a script to add up all the batches in each jurisdiction's manifest and report those totals out by jurisdiction - we've pulled this data out several times to help QA jurisdictions' manifests' and make sure they don't have totally nonsensical data in there. Now we'd like to standardize that process by providing that file programmatically on the AA's audit setup review screen. This feature should be data available at the time the download is generated, so while jurisdictions are still uploading data, downloading multiple times should produce different data and that is desirable.

CoA's:
- Add a file download option to the AA audit setup review screen labeled "Ballot totals by jurisdiction"
- This file should total the ballot counts in various types of file uploads we get by jurisdiction (which files depends on the type of audit) and dump them into a single CSV with the following fields:
-- "Jurisdiction Name" (all audit types)
-- "Total Ballots in Manifest" (all audit types except hybrid, add up all the ballot counts for the batches in each jurisdiction's manifest)
-- "Total Ballots in CVR" (ballot comparison and hybrid, count the ballot rows in the CVR file for each jurisdiction)
-- "Valid Voted Ballots in Batches (Minimum)" (batch comparison, add up all the vote counts for each batch - divided the number of votes allowed - and total for each jurisdiction's )
-- "Totally Ballots in Manifest (CVR)" (hybrid)
-- "Total Ballots in Manifest (No-CVR)" (hybrid)
- This data should be pulled based on the data in the db at the point the download is triggered, point-in-time snapshot style

That should give us reports that look something like this for each audit type:

Ballot Polling:
![Screen Shot 2021-06-02 at 1.06.38 PM.png](https://zube.io/files/votingworks/79e2756a693a4db0e1c7620b35278bbf-screen-shot-2021-06-02-at-1-06-38-pm.png)
Batch Comparison:
![Screen Shot 2021-06-02 at 1.06.45 PM.png](https://zube.io/files/votingworks/e4591df8d331325b62b13d97323f1f31-screen-shot-2021-06-02-at-1-06-45-pm.png)
Ballot Comparison:
![Screen Shot 2021-06-02 at 1.07.04 PM.png](https://zube.io/files/votingworks/bea51cf88b7d754310234810d15809a6-screen-shot-2021-06-02-at-1-07-04-pm.png)
Hybrid: 
[...]

PR #1227, opened 2021-06-14
Audit progress download button

jonahkagan:
Task: #1214 

Adds a button to download the progress table as a CSV. In order to get the data out of the table to download, we read it directly from the DOM. While it would be a bit cleaner and more flexible to generate the CSV directly from React state data, there's not a reasonably straightforward way to access the rendered state of the table component, so we'd have to duplicate the logic that renders the columns (the status column in particular is thorny). The upside of this approach is that users shouldn't be surprised at all - what they see on the page is exactly what they'll get in the CSV (minus formatting).

Screen Shot 2021-06-14 at 4 42 47 PM
Screen Shot 2021-06-14 at 4 42 58 PM

### DIFF
--- a/client/src/components/Atoms/Table.tsx
+++ b/client/src/components/Atoms/Table.tsx
@@ -1,7 +1,8 @@
 import React from 'react'
 import { useTable, useSortBy, Column, Row } from 'react-table'
 import styled from 'styled-components'
-import { Icon, HTMLTable } from '@blueprintjs/core'
+import { Icon, HTMLTable, Button } from '@blueprintjs/core'
+import { downloadFile } from '../utilities'
 
 const StyledTable = styled.table`
   width: 100%;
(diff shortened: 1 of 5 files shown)


MorganLove:
Looks awesome! Do you think it's possible to make it export only the data visible during filtering, or would that be too difficult with this approach? Or it may not be a useful feature either. I was just thinking that if the focus is on printing exactly what's seen in the table, then people might expect the filter feature to be reflected in the download as well.

jonahkagan:
@MorganLove I think it will only export the data visible during filtering already, since the filtered out data won't be rendered in the DOM.

MorganLove:
> @MorganLove I think it will only export the data visible during filtering already, since the filtered out data won't be rendered in the DOM.

I tested it and it rendered the rows that were filtered out. I checked before commenting, because I was curious if it would behave that way.

MorganLove (approved):

Issue #1208, opened 2021-06-01
Add JA activity events to Slack channel


mcchilders:
Right now the Arlo-activity channel is showing only AA activity - unclear whether this was a bug or just a miscommunication about the original scope. Let's add JA events, too!

JA's
- File uploads (manifest, batch totals, CVRs)
- Set number of audit boards
- For batch or offline audits: entering data

AB's
- Submit final ballot data

PR #1228, opened 2021-06-15
Jurisdiction activity notifications

jonahkagan:
Task: #1208 

Add activity log notifications for jurisdiction level events.

### DIFF
--- a/server/activity_log/activity_log.py
+++ b/server/activity_log/activity_log.py
@@ -54,6 +54,33 @@ class CalculateSampleSizes(Activity):
     pass
 
 
+@dataclass
+class JurisdictionActivity(Activity):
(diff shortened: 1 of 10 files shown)


benadida (approved):
the use of `isinstance` is making me a little sad, though I suppose going OO here would be... overkill? anyways I can live with this :)

jonahkagan:
I wanted all Slack specific code to be in its own module, and I also wanted the different activities to have a structured type. So I needed a basic datatype that could have different variants (aka a tagged union or sum type). Python doesn't have that baked in, but this seemed like a good approximation. https://stackoverflow.com/questions/16258553/best-way-to-define-algebraic-data-types-in-python

Issue #1179, opened 2021-04-28
Handle JA states when round start has errored for some reason


mcchilders:
- Let's figure out what we show now - right now it's undefined
- If the AA round start errors, or while the sample is being drawn for the next round, JA view should be whatever the prior state was (just don't move them forward until the round is ready/complete

jonahkagan:
Also when drawing the sample is in progress

PR #1229, opened 2021-06-16
Show JA setup screen while drawing sample/if sample errors

jonahkagan:
Task: #1179 

When the sample is being drawn or if drawing the sample errors, the audit hasn't started yet, and the JA can't begin their auditing. So we keep them on the setup screen.


MorganLove (approved):
@jonahkagan this is a neat and elegant way of doing this!

Issue #1219, opened 2021-06-03
Add "undo launch" button if audit has launched the 1st round but no JA action has been taken yet


mcchilders:
JA actions that would prevent the "unlaunch" (at which point we should hide the button):
- Setting audit boards

This button should also appear with the error message if launching a round errors for some reason

PR #1231, opened 2021-06-16
Undo audit launch button

jonahkagan:
Task: #1219 

Adds a button to undo launching the audit as long as no jurisdiction has started auditing yet.

https://user-images.githubusercontent.com/530106/122292420-d5b36c80-ceaa-11eb-9420-41d64e95fbd6.mov

### DIFF
--- a/server/api/rounds.py
+++ b/server/api/rounds.py
@@ -1056,3 +1056,37 @@ def list_rounds_jurisdiction_admin(
     election: Election, jurisdiction: Jurisdiction  # pylint: disable=unused-argument
 ):
     return jsonify({"rounds": [serialize_round(r) for r in election.rounds]})
+
+
+@api.route("/election/<election_id>/round/<round_id>", methods=["DELETE"])
+@restrict_access([UserType.AUDIT_ADMIN])
+def undo_create_round(election: Election, round: Round):
+    current_round = get_current_round(election)
+    if not current_round or current_round.id != round.id:
+        raise Conflict(
+            "Cannot undo starting this round because it is not the current round."
+        )
+
+    if len(list(round.audit_boards)) > 0:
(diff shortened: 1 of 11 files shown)


server/api/rounds.py
```diff
@@ -1033,3 +1033,37 @@ def list_rounds_jurisdiction_admin(
...
+            SampledBallot.query.join(Batch)
+            .join(Jurisdiction)
+            .filter_by(election_id=election.id)
+            .filter(not_(SampledBallot.draws.any()))
+            .with_entities(SampledBallot.id)
+            .subquery()
+        )
+    ).with_entities(SampledBallot).delete(synchronize_session=False)
+
+    db_session.commit()
```
benadida, line 1067: I *think* there is a potential race condition - albeit an unlikely one - if a jurisdiction creates audits boards right after the check for audit boards but before the DB call to delete the round.

The only way to make this strictly work is to ensure both are locking on the same entity and do a `select for update` here on that/those entity/ies.

Thoughts?

MorganLove:
@jonahkagan should the warning text on the launch popup change to "this cannot be undone once jurisdictions have started auditing"?

server/api/rounds.py
jonahkagan: I agree there's a race condition, and that it's fairly unlikely. I believe if this occurred, the audit boards that had just gotten created would instantly be deleted by cascade, since they reference the round. I think that's ok behavior in this case, since the jurisdiction wouldn't have been able to actually start auditing in that very short moment. It might even be better behavior because then the audit admin that was trying to undo launch can continue doing that, instead of getting blocked.

benadida: 👍

Issue #1230, opened 2021-06-16
Ensure that ballots "not found" in ballot polling are counted as zombie ballots


mcchilders:
Now that the "not found" button is an option in ballot polling as well as ballot comparison, we need to handle phantom ballots as zombies in ballot polling, as well.

To do this, ballots selected for audit that are marked "not found" by the auditors should contribute 1 vote for *every loser* to the total vote counts.

PR #1232, opened 2021-06-18
Ballot polling: count not found ballots for the loser

jonahkagan:
Task: #1230 

This change only applies to ballot polling audits with online data entry. We count the number of ballots not found by the audit boards and pass that into the ballot polling math, which counts those ballots as a vote for every loser. We also add a note to the audit report about not found ballots.


benadida (approved):
looks good to me, but would like to see @umbernhard review the math, too.

umbernhard (approved):

Issue #1212, opened 2021-06-02
Better totals row detection/rejection


mcchilders:
- Make sure the last row doesn't magically add up to the number of ballots in all the prior rows; if so, throw an error "looks like you might have a totals row, if so, delete"
- Update the error message to ask ppl to remove the entire totals row
- Look for the word "Total" or "totals" or "subtotal" (any capitalization, any position, but not within another word) - doesn't have to be the whole contents, just contained within the cell.

PR #1233, opened 2021-06-21
csv_parse: improve total row detection

jonahkagan:
Task: #1212 

- Check for the words total(s) or subtotal(s) anywhere in the cell
- Check if the values in a column sum up to the value in the last row in the column

### DIFF
--- a/server/util/csv_parse.py
+++ b/server/util/csv_parse.py
@@ -1,4 +1,5 @@
 # pylint: disable=stop-iteration-return
+from collections import defaultdict
 from enum import Enum
 from typing import List, Iterator, Dict, Any, NamedTuple, Tuple
 import csv as py_csv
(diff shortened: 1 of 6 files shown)


server/util/csv_parse.py
```diff
@@ -258,14 +261,37 @@ def reject_duplicate_values(
...
+
+
 def reject_total_rows(csv: CSVDictIterator) -> CSVDictIterator:
     for r, row in enumerate(csv):  # pylint: disable=invalid-name
         for value in row.values():
-            if value.lower() in ["total", "totals", "total ballots", "county totals"]:
-                raise CSVParseError(f"Remove total row (row {r+2})")
+            if TOTAL_REGEX.search(value):
+                raise CSVParseError(
+                    f"It looks like you might have a total row (row {r+2})."
```
eventualbuddha, line 272: Should it include the name of the row?

server/util/csv_parse.py
```diff
@@ -258,14 +261,37 @@ def reject_duplicate_values(
...
+
+    for row in csv:
+        for column in columns:
+            value = row.get(column.name)
+            if column.value_type == CSVValueType.NUMBER and value is not None:
+                column_values[column.name].append(value)
         yield row
 
+    for values in column_values.values():
+        if sum(values[:-1]) == values[-1]:
```
eventualbuddha, line 289: Is there a chance for this to be a false positive?

server/util/csv_parse.py
jonahkagan: That would require some knowledge about which column is the "name" of the row. We could decide that, but I don't know that it's necessary

server/util/csv_parse.py
jonahkagan: For sure, but I guess that's a risk we're willing to take? If it's a false positive, a possible fix is swapping that row with a different row (since it only checks the last row)

server/util/csv_parse.py
eventualbuddha: I think I was thinking "column" when I wrote this. Nevermind!

benadida (approved):

PR #1234, opened 2021-06-23
Support tools API and test cleanup

jonahkagan:
- When a query errors, catch it using the react-query `onError` handler instead of using the `isError` flag while rendering. Triggering a toast side-effect while rendering can cause duplicate toasts, since rendering might happen more than once.
- Refactor the tests so that we can type check the mock data
- Reorder tests so the error cases are paired with the success cases


MorganLove (approved):
This makes sense to me, and is a lot cleaner. Nice work!

Issue #1222, opened 2021-06-07
Add expected batch totals and discrepancy flag to audit report for batch comp


mcchilders:
In the audit report for ballot comparison, we include auditor input for all sampled ballots, the expected votes for those ballots, a flag for whether there's a discrepancy between the two, and an indication of the discrepancy (+/- 1/2).

In batch comparison audits, however, our audit report is currently showing only the auditor entries for a sampled batch. We'd like to add the rest of the data to the audit report, so that it also shows:
- expected totals for each sampled batch
- a flag if there's a discrepancy between a particular batch
-- if possible, some indication of what the max possible error/taint per batch w/discrepancies ... the math calculates this value so we should be able to pull it in

jonahkagan:
started here: https://github.com/votingworks/arlo/compare/batch-tallies-in-report?expand=1

PR #1240, opened 2021-07-06
Reported batch results and discrepancies in audit report

jonahkagan:
Task: #1222 

Reworks `macro.compute_error` to return a `BatchError` dict so that we can return both the raw error in votes and the weighted error (mirroring what we do in `supersimple` for ballot comparison). Then adds the raw error to the audit report, along with the reported results from the batch tallies file. We count and report negative errors as 0, while batches with no error are counted as 0 but reported as blank.

### DIFF
--- a/server/api/reports.py
+++ b/server/api/reports.py
@@ -692,7 +692,7 @@ def sampled_batch_rows(election: Election, jurisdiction: Jurisdiction = None):
         }
 
         audit_results = {
-            choice.name: next(
+            choice: next(
                 (
                     result.result
                     for result in batch.results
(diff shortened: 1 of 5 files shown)


server/api/reports.py
```diff
@@ -686,25 +671,73 @@ def sampled_batch_rows(election: Election, jurisdiction: Jurisdiction = None):
...
+            for choice in contest.choices
+        }
+
+        is_audited = len(batch.results) > 0
+
+        error = (
+            macro.compute_error(
+                {contest.name: reported_results},
+                {contest.name: audit_results},
+                # Construct the contest object by hand instead of using
```
umbernhard, line 712: Would it make more sense to add another `from_db_contest` function or add a flag to the existing one that does this? Are we likely to have to do this again?

umbernhard (approved):
It doesn't look like the "not found counted as vote for the loser" is tested here, is it tested somewhere else already?

jonahkagan:
> It doesn't look like the "not found counted as vote for the loser" is tested here, is it tested somewhere else already?

I think in batch comparison we have to rely on the auditors to count not found ballots for the loser, since they enter a count for the entire batch.

umbernhard:
doh, that's right. My bad!

server/api/reports.py
jonahkagan: After reading this comment, I ended up figuring out another way to do this that just uses `from_db_contest` instead

PR #1242, opened 2021-07-07
Switch from pipenv to poetry

jonahkagan:
pipenv cannot update a single package at a time, instead forcing you to update every package to its latest version, which is blocking us from upgrading some packages we want upgraded (e.g. sentry-sdk) and fixing security vulnerabilities from dependabot (e.g. #1211). This is a long running issue with pipenv (see https://github.com/pypa/pipenv/issues/2412, among many other issues filed on the project). After trying many workarounds, I decided to try switching to another leading package manager, poetry, which seems to work very smoothly so far!


server/audit_math/sampler.py
```diff
@@ -121,16 +121,19 @@ def draw_ppeb_sample(
 
     # Map each batch to its weighted probability of being picked
     weighted_errors = [
-        macro.compute_max_error(batch_results[batch], contest) / U
+        float(macro.compute_max_error(batch_results[batch], contest) / U)
```
umbernhard, line 124: Why this change?

server/audit_math/sampler.py
```diff
@@ -121,16 +121,19 @@ def draw_ppeb_sample(
...
         for batch in batch_results
     ]
 
-    sample: List[Tuple[Any, Any]] = (
-        generator.choice(
-            list(batch_results.keys()),
-            sample_size + num_sampled,
-            p=weighted_errors,
-            replace=True,
+    sample: List = (
```
umbernhard, line 128: Same comment as above. I'm guessing something about typechecking changed with poetry?

umbernhard (approved):

server/audit_math/sampler.py
jonahkagan: Yeah, I let numpy get upgraded to it's latest version, which slightly changed the typechecking for that library. I could just lock it back to the version we had it at before, but it seemed like a minor enough change to let it slide

PR #1243, opened 2021-07-08
Show all lines of imprinted ID in placeholders

jonahkagan:
Certain long imprinted ids were getting cut off. The splitTextToSize function was splitting on the space between "Imprinted ID:" and the long imprinted id, causing the id to be moved to the second line (which we then didn't print).

It's still not totally clear to me why splitTextToSize did that (especially because it did the right thing and cut off the imprinted id in the middle of the id for the labels on the same audit), but this is a viable fix since we have plenty of space on the placeholders.


MorganLove (approved):
Ah! That would do it. I agree this makes sense as a solution.

PR #1244, opened 2021-07-08
Build Heroku app with poetry

jonahkagan:
We have to add a specific community buildpack.


pyproject.toml
```diff
@@ -25,6 +26,7 @@ sentry-sdk = {version = "^1.2.0", extras = ["flask"]}
 athena = {git = "https://github.com/filipzz/athena.git", rev = "v0.8.3"}
 psycopg2-binary = "~2.8.6"
 consistent-sampler = {git = "https://github.com/votingworks/consistent_sampler.git"}
+scipy = "~1.6.0"
```
jonahkagan, line 29: this was missing even though we rely on it

umbernhard: huh.

umbernhard (approved):

PR #1246, opened 2021-07-10
Statusbox fix for batch comparison

### LINKED ISSUE #1238: Bug: AA status bar shows incorrect status on batch audits?
For some reason, the status bar at the top of the AA interface shows that all audit boards are complete immediately after you start a batch audit (when no audit boards are complete?) Can we correct or remove that sentence?

MorganLove:
**Description**

Closes #1238 

**Testing**

- Currently no test updates

**Notes**

So the bug was in JA login after Batch Audit is launched. Basically, for Batch Comparison, as it's about batches, the Audit Board API has numSampledBallots & numAuditedBallots both always as 0. Due to that it shows "1 of 1 Audit Boards complete."

For now, this is a frontend fix by popping "1 of 1 Audit Boards complete" from the details & not thus showing this line for Batch Comparison.

As a backend fix, we would need to update the API to calculate numSampledBallots based on total numBallots across all Batches.

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
+++ b/client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
@@ -340,6 +340,7 @@ describe('StatusBox', () => {
             cvrs={{ file: null, processing: null }}
             auditType="BALLOT_POLLING"
             auditName="Test Audit"
+            isAuditOnline
           />
         </Router>
       )
@@ -366,6 +367,7 @@ describe('StatusBox', () => {
             cvrs={{ file: null, processing: null }}
             auditType="BALLOT_POLLING"
             auditName="Test Audit"
+            isAuditOnline
           />
         </Router>
       )
@@ -388,6 +390,7 @@ describe('StatusBox', () => {
             cvrs={{ file: null, processing: null }}
             auditType="BALLOT_POLLING"
             auditName="Test Audit"
+            isAuditOnline
           />
         </Router>
       )
@@ -409,6 +412,7 @@ describe('StatusBox', () => {
             cvrs={{ file: null, processing: null }}
             auditType="BALLOT_POLLING"
             auditName="Test Audit"
+            isAuditOnline
           />
         </Router>
       )
@@ -430,6 +434,7 @@ describe('StatusBox', () => {
             cvrs={{ file: null, processing: null }}
             auditType="BALLOT_POLLING"
             auditName="Test Audit"
+            isAuditOnline
           />
         </Router>
       )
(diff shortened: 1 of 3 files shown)


client/src/components/MultiJurisdictionAudit/StatusBox/index.tsx
```diff
@@ -367,6 +367,10 @@ export const JurisdictionAdminStatusBox = ({
     const details = [
       `${numCompleted} of ${auditBoards.length} audit boards complete.`,
```
jonahkagan, line 368: thinking about this more, this sentence really only makes sense for online data entry audits, since in all offline audits, all of the audit board results are entered at once

MorganLove: @jonahkagan yeah, good point. Which means this should work fine for this, but we could also just filter it out for all offline audits. I'm not seeing that online/offline is available in the status box yet, though. I wonder how much it would take to add that in...

jonahkagan: it's in the audit settings, so should be easy

MorganLove: @jonahkagan okay, let me know what you think of my latest update.

client/src/components/MultiJurisdictionAudit/index.tsx
```diff
@@ -225,6 +225,7 @@ export const JurisdictionAdminView: React.FC = () => {
           auditBoards={auditBoards}
           auditType={auditSettings.auditType}
           auditName={auditSettings.auditName}
+          isAuditOnline={!!auditSettings.online}
```
jonahkagan, line 228: `auditSettings.online` should already be a boolean, so i don't think the `!!` is necessary. also at this point i feel like we might as well pass in the whole `auditSettings` object, though this way is fine too

client/src/components/MultiJurisdictionAudit/StatusBox/index.test.tsx
```diff
@@ -459,6 +465,34 @@ describe('StatusBox', () => {
...
+            batchTallies={{ file: null, processing: null }}
+            cvrs={{ file: null, processing: null }}
+            auditType="BALLOT_POLLING"
+            auditName="Test Audit"
+            isAuditOnline={false}
+          />
+        </Router>
+      )
+      screen.getByText('Round 1 of the audit is in progress.')
+      screen.queryByText('1 of 1 audit boards complete.')
```
jonahkagan, line 487: remove this line

jonahkagan (approved):

client/src/components/MultiJurisdictionAudit/index.tsx
MorganLove: @jonahkagan it's actually `boolean | null` which is why I went this route.

PR #1252, opened 2021-07-26
One-time code for JA login

jonahkagan:


client/src/components/HomeScreen.tsx
```diff
@@ -97,6 +100,57 @@ const LoginScreen: React.FC = () => {
...
+  } = useForm<{
+    code: string
+  }>()
+  const [submittedEmail, setSubmittedEmail] = useState<string | null>(null)
+
+  const onSubmitEmail = async ({ email }: { email: string }) => {
+    const response = await fetch('/auth/jurisdictionadmin/code', {
+      method: 'POST',
+      headers: { 'Content-type': 'application/json' },
+      body: JSON.stringify({ email }),
```
benadida, line 125: does this have CSRF protection?

client/src/components/HomeScreen.tsx
```diff
@@ -97,6 +100,57 @@ const LoginScreen: React.FC = () => {
...
+    if (response.status === 400)
+      setErrorEmail('email', { message: error.message })
+    else toast.error(error.message)
+  }
+
+  const onSubmitCode = async ({ code }: { code: string }) => {
+    const response = await fetch('/auth/jurisdictionadmin/login', {
+      method: 'POST',
+      headers: { 'Content-type': 'application/json' },
+      body: JSON.stringify({ email: submittedEmail, code }),
```
benadida, line 142: same, CSRF

server/auth/routes.py
```diff
@@ -188,24 +182,87 @@ def auditadmin_login_callback():
...
+    if user.login_code_requested_at is None or (
+        # Reuse the existing login code if it hasn't expired yet. That way if
+        # they request a new code while waiting for a slow email, we won't wipe
+        # out the code we sent when the email does come through.
+        datetime.now(timezone.utc) - user.login_code_requested_at
+        > LOGIN_CODE_LIFETIME
+    ):
+        user.login_code_requested_at = datetime.now(timezone.utc)
+
+    user.login_code_attempts = 0
```
benadida, line 217: do you need to `.add(user)` somewhere to store in DB?

server/auth/routes.py
```diff
@@ -188,24 +182,87 @@ def auditadmin_login_callback():
...
+
+
+@auth.route("/auth/jurisdictionadmin/login", methods=["POST"])
+def jurisdiction_admin_login():
+    body = request.get_json()
+    user = User.query.filter_by(email=body.get("email")).one_or_none()
+    if user is None:
+        raise BadRequest("Invalid email address.")
+
+    if user.login_code_attempts > 20:
```
benadida, line 245: I'd say 10 is fine here. If you're typo'ing it 10 times... that seems problematic. And we do want to prevent brute force :)

server/auth/routes.py
```diff
@@ -188,24 +182,87 @@ def auditadmin_login_callback():
...
+
+    db_session.commit()
+
+    return jsonify(status="ok")
+
+
+@auth.route("/auth/jurisdictionadmin/login", methods=["POST"])
+def jurisdiction_admin_login():
+    body = request.get_json()
+    user = User.query.filter_by(email=body.get("email")).one_or_none()
```
benadida, line 241: There's a race condition here that could be security problematic: if you don't lock the row, then multiple parallel requests could trounce each other and allow an attacker to brute force more than the limit of tries in parallel.

Can we do a select for update on the `user`?

benadida (commented):
overall I'm good with the direction.

A few comments.

client/src/components/HomeScreen.tsx
jonahkagan: whoops, looks like i forgot to base this branch on the csrf branch, and so i forgot to add the csrf token here. if i had based it properly, i imagine this request would have failed without the token. nice catch!

server/auth/routes.py
jonahkagan: no, when you use SQLAlchemy to load an object from the database, it remembers it in `db_session`, so when you call `db_session.commit()` it commits any changes to any objects loaded in that session

server/auth/routes.py
jonahkagan: yep, will do

server/auth/routes.py
jonahkagan: ok will change

jonahkagan:
@benadida ready for review (except that I haven't updated Cypress tests yet)

jonahkagan:
@benadida alright Cypress tests are updated and passing now!

benadida (approved):

PR #1253, opened 2021-07-28
CSRF protection

jonahkagan:
Use the flask_seasurf library (recommended by the docs of flask_talisman, which we already use in Arlo) to implement CSRF protection for all POST/PUT/DELETE API endpoints using a token set in a cookie.

### DIFF
--- a/client/src/components/testUtilities.tsx
+++ b/client/src/components/testUtilities.tsx
@@ -74,11 +74,31 @@ interface FetchRequest {
   url: string
   options?: RequestInit
   response: object | null
-  skipBody?: boolean
   error?: {
     status: number
     statusText: string
   }
+}
+
+const isEqualRequestBody = (
+  body1?: BodyInit | null,
+  body2?: BodyInit | null
+) => {
+  const formDataToObject = (body?: BodyInit | null) =>
+    body instanceof FormData
+      ? Object.fromEntries(
+          [...body.entries()].map(([key, value]) => [
+            key,
+            value instanceof File
+              ? {
+                  name: value.name,
+                  type: value.type,
+                }
+              : value,
+          ])
+        )
+      : body
+  return equal(formDataToObject(body1), formDataToObject(body2))
 }
 
 export const withMockFetch = async (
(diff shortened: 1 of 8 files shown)


client/src/components/testUtilities.tsx
```diff
@@ -140,7 +137,7 @@ export const withMockFetch = async (
       // eslint-disable-next-line no-console
       console.error('Unexpected extra fetch request:\n', { url, options })
     }
-    return new Response(JSON.stringify({}))
+    return new Response(JSON.stringify(null))
```
jonahkagan, line 140: the changes in this file are just cleanup that i ended up doing while trying to figure out why some tests were passing after i modified them to fail. they ended up not being necessary since I decided to not have the frontend send the CSRF token header if no CSRF token cookie was set, but I figured I'd include them anyway

umbernhard (approved):
Looks pretty good to me, though I'm not super comfortable with typescript.

client/src/components/utilities.ts
```diff
@@ -11,12 +11,27 @@ export const tryJson = (responseText: string) => {
...
+  Object.fromEntries(
+    document.cookie.split(';').map(pair => pair.trim().split('='))
+  )
+
+export const addCSRFToken = (options?: RequestInit) => {
+  const token = parseCookies()._csrf_token
+  if (token && options && ['POST', 'PUT', 'DELETE'].includes(options.method!))
+    return {
+      ...options,
+      headers: { ...options.headers, 'X-CSRFToken': token },
```
benadida, line 24: I recently learned that `X-` for HTTP headers is the deprecated naming scheme, and more accepted is something like `anti-csrf-token` or the like. I don't care *that* much, but thought I'd mention it in case you want to do the more modern way.

client/src/components/utilities.ts
```diff
@@ -11,12 +11,27 @@ export const tryJson = (responseText: string) => {
   }
 }
 
+const parseCookies = () =>
+  Object.fromEntries(
+    document.cookie.split(';').map(pair => pair.trim().split('='))
```
benadida, line 16: hmmm, so this is not as secure as it could be.

The most secure configuration would be to make the cookie `httpOnly`, which then means the cookie is not accessible from JavaScript, which is an XSS defense.

The CSRF token is then delivered via a call to `/csrf` or the like, that explicitly hands the CSRF cookie to the JS client.

benadida (approved):
Noting one additional security improvement we could make.

I'm good with merging this and filing a followup ticket for securing the cookie even more.

client/src/components/utilities.ts
jonahkagan: I'm fine with just using the library default here I think

client/src/components/utilities.ts
jonahkagan: Based on the [docs](https://flask-seasurf.readthedocs.io/en/latest/#ajax-usage) I have found for the library, it seems like reading the token from the cookie is the standard way to do it. It does have one [recommendation](https://docs.djangoproject.com/en/dev/ref/csrf/#acquiring-the-token-if-csrf-use-sessions-or-csrf-cookie-httponly-is-true) for when the cookie is httpOnly, which is to "...include the CSRF token in your HTML and read the token from the DOM with JavaScript." Does that sound better than reading it from the cookie to you?

I also saw this discussion on StackExchange where many people think that CSRF token cookies don't require httpOnly: https://security.stackexchange.com/questions/175536/does-a-csrf-cookie-need-to-be-httponly.

Curious to discuss

benadida: @jonahkagan after some thought and consulting with websec colleagues, I concur with your conclusion: reading `document.cookie` is best.

One thing you're doing particularly well which I want to make sure we don't lose in some future optimization: you're reading the CSRF cookie every time you make a request. That's the right thing to do. Let's not lose that by caching it in local state or something like that. (Reason: CSRF token can change for a number of reasons. Best to re-read it and use the freshest every time.)

PR #1254, opened 2021-08-02
Updating label generation

### LINKED ISSUE #987: Look at optimizing text space on ballot labels
Particularly 4 lines of text on labels - can we bump the font size down to accommodate that when necessary (and not when it's not?) Or is detecting when it's 4 lines hard?

MorganLove:
**Description**

Closes #987 

- Tweaks the font size to fit one more line by default
- Puts both the audit board name and ballot number on the first line since we know there will always be room for those two there
- Refactors label generation to support wrapping longer lines as long as there is room for more, then truncating the rest as needed, while also shrinking the font size by one more step if needed as well

**Testing**

- Updates the pdf snapshots

**Progress**

- Manually tested to check behavior with multiple pages of labels and various combinations of line lengths. I'd like to add unit tests for those though.


benadida (approved):

client/src/components/MultiJurisdictionAudit/RoundManagement/generateSheets.ts
```diff
@@ -12,6 +12,44 @@ const LABEL_MARGIN_X = 9
...
+          )
+        } else {
+          finalLines.push(
+            ...labels.splitTextToSize(
+              line,
+              LABEL_WIDTH - LABEL_PADDING_X * 2
+            )[0]
+          )
+        }
+      })
```
benadida, line 41: since you're using functional style in the outer loop with `reduce`, would be cleaner to stick to that and use a `map` on `lines`.

benadida (approved):
one suggestion, otherwise lgtm

Issue #1023, opened 2021-02-01
Convert file processing to use background task framework


jonahkagan:
The background worker currently has custom code specific to processing files (e.g. ballot manifests, jurisdiction files). Now that we have a new framework for running background tasks that we use for drawing the sample in the background, we should convert the existing file processing code to use this framework so that there aren't different ways of doing things in the background worker.

jonahkagan:
Also would be useful to change the file processing functions to clear out any previously processed data for that file so that they can also be used for reprocessing files.

PR #1261, opened 2021-08-25
Refactor: Use background task framework for file processing 

jonahkagan:
Task: #1023 

Migrate file processing to the background task framework. Now there will be only one way that background work is done. Precursor to adding a progress bar for file processing.

Some bonus effects:
- Tasks run automatically in tests
- The file processing code is now type-checked (it wasn't before)
- Code coverage bumps up
- Reprocessing files gets simpler - we just create a new task

I also moved the code for processing the jurisdictions file into the jurisdictions module so that it would match the way we structure code for all other file uploads.


server/api/batch_tallies.py
```diff
@@ -1,34 +1,34 @@
...
+def process_batch_tallies_file(jurisdiction_id: str):
+    jurisdiction = Jurisdiction.query.get(jurisdiction_id)
+
+    def process() -> None:
         # We only support one contest for batch audits, so we can just take the
         # first contest from the jurisdiction's universe.
-        assert len(jurisdiction.contests) == 1
-        contest = jurisdiction.contests[0]
+        assert len(list(jurisdiction.contests)) == 1
+        contest = list(jurisdiction.contests)[0]
```
benadida, line 31: are we doing the query twice here? Should we store `list(jurisdiction.contest)` and then work off that?

server/migrations/versions/8b6b89bfbc97_backgroundtask.py
```diff
@@ -18,31 +18,7 @@
...
-    op.create_foreign_key(
-        op.f("round_draw_sample_task_id_fkey"),
-        "round",
-        "background_task",
-        ["draw_sample_task_id"],
-        ["id"],
-        ondelete="cascade",
-    )
-
+def backfill():  # pragma: no cover
```
benadida, line 21: I'm confused as to why this existing migration is changing. Is that expected?

benadida (approved):
a couple of small questions.

server/api/batch_tallies.py
jonahkagan: sqlalchemy caches relationship query results, so there's no performance hit, it's just for making the typechecker not complain

server/migrations/versions/8b6b89bfbc97_backgroundtask.py
jonahkagan: i was just doing a small refactor to increase code coverage, but looking at it again i may have made a small mistake

jonahkagan: just pushed a fix

Issue #1262, opened 2021-08-26
File reprocessing errors don't show up unless you reload page


jonahkagan:
If a JA uploads a valid manifest, then a valid CVR file, then a new manifest that causes an error when reprocessing the CVR, it appears that there is no error. We decided that we wanted the error to show up under the CVR file, not the manifest. However, the frontend doesn't know about reprocessing, so it doesn't reload the CVR file status. If you refresh the page, the error then appears.

Solution ideas:
- Rework the frontend to know about reprocessing (potentially switching to the react-query library would give us an easy way to express the dependency)
- Get rid of complicated reprocessing - when a JA changes the manifest after uploading a CVR, tell them they have to reupload the CVR file afterwards.

Note that this same issue exists for other reprocessing situations:
- manifest + batch tallies
- jurisdictions file + standardized contests

PR #1265, opened 2021-08-26
Reload files on frontend after reprocessing

jonahkagan:
Task: #1262 

After a file gets reprocessed due to a file it depends on being changed, we need to reload that file to see what happened (specifically, we need to show errors).

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/useCSV.ts
+++ b/client/src/components/MultiJurisdictionAudit/useCSV.ts
@@ -54,7 +54,8 @@ const deleteCSVFile = async (url: string): Promise<boolean> => {
 const useCSV = (
   url: string,
   formKey: string,
-  shouldFetch: boolean = true
+  shouldFetch: boolean = true,
+  dependencyFile?: IFileInfo | null
 ): [
   IFileInfo | null,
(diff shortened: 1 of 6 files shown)


client/src/components/MultiJurisdictionAudit/useCSV.ts
```diff
@@ -64,9 +65,14 @@ const useCSV = (
 
   useEffect(() => {
     ;(async () => {
-      setCSV(await loadCSVFile(url, shouldFetch))
+      // Reload this file whenever a file it depends finishes processing. This is
+      // useful when one file gets reprocessed on the backend after its dependency
+      // processes (e.g. CVR depends on ballot manifest).
+      if (dependencyFile === undefined || dependencyFile) {
```
MorganLove, line 71: The concept makes sense, but I'm struggling to grok exactly how this line does it. Maybe clarify more how passing in the dependencyFile object informs this of when a reload is needed?

jonahkagan: oh yeah i see now how that line is confusing. i just simplified it, hopefully that makes it clearer

MorganLove (approved):
LGTM!

PR #1266, opened 2021-08-30
Refactor CSVFile component  to prevent extra rerenders 

jonahkagan:
Previously, the CSVFile component was set up to use a key to force the form to rerender every time we reloaded a new file from the server. This was proving to be problematic in some other changes to the form I was working on, so I figured out a way to preserve the existing behavior without rerendering the entire form on every server file reload.

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/CSVForm/index.tsx
+++ b/client/src/components/MultiJurisdictionAudit/CSVForm/index.tsx
@@ -86,9 +86,9 @@ const CSVFile = ({
                 )}
               </FormSectionDescription>
             </FormSection>
-            <FormSection>
-              {isEditing || !file || isProcessing ? (
-                <>
+            {isEditing || !file || isProcessing ? (
+              <>
+                <FormSection>
                   <FileInput
                     inputProps={{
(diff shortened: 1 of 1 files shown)


client/src/components/MultiJurisdictionAudit/CSVForm/index.tsx
```diff
@@ -131,43 +140,35 @@ const CSVFileForm = ({
...
-                >
-                  Upload File
-                </FormButton>
-              ) : (
-                // We give these buttons a key to make sure React doesn't reuse
-                // the submit button for one of them.
-                <>
+                </FormSection>
+                <div>
+                  {/* We give these buttons a key to make sure React doesnt
```
MorganLove, line 145: "doesn't"

MorganLove (approved):

client/src/components/MultiJurisdictionAudit/CSVForm/index.tsx
jonahkagan: lol i had to take out the apostrophe because the linter wanted me to escape it

Issue #1009, opened 2021-01-19
Better UX feedback while CVR file is uploading/processing


mcchilders:
For large CVR file uploads, be able to show a progress bar/indication of if the upload has timed out/errored etc.

MorganLove:
@mcchilders What kind of feedback were you envisaging?

PR #1275, opened 2021-09-08
CVR processing progress

jonahkagan:
Task: #1009 

Show a progress bar when CVR is processing.


server/api/offline_results.py
```diff
@@ -2,6 +2,7 @@
 from typing import List, Optional
 from flask import jsonify, request
 from werkzeug.exceptions import BadRequest, Conflict
+from sqlalchemy import func
```
benadida, line 5: I'm a little confused about these `import` statements in a few files that don't seem to come with any other changes. What's going on there?

benadida (approved):

server/api/offline_results.py
jonahkagan: Oh right, should have explained that. Previously, there was a wildcard import `import * from sqlalchemy` in models.py, so all of these names like `func` were imported implicitly and then exported to every file that imported models.py. That caused a conflict when importing the variable `engine` from database.py, since sqlalchemy also has a name `engine` that's something different. I decided to get rid of the wildcard import and be more explicit to avoid issues like this in the future.

PR #1278, opened 2021-09-09
CVR upload progress

jonahkagan:
Show a progress bar during large CVR file uploads (> 1MB). To do this, we have to use `axios` (which uses XHR under the hood), since `fetch` doesn't support this feature.

https://user-images.githubusercontent.com/530106/132766937-b1b706c6-fb6f-43bf-8a59-208cef8084cb.mov

### DIFF
--- a/client/src/components/MultiJurisdictionAudit/useCSV.ts
+++ b/client/src/components/MultiJurisdictionAudit/useCSV.ts
@@ -1,5 +1,7 @@
+import axios, { AxiosRequestConfig } from 'axios'
+import { toast } from 'react-toastify'
 import { useEffect, useState } from 'react'
-import { api, useInterval } from '../utilities'
+import { api, useInterval, addCSRFToken } from '../utilities'
 import { IAuditSettings } from './useAuditSettings'
 
 export enum FileProcessingStatus {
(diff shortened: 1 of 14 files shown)


client/src/components/MultiJurisdictionAudit/useCSV.ts
```diff
@@ -63,24 +79,44 @@ const useCSV = (
   () => Promise<boolean>
 ] => {
   const [csv, setCSV] = useState<IFileInfo | null>(null)
-
+  const [upload, setUpload] = useState<IUpload | null>(null)
+
+  // Load (or reload) this file whenever a file it depends on changes. This is
```
jonahkagan, line 84: Had to refine this logic because it wasn't working correctly once I added the upload progress to the hook state. Most likely the issue was that we were using an object (`dependencyFile`) in the useEffect dependency list, which is not very reliable because the objects are checked for changes using reference equality, not structural equality. By adding the upload progress to these objects, it changed something about the way this useEffect hook noticed changes, and started causing issues.

The fix was to pull out the specific pieces of info from the object that we wanted to check.

benadida (approved):

PR #1285, opened 2021-09-15
Delete unused CvrBallots after audit completes

jonahkagan:
During a ballot comparison or hybrid audit, we need every record from the CVRs on hand in the database in case the corresponding ballot gets sampled. Once the audit completes, we only need the CVRs for the ballots that got sampled. In this change, at the end of the audit, we delete all CvrBallots that don't correspond to a sampled ballot. Since CvrBallots are the largest use of database storage by an order of magnitude, this will hopefully significantly slow database growth.


jonahkagan:
Closing for now until we decide to do this

PR #1290, opened 2021-09-24
Fixing bug where we report stopping size as round size

umbernhard:
**Description**

Currently subsequent rounds in ballot comparison audits report the estimated stopping size, not the round size, which is (stopping size) - (ballots sampled so far). This fixes that. 

**Testing**

Many test cases are broken, need @jonahkagan's help to fix them.

**Progress**

Tests need to be fixed.

### DIFF
--- a/server/api/sample_sizes.py
+++ b/server/api/sample_sizes.py
@@ -153,9 +153,19 @@ def sample_size_options(
                 "2-over": discrepancy_counter[2],
             }
 
-            sample_size = supersimple.get_sample_sizes(
+            # Since this computation comptues the total stopping size, We need
+            # to subtract the number of samples drawn already to get to a round
+            # size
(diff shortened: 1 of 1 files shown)


jonahkagan (changes requested):
I think this change would make more sense as part of `supersimple.get_sample_size`, so that the sample size function from supersimple behaves the same as all the other audit math modules. I sketched that out here: https://github.com/votingworks/arlo/pull/1292. If you agree, I think we should close this and switch to that PR.

umbernhard:
@jonahkagan The reason I wanted to implement this way was for parity with Stark's tool, but I suppose that may not be a sufficiently good reason. I like your point about keeping this consistent with our other math modules, so I'm closing and will move to the other PR.

PR #1292, opened 2021-10-04
Subtract previously sampled ballots from ballot comparison sample size

jonahkagan:
Previously, the supersimple module was reporting the stopping size (number of total ballots necessary to end the audit) as the sample size. This change subtracts out the previously sampled ballots so that the sample size is only the additional ballots that should be sampled.


umbernhard (approved):
Looks good!

PR #1296, opened 2021-10-06
Parse ClearBallot CVRs

jonahkagan:
Task: #1250 

- Refactors the CVR processing function to allow for different parsers for different file types
- Adds Jurisdiction.cvr_file_type to the db
- Adds a CVR file type selector to the CSV upload form, only displayed for CVRs
- Show the file type to the AA in the jurisdiction progress screen modal

Note that there may be increased memory usage due to the refactor, but I'll be investigating that separately and fixing in a later PR if necessary.

Screen Shot 2021-10-06 at 12 45 01 PM

Screen Shot 2021-10-06 at 12 45 44 PM

### DIFF
--- a/scripts/manifest-for-cvr.py
+++ b/scripts/manifest-for-cvr.py
@@ -4,29 +4,51 @@ import sys
 from collections import defaultdict
 
 if __name__ == "__main__":
-    if len(sys.argv) != 3:
-        print("Usage: python -m scripts.manifest-for-cvr <cvr-path> <manifest-path>")
+    if len(sys.argv) != 4:
+        print(
+            "Usage: python -m scripts.manifest-for-cvr <cvr-path> <cvr-file-type> <manifest-path>"
+        )
         sys.exit(1)
 
     with open(sys.argv[1], "r") as cvr_file:
         cvr = csv.reader(cvr_file, delimiter=",")
-
-        _election_name_row = next(cvr)
-        _contest_row = next(cvr)
-        _contest_choices_row = next(cvr)
-        headers_and_affiliations = next(cvr)
-
-        assert headers_and_affiliations[1] == "TabulatorNum"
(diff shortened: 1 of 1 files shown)


client/src/components/MultiJurisdictionAudit/Progress/JurisdictionDetail.tsx
```diff
@@ -35,6 +35,12 @@ const FileStatusTag = ({
   }
 }
 
+const prettyCvrFileType = (cvrFileType: CvrFileType) =>
+  ({
+    [CvrFileType.DOMINION]: 'Dominion',
```
benadida, line 40: I'm confused by the square brackets around the `CvrFileType.DOMINION`... You can't key on the raw CVR file type?

client/src/components/MultiJurisdictionAudit/index.test.tsx
```diff
@@ -567,26 +543,57 @@ describe('JA setup', () => {
...
       await screen.findByText('Uploaded at 11/18/2020, 9:39:14 PM.')
+      expect(fileTypeSelect).toBeDisabled()
+      within(fileTypeSelect).getByRole('option', {
+        name: 'ClearBallot',
+        selected: true,
+      })
+    })
+  })
+
+  it('after deleting CVRs, keeps last selected CVR file type ', async () => {
```
benadida, line 580: nice touch.

benadida (approved):

client/src/components/MultiJurisdictionAudit/Progress/JurisdictionDetail.tsx
jonahkagan: square brackets are the notation for using a dynamic value as a property name. I can't remember why I did it this way but it looks like TS can also figure it out if i just use the CVR file type strings directly, so i'll change it

PR #1313, opened 2021-10-21
Support Tools: only toast once on error

jonahkagan:
- Since we added a default onError handler to the QueryClient that toasts errors, the individual error toasting on each query are no longer needed (and are in fact causing duplicate toasts)
- Also fixes tests to close toasts after checking them to make sure there are no state updates happening after the test is complete (which can cause weird bugs)

### DIFF
--- a/client/src/components/SupportTools/SupportTools.tsx
+++ b/client/src/components/SupportTools/SupportTools.tsx
@@ -101,13 +101,8 @@ const Organizations = () => {
 
   if (!organizations.isSuccess) return null
 
-  const onSubmitCreateOrganization = async ({ name }: { name: string }) => {
-    try {
-      await createOrganization.mutateAsync({ name })
-      reset()
-    } catch (error) {
-      toast.error(error.message)
-    }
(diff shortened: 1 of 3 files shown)


client/src/components/SupportTools/SupportTools.tsx
```diff
@@ -203,12 +196,8 @@ const Organization = ({ organizationId }: { organizationId: string }) => {
...
       yesButtonLabel: 'Delete',
       onYesClick: async () => {
-        try {
-          await deleteOrganization.mutateAsync({})
-          toast.success(`Deleted organization ${name}`)
-        } catch (error) {
-          toast.error(error.message)
-        }
+        await deleteOrganization.mutateAsync()
+        toast.success(`Deleted organization ${name}`)
```
jonahkagan, line 200: in `onYesClick` handlers for `Confirm` dialogs, we need to use promises, because we want errors to bubble up to the dialog so that the dialog knows to stay open (vs closing on success)

PR #1324, opened 2021-10-28
Activity log for finalizing batch results

jonahkagan:


benadida (approved):
assuming this needs to be rebased, I only reviewed the last commit. LGTM.

PR #1325, opened 2021-10-28
Support Tools: Remove "clear offline results" for batch audits

jonahkagan:
It won't be necessary now that JAs can edit results


benadida (approved):
assuming rebase, only reviewed last commit.

PR #1326, opened 2021-10-28
Button for AA to unfinalize batch results

jonahkagan:


server/api/batches.py
```diff
@@ -241,3 +241,28 @@ def finalize_batch_results(
...
+    round: Round,
+):
+    results_finalized = BatchResultsFinalized.query.filter_by(
+        jurisdiction_id=jurisdiction.id, round_id=round.id
+    ).one_or_none()
+    if results_finalized is None:
+        raise Conflict("Results have not been finalized")
+
+    if round.ended_at is not None:
+        raise Conflict("Results cannot be unfinalized after the audit round ends")
```
benadida, line 263: there's a race condition possibility here: if the round is loaded for this function call first, then the round is finalized / ends in a separate process/thread, then `ended_at` is checked here, it's null because loaded earlier, then the results are removed, leaving the system in an inconsistent state.

I would recommend joining against round in the `BatchResultsFinalized` query with the condition on `ended_at` in the query, so that the `delete` is issued conditionally. If the round is finalized in parallel, the delete won't delete anything.

You could then shift the error reporting to just looking at whether `delete` affected any rows.

benadida (changes requested):

server/api/batches.py
jonahkagan: @benadida fixed with a `for update` lock here: eb0694040ba91ec360ab7173dabf75304c81bf28

benadida (approved):

PR #1328, opened 2021-10-29
Fixing headers for DataEntry

### LINKED ISSUE #1319: Adjust data entry UI labels
Since the data entry UI redesign we've had a mismatch between how we used to describe ballot locations (and currently still do in other places in the UI) and what the data entry screen now shows, which is confusing for users. Right now it looks like this:

We'd like it to instead have it be in this order, left to right, top to bottom, and this wording:
Container:
Tabulator:
Batch:
Ballot Number:
Imprinted ID:

That will mean adjusting:
- the order the fields appear on the page, 
- changing the labels slightly (back to what they were)
- re-capitalizing the labels

We'd also like to make a similar change on the ballot list in the data entry UI, which currently uses "Position" - let's change that to "Ballot Number", too.

Current UI that we should adjust:

MorganLove:
**Description**

Closes #1319 

**Testing**

- Snapshots updated

**Progress**

Ready for review

### DIFF
--- a/client/src/components/DataEntry/Ballot.tsx
+++ b/client/src/components/DataEntry/Ballot.tsx
@@ -245,29 +245,29 @@ const Ballot: React.FC<IProps> = ({
               </TopRow>
               <FlushDivider />
               <BallotMainRow>
-                <div>
-                  <SubTitle>batch</SubTitle>
-                  <BallotRowValue>{ballot.batch.name}</BallotRowValue>
-                </div>
-                <div>
-                  <SubTitle>position</SubTitle>
-                  <BallotRowValue>{ballot.position}</BallotRowValue>
-                </div>
                 {ballot.batch.container && (
                   <div>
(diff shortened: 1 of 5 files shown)


client/src/components/DataEntry/Ballot.tsx
```diff
@@ -245,29 +245,29 @@ const Ballot: React.FC<IProps> = ({
...
-                  <BallotRowValue>{ballot.batch.name}</BallotRowValue>
-                </div>
-                <div>
-                  <SubTitle>position</SubTitle>
-                  <BallotRowValue>{ballot.position}</BallotRowValue>
-                </div>
                 {ballot.batch.container && (
                   <div>
-                    <SubTitle>container</SubTitle>
+                    <SubTitle>cCntainer</SubTitle>
```
benadida, line 250: Typo

benadida (approved):
Typo to fix otherwise lgtm

PR #1330, opened 2021-11-01
Bugfix: Remove contests correctly

jonahkagan:


server/tests/api/test_contests.py
```diff
@@ -100,7 +100,7 @@ def test_contests_create_get_update_multiple(
 
     json_contests[0]["name"] = "Changed name"
     json_contests[1]["totalBallotsCast"] = 600
-    json_contests[2]["jurisdictionIds"] = jurisdiction_ids[1:]
+    json_contests = json_contests[:2]
```
jonahkagan, line 103: wanted to add a test for this case to the backend as well just to be safe

benadida (approved):

MorganLove (approved):
This looks like a great way to handle this!

Issue #1329, opened 2021-10-29
error when creating an audit admin that already exists with mismatched case


benadida:
repro:
- have a user that exists in DB with all lowercase email address
- try adding that same email, but with some uppercase letters, as an admin of an org
- see SQL failure

PR #1333, opened 2021-11-01
Bugfix: Create audit admins with email case-insensitity

jonahkagan:
Task: #1329 

While we were saving the email as lowercase (due to the validator on User.email), we weren't finding existing audit admins with lowercase emails.

### DIFF
--- a/server/api/support.py
+++ b/server/api/support.py
@@ -189,7 +189,7 @@ def create_audit_admin(organization_id: str):
     audit_admin = request.get_json()
     validate(audit_admin, AUDIT_ADMIN_SCHEMA)
 
-    user = User.query.filter_by(email=audit_admin["email"]).one_or_none()
+    user = User.query.filter_by(email=audit_admin["email"].lower()).one_or_none()
     if not user:
         user = User(
             id=str(uuid.uuid4()),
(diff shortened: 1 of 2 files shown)


benadida (approved):
A better fix for this, IMO:

- create the unique index on `lower(email)`, which is doable according to [docs](https://www.postgresql.org/docs/14/sql-createindex.html#:~:text=on%20the%20expression-,lower(title),-%2C%20allowing%20efficient%20case)
- attempt to create the user and, on duplicate, run with the existing user_id

Not going to block my review on this since this fixes a clear bug, but suggestion for improvement.

jonahkagan:
@benadida thinking through your suggestion, if we had a unique index on `lower(email)` and we hit the unique index violation on insert, we'd then still need a way to query for the existing user with that email in a case-insensitive way. So wouldn't that still require the same logic we have here? Or am I missing something?

benadida:
> @benadida thinking through your suggestion, if we had a unique index on `lower(email)` and we hit the unique index violation on insert, we'd then still need a way to query for the existing user with that email in a case-insensitive way. So wouldn't that still require the same logic we have here? Or am I missing something?

oooh, yeah good point. Yeah, fair.

PR #1337, opened 2021-11-02
Incorporate support for RAIRE assertions and auditing of RCV/IRV elections

michelleblom:
**Description**

Implementation of the RAIRE algorithm for generating assertions that need to be checked during an audit of a RCV/IRV election. Main RAIRE algorithm is located in server/audit_math/raire.py with utilities in server/audit_math/raire_utils.py.

**Testing**

Test election instances are located in RaireData/Input with the expected assertions to be formed for each instance located in RaireData/Output.

Tests can be run using run_raire_tests.sh. The file test_raire.py exists only for the purpose of testing.

**Progress**


umbernhard:
This was already merged via another PR, so I'm closing.

PR #1339, opened 2021-11-02
Bugfix: detect ballot comparison full hand tally based on sample size, not unique ballots sampled

jonahkagan:


server/audit_math/supersimple.py
```diff
@@ -288,8 +288,8 @@ def compute_risk(
     if 0 < p < alpha:
         result = True
 
-    if len(sample_cvr) >= N:
-        # We've done a full hand recount
+    # Special case if the sample size equals all the ballots (i.e. a full hand tally)
+    if sum(ballot["times_sampled"] for ballot in sample_cvr.values()) >= N:
```
umbernhard, line 292: @jonahkagan just to clarify here, this will trigger even if not all the ballots are being sampled. That's what we want, yes?

umbernhard (approved):

server/audit_math/supersimple.py
jonahkagan: I think so, due to sampling with replacement. My thinking was that a "full hand tally" should be defined as: sample size = total ballots cast, regardless of whether you sample with or without replacement. This line of code is just reconstructing the sample size from the actual sample drawn.

Does that check out?

umbernhard: Yeah. I think functionally it should be a lower threshold than that, but that works. It just leaves us with the case where not every unique ballot is sampled but we still call it a full hand count, which is maybe a little incoherent but if it works...

jonahkagan: That's true it is a little weird. The problem this was solving was that you could get to a weird state in round 2 if the round 1 sample size was "all the ballots" but some ballots were sampled multiple times. The p-value would be non-zero, but the round 2 sample size would be 0 - essentially a dead-end. I wanted to make the risk calculation code match the sample size estimation code in its assessment of when the audit was a full hand tally and there was no more work to be done.

That being said, we're also going to make Arlo disallow starting ballot comparison audits that sample all ballots for now, so we shouldn't be reaching this case often (only in escalated later rounds).

umbernhard: Yeah, like I said it'd be easier if we just made it a threshold lower than "nearly all  the ballots." This is okay by me for now though (which is why I approved the PR).

PR #1342, opened 2021-11-03
Poetry: Less specific python version

jonahkagan:
Requires setting DISABLE_POETRY_CREATE_RUNTIME_FILE=1 on Heroku


umbernhard (changes requested):
So I don't think this will fix my specific problem, as Ubuntu 20.04 ships with 3.8.10. Could we bump the version down to that?

jonahkagan:
Whoops, my bad. Changed to just `^3.8` - how about that?

umbernhard (approved):

PR #1343, opened 2021-11-03
Fix makefile to work on ubuntu

umbernhard:
Fixes nodejs and poetry install process to work correctly on new Ubuntu installs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ deps:
 	curl -fsSL https://deb.nodesource.com/setup_12.x | sudo -E bash -
 	sudo apt-get install -y nodejs
 	curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python3.8 -	
-	export PATH="${PATH}:~/.local/bin/"
+	export PATH="${PATH}:${HOME}/.local/bin/"
 	sudo npm install -g yarn
 	sudo apt install postgresql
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,6 +1,9 @@
 deps:
-	sudo apt install python3.8 python3-pip nodejs libpython3.8-dev libpq-dev graphicsmagick
-	python3.8 -m pip install poetry
+	sudo apt install python3.8 python3-pip libpython3.8-dev libpq-dev graphicsmagick
+	curl -fsSL https://deb.nodesource.com/setup_12.x | sudo -E bash -
+	sudo apt-get install -y nodejs
+	curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python3.8 -
```
jonahkagan, line 5: wanna add comments linking back to the documentation where you got the node/poetry install commands from? makes it easy to go back and reference

Makefile
```diff
@@ -1,6 +1,9 @@
 deps:
-	sudo apt install python3.8 python3-pip nodejs libpython3.8-dev libpq-dev graphicsmagick
-	python3.8 -m pip install poetry
+	sudo apt install python3.8 python3-pip libpython3.8-dev libpq-dev graphicsmagick
```
jonahkagan, line 2: maybe try removing `python3-pip` now that we're not using pip to install poetry

jonahkagan (approved):

PR #1348, opened 2021-11-08
Fix jurisdiction progress status text and colors

jonahkagan:
- A couple of the classes were incorrect
- We were missing one state for when file have started but are not complete yet


server/migrations/versions/f400f19f7a35_file_background_task.py
```diff
@@ -1,5 +1,6 @@
 # pylint: disable=invalid-name
 """File background task
+
```
jonahkagan, line 3: this was just a random comment that was misformatted that i ran into

MorganLove (approved):
LGTM!

PR #1350, opened 2021-11-09
Rename "sampled all ballots" to "is full hand tally"

jonahkagan:


server/api/offline_results.py
```diff
@@ -69,18 +69,11 @@ def validate_offline_results(
...
 
     for contest in jurisdiction.contests:
-        num_ballots: int
-        # Special case: if we sampled all ballots, then the max results allowed
-        # should be based on the ballot manifest, since we don't have any
-        # sampled ballots in the db
-        if sampled_all_ballots(round, election):
-            assert jurisdiction.manifest_num_ballots is not None
-            num_ballots = jurisdiction.manifest_num_ballots
-        elif contest.is_targeted:
```
benadida, line 79: what's the reasoning for removing this condition here?

benadida (approved):
one quick question about a removed conditional.

server/api/offline_results.py
jonahkagan: we don't actually use the offline results flow for full hand tallies. i think we may have put this condition in right before we built the full hand tally interface, thinking we would use this, and then forgot to take it out

PR #1355, opened 2021-11-10
Adding confirmation toast for ballot data entry

MorganLove:
**Description**

Added a green confirmation toast for when a ballot data entry is completed and when the next ballot is shown to provide clear and unambiguous direction to audit board members that the next ballot is there to be audited.

**Testing**

- Updated cypress to account for the toast

**Progress**

Manually tested and works on web and mobile.

### DIFF
--- a/client/cypress/end-to-end/ballot-comparison.spec.js
+++ b/client/cypress/end-to-end/ballot-comparison.spec.js
@@ -247,9 +247,9 @@ describe('Ballot Comparison Test Cases', () => {
         .click({ force: true })
       cy.findByRole('button', { name: 'Submit Selections' }).click()
       cy.findByText('Confirm Selections').click()
-      cy.findByText('Success! Now showing the next ballot to audit.')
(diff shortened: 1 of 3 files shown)


MorganLove:
(The cypress test appears to be being flaky, and the error doesn't appear related to this PR, so I'm not sure what's up with it.)

jonahkagan (commented):
Do the unit tests need to be updated as well?

MorganLove:
@jonahkagan I didn't add any because we were trying to get this out asap during the currently running audit, and just wanted it to pass current tests. Of course now I'm stalled on the cypress flakiness on this branch though.

jonahkagan:
I added a `findAndCloseToast` helper to the unit tests, so it might just be as simple as adding a couple of those to existing unit tests

MorganLove:
@jonahkagan I got the cypress to work, and I tried putting in the `findAndCloseToast` into the unit tests but for some reason it wasn't working. I thought I might be using it wrong so I left it in there commented out so you could take a look.

jonahkagan (approved):

PR #1362, opened 2021-11-16
Extra tests

umbernhard:
**Description**

Adds an arlo-extra-tests submodule to hold other Arlo test vectors that are too slow for CI. 

**Testing**

For now, a bunch of other RAIRE tests. 

**Progress**

Ready for review.


jonahkagan (approved):
In my (limited) experience, submodules have been kind of annoying to use. I doubt these tests will be used often, so I was imagining something where you just download the files and copy them into the test directory manually (and then delete them afterward).

That being said, if you think this is a nicer way to do it, I'm totally fine with that.

PR #1364, opened 2021-11-17
Tweak to run-dev

MorganLove:
**Description**

Adding more interaction for easier setup and launch

### DIFF
--- a/run-dev.sh
+++ b/run-dev.sh
@@ -1,8 +1,31 @@
 #!/usr/bin/env bash
 
-export ARLO_AUDITADMIN_AUTH0_BASE_URL=${ARLO_AUDITADMIN_AUTHO_BASE_URL:-http://localhost:8080}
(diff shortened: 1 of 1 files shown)


jonahkagan:
I run this script pretty often so I'd rather it not be interactive. Some ideas:
- I have environment variables configured via a separate script that runs every time I open a new terminal. That could be a good method for individual engineers to set up env vars.
- If we want to support default values for more env vars in `development`, there are two ways to go about it: 1) In config.py, you can specify default values for different environments. 2) In run-dev.sh, there's a pattern already established for setting default values. Though I think it would be better to consolidate and use the config.py approach.
- For setting up data in the database, I think a separate script makes more sense. However, the scripts for creating and org and admin are outdated, now that I think about it, since we moved that functionality to Support Tools. So really we should remove those scripts, since they are not fully accurate anymore.
- I load data into the database using pg_dump and pg_restore. We could potentially create some tooling around that if we wanted.

PR #1367, opened 2021-11-20
Clarify cypress comments

MorganLove:
Documentation and one tweak to correct a piece of the end to end test


MorganLove:
The test tweak that is failing currently was commented out prior, without explanation. I'm not sure why it's failing yet, but if it's supposed to actually be commented out, we should document why.

jonahkagan (approved):

PR #1369, opened 2021-12-06
Fixing a few bugs

umbernhard:
**Description**

Fixes bug where we don't return the lowest possible p-value in batch comparison audits. Also fixes another bug where we aren't properly filtering audited batches that don't contain the contest we're interested in. 

**Testing**

Modified the MACRO tests to include the lowest possible p-values.

**Progress**

ready.


server/audit_math/macro.py
```diff
@@ -304,7 +310,8 @@ def compute_risk(
         else:
             p *= ((1 - 1 / U) / (1 - taint)) ** times_sampled[batch]
 
-        if p <= alpha:
-            return float(p), True
+        # Make sure we're returning the lowest p-value in the sequence
+        if p <= low_p:
+            low_p = p
```
jonahkagan, line 315: Does this logic depend on the order of the batches in `sample_results`? It seems like the p-value gets multiplied each iteration of the loop, which means the order of the multiplication would determine what the lowest p-value is.

If so, I'm wondering what order we expect them to be in

umbernhard: The order should be the order they were drawn.

jonahkagan:
@umbernhard can we close this for now? I think #1371 covers part of the issue here.

umbernhard:
Yeah I think that's right.

PR #1370, opened 2021-12-07
Remove CSV streaming obstacles

jonahkagan:
While most of our CSV parsing code already uses a streaming framework, there were a few places where we cheated and loaded the whole file into memory, namely:
- Decoding
- Validating that the file is not empty
- Tracking parsing progress (getting the total lines)

This PR changes each of these pieces to be streaming-compatible (but doesn't actually stream the file contents, since there's no way to stream it out of one field in the database). This sets us up for streaming file contents from external storage.

### DIFF
--- a/server/api/cvrs.py
+++ b/server/api/cvrs.py
@@ -13,8 +13,6 @@ from werkzeug.exceptions import BadRequest, NotFound, Conflict
 from sqlalchemy import func, and_
 from sqlalchemy.orm import Session
 
-from server.util.csv_parse import decode_csv_file
-
 from . import api
 from ..database import db_session, engine as db_engine
 from ..models import *  # pylint: disable=wildcard-import
@@ -27,7 +25,14 @@ from ..worker.tasks import (
 )
 from ..util.file import serialize_file, serialize_file_processing
 from ..util.csv_download import csv_response
-from ..util.csv_parse import decode_csv_file, validate_csv_mimetype
+from ..util.csv_parse import (
+    CSVIterator,
+    decode_csv,
+    decode_csv_file,
+    validate_comma_delimited,
+    validate_csv_mimetype,
+    validate_not_empty,
+)
 from ..util.jsonschema import JSONDict
 from ..audit_math.suite import HybridPair
 from ..activity_log.activity_log import UploadFile, activity_base, record_activity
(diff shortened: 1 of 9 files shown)


server/util/csv_parse.py
```diff
@@ -61,14 +68,51 @@ def parse_csv(csv_string: str, columns: List[CSVColumnType]) -> CSVDictIterator:
...
+    try:
+        contents = file.read()
+        if contents == b"":
+            return ""
+        return decode_csv(io.BytesIO(contents)).read()
+    except CSVParseError as err:  # pragma: no cover
+        raise BadRequest(INVALID_CSV_ERROR) from err
+
+
+def validate_not_empty(file: BinaryIO):
```
eventualbuddha, line 104: Should this function be responsible for seeking back to 0?

eventualbuddha (approved):

server/util/csv_parse.py
jonahkagan: oh definitely! i intended to do that and forgot. i'm actually kind of shocked this didn't break everything...

jonahkagan: Ah, I think I know why it didn't break - because `validate_comma_delimited` later seeks back to 0. That was just luck though, I will fix

server/api/cvrs.py
```diff
@@ -381,12 +394,11 @@ def process_cvr_file(
...
-        header_lines = 4 if jurisdiction.cvr_file_type == CvrFileType.DOMINION else 1
-        total_lines = len(jurisdiction.cvr_file.contents.splitlines()) - header_lines
-        emit_progress(0, total_lines)
-
-        if jurisdiction.cvr_file.contents == "":
-            raise UserError("CVR file cannot be empty.")
+        # Ideally, the CVR should have the same number of ballots as the
+        # manifest, so we can use that as an approximation of the file parsing
+        # progress since we're streaming the file and don't know the size up front.
+        total_records = jurisdiction.manifest_num_ballots
```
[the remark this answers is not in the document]
jonahkagan: It seems likely that we could get the file size in bytes, but I'm not sure how we would get it in lines, which is the current progress denominator. It might be possible to figure out a way to make the stream also report its progress in bytes, but since it's a few layers deep in generators, it seems tricky.

Given that all this functionality does is show a progress bar to the user, an approximate level of accuracy should be sufficient IMO

jonahkagan:
@eventualbuddha I caught one other mistake I made. See 18d8065a7246ea737a434d7caa493ffd5504f05a. Basically `file.readlines` is a terrible choice for encoding detection because (1) it's eager, returning a list instead of an iterator and (2) it uses `b'\n'` to detect newlines, which isn't guaranteed to work at all until we know the encoding

eventualbuddha:
> @eventualbuddha I caught one other mistake I made. See [18d8065](https://github.com/votingworks/arlo/commit/18d8065a7246ea737a434d7caa493ffd5504f05a). Basically `file.readlines` is a terrible choice for encoding detection because (1) it's eager, returning a list instead of an iterator and (2) it uses `b'\n'` to detect newlines, which isn't guaranteed to work at all until we know the encoding

I did wonder about that. I should have asked! Thanks for following up.

PR #1380, opened 2021-12-13
Store file uploads externally on s3 or filesystem

jonahkagan:
Task: #1305

Adds some utility functions to store files on s3 or on the local filesystem based on the env variable `ARLO_FILE_UPLOAD_STORAGE_PATH`. Adds a new db field `File.storage_path` to save the path where a file upload is stored. Uses this new functionality for all new file uploads, maintaining backwards compatibility for files that were previously stored in the database.

### DIFF
--- a/server/api/ballot_manifest.py
+++ b/server/api/ballot_manifest.py
@@ -7,7 +7,6 @@ from werkzeug.exceptions import BadRequest, NotFound
 from sqlalchemy import func
 from sqlalchemy.orm import Session
 
-
 from . import api
 from ..database import db_session, engine
 from ..models import *  # pylint: disable=wildcard-import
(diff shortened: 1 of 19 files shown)


server/api/ballot_manifest.py
```diff
@@ -116,9 +120,7 @@ def process() -> None:
             CSVColumnType(CVR, CSVValueType.YES_NO, required=use_cvr),
         ]
 
-        # Temporarily wrap file contents in a buffer so we can "stream" it until
-        # we have actual file streaming from storage
-        manifest_file = io.BytesIO(jurisdiction.manifest_file.contents.encode("utf-8"))
+        manifest_file = retrieve_file_contents(jurisdiction.manifest_file)
```
benadida, line 123: as a first-time reader of this code, my instinct is that `retrieve_file_contents` is returning the whole file contents, not a stream. Maybe naming could indicate this more clearly?

PR #1381, opened 2021-12-14
Script to migrate files from database to s3

jonahkagan:
Doing this separately from our regular database migrations since it relies on an external service, so it doesn't really make sense to run it automatically for anyone who is using the migrations.

Ran locally and completed in a few minutes with no errors. Notably, found 10 orphan files that were no longer attached to any audit.

### DIFF
--- a/scripts/migrate-files-to-s3.py
+++ b/scripts/migrate-files-to-s3.py
@@ -1,0 +1,92 @@
+# pylint: disable=invalid-name
+import io
+from server.models import *  # pylint: disable=wildcard-import
+from server.database import db_session
+from server.util.file import store_file, timestamp_filename
+
+
+def migrate_file(file, path):
+    if not file.storage_path:
+        storage_path = store_file(io.BytesIO(file.contents.encode("utf-8")), path)
+        file.storage_path = storage_path
+        return storage_path
+    else:
+        print("Already migrated")
+        return None
+
+
+if __name__ == "__main__":
+    to_migrate = File.query.filter(File.storage_path.is_(None)).count()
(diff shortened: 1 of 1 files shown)


scripts/migrate-files-to-s3.py
```diff
@@ -0,0 +1,92 @@
+# pylint: disable=invalid-name
+import io
+from server.models import *  # pylint: disable=wildcard-import
+from server.database import db_session
+from server.util.file import store_file, timestamp_filename
+
+
+def migrate_file(file, path):
+    if not file.storage_path:
+        storage_path = store_file(io.BytesIO(file.contents.encode("utf-8")), path)
```
benadida, line 10: if this fails, can you just re-run the script? I think so from reading it, just checking.

scripts/migrate-files-to-s3.py
```diff
@@ -0,0 +1,92 @@
...
+        file.storage_path = storage_path
+        return storage_path
+    else:
+        print("Already migrated")
```
benadida, line 14: suggestion: what if this pulled the file from S3 and compared content? that way the script could be used to "check" that everything worked well.

benadida (approved):

scripts/migrate-files-to-s3.py
jonahkagan: yep, it's built to be re-run on failures. one transaction per audit

scripts/migrate-files-to-s3.py
jonahkagan: sure! will add

PR #1382, opened 2021-12-14
Remove File.contents

jonahkagan:
Replace with File.storage_path


benadida (approved):
ahah, there's the renaming I was looking for earlier :)

eventualbuddha (approved):

Issue #1249, opened 2021-07-21
ES&S CVR parser


mcchilders:
Sample files: https://drive.google.com/drive/u/0/folders/1P01lfpvYjueG7GKj8fohiyrahoVrBbjc

ES&S exports the data we need in two different files, their actual CVR file and what they call the 'ballot manifest' (different from RLA ballot manifest). 
- The ballot manifest file only allows export of 20k rows at a time, so there may be multiple of these exports that correspond to a single CVR file

To get the data we need manually, I:
- Combine two ES&S exports into one, since half the data we need is in one, and half is in the other (CVR ID is the key that can be used to match ballot data across files)
- Deriving the sequential position of each ballot within a scanning batch and creating a new field for that data, since that is not in the files anywhere but you can infer it using the data that is present (Batch Name, combined with the order from the Tabulator CVR field, can give you the sequential position of the ballots within a batch e.g. 1,2,3,4,5 ...). 
- Pulling the Tabulator ID into its own field (first four digits of the Tabulator CVR field)
- Use TabulatorCVR as ImprintedID
- Overvotes need to be a "candidate" rather than just derived within Arlo

Note that CVR ID and TabulatorCVR are ordered, but not necessarily sequential (missing numbers in the order does happen)

PR #1391, opened 2021-12-21
ES&S CVR parsing

jonahkagan:
Task: #1249 

- Adds an ES&S CVR file type (to the database and frontend)
- Updates the frontend to allow multiple file selection when uploading ES&S CVRs
- Updates the backend to support multi-file uploads by zipping them
- Parses ES&S CVRs

### DIFF
--- a/server/api/cvrs.py
+++ b/server/api/cvrs.py
@@ -219,7 +219,9 @@ def column_value(
 ):
     index = header_indices.get(header)
     if index is None:
-        raise UserError(f"Missing required column {header}")
+        if required:
+            raise UserError(f"Missing required column {header}")
+        return None
     value = row[index] if index < len(row) else None
     if required and (value is None or value == ""):
         raise UserError(f"Missing required column {header} in row {row_number}.")
--- a/server/api/cvrs.py
+++ b/server/api/cvrs.py
@@ -588,6 +588,8 @@ def parse_ess_cvrs(
 
         return {
             contest_name: dict(
+                # Until we know how vote-for-n contests are serialized in the
+                # ES&S CVR, we assume vote-for-1
                 votes_allowed=1,
                 choices={
                     choice: dict(
--- a/server/api/cvrs.py
+++ b/server/api/cvrs.py
@@ -556,16 +556,6 @@ def parse_ess_cvrs(
                     " inconsistency in other rows in the CVR file."
                 )
 
-    def parse_all_ballots() -> Iterable[CvrBallot]:
-        for name, ballots_file in ballots_files.items():
-            try:
-                validate_comma_delimited(ballots_file)
-                ballots_csv = csv.reader(ballots_file, delimiter=",")
-                yield from parse_ballots_csv(ballots_csv)
-                ballots_file.close()
-            except UserError as error:
-                raise UserError(f"{name}: {error}") from error
-
     def parse_contest_metadata(cvr_csv: CSVIterator) -> CVR_CONTESTS_METADATA:
         headers = next(cvr_csv)
         # Based on files we've seen, the first 3 columns are Cast Vote Record,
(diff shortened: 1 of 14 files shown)


server/api/cvrs.py
```diff
@@ -408,6 +421,274 @@ def parse_cvr_rows() -> Iterable[CvrBallot]:
...
+                },
+                total_ballots_cast=0,  # Will be counted while parsing rows
+            )
+            for contest_name, choices in contest_choices.items()
+            if len(choices) > 0
+        }
+
+    def parse_interpretations(
+        cvr_csv: CSVIterator, contests_metadata: CVR_CONTESTS_METADATA
+    ) -> Iterable[Tuple[str, str]]:  # (CVR number, interpretations)
```
eventualbuddha, line 609: In TypeScript I sometimes find it useful to alias common types like `string` with a more specific name. Perhaps you could do that here and ditch the comment.

server/api/cvrs.py
```diff
@@ -408,6 +421,274 @@ def parse_cvr_rows() -> Iterable[CvrBallot]:
...
+
+    def parse_and_concat_ballots_files(
+        ballots_files: Dict[str, TextIO]
+    ) -> Iterable[CvrBallot]:
+        for name, ballots_file in ballots_files.items():
+            try:
+                validate_comma_delimited(ballots_file)
+                ballots_csv = csv.reader(ballots_file, delimiter=",")
+                yield from parse_ballots_csv(ballots_csv)
+                ballots_file.close()
```
eventualbuddha, line 654: It's a bit odd that `parse_and_concat_ballots_files` is responsible for closing the files. It looks like the opening happens in `unzip_files`? I'm unsure how to make this clearer, but I first noticed this because I wondered what would happen to the files given an exception. Maybe we don't care as we're not handling it and it'll just bubble up.

eventualbuddha (approved):
I haven't conducted a thorough review in the sense of checking out the code and trying it. I've tried to read the code fairly closely and understand the overall structure, and that much seems pretty reasonable.

server/api/cvrs.py
jonahkagan: Yeah I'm not sure what a clear pattern for this would be. The only Python idiom I know is to use context managers for opening/closing files:

```python
with open(path) as file:
  ... do stuff ...
```

I started out trying to use this pattern, but if you want to `return` the file inside the `with`, it doesn't work, because then the file gets closed when the function returns. 

It's possible to make `unzip_files` into a context manager so that it can be used in a `with` block, but then everything has to be indented, which I think is a major knock on readability.

I guess one pattern could be:

```python
def do_all_the_things(zip_file):
  ... do stuff ...

with unzip_files(path) as zip_file:
  do_all_the_things(zip_file)
```

Do you think that would be more readable? Again, adds a level of indentation to make the helper function.

As far as what happens if you don't close files, I haven't been able to find evidence that there are any real consequences. My cursory research seemed to indicate that they just get closed when they are garbage collected. But might be good to put the `close` calls into a `finally` block or something.

eventualbuddha: I don't feel strongly about it since this is all contained within the parent function.

server/api/cvrs.py
```diff
@@ -408,6 +421,274 @@ def parse_cvr_rows() -> Iterable[CvrBallot]:
...
+            return decode_csv(file)
+        except UserError as error:
+            raise UserError(f"{name}: {error}") from error
+
+    text_files = {name: decode_file(file, name) for name, file in files.items()}
+
+    def is_ballots_file(file: TextIO):
+        first_line = file.readline()
+        file.seek(0)
+        return first_line.startswith("Ballots")
```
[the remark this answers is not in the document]
jonahkagan: technically, yes, but doesn't seem like it will have any impact, so gonna just leave it

server/api/cvrs.py
jonahkagan: Good idea. I think in this case since it's just one spot, it's not worth aliasing, but I like the idea in general and can see it being useful.

server/api/cvrs.py
jonahkagan: I decided to at least move the `close` call to a `finally` block. And also realized I never closed `cvr_file`, so added that. 4f77bf93

PR #1392, opened 2021-12-21
Change required column error message

jonahkagan:
Make it less repetitive and easier to read

### DIFF
--- a/server/tests/ballot_comparison/test_cvrs.py
+++ b/server/tests/ballot_comparison/test_cvrs.py
@@ -646,7 +646,7 @@ CvrNumber,TabulatorNum,BatchId,RecordId,CountingGroup,PrecinctPortion,BallotType
 CvrNumber,TabulatorNum,BatchId,RecordId,ImprintedId,CountingGroup,PrecinctPortion,BallotType,REP,DEM
 ,TABULATOR1,BATCH1,1,1-1-1,Election Day,12345,COUNTY,0,1
 """,
-            "CvrNumber is required. Missing CvrNumber in row 1.",
+            "Missing required column CvrNumber in row 1.",
             "DOMINION",
         ),
         (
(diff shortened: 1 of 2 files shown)


server/tests/ballot_comparison/test_cvrs.py
```diff
@@ -646,7 +646,7 @@ def test_invalid_cvrs(
 CvrNumber,TabulatorNum,BatchId,RecordId,ImprintedId,CountingGroup,PrecinctPortion,BallotType,REP,DEM
 ,TABULATOR1,BATCH1,1,1-1-1,Election Day,12345,COUNTY,0,1
 """,
-            "CvrNumber is required. Missing CvrNumber in row 1.",
+            "Missing required column CvrNumber in row 1.",
```
umbernhard, line 649: This is maybe an unnecessary nitpick, but it'd be great if we could demarcate that the column name must match that string a little better. Like, put it in quotes or something: `"Missing required column \"CvrNumber\" in row 1."`

umbernhard (approved):
Aside from nitpick, this looks good.

server/tests/ballot_comparison/test_cvrs.py
jonahkagan: Not an unreasonable idea. Currently, this styling is consistent with the error messages from our other csv_parse module, so we'd probably want to change both of them if we want to make that change. I don't feel passionately about changing it, but if you do feel free! (Maybe in a separate PR)

umbernhard: Heh, that sounds like an after-Christmas change to me. I should probably get @mcchilders and @ginvdr's suggestions.

ginvdr: I literally finished the Troubleshooting guide and updating associated Front message templates this morning...  so I'm inclined to vote no for completely selfish reasons.  No strong feelings otherwise!

PR #1395, opened 2021-12-22
Case insensitive JA login

jonahkagan:


server/auth/routes.py
```diff
@@ -186,7 +186,7 @@ def is_code_expired(timestamp: datetime):
 def jurisdiction_admin_generate_code():
     body = request.get_json()
     user = (
-        User.query.filter_by(email=body.get("email"))
+        User.query.filter_by(email=body.get("email").lower())
```
eventualbuddha, line 189: Is this column case-sensitive? Are we ensuring they're all lowercase somewhere else? Should the column be case-insensitive?

jonahkagan: The column is case-sensitive. There's code in the model to make sure values are always converted to lowercase before being written to the db. We've used this approach to ensure case-insensitive lookups elsewhere.

As to the last question: I didn't even know there was a case-insensitive datatype available until you asked! Found this: https://www.postgresql.org/docs/12/citext.html. Might be something to consider

umbernhard (approved):

Issue #1389, opened 2021-12-16
Use UniqueVotingIdentifier field in Dominion CVRs


jonahkagan:
If ImprintedId isn't present for some ballots, fall back to UniqueVotingIdentifier.

PR #1396, opened 2021-12-22
Dominion CVRs: fall back to UniqueVotingIdentifier if no ImprintedId

jonahkagan:
Task: #1389

### DIFF
--- a/server/tests/ballot_comparison/snapshots/snap_test_cvrs.py
+++ b/server/tests/ballot_comparison/snapshots/snap_test_cvrs.py
@@ -128,7 +128,7 @@ snapshots["test_clearballot_cvr_upload 2"] = {
     },
 }
 
-snapshots["test_cvrs_counting_group 1"] = [
+snapshots["test_dominion_cvr_unique_voting_identifier 1"] = [
     {
         "ballot_position": 1,
         "batch_name": "BATCH1",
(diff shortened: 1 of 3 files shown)


server/tests/ballot_comparison/snapshots/snap_test_cvrs.py
```diff
@@ -181,62 +181,62 @@
     {
         "ballot_position": 2,
         "batch_name": "BATCH1",
-        "imprinted_id": "2-1-2",
+        "imprinted_id": "56_083-212",
```
umbernhard, line 184: I assume this change is just to match the actual Dominion style?

server/tests/ballot_comparison/snapshots/snap_test_cvrs.py
```diff
@@ -256,7 +256,128 @@
     },
 }
 
-snapshots["test_cvrs_newlines 1"] = [
+snapshots["test_dominion_cvr_upload 1"] = [
+    {
+        "ballot_position": 1,
+        "batch_name": "BATCH1",
+        "imprinted_id": "1-1-1",
```
umbernhard, line 263: If the answer to my above question is yes, why not use the same format here?

server/tests/ballot_comparison/snapshots/snap_test_cvrs.py
```diff
@@ -459,48 +580,55 @@
...
         "choices": {
-            "Choice 2-1": {"column": 2, "num_votes": 13},
-            "Choice 2-2": {"column": 3, "num_votes": 6},
-            "Choice 2-3": {"column": 4, "num_votes": 7},
+            "Choice 2-1": {"column": 2, "num_votes": 15},
+            "Choice 2-2": {"column": 3, "num_votes": 7},
+            "Choice 2-3": {"column": 4, "num_votes": 8},
         },
-        "total_ballots_cast": 14,
+        "total_ballots_cast": 15,
```
umbernhard, line 631: Why would this change?

umbernhard (approved):
Just a few outstanding questions, but I think this is good.

server/tests/ballot_comparison/snapshots/snap_test_cvrs.py
jonahkagan: this diff is misleading. the snapshots got reordered, i think because i renamed some tests. let me see if i can fix

umbernhard: oh I gotcha. It's not a big deal.

jonahkagan: ok check now, should be clearer

umbernhard: Yep!

PR #1401, opened 2022-01-04
Fix Husky

jonahkagan:
When we upgraded the version, we didn't make the necessary changes (moving from package.json to separate scripts).


jonahkagan:
Yeah the hooks weren't running anymore. I found migration instructions here: https://typicode.github.io/husky/#/?id=migrate-from-v4-to-v7

umbernhard (approved):

PR #1402, opened 2022-01-05
ES&S CVRs: ensure ballots files are parsed in order of CVR number

jonahkagan:
Previously, we just concatenated the ballots files in the order they were uploaded. Instead, we should order the files by the first CVR number in each file (since the ballots are ordered by CVR number within the file).


benadida (approved):
I like this implementation :)

PR #1403, opened 2022-01-06
Implements MACRO with RAIRE for ballot comparison audits of IRV elections. 

umbernhard:
Adds support for IRV elections in ballot comparison audits by utilizing the RAIRE algorithm for generating auditable assertions. Also implements a CVR validation scheme based on San Francisco's IRV ruleset.


server/tests/audit_math/test_supersimple_raire.py
```diff
@@ -0,0 +1,1146 @@
...
+    }
+
+    discrepancies = supersimple_raire.compute_discrepancies(
+        cvrs, sample_cvr, assertions["Contest F"][0]
+    )
+
+    assert discrepancies
+    assert discrepancies[0]["counted_as"] == 1
+    # Again, this should be impossible.
+    assert discrepancies[0]["weighted_error"] == Decimal("inf")
```
jonahkagan, line 193: why is this infinite?

server/tests/audit_math/test_supersimple_raire.py
```diff
@@ -0,0 +1,1146 @@
...
+        0: {"times_sampled": 1, "cvr": {"Contest D": {"winner": 1, "loser": 2}}}
+    }
+
+    discrepancies = supersimple_raire.compute_discrepancies(
+        cvrs, sample_cvr, assertions["Contest D"][0]
+    )
+
+    assert discrepancies
+    assert discrepancies[0]["counted_as"] == 2
+    # This should actually be impossible, unless we're already doing a recount.
```
jonahkagan, line 243: I would think this is possible if a ballot is in the manifest but not the CVR for some reason

server/tests/audit_math/test_supersimple_raire.py
```diff
@@ -0,0 +1,1146 @@
...
+
+    # Idempotent for correct cvrs
+    assert expected == supersimple_raire.validate_cvr(
+        {"contest": {"Alice": 1, "Bob": 2, "Charlie": 0}}
+    )
+
+    # Correctly fixes gaps
+    assert expected == supersimple_raire.validate_cvr(
+        {"contest": {"Alice": 2, "Bob": 4, "Charlie": 0}}
+    )
```
jonahkagan, line 460: I think this function could benefit from some more tests, since it's fairly complex logic. Currently there's no testing for the overvote adjustment logic, and only one case covered for the gap-fixing logic.

jonahkagan: Though I did just notice the next test function below which is awesome! Love that you tested from the discrepancy perspective

jonahkagan (commented):
This looks great! Love the thorough test coverage. I left a few questions and suggestions.

Also, is there any way we can speed up those tests at all? They take 30s on my machine, which is a good chunk of the time of the entire server test suite. For instance, could we pre-compute the value of `compute_raire_assertions` and then make that a test fixture so we don't have to recompute it every time we run the tests (we'd need some way to force ourselves to recompute it whenever the original test data changes or the code changes...). Or maybe they only run in CI or something?

server/audit_math/raire.py
```diff
@@ -70,10 +66,11 @@ def make_frontier(
     # Our frontier initially has a node for each alternate election outcome
     # tail of size two. The last candidate in the tail is the ultimate winner.
     for cand in contest.candidates:
-        if cand == winner:
+        if cand in contest.winners:
             continue
 
         for other in contest.candidates:
+            # We don't care about other winners
```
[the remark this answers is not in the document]
umbernhard: Yep, don't know how it got there.

server/audit_math/supersimple_raire.py
```diff
@@ -0,0 +1,378 @@
...
+    for contest in output:
+        next_highest = 2
+        for cand, rank in sorted(output[contest].items(), key=lambda item: item[1]):
+            if rank <= 1:
+                continue
+            if rank > next_highest:
+                output[contest][cand] = next_highest
+                next_highest += 1
+
+    return output
```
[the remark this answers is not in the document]
umbernhard: Yeah, I figured there was a cleaner way to do a lot of it. I'll make this change.

server/audit_math/supersimple_raire.py
```diff
@@ -0,0 +1,378 @@
...
+        a_w, a_l = (
+            assertion.is_vote_for_winner(audited_v),
+            assertion.is_vote_for_loser(audited_v),
+        )
+        error = (v_w - a_w) - (v_l - a_l)
+
+    if error == 0:
+        return None
+
+    # TODO: this seems wrong - V_wl here is defined as the first-round tallies for winner and loser
```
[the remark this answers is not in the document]
umbernhard: ope, this was already TODONE. Will remove.

server/audit_math/supersimple_raire.py
```diff
@@ -0,0 +1,378 @@
...
+        total_ballots  - the total number of ballots cast in the election
+        sample_results - if a sample has already been drawn, this will
+                         contain its results, of the form:
+                         {
+                            'sample_size': n,
+                            '1-under':     u1,
+                            '1-over':      o1,
+                            '2-under':     u2,
+                            '2-over':      o2,
+                         }
```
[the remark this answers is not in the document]
umbernhard: Yes, good catch.

server/audit_math/supersimple_raire.py
```diff
@@ -0,0 +1,378 @@
...
+
+    return stopping_size
+
+
+def compute_margin_for_assertion(cvrs: CVRS, assertion: RaireAssertion) -> Decimal:
+    v_w, v_l = 0, 0
+    for cvr in cvrs:
+        # Typechecker shenanigans
+        val = cvrs[cvr]
+        assert val is not None
```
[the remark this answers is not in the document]
umbernhard: I believe this function will only ever be computed over the CVR, not over the audited results, so yes, it should never be none.

server/tests/audit_math/test_supersimple_raire.py
umbernhard: Hmm, it shouldn't be. Another good catch, I think I thought this contest was a tie but it's not.

server/tests/audit_math/test_supersimple_raire.py
umbernhard: Yeah, you're right. I will fix.

server/tests/audit_math/test_supersimple_raire.py
umbernhard: I realized it was easier to test the discrepancies than doing it here, so that's what I went with. I can beef this one up too.

server/audit_math/supersimple_raire.py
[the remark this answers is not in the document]
jonahkagan: Great, maybe update the comment to say that?

server/audit_math/raire.py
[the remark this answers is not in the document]
umbernhard: oh, no actually this was in the right place. We don't care if `cand` is a winner, and we also don't care if `other` is a winner.

server/tests/audit_math/test_supersimple_raire.py
umbernhard: oh, I see. If the audit board finds a race on the ballot that wasn't reported in the CVR I was thinking that should trigger a recount, but maybe not? I will ping @mcchilders.

server/audit_math/supersimple_raire.py
```diff
@@ -0,0 +1,365 @@
...
+                          only.
+                        'ballot_id': {
+                            'counted_as': -1, # The maximum value for a discrepancy
+                            'weighted_error': -0.0033 # Weighted error used for p-value calculation
+                        }
+                    }
+    """
+    V = compute_margin_for_assertion(cvrs, assertion)
+    print(V)
+    print(assertion)
```
jonahkagan, line 91: remove prints

server/tests/audit_math/test_supersimple_raire.py
```diff
@@ -0,0 +1,1153 @@
...
+
+    assert discrepancies
+    assert discrepancies[0]["counted_as"] == 1
+    assert discrepancies[0]["weighted_error"] == Decimal(1) / Decimal(2000)
+
+
+def test_ballot_not_found_discrepancy(cvrs, assertions):
+    sample_cvr = {0: {"times_sampled": 1, "cvr": None}}
+
+    print(assertions["Contest D"])
```
jonahkagan, line 221: remove

jonahkagan (approved):

jonahkagan:
@umbernhard is this ready to merge?

umbernhard:
It is! Good call.

PR #1413, opened 2022-02-16
Disallow generating audit report before audit starts

jonahkagan:


server/api/reports.py
```diff
@@ -742,6 +743,9 @@ def sampled_batch_rows(election: Election, jurisdiction: Jurisdiction = None):
 @api.route("/election/<election_id>/report", methods=["GET"])
 @restrict_access([UserType.AUDIT_ADMIN])
 def audit_admin_audit_report(election: Election):
+    if len(list(election.rounds)) == 0:
```
carolinemodic, line 746: is election.rounds the current round of an audit we are on or something?

server/tests/api/test_reports.py
```diff
@@ -77,3 +77,35 @@ def test_jurisdiction_admin_report(
         == 'attachment; filename="audit-report-J1-Test-Audit-test-jurisdiction-admin-report-DATETIME.csv"'
     )
     assert_match_report(rv.data, snapshot)
+
+
+def test_report_before_audit_starts(
```
carolinemodic, line 82: just curious do we have something like jest coverage that tracks the test coverage in ARLO?

carolinemodic (approved):

server/api/reports.py
jonahkagan: `election.rounds` is all of the rounds of the audit. so if it's empty, that means the audit hasn't started yet

server/tests/api/test_reports.py
jonahkagan: yeah, it's configured in .coveragerc. currently at 99.6%. i'd like to get to 100% - would probably only take a few hours

Issue #1371, opened 2021-12-07
Batch comparison: order sample results by ticket number


jonahkagan:
When computing the p-value, we should process the sample results in the order the ballots were sampled.

PR #1418, opened 2022-03-01
Macro: Compute p-value using sample order

jonahkagan:
Task: #1371 

Change `macro.compute_risk` to take an ordering of the sampled batches by ticket number instead of just the number of times each batch got sampled. Compute the p-value by evaluating batch taints in sample order, bailing once we get below the risk limit.

Also added type checking to test_macro.py by adding a `None` return type to the test functions.

### DIFF
--- a/server/api/rounds.py
+++ b/server/api/rounds.py
@@ -241,17 +241,17 @@ def sampled_batch_results(election: Election,) -> BatchTallies:
     }
 
 
-def batches_times_sampled(election: Election) -> Dict[Tuple[str, str], int]:
-    sampled_batch_draw_counts = (
+def sampled_batches_by_ticket_number(election: Election) -> Dict[str, Tuple[str, str]]:
+    batches_by_ticket_number = (
         SampledBatchDraw.query.join(Batch)
         .join(Jurisdiction)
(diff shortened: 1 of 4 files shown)


server/api/rounds.py
```diff
@@ -241,17 +241,17 @@ def sampled_batch_results(election: Election,) -> BatchTallies:
...
-    sampled_batch_draw_counts = (
+def sampled_batches_by_ticket_number(election: Election) -> Dict[str, Tuple[str, str]]:
+    batches_by_ticket_number = (
         SampledBatchDraw.query.join(Batch)
         .join(Jurisdiction)
         .filter_by(election_id=election.id)
-        .group_by(Jurisdiction.name, Batch.name)
-        .values(Jurisdiction.name, Batch.name, func.count())
+        .order_by(SampledBatchDraw.ticket_number)
+        .values(SampledBatchDraw.ticket_number, Jurisdiction.name, Batch.name)
```
umbernhard, line 250: This is a really good catch, I hadn't considered that we weren't using sampling order.

umbernhard (approved):
This looks good to me.

PR #1425, opened 2022-03-08
Support Tools: button to permanently delete audit

jonahkagan:
- Add a button for deleted ("archived") audits to be permanently deleted by Support Users
- Also reorganize audits into two sections: active and deleted

### DIFF
--- a/client/src/components/SupportTools/SupportTools.tsx
+++ b/client/src/components/SupportTools/SupportTools.tsx
@@ -35,6 +35,8 @@ import {
   useDeleteOrganization,
   useRenameOrganization,
   useRemoveAuditAdmin,
+  useDeleteElection,
+  IElectionBase,
 } from './support-api'
 import { useConfirm, Confirm } from '../Atoms/Confirm'
 
@@ -90,6 +92,7 @@ const ButtonList = styled(ButtonGroup).attrs({
       border-top: 1px solid ${Colors.LIGHT_GRAY3};
     }
   }
+  margin-bottom: 30px;
 `
 
 const Organizations = () => {
@@ -168,6 +171,7 @@ const Organization = ({ organizationId }: { organizationId: string }) => {
   const removeAuditAdmin = useRemoveAuditAdmin(organizationId)
   const deleteOrganization = useDeleteOrganization(organizationId)
   const renameOrganization = useRenameOrganization(organizationId)
+  const deleteElection = useDeleteElection()
   const { confirm, confirmProps } = useConfirm()
 
   const {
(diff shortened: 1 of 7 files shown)


client/src/components/SupportTools/SupportTools.tsx
```diff
@@ -235,6 +239,18 @@ const Organization = ({ organizationId }: { organizationId: string }) => {
...
+      title: 'Confirm',
+      description: `Are you sure you want to permanently delete ${auditName}?`,
+      yesButtonLabel: 'Delete',
+      onYesClick: async () => {
+        await deleteElection.mutateAsync({ electionId: id, organizationId })
+        toast.success(`Deleted ${auditName}`)
+      },
+    })
+  }
+
```
benadida, line 253: does this provide some visual feedback while it's deleting, or does the screen freeze until success?

client/src/components/SupportTools/SupportTools.tsx
```diff
@@ -261,16 +277,40 @@ const Organization = ({ organizationId }: { organizationId: string }) => {
...
+                >
+                  {election.auditName}
+                </LinkButton>
+              ))}
           </ButtonList>
+          <H3>Deleted Audits</H3>
+          <Table striped>
+            <tbody>
+              {elections
+                .filter(election => !election.deletedAt)
```
benadida, line 296:
```suggestion
                .filter(election => election.deletedAt)
```

shouldn't it be this ^^ ?

server/util/file.py
```diff
@@ -71,6 +71,17 @@ def retrieve_file(storage_path: str) -> BinaryIO:
         return open(storage_path, "rb")
 
 
+def delete_file(storage_path: str):
+    if config.FILE_UPLOAD_STORAGE_PATH.startswith("s3://"):
+        assert storage_path.startswith("s3://")
+        parsed_path = urlparse(storage_path)
+        bucket_name = parsed_path.netloc
+        key = parsed_path.path[1:]
+        s3().delete_object(bucket_name, key)
```
benadida, line 80: if this fails, then we'll be left with an election where some files have been deleted, right?

Probably okay, though may want to improve this at some point.

benadida (approved):
I'm a little bit wary of the cascading deletes... I know they make the code cleaner than having to do all the deletes ourselves but, it does mean that a simple mistake at the `psql` prompt could be quite painful.

I'm okay moving forward here, though if my concern resonates, maybe let's talk about it?

client/src/components/SupportTools/SupportTools.tsx
jonahkagan: whoops, fixed and covered with a test

client/src/components/SupportTools/SupportTools.tsx
jonahkagan: there's a spinner on the button and it disables

server/util/file.py
jonahkagan: Yes, that's true. Given that the goal is deleting everything, a failure that only partially completes feels like a not-so-bad outcome, so I'm also ok with leaving it for now

PR #1447, opened 2022-04-18
Fix ballot comparison round 2 sample size

jonahkagan:
`supersimple.nMin` expects the number of expected discrepancies, not the rate. We were passing the number of expected one-vote discrepancies but the rate of expected two-vote discrepancies.

### DIFF
--- a/server/tests/audit_math/test_supersimple.py
+++ b/server/tests/audit_math/test_supersimple.py
@@ -643,7 +643,7 @@ o2_stopping_size = {
     "Contest B": 60000,
     "Contest C": 36000,
     "Contest D": 15000,
-    "Contest E": 6,
+    "Contest E": 33,
     "Contest F": 15,
     "Two-winner Contest": 1000,
 }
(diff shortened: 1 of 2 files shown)


umbernhard (changes requested):
The math looks good, would you like me to update the tests or can you do it?

jonahkagan:
I updated one existing test that covered this. If you have a good idea for adding more tests, feel free, otherwise I think it's good to go.

server/tests/audit_math/test_supersimple.py
```diff
@@ -643,7 +643,7 @@ def test_multiplicity():
     "Contest B": 60000,
     "Contest C": 36000,
     "Contest D": 15000,
-    "Contest E": 6,
+    "Contest E": 33,
```
umbernhard, line 646: Wow, I kinda can't believe it only affected this one contest.

umbernhard (approved):

server/tests/audit_math/test_supersimple.py
umbernhard: I guess that's more a testament to my testing strategy than anything, huh :trollface:

Issue #1248, opened 2021-07-21
Hart CVR parser in Arlo (we have a script)


mcchilders:
Conditions of Acceptance:
- Allow JA to specify that a CVR is a Hart CVR at time of upload
-  Accept a (likely zipped file folder) collection of individual CVR files as the file upload and parse into our data format (e.g. having Arlo run the script that we currently run outside Arlo for Hart files). 
- Store both the original and the parsed versions for posterity, in case we need to prove they're the same (or troubleshoot parsing in future)

Sample files:
https://drive.google.com/drive/u/0/folders/1HEeeXSVJhjgVKAQIAE-E3GwTCypHWwOq

mcchilders:
At this point we have not found any combination of Hart exports (individual CVR files plus add'l reports) that allows us to know which tabulator a ballot was scanned by, and since batch names are duplicated across tabulators, we don't have enough data to use these files absent a human adding data.

rsmith1302:
I'm not familiar with this project, so I might be misinterpreting the issue, but I've written a Hart CVR parser. The expoted PDF (but not the XML) contains a DeviceSerial field that will identify the scanner. I use pdfgrep to extract the metadata from the PDF.

jonahkagan:
Thanks for the tip @rsmith1302!

PR #1448, opened 2022-04-19
Hart CVR parsing

jonahkagan:
Task: #1248 

Adds Hart CVR parsing to the worker and enables Hart as an option on the JA file upload screen. Hart CVRs are uploaded as a single zip archive containing an XML file for each ballot.

Also tweaks the instructions language on that screen to be more ambiguous about the exact format of the file.


client/src/components/JurisdictionAdmin/JurisdictionAdminView.test.tsx
```diff
@@ -350,6 +350,44 @@ describe('JA setup', () => {
     })
   })
 
+  it('allows zip CVR files to be uploaded for ES&S', async () => {
```
carolinemodic, line 353: I assume you mean hart here not es&s?

server/api/cvrs.py
```diff
@@ -742,6 +744,161 @@ def join_ballots_to_interpretations(
...
+    # 3. Next, parse out the interpretations.
+
+    zip_file = retrieve_file(jurisdiction.cvr_file.storage_path)
+    files = unzip_files(zip_file)
+    zip_file.close()
+
+    # Remove excess files (e.g. WriteIn directory)
+    files = {name: file for name, file in files.items() if name.endswith(".xml")}
+
+    namespace = "http://tempuri.org/CVRDesign.xsd"
```
carolinemodic, line 765: what is this lol

carolinemodic (approved):

server/api/cvrs.py
jonahkagan: I have no idea but you have to do it and @benadida said it's good security practice or something.

PR #1452, opened 2022-04-25
Change JA file upload form copy to "file" instead of "csv"

jonahkagan:
Previously, all files uploaded were CSVs, now for Hart CVRs, we also
accept a zip, so we just change the copy to be generic.


arsalansufi (approved):
LGTM! 📁

PR #1455, opened 2022-04-25
Use approximate comparison in Minerva tests

jonahkagan:
I set up a new computer and some of the test assertions started failing due to tiny differences in floats. I saw that some tests already used `approx` to deal with this, so I added that where necessary to pass.


umbernhard (approved):
Well that's nifty. I assume `approx` lets us configure the number of digits of certainty?

jonahkagan:
Yeah looks like it!

Issue #1438, opened 2022-03-29
When presenting option to start the next round, show sample size


benadida:
We don't want it to be a surprise when the next round is started, especially in pilot situations where how large the next round is may determine whether they even go through with a second round.

Note that, in case of batch audits, it should show the number of batches to be audited.

PR #1456, opened 2022-04-25
Move sample size options to their own table

jonahkagan:
Task: #1438 

In order to showing sample size options for rounds other than the first round to the user, we need some place to store them. So instead of just storing first round sample size options with each Election, we create a separate table for sample size options, keyed on election_id+round_number.

This PR implements this new table without adding any new functionality yet.


server/migrations/versions/cec7ecc73bd8_samplesizeoptions_table.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+
+    op.execute(
+        """
+        INSERT INTO sample_size_options
+            (created_at, updated_at, election_id, round_num, task_id, sample_size_options)
+        SELECT created_at, updated_at, id, 1, sample_size_options_task_id, sample_size_options
+        FROM election
+        """
+    )
```
arsalansufi, line 53: Nice, so this is migrating existing sample size data in the election table to the new dedicated sample size table. Is this necessary because we persist data from past audits in our databases? (Indefinitely?) Or do we need to do this because we have some in-flight audits right now?

server/api/sample_sizes.py
```diff
@@ -220,31 +223,40 @@ def serialize_sample_size_options(sample_size_options):
 @api.route("/election/<election_id>/sample-sizes", methods=["GET"])
 @restrict_access([UserType.AUDIT_ADMIN])
 def get_sample_sizes(election: Election):
+    sample_sizes = SampleSizeOptions.query.filter_by(
+        election_id=election.id, round_num=1,
+    ).one_or_none()
+    # If we've never queried sample sizes before for this audit, create a row in
+    # the database to store them.
```
arsalansufi, line 230: Alternatively, could we create the sample sizes record when we create the election (i.e. audit) record? (The current approach seems very reasonable, just curious to hear the reasoning that led to it)

arsalansufi (approved):
LGTM! Left some questions for my own understanding

server/migrations/versions/cec7ecc73bd8_samplesizeoptions_table.py
jonahkagan: Currently we persist data until audit admins delete it. Setting an expiration date is an idea in our product backlog.

Migrations auto-run during the deploy process: https://devcenter.heroku.com/articles/release-phase#when-does-the-release-command-run. I generally try not to release during an active audit (especially for migrations), but even when I do it's usually pretty seamless.

server/api/sample_sizes.py
jonahkagan: Yeah, I considered that. Two reasons I went this direction:
- Next up, we'll be computing sample sizes for later rounds in this same endpoint, so if this endpoint wasn't creating the sample size record for later rounds, it would have to be created elsewhere (e.g. where we finish the previous round). Then we'd have two places that create these records
- It seemed easier to read the code by having all this logic in one place

server/migrations/versions/cec7ecc73bd8_samplesizeoptions_table.py
arsalansufi: Nice, thanks for that context! And awesome to hear that even releasing during an active audit is pretty seamless

PR #1457, opened 2022-04-25
Move Minerva doctests to test file to use approx

jonahkagan:
Some Minerva doctests need pytest.approx to account for precision issues. Doctests don't support this, so we move them to the test_minerva.py test file.

### DIFF
--- a/server/audit_math/minerva.py
+++ b/server/audit_math/minerva.py
@@ -177,15 +177,6 @@ def collect_risks(
 
     Outputs:
         risks - Mapping of (winner, loser) pairs to their risk levels
-
-    >>> c3 = make_arlo_contest({"a": 600, "b": 400, "c": 100, "_undervote_": 100})
-    >>> collect_risks(0.1, c3, [120], make_sample_results(c3, [[56, 40, 3]]))
-    {('winner', 'loser'): 0.0933945799801079}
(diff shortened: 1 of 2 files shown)


server/audit_math/minerva.py
```diff
@@ -177,15 +177,6 @@ def collect_risks(
...
         risks - Mapping of (winner, loser) pairs to their risk levels
-
-    >>> c3 = make_arlo_contest({"a": 600, "b": 400, "c": 100, "_undervote_": 100})
-    >>> collect_risks(0.1, c3, [120], make_sample_results(c3, [[56, 40, 3]]))
-    {('winner', 'loser'): 0.0933945799801079}
-    >>> collect_risks(0.1, c3, [83], make_sample_results(c3, [[40, 40, 3]]))
-    {('winner', 'loser'): 0.5596434615209632}
-    >>> collect_risks(0.1, c3, [82], make_sample_results(c3, [[40, 40, 3]]))
-    Traceback (most recent call last):
-    ValueError: Incorrect number of valid ballots entered
```
arsalansufi, line 188: TIL about Python doctests!

jonahkagan: Yeah, they are cool. We don't use them generally though - these were added by an external contributor

PR #1460, opened 2022-04-26
Show sample sizes to audit admin in later rounds

jonahkagan:
Task: #1438 

Best reviewed commit-by-commit.

- First, we add the ability for the /sample-sizes endpoint to compute later round sample sizes
- Second, we use that new ability to show the sample sizes in the AA status box before starting another round
- Last, we remove the code in the backend that computed and chose the sample size for later rounds while drawing the sample, instead sending it from the frontend (this matches the round 1 sample size data flow and also sets us up to allow the user to modify later round sample sizes if we ever want that)

Due to some constraints with what the StatusBox component currently supports, I didn't do much fancy design for now. Ultimately, I think it would make sense to rethink this UI entirely so that viewing/editing sample sizes looks the same in the first round and later rounds, so I didn't think it was worth putting too much effort in here.
Screen Shot 2022-04-26 at 3 57 21 PM

### DIFF
--- a/client/src/components/Atoms/StatusBox.tsx
+++ b/client/src/components/Atoms/StatusBox.tsx
@@ -12,10 +12,12 @@ import {
   IRound,
   drawSampleError,
   isAuditStarted,
+  ISampleSizes,
 } from '../AuditAdmin/useRoundsAuditAdmin'
 import { IAuditBoard } from '../useAuditBoards'
 import { IContest } from '../../types'
 import useSampleSizes from '../AuditAdmin/Setup/Review/useSampleSizes'
+import { mapValues } from '../../utils/objects'
 
 const SpacedH3 = styled(H3)`
   &.bp3-heading {
@@ -138,7 +140,7 @@ export const isSetupComplete = (
 
 interface IAuditAdminProps {
   rounds: IRound[]
-  startNextRound: () => Promise<boolean>
+  startNextRound: (sampleSizes: ISampleSizes) => Promise<boolean>
   undoRoundStart: () => Promise<boolean>
   jurisdictions: IJurisdiction[]
   contests: IContest[]
@@ -293,7 +295,7 @@ interface IAuditAdminAnotherRoundStatusBoxProps {
   auditSettings: IAuditSettings
   contests: IContest[]
   roundNum: number
-  startNextRound: () => Promise<boolean>
+  startNextRound: (sampleSizes: ISampleSizes) => Promise<boolean>
   children?: ReactElement
 }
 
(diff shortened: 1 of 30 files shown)


server/api/rounds.py
```diff
@@ -710,16 +722,6 @@ def draw_sample(round_id: str, election_id: str):
...
-    # For rounds after the first round, automatically select a sample size
-    if round.round_num > 1:
-
-        sample_size_options = sample_sizes_module.sample_size_options(election)
-        for round_contest in round.round_contests:
-            if round_contest.contest_id in sample_size_options:
-                round_contest.sample_size = sample_sizes_module.autoselect_sample_size(
-                    sample_size_options[round_contest.contest_id], election.audit_type
-                )
-
```
benadida, line 734: Given this removal, might there be a reason to check that the sample sizes have indeed been computed before trying to draw the sample?

benadida (approved):

server/api/sample_sizes.py
```diff
@@ -214,6 +213,20 @@ def next_round_sample_size_options(election_id: str):
     sample_sizes.sample_size_options = sample_size_options(election)
 
 
+# In rounds other than the first round, we want to automatically select a sample
+# size from the generated options instead of letting the user pick.
```
arsalansufi, line 209: Just curious why we let users pick from a set of options for the first round but not for later rounds. Is it because, if we can't complete the audit in one round, we have to start getting more prescriptive in subsequent rounds to ensure that the audit completes?

arsalansufi (approved):
LGTM! Left some questions for my understanding and a minor comment

server/api/sample_sizes.py
```diff
@@ -244,13 +252,13 @@ def get_sample_sizes(election: Election):
             datetime.now(timezone.utc) - sample_sizes.task.completed_at
             > timedelta(seconds=5)
         )
-        # Don't need to recompute after the audit launches
-        and len(list(election.rounds)) == 0
     )
-    if not sample_sizes.task or existing_options_expired:
+    if round_num == next_round_num and (
+        not sample_sizes.task or existing_options_expired
+    ):
```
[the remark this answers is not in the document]
jonahkagan: There's an explanatory comment a few lines above. Can you read that and tell me if it explains? I know this code is confusing and want to make it as clear as possible

server/api/sample_sizes.py
jonahkagan: I'm not sure why to be honest

client/src/components/Atoms/StatusBox.tsx
```diff
@@ -306,6 +308,12 @@ const AuditAdminAnotherRoundStatusBox = ({
   children,
 }: IAuditAdminAnotherRoundStatusBoxProps) => {
   const sampleSizesResponse = useSampleSizes(electionId, roundNum + 1, true)
+  // The server should autoselect one option per contest, so we pick the first
+  // item in the options array
```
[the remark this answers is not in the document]
jonahkagan: Thanks, I love getting feedback on what's confusing :)

server/api/rounds.py
jonahkagan: I don't think we need to, since the sample size is now a required parameter in the request to create the round (which in turn runs the task to draw the sample).

Issue #1430, opened 2022-03-23
Bug: clear choice selections when 'ballot not found' marked


mcchilders:
If candidates are selected in the audit board data entry screen and "ballot not found" is subsequently selected, the candidate selections are not cleared as they should be.

PR #1462, opened 2022-04-28
Clear audit board choices in UI when "Ballot Not Found" is selected

arsalansufi:
# Overview

Issue link: https://github.com/votingworks/arlo/issues/1430

This PR clears audit board choices in the UI when "Ballot Not Found" is selected. We already clear the choices from the relevant API request, just not in the UI. The current UI behavior may cause audit board users concern.

### Before

https://user-images.githubusercontent.com/12616928/165802096-382f8260-75a9-4561-ad41-8bdadfadc94b.mov

### After

https://user-images.githubusercontent.com/12616928/165802111-9e936fac-a5c6-4401-a89e-44f97462c978.mov

# Testing

- [x] Tested manually (see the above screenshots)
- [x] Added automated tests


client/src/components/AuditBoard/BallotAudit.tsx
```diff
@@ -57,7 +59,14 @@ const BallotAudit: React.FC<IProps> = ({
...
         <Formik
-          initialValues={{ interpretations }}
+          initialValues={{
+            interpretations,
+            // Formik only reinitializes if `enableReinitialize` is true and `initialValues`
+            // changes as determined by a deep equality check (reference changes don't trigger
+            // reinitialization). Changing `interpretationsInitializedAt` while leaving
+            // `interpretations` intact accordingly allows consumers to reset the form
+            interpretationsInitializedAt,
+          }}
```
arsalansufi, line 69: I experimented with a few different ideas, and this felt like the simplest and clearest. There are more details on the deep comparison behavior of Formik [here](https://github.com/jaredpalmer/formik/issues/1808). Formik suggests using the `resetForm` prop, but our reset button (i.e. the "Ballot Not Found" button) lives outside the form.

client/src/components/AuditBoard/AuditBoardView.test.tsx
```diff
@@ -472,5 +472,68 @@ describe('AuditBoardView', () => {
...
+        userEvent.click(
+          within(dialog).getByRole('button', { name: 'Confirm Selections' })
+        )
+
+        await waitFor(() => {
+          expect(dialog).not.toBeInTheDocument()
+        })
+        await screen.findByText('Audit Ballot Selections')
+      })
+    })
```
jonahkagan, line 537: great test!

client/src/components/AuditBoard/BallotAudit.tsx
jonahkagan: I agree given the constraints this seems like a viable and clear solution. Did you experiment at all with moving the Ballot Not Found button inside the form? 

Also just as a note, I've been slowly trying to move away from Formik because it's often got annoying quirks. I've been building all new forms with [react-hook-form](https://react-hook-form.com/). I don't think that will help here but might be worth looking into briefly.

arsalansufi: I did -- the key challenge was that I'd either have to change the layout of the page, bringing the "Ballot Not Found" down out of the row it's in, _or_ move all of [this content](https://github.com/votingworks/arlo/blob/d154ab646120434cbb04447800c2c2884ab6218a/client/src/components/AuditBoard/Ballot.tsx#L249-L286) nto the `BallotAudit` component and its `form`

Possible! But a little clunky

jonahkagan: I think it's clearer to include the Ballot Not Found button in the form since it conceptually _is_ submitting the same type of response that the form is submitting ("what's the audit result for this ballot?"). So even though there would be some extra content in there to make the layout work, that seems ok to me.

arsalansufi: Good point, even if some of the header content feels auxiliary to the form, the "Ballot Not Found" button does feel core to it. Working on that now!

client/src/components/AuditBoard/BallotAudit.tsx
```diff
@@ -38,84 +68,129 @@ const SubmitButton = styled(FormButton)`
...
   interpretations: IBallotInterpretation[]
-  interpretationsInitializedAt: Date
   setInterpretations: (interpretations: IBallotInterpretation[]) => void
-  previousBallot: () => void
   confirmSelections: (interpretations: IBallotInterpretation[]) => void
+  submitBallotNotFound: () => void
+  previousBallot: () => void
 }
 
 const BallotAudit: React.FC<IProps> = ({
```
arsalansufi, line 80: Much easier to review the changes to this component ignoring whitespace with [?w=1](https://github.com/votingworks/arlo/pull/1462/commits/c0f54487faaf487fd2b4792d6606bff446cd9909?w=1)!

client/src/components/AuditBoard/AuditBoardView.test.tsx
```diff
@@ -535,5 +536,71 @@ describe('AuditBoardView', () => {
         await screen.findByText('Audit Ballot Selections')
       })
     })
+
+    it('clears all choices when "Ballot Not Found" is selected, even if previous choices were almost submitted', async () => {
```
arsalansufi, line 540: There are ways to write the choice clearing code such that you pass the above test (clears all choices when "Ballot Not Found" is selected) but not this one (clears all choices when "Ballot Not Found" is selected, **even if previous choices were almost submitted**) and vice versa

arsalansufi:
@jonahkagan This is now ready for a follow-up look!

client/src/components/AuditBoard/BallotAudit.tsx
```diff
@@ -38,75 +68,129 @@ const SubmitButton = styled(FormButton)`
...
+  ballot,
   contests,
   interpretations,
   setInterpretations,
   confirmSelections,
+  submitBallotNotFound,
```
jonahkagan, line 86: what do you think about renaming this `confirmBallotNotFound` to match `confirmSelections`?

client/src/components/AuditBoard/Ballot.tsx
```diff
@@ -143,7 +110,7 @@ const Ballot: React.FC<IProps> = ({
 
   const [interpretations, setInterpretations] = useState<
     IBallotInterpretation[]
-  >(contests.map(emptyInterpretation))
+  >(contests.map(constructEmptyInterpretation))
```
jonahkagan, line 113: If I'm reading this code correctly, I think the `interpretations` state is actually unnecessary. On initial render, it's derived from `ballot.interpretations` or set to a default empty state. And then it only gets set before submitting. I think removing it might make this all even simpler.

client/src/components/AuditBoard/AuditBoardView.test.tsx
jonahkagan: Nice!

client/src/components/AuditBoard/Ballot.tsx
jonahkagan: Tried it out to be sure I wasn't missing something: https://github.com/votingworks/arlo/pull/1463

jonahkagan: I know it's expanding the scope a bit but I didn't notice til just now and it feels like it might help prevent bugs like this in the future.

arsalansufi: Ooh interesting, taking a sec to process and incorporate 🤔

client/src/components/AuditBoard/BallotAudit.tsx
arsalansufi: :100:

client/src/components/AuditBoard/Ballot.tsx
arsalansufi: Okay, it's clicking!

I see now that we were previously copying to state (and occasionally letting that state diverge from `ballot.interpretations`, e.g. after a user clicked submit but didn't confirm) when we could just consider `ballot.interpretations` directly (and avoid any possibility of divergence altogether)

:100: find!

https://github.com/votingworks/arlo/pull/1462/commits/135f7e02246b82c68d3c4e8f476c390c92311b13

jonahkagan (approved):
Great work! Thanks for sticking through all the rounds of review

client/src/components/AuditBoard/__snapshots__/Ballot.test.tsx.snap
```diff
@@ -144,6 +144,7 @@ exports[`Ballot renders correctly with an audited ballot 1`] = `
                           class="bp3-control bp3-checkbox sc-chPdSV gVsrQj"
                         >
                           <input
+                            checked=""
```
arsalansufi, line 147: Taking a sec to sanity check why this snapshot diff arose 🤔

arsalansufi: Okay! This makes sense. We expect this checkbox to be checked. In fact, the relevant test has explicit assertions that this is the case:

https://github.com/votingworks/arlo/blob/d154ab646120434cbb04447800c2c2884ab6218a/client/src/components/AuditBoard/Ballot.test.tsx#L55-L57

It's just that, back when we were pulling `interpretations` from state, we relied on a `useEffect` hook to properly initialize that state:

https://github.com/votingworks/arlo/blob/d154ab646120434cbb04447800c2c2884ab6218a/client/src/components/AuditBoard/Ballot.tsx#L221-L232

This meant that, on the very first render (which is what our snapshots typically consider), this checkbox wasn't checked. It was only after an (almost immediate) follow-up render due to the `useEffect` hook that the checkbox became checked.

In our new cleaner setup, the checkbox is checked on very first render :tada:

jonahkagan: Great investigation. That all makes sense.

In general, I'm trying to move Arlo away from snapshot tests to evaluate UI interactions (for exactly reasons like this). If you come across a snapshot, feel free to replace it with meaningful assertions and remove the snapshot.

Issue #1442, opened 2022-03-29
Allow conducting ballot-comparison audits with mixed ES&S and Dominion CVRs


benadida:
ES&S and Dominion CVRs treat overvotes differently, which leads us to having different data entry patterns for both, and thus to not be able to handle states that have mixed-vendor CVRs. We need to fix that.

TBD: how we approach this. Needs a discussion with @jonahkagan and @ginvdr

jonahkagan:
Idea for how to approach this: Special-case ES&S CVRs in the audit math. 
- The audit board audits the ballot normally - e.g. when they see an overvote, they just record all the choices that were voted for
- In the audit math, when we see "overvote" was recorded in the CVR, check to see if the audit board interpretation is also an overvote (we can easily tell because we only support vote-for-1 contests for ES&S).
- Record a discrepancy if the audit board interpretation is not an overvote (maybe +2 to be conservative)
- Do the same for undervotes

umbernhard:
I think this is sound. Idk if there's a reason to hardcode +2, especially if we are taking the audit board's input to be the source of truth. We'll know whether it's +2, +1, or 0 based on who they audit the ballot for.

jonahkagan:
Ah yeah that could work!

PR #1464, opened 2022-04-28
Handle ES&S CVR overvotes and undervotes in audit math

jonahkagan:
Task: #1442 

When a voter overvotes or undervotes, ES&S CVRs record the strings "overvote" and "undervote" respectively. This doesn't tell us what choices the voter actually selected. Currently, our audit math for ballot comparison audits compares the choices the voter selected according to the CVR with the choices the voter selected according to the audit board, so it doesn't work in the ES&S case.

This PR makes a few changes to support this case:
1. When parsing ES&S CVRs, convert the "overvote" for a ballot into an "o" for each choice, and an "undervote" for a ballot into a "u" for each choice (for compact storage that matches our existing one-character-per-choice storage mechanism)
2. Move the actual parsing of these interpretation characters into the audit math so that we can easily pass these strings along (previously, we parsed them as soon as we pulled them from the database)
3. Add a special case into the discrepancy calculation to handle ES&S overvotes and undervotes

While making this change, I also realized that the tests in test_supersimple.py used a number for the ballot id instead of a string (the expected type), so in order to get type checking to pass, I went ahead and fixed that.


server/audit_math/sampler_contest.py
```diff
@@ -38,7 +38,7 @@ class Contest:
 
     candidates: Dict[str, int]  # Dict mapping candidates to their vote totals
     num_winners: int  # How many winners this contest had
-    votesAllowed: int  # How many voters are allowed in this contest
+    votes_allowed: int  # How many voters are allowed in this contest
```
jonahkagan, line 41: this was just a typo I found along the way

server/tests/audit_math/test_supersimple.py
```diff
@@ -211,48 +229,240 @@ def test_race_not_in_sample_discrepancy(contests, cvrs):
...
     )
 
     assert discrepancies
-    assert discrepancies[0]["counted_as"] == 2
-    assert discrepancies[0]["weighted_error"] == Decimal(2) / Decimal(2000)
+    assert discrepancies["ballot-0"]["counted_as"] == 2
+    assert discrepancies["ballot-0"]["weighted_error"] == Decimal(2) / Decimal(2000)
+
+
+def test_ess_discrepancies(contests) -> None:
```
jonahkagan, line 279: Adding this test case is the only meaningful change in this file.

server/audit_math/supersimple.py
```diff
@@ -116,25 +115,46 @@ def discrepancy(
...
+        has_undervote = "u" in reported.get(contest.name, {}).values()
+        audited_votes = sum(map(int, (audited.get(contest.name, {}).values())))
+        # If the audited result correctly identified overvote/undervote, return
+        # 0 error.  Otherwise, return an error using the standard formula, but
+        # substituting in the appropriate overvotes/undervotes.
+        if has_overvote:
+            if audited_votes > 1:
+                return 0
+            else:
+                return (1 - a_w) - (1 - a_l)
```
umbernhard, line 142: I think the 1 here should be a zero. An overvote isn't a valid vote for any candidate, so v_w and v_l should be zero.

umbernhard (changes requested):
I think there's a math thing to clean up, but otherwise this looks good. Unfortunately I suspect fixing the math thing will require rewriting some tests.

server/audit_math/supersimple.py
jonahkagan: I was thinking about what would happen in a "normal" overvote case (e.g. for another CVR type), and I was thinking that the CVR would show exactly the marks made by the voter (multiple 1s), not the interpretation made of an overvote (which would be all zeros). I don't think there's anywhere that we convert the multiple 1s to all 0s, so I think this is the most faithful representation of that case. 

Luckily, whether it's 1s or 0s, I think those two parts of this equation cancel out anyway, so I don't think it matters at the end of the day:
```
(1 - a_w) - (1 - a_l)
= 1 - a_w - 1 + a_l
= 1 - 1 - a_w + a_l
= a_l - a_w
```

I just wrote it out the long way instead of simplifying so that it would be clearer to read.

umbernhard: ah I see. Yeah, I think that makes sense to me then.

umbernhard (approved):

PR #1469, opened 2022-05-03
find_first_duplicate util function

jonahkagan:
Adds a util function `find_first_duplicate` to return the first duplicate value in a collection. Also refactors the utils to combine the `group_by` utils with this one into a new module `utils.colllection`. Last, uses `find_first_duplicate` everywhere it would have been applicable up til now.


server/util/collections.py
```diff
@@ -0,0 +1,22 @@
...
+import itertools
+
+T = TypeVar("T")
+
+
+def group_by_iter(items: Iterable[T], key=None):
+    return itertools.groupby(sorted(items, key=key), key=key)
+
+
+def group_by(items: Iterable[T], key=None):
```
arsalansufi, line 12: I know this code was already written before this PR, but a quick comment with some sample inputs and outputs could be helpful. Even pointing readers to the test (which has sample inputs and outputs) could be sufficient.

I also wonder if we should rename `key` `key_fn` for clarity, since it's a function

server/api/audit_boards.py
```diff
@@ -49,7 +51,7 @@ def validate_audit_boards(
         audit_boards, {"type": "array", "items": CREATE_AUDIT_BOARD_REQUEST_SCHEMA}
     )
 
-    if len(set(ab["name"] for ab in audit_boards)) != len(audit_boards):
+    if find_first_duplicate(ab["name"] for ab in audit_boards):
```
arsalansufi, line 54: Nice! Much clearer now

arsalansufi (approved):
LGTM!

server/util/collections.py
jonahkagan: Thanks, I'll add some documentation. `key` is the Python convention (see `sorted` and `itertools.groupby` that are used in this file), so I'll probably leave that

PR #1474, opened 2022-05-09
CSV parse: don't detect total row for column that sums to 0

jonahkagan:
We have some logic to auto-detect total rows (which aren't allowed) if the last row in the CSV has a value that is the sum of all the other rows for that column. This logic shouldn't apply for columns with all 0s, since that doesn't indicate a total row. This case won't happen often (probably only for batch tallies where a candidate got 0 votes - very rare). But still good to guard against.


arsalansufi (approved):
0️⃣

Issue #1437, opened 2022-03-29
Generate batch tally sheets for printing


benadida:
In batch audits, make it easy to print a batch tally sheet that will be filled out by hand, much like:
https://www.google.com/url?q=https://docs.voting.works/arlo/full-manual-tally/audit-board-guide

PR #1475, opened 2022-05-09
Allow jurisdiction managers to generate and print batch tally sheets for record keeping by hand

arsalansufi:

# Overview

Issue link: https://github.com/votingworks/arlo/issues/1437

This PR allows jurisdiction managers to generate and print batch tally sheets for record keeping by hand as audit boards complete batch audits ✍️

This PR also pioneers the use of [@react-pdf/renderer](https://react-pdf.org) in Arlo!

There are still some improvements to make to this batch tally sheet functionality, but those can happen at a later time. I'll be documenting these in follow-up issues. These include:

- Adding container and tabulator to batch tally sheets
- ~Automatically opening the print dialog when the PDF is opened~ Decided this isn't actually necessary
- ~Giving audit admins the ability to print batch tally sheets~ Incorporating into this PR

# Screenshots / screencaptures

### In action

https://user-images.githubusercontent.com/12616928/167498883-78089088-6c50-415b-a3ae-1e7da1ea59c2.mov

### New button

02-button

### Sample batch tally sheet, common case

03-batch-tally-sheet-common-case

### Sample batch tally sheet, extreme case

04-batch-tally-sheet-extreme-case-1 05-batch-tally-sheet-extreme-case-2

### Sample batch tally sheets in PDF form

Note that, when actually downloading batch tally sheets, the generated PDF will contain tally sheets for all batches, not just one.

[06-batch-tally-sheet-common-case.pdf](https://github.com/votingworks/arlo/files/8655445/06-batch-tally-sheet-common-case.pdf)
[07-batch-tally-sheet-extreme-case.pdf](https://github.com/votingworks/arlo/files/8655446/07-batch-tally-sheet-extreme-case.pdf)

### PDF generation failure error case

08-error-case

# Testing

- [x] Added comprehensive tests
- [x] Tested manually (see the screenshots and screencaptures above, notably the extreme case with very long text)


client/src/components/JurisdictionAdmin/RoundManagement.test.tsx
```diff
@@ -68,6 +89,15 @@ const apiCalls = {
 }
 
 describe('RoundManagement', () => {
+  beforeEach(() => {
+    // Clear mock call counts, etc.
+    jest.clearAllMocks()
+
+    // Reset the BlobProvider mock's implementation since some tests override the default mock
+    // implementation
+    ;(BlobProvider as jest.Mock).mockImplementation(() => null)
```
arsalansufi, line 98: This funky leading semicolon is a side effect of the fact that Arlo's auto-formatting prefers no trailing semicolons. The leading semicolon here is necessary to distinguish from a function call (`jest.clearAllMocks()(BlobProvider as jest.Mock)...`)

client/src/components/JurisdictionAdmin/RoundManagement.test.tsx
```diff
@@ -165,20 +195,101 @@ describe('RoundManagement', () => {
...
+
     await withMockFetch(expectedCalls, async () => {
       renderView({
-        round: roundMocks.incomplete,
         auditBoards: auditBoardMocks.unfinished,
         createAuditBoards: jest.fn(),
+        round: roundMocks.incomplete,
+      })
+      await screen.findByRole('button', {
+        name: /Download Aggregated Batch Retrieval List/,
```
arsalansufi, line 227: I know we discussed removing the "Aggregated" in this copy. Taking care of that in a separate PR!

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/DownloadBatchTallySheetsButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { toast } from 'react-toastify'
+
+import BatchTallySheet from './BatchTallySheet'
+import useContestsJurisdictionAdmin from '../useContestsJurisdictionAdmin'
+import { useBatches } from '../useBatchResults'
+
+const AnchorButtonErrorState = (): JSX.Element => {
+  useEffect(() => {
+    toast.error('Error preparing batch tally sheets for download')
+  }, [])
```
arsalansufi, line 13: It may also make sense to log this error to Sentry. I know client-side Sentry automatically reports unhandled client-side errors. This case is interesting, though, because the error is handled, so we can't rely on that behavior. We could explicitly report to Sentry here

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/DownloadBatchTallySheetsButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+const DownloadBatchTallySheetsButton = ({
+  electionId,
+  jurisdictionId,
+  jurisdictionName,
+  roundId,
+}: IProps): JSX.Element | null => {
+  const batchesResp = useBatches(electionId, jurisdictionId, roundId)
+  const contests = useContestsJurisdictionAdmin(electionId, jurisdictionId)
+
+  if (!batchesResp || !batchesResp.data || !contests || contests.length === 0) {
```
jonahkagan, line 38: `useBatches` will always return a value. The pattern I've been starting to use for data loading using react-query looks like this:
```ts
const batchesQuery = useBatches(electionId, jurisdictionId, roundId)
if (!batchesQuery.isSuccess) return null
const { batches } = batchesQuery.data
```

It also should be impossible for `contests.length` to be 0 at this point, since the audit has launched and that requires configuring contests.

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/DownloadBatchTallySheetsButton.tsx
jonahkagan: Hmm yeah failure here would be unexpected and we'd probably want to know. I like the idea of explicitly reporting

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/DownloadBatchTallySheetsButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+          <AnchorButton
+            href={url || undefined}
+            icon="th"
+            loading={loading}
+            target="_blank"
+          >
+            Download Batch Tally Sheets
+          </AnchorButton>
+        )
+      }}
```
jonahkagan, line 76: So this UX is a little different from the existing buttons, which generate their PDFs on click. Pre-generating seems nice because I assume it will be faster to download on click (though I wonder if it will impact page rendering performance at all - did you test on an election with many sampled batches?)

I don't know that it makes sense to make pre-rendering transparent to the user. I think it makes more sense for them to think that nothing happens until they click the button, since they might load the page and not choose to download those PDFs.

A few thoughts:
- Is it possible to not pre-generate the PDFs and only generate them on button click?
- Could we pre-generate in the background, but then if they click before we're done, have the button spin then? (Or if there was an error, toast the error then)

client/src/components/JurisdictionAdmin/RoundManagement.tsx
```diff
@@ -195,6 +199,18 @@ export const JAFileDownloadButtons = ({
...
+      // DownloadBatchTallySheetsButton to make audit admin API calls instead of jurisdiction
+      // manager API calls depending on the user type
+      userType === 'jurisdiction_admin' && (
+        <DownloadBatchTallySheetsButton
+          electionId={electionId}
+          jurisdictionId={jurisdictionId}
+          jurisdictionName={jurisdictionName}
+          roundId={round.id}
+        />
+      )}
```
jonahkagan, line 213: I think audit admins actually can already do this: https://github.com/votingworks/arlo/blob/main/server/api/batches.py#L91-L95. I think I misspoke when giving the Arlo architecture overview to say that the user types had entirely separate APIs - it started that way but as we've given the audit admin more abilities to act like jursidiction admins, that has shifted.

So I think you can just enable it on the frontend for audit admins as well

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/BatchTallySheet.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+      batchName="Batch 1"
+      choices={[
+        { id: 'choice-1', name: 'Choice 1', numVotes: 0 },
+        { id: 'choice-2', name: 'Choice 2', numVotes: 0 },
+        { id: 'choice-3', name: 'Choice 3', numVotes: 0 },
+      ]}
+      jurisdictionName="Jurisdiction 1"
+    />
+  )
+  expect(container).toMatchSnapshot()
```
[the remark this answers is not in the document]
jonahkagan: Agreed! We even have a way to snapshot the resulting PDF: https://github.com/votingworks/arlo/blob/1dfea479613546d5ee8210d06d297958f895b5aa/client/src/components/JurisdictionAdmin/generateSheets.test.tsx#L60. I think it's nice to have a visual snapshot diff for PDFs so that you can really make sure the styling works as expected

jonahkagan (commented):
Nice work!

Interesting to see how react-pdf works. I didn't realize beforehand that you had to use all these pdf-specific building blocks - I thought it would just take HTML. Seems like a bit of a bummer to have to reinvent HTML using their building blocks.

arlo.code-workspace
```diff
@@ -17,5 +17,8 @@
     ],
     "python.formatting.provider": "black",
     "typescript.tsdk": "client/node_modules/typescript/lib",
+    "editor.codeActionsOnSave": {
+      "source.fixAll": true
+    },
```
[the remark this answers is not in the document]
arsalansufi: Good q, I was following the example set in VxSuite ([relevant config](https://github.com/votingworks/vxsuite/blob/7ae6e124188d0c3e214fbcf3fe2c1f505a697faf/vxsuite.code-workspace#L105-L107) + [relevant Slack thread](https://votingworks.slack.com/archives/C014MGC0WRE/p1648748528735429)).

While `codeActionsOnSave` can do [more than formatting](https://stackoverflow.com/questions/60718493/difference-between-codeactionsonsave-and-formatonsave-in-vs-code), `source.fixAll` doesn't seem to be doing anything other than formatting for me. It's not auto-fixing all eslint warnings / errors, e.g. it doesn't remove unused imports. I admittedly haven't been able to find super clear documentation on what exactly `source.fixAll` does

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/BatchTallySheet.tsx
```diff
@@ -0,0 +1,145 @@
...
+        <Text style={styles.pLastInSection}>
+          Batch Type (Optional): {blankLine(20)}
+        </Text>
+      </View>
+      <View style={styles.pageSection}>
+        <PdfCheckbox
+          fontSize={styles.page.fontSize}
+          textBeforeCheckbox="Was the container sealed when received by the audit board?"
+          textAfterCheckbox="Yes"
+        />
```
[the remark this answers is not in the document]
arsalansufi: Good q - I was finding that, for every checkbox, I wanted associated text to sit inline with the checkbox. To accomplish this, I needed to wrap every checkbox and associated text in a `View` with flex styling.

https://github.com/votingworks/arlo/blob/950c9fced920c1d03e99cdd99e052c371f5aa01e/client/src/components/Atoms/pdf/PdfCheckbox.tsx#L18-L31

So I went ahead and baked that wrapping into the component. This also allowed me to ensure consistent spacing between checkboxes and their associated text

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/BatchTallySheet.tsx
```diff
@@ -0,0 +1,145 @@
...
+          seal the container.
+        </Text>
+        <PdfCheckbox
+          fontSize={styles.page.fontSize}
+          textBeforeCheckbox="Was the container resealed by the audit board?"
+          textAfterCheckbox="Yes"
+        />
+      </View>
+      <View style={[styles.pageSection, styles.flexRow]}>
+        <PdfSignatureLine label="(Audit Board Member)" marginRight={9} />
```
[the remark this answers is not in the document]
arsalansufi: Yeah this crossed my mind - will update!

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/DownloadBatchTallySheetsButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  const { batches } = batchesResp.data
+  // Batch comparison audits only support a single contest
+  const contest = contests[0]
+
+  return (
+    <BlobProvider
+      document={
+        <Document title="batch-tally-sheets.pdf">
```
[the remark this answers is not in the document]
arsalansufi: Actually, describing things more clearly, the existing download buttons don't immediately open a print dialog. They do download the file. Then, when the downloaded file is opened by the user, a print dialog is auto-opened.

After https://github.com/votingworks/arlo/pull/1475/commits/e793e2808e144c8628b5a273ecf40569ae3a7cd2, existing download buttons _and_ the button that I've added all download the file. When the downloaded file is opened, a print dialog is no longer auto-opened.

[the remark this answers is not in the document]
jonahkagan: Wow that is a weird feature! I never noticed that. Didn't realize PDFs could do that

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/BatchTallySheet.tsx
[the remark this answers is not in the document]
jonahkagan: Gotcha. Slight preference for using composition here but not a blocker

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/BatchTallySheet.test.tsx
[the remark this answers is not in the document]
arsalansufi: Ooh very cool!

I just did a quick investigation into generating visual snapshots for PDFs generated by `@react-pdf/renderer` but was reminded why I did my [HTML mocking](https://github.com/votingworks/arlo/blob/950c9fced920c1d03e99cdd99e052c371f5aa01e/client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/BatchTallySheet.test.tsx#L9-L13) to begin with. When trying to render `@react-pdf/renderer` PDF content in Jest tests, I encounter errors like:

```
Error: BlobProvider is a web specific API. You're either using this component on Node, or your bundler is not loading react-pdf from the appropriate web build.
```

[Others do too and recommend mocking](https://github.com/diegomura/react-pdf/issues/710) as I've done

I created a follow-up issue for generating visual snapshots for PDFs generated by `@react-pdf/renderer`: https://github.com/votingworks/arlo/issues/1476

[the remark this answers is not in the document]
jonahkagan: Ah that's too bad

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/DownloadBatchTallySheetsButton.tsx
arsalansufi: https://github.com/votingworks/arlo/pull/1475/commits/34871cae983dfeede2215d717d06636a6377f1bf

arlo.code-workspace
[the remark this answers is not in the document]
arsalansufi: Sounds good! https://github.com/votingworks/arlo/pull/1475/commits/4047af0cb2b6a457292cd608d2a2a88d2624078d

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/BatchTallySheet.tsx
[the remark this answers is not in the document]
arsalansufi: 👍 Incorporated into https://github.com/votingworks/arlo/pull/1475/commits/872042ec16bf01b852d3f51679395fc0eb9b1737

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/BatchTallySheet.tsx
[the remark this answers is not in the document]
arsalansufi: Taken care of in https://github.com/votingworks/arlo/pull/1475/commits/872042ec16bf01b852d3f51679395fc0eb9b1737! The main reason I'd held off earlier was a challenge with properly typing a general-purpose style prop, but [I got that sorted out](https://github.com/votingworks/arlo/blob/872042ec16bf01b852d3f51679395fc0eb9b1737/client/src/components/Atoms/pdf/styles.ts)

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/BatchTallySheet.tsx
```diff
@@ -0,0 +1,145 @@
...
+    fontSize: 18,
+  },
+  subHeading: {
+    fontSize: 16,
+    marginTop: 10,
+  },
+  p: {
+    marginBottom: 10,
+    overflow: 'hidden',
+  },
```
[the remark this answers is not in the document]
arsalansufi: Yeah it was a combination of complexity and generalizability that made me move certain styles to custom components (more complex and more generalizable --> custom component). In retrospect, making reusable/shared `PdfP`, `PdfHeading`, and `PdfSubHeading` components makes sense. Thanks for the nudge!

Note that I still didn't create a reusable/shared `PdfPageSection` component - that felt specific to this particular PDF

https://github.com/votingworks/arlo/pull/1475/commits/872042ec16bf01b852d3f51679395fc0eb9b1737

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/DownloadBatchTallySheetsButton.tsx
arsalansufi: https://github.com/votingworks/arlo/pull/1475/commits/a1aa0361379dbdbafacc01bb582b6dfaa99c0927

client/src/components/JurisdictionAdmin/RoundManagement.tsx
arsalansufi: Ooh so I did initially try enabling this for audit admins, but I encountered this permission error

error-ui

Taking a closer look, while batch retrieval works fine for both user types, contest retrieval does not. So there's still some API work to be done that feels like it belongs in its own follow-up issue / PR

error

jonahkagan: Fixing that should be as easy as applying this patch:

```diff --git a/server/api/contests.py b/server/api/contests.py
index 35523cf8..bbcbf05d 100644
--- a/server/api/contests.py
+++ b/server/api/contests.py
@@ -223,7 +223,7 @@ def list_contests(election: Election):
 @api.route(
     "/election/<election_id>/jurisdiction/<jurisdiction_id>/contest", methods=["GET"]
 )
-@restrict_access([UserType.JURISDICTION_ADMIN])
+@restrict_access([UserType.AUDIT_ADMIN, UserType.JURISDICTION_ADMIN])
 def list_jurisdictions_contests(
     election: Election, jurisdiction: Jurisdiction,  # pylint: disable=unused-argument
 ):```

Not trying to expand scope here but I do think we generally want AAs to be able to click the same buttons as JAs, and it should be easy to facilitate. If the above doesn't work then I'm ok leaving it for a follow up task.

File: client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/DownloadBatchTallySheetsButton.tsx (hunk 5 above)
Comment (arsalansufi) [reply client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/DownloadBatchTallySheetsButton.tsx:90]: Great call re `setImmediate`! Will incorporate that!

And yup, I've moved the error handling into the `onClick` in my actual cleaned up code

Also agree that caching is probably riskier than it's worth for now

File: client/src/components/JurisdictionAdmin/RoundManagement.test.tsx [hunk 13]
```diff
@@ -165,20 +200,103 @@ describe('RoundManagement', () => {
...
+      expect(within(pdf).queryAllByText('Candidates/Choices')).toHaveLength(3)
+      expect(within(pdf).queryAllByText('Enter Stack Totals')).toHaveLength(3)
+      expect(within(pdf).queryAllByText('Choice One')).toHaveLength(3)
+      expect(within(pdf).queryAllByText('Choice Two')).toHaveLength(3)
+
       screen.getByRole('table') // Tested in BatchRoundDataEntry.test.tsx
     })
   })
 
+  it('handles failures to generate batch tally sheets PDF whoa', async () => {
```
jonahkagan, line 261: whoa!

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/BatchTallySheet.tsx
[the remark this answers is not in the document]
jonahkagan: Nicely done!

client/src/components/JurisdictionAdmin/RoundManagement.test.tsx
```diff
@@ -234,31 +226,18 @@ describe('RoundManagement', () => {
...
-      expect(within(pdf).queryAllByText('Enter Stack Totals')).toHaveLength(3)
-      expect(within(pdf).queryAllByText('Choice One')).toHaveLength(3)
-      expect(within(pdf).queryAllByText('Choice Two')).toHaveLength(3)
-
       screen.getByRole('table') // Tested in BatchRoundDataEntry.test.tsx
+
+      act(() => {
+        userEvent.click(downloadBatchTallySheetsButton)
+      })
+      // PDF contents are tested in BatchTallySheet.test.tsx
```
arsalansufi, line 234: Still trying to see if I can do some rendering in this test as I was earlier - haven't been able to get things working quite yet

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/DownloadBatchTallySheetsButton.tsx
arsalansufi: And here! https://github.com/votingworks/arlo/pull/1475/commits/ccdf21d58ca16d2c0d6d2a71522ee7f10236c1a9

With the minimal update to get tests working in https://github.com/votingworks/arlo/pull/1475/commits/b484c60659b9bdb669c4a4bdf8aa574e8b46a5ba. Still trying to make some improvements there (https://github.com/votingworks/arlo/pull/1475#discussion_r870626541)

client/src/components/JurisdictionAdmin/RoundManagement.test.tsx
arsalansufi: Sorted this out! https://github.com/votingworks/arlo/pull/1475/commits/4981c6821a6c90a55f3a8fb66d5ce7501e96a6bd

client/src/components/Atoms/pdf/styles.ts
```diff
@@ -0,0 +1,14 @@
+import ReactPdf from '@react-pdf/renderer'
+
+/**
+ * Since @react-pdf/renderer doesn't export its Style type, we have to do some extra work to get to
+ * it
+ *
+ * ReactPdf.NodeProps['style']         --> Style | Style[] | undefined
```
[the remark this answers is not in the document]
arsalansufi: Ah nice! I'll stick with what I have for now to avoid having to install the additional `@react-pdf/types`

client/src/components/JurisdictionAdmin/RoundManagement.tsx
arsalansufi: Nice! That did in fact just work! Working on adding an audit admin test now, though, which may take me a sec to figure out

client/src/components/Atoms/pdf/styles.ts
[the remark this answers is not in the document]
jonahkagan: I think it's already a dependency of @react-pdf/renderer

arsalansufi:
> Interesting to see how react-pdf works. I didn't realize beforehand that you had to use all these pdf-specific building blocks - I thought it would just take HTML. Seems like a bit of a bummer to have to reinvent HTML using their building blocks.

Definitely. While this works fine for Arlo's fairly simple PDFs, using this library in VxSuite for ballots could pose challenges. Though it's not nearly as popular as @react-pdf/renderer ([10.6K stars on GitHub](https://github.com/diegomura/react-pdf)), @progress/kendo-react-pdf ([121 stars on GitHub](https://www.npmjs.com/package/@progress/kendo-react-pdf)) does support [converting React HTML to a PDF](https://www.telerik.com/blogs/generating-pdf-react-easy-as-1-2-3). Another option to explore down the road. But I'm happy with @react-pdf/renderer for now in Arlo! Still an improvement over jsPDF's imperative style

-----

To round things out, here are some screenshots of the final UI, both for jurisdiction managers and audit admins. (cc @ginvdr)

_Jurisdiction manager_

https://user-images.githubusercontent.com/12616928/167956992-e12109de-d0c3-4833-9876-106f531b52c1.mov

_Audit admin_

https://user-images.githubusercontent.com/12616928/167957004-7cab8b42-bf58-43d2-bfa0-346e22934e75.mov

And here's a sample PDF (all batches): [batch-tally-sheets.pdf](https://github.com/votingworks/arlo/files/8673731/batch-tally-sheets.pdf)

-----

@jonahkagan Ready for a final look!

client/src/components/AuditAdmin/Progress/Progress.test.tsx
```diff
@@ -738,6 +786,69 @@ describe('Progress screen', () => {
...
+      userEvent.click(
+        await within(modal).findByRole('button', {
+          name: /Download Aggregated Batch Retrieval List/,
+        })
+      )
+      expect(window.open).toHaveBeenCalledWith(
+        '/api/election/1/jurisdiction/jurisdiction-id-1/round/round-1/batches/retrieval-list'
+      )
+
+      await act(async () => {
```
jonahkagan, line 836: What's this `await act(...)` for? I thought `userEvent.click` had `act` built in

client/src/components/Atoms/pdf/styles.ts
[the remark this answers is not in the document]
arsalansufi: Ah nice, switched over: https://github.com/votingworks/arlo/pull/1475/commits/888f072ad33d051d7fdfce973972d5174e8b1a56

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/DownloadBatchTallySheetsButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { Document, pdf } from '@react-pdf/renderer'
+import { toast } from 'react-toastify'
+import * as FileSaver from 'file-saver'
+import * as Sentry from '@sentry/react'
+
+import BatchTallySheet from './BatchTallySheet'
+import useContestsJurisdictionAdmin from '../useContestsJurisdictionAdmin'
+import { useBatches } from '../useBatchResults'
+
+const FILE_NAME = 'batch-tally-sheets.pdf'
```
jonahkagan, line 12: This doesn't really match the style of some of the other file names, but they aren't super consistent so I think it's fine. Added a follow up task to fix: https://github.com/votingworks/arlo/issues/1479

client/src/components/JurisdictionAdmin/RoundManagement.test.tsx
```diff
@@ -165,17 +211,94 @@ describe('RoundManagement', () => {
...
+      })
+      const downloadBatchTallySheetsButton = screen.getByRole('button', {
+        name: /Download Batch Tally Sheets/,
       })
-      await screen.findByText('Download Aggregated Batch Retrieval List')
       screen.getByRole('table') // Tested in BatchRoundDataEntry.test.tsx
+
+      act(() => {
+        userEvent.click(downloadBatchTallySheetsButton)
+      })
```
jonahkagan, line 242: unnecessary act?

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/DownloadBatchTallySheetsButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+      } catch (err) {
+        toast.error('Error preparing batch tally sheets for download')
+        Sentry.captureException(err)
+      }
+      setIsDownloading(false)
+    })
+  }
+
+  return (
+    <Button icon="th" loading={isDownloading} onClick={onClick}>
```
jonahkagan, line 71: nit: switch to `icon="document"` to differentiate from the retrieval list

jonahkagan (approved):
Awesome work!

client/src/components/AuditAdmin/Progress/Progress.test.tsx
arsalansufi: Ooh very interesting. As I was building this test out and this click was its last line (because I hadn't yet written the assertions that came after), I was receiving the following warning from Jest:

warning

Adding the `act` resolved the warning. I assumed it was needed because of the newly added `setImmediate` triggering state updates with a delay.

But now that I have some assertions after this line, the `act` no longer appears to be necessary. Removing the assertions afterwards results in the above error resurfacing.

Will remove the `act` calls but still curious to find an explanation for the above behavior (a task for tomorrow)

client/src/components/JurisdictionAdmin/RoundManagement.test.tsx
arsalansufi: See https://github.com/votingworks/arlo/pull/1475#discussion_r870817300

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/DownloadBatchTallySheetsButton.tsx
arsalansufi: Nice, thanks for creating the follow-up task!

client/src/components/AuditAdmin/Progress/Progress.test.tsx
jonahkagan: There's a short explanation of the act warning here: https://testing-library.com/docs/react-testing-library/faq/. And it also links to a longer blog post with a good explanation. It basically means that there was code trying to make React state updates after the test already finished.

arsalansufi: Ah perfect, thanks for that!

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton/BatchTallySheet.tsx
```diff
@@ -0,0 +1,145 @@
...
+        <Text style={styles.subHeading}>
+          Batch Name: <Text style={styles.bold}>{batchName}</Text>
+        </Text>
+      </View>
+      <View style={styles.pageSection}>
+        <Text style={styles.p}>Jurisdiction: {jurisdictionName}</Text>
+        <Text style={styles.p}>Audit Board: {auditBoardName}</Text>
+        <Text style={styles.pLastInSection}>
+          Batch Type (Optional): {blankLine(20)}
+        </Text>
```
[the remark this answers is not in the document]
arsalansufi: GitHub issue: https://github.com/votingworks/arlo/issues/1491

PR #1477, opened 2022-05-10
Create utils/api

jonahkagan:
Create utils/api.ts for `fetchApi` and related functions.

This was originally created in support-api.ts to try out using react-query. Now that we're moving towards that as the standard approach, it should have its own file. I didn't bother moving over the deprecated `api `util function since it will eventually be phased out.

Also removes unused checkAndToast function.


arsalansufi (approved):
LGTM!

PR #1478, opened 2022-05-11
react-query v3.39.0

jonahkagan:
Needed some of the changes for upcoming work. I had to modify a couple small things to get the existing tests to pass.


client/src/components/SupportTools/SupportTools.test.tsx
```diff
@@ -175,20 +175,9 @@ const apiCalls = {
...
-      client={
-        new QueryClient({
-          defaultOptions: {
-            ...queryClientOptions,
-            queries: { ...queryClientOptions.queries, retry: false },
-          },
-        })
-      }
-    >
+    <QueryClientProvider client={queryClient}>
```
jonahkagan, line 180: This change is just to use our main query client like we do in all other tests. This test never got updated because it was the first to use react-query

arsalansufi (approved):
LGTM!

PR #1480, opened 2022-05-11
Use react-query for useJurisdictions hook on progress screen

jonahkagan:
This will enable file uploads/downloads on the progress screen (coming up in the next PR) to invalidate and reload the jurisdictions status using react-query's query invalidation functionality. It's also a step towards the bigger project of moving to react-query. In this case, to keep scope small, I only migrated the necessary component and left the old hook around but marked it deprecated.

### DIFF
--- a/client/src/components/AuditAdmin/AuditAdminView.test.tsx
+++ b/client/src/components/AuditAdmin/AuditAdminView.test.tsx
@@ -1,11 +1,13 @@
 import React from 'react'
-import { waitFor, fireEvent, render, screen } from '@testing-library/react'
+import { waitFor, fireEvent, screen } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import {
   BrowserRouter as Router,
   Router as RegularRouter,
   useParams,
+  Route,
 } from 'react-router-dom'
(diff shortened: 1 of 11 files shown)


client/src/components/AuditAdmin/AuditAdminView.test.tsx
```diff
@@ -23,43 +20,40 @@ import {
...
-afterEach(() => {
-  paramsMock.mockReturnValue({
-    electionId: '1',
-    view: 'setup',
-  })
-})
+// AuditAdminView will only be rendered once the user is logged in, so
+// we simulate that.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const AuditAdminViewWithAuth = (props: any) => {
```
arsalansufi, line 38: Could we type this to indicate that we expect the provided props to be of type `RouteProps` and just cast to `any` when we pass to `AuditAdminView` (which I see [doesn't list any props](https://github.com/votingworks/arlo/blob/77785622733b49e705018ee4260566bcc4a97a14/client/src/components/AuditAdmin/AuditAdminView.tsx#L27))?

arsalansufi (approved):
Thanks for the smaller incremental PRs!

PR #1482, opened 2022-05-16
Allow JMs and AAs to generate and print batch tally sheets, take 2

arsalansufi:
_Commit-by-commit reviewing recommended_

# Overview

Issue link: https://github.com/votingworks/arlo/issues/1437

Take 2! We tried rolling this out previously (https://github.com/votingworks/arlo/pull/1475) but had to roll back (https://github.com/votingworks/arlo/pull/1481) because the new PDF library that we were piloting (@react-pdf/renderer) would've required us to relax our CSP. We're now back to using the existing jsPDF library in Arlo.

# Screenshots / screencaptures

### Jurisdiction manager UI

https://user-images.githubusercontent.com/12616928/168661966-2f7cd573-5ca8-438f-be7e-0e297f5ed294.mov

batch-tally-sheets-jm

### Audit admin UI

https://user-images.githubusercontent.com/12616928/168661971-adf35263-0e36-4a1a-b0ab-360dcf774d99.mov

batch-tally-sheets-aa

### Error case

batch-tally-sheets-error

### Sample batch tally sheets

_PDF, all sheets_

[Batch Tally Sheets.pdf](https://github.com/votingworks/arlo/files/8702803/Batch.Tally.Sheets.pdf)

_Screenshot, single sheet, standard case_

example

_Screenshot, single sheet, long text_

example-long-text

See the PDF snapshots in the PR for many more examples!

# Testing

- [x] Added automated tests at various levels (unit, page, and PDF snapshot)
- [x] Through the above, covered rendering edge cases, including long text and all possible page break points
- [x] Tested JM and AA UIs manually


client/package.json
```diff
@@ -95,6 +95,7 @@
     "history": "^4.9.0",
     "http-proxy-middleware": "^0.19.1",
     "jspdf": "^2.3.1",
+    "jspdf-autotable": "^3.5.23",
```
arsalansufi, line 98: While `jsPDF` has table support built-in, it's limited. `jspdf-autotable` is a widely used extension. (It has [no dependencies](https://github.com/simonbengtsson/jsPDF-AutoTable/blob/2022d098b27c74d666f13a96fa08b99b35658413/package.json#L68), so no need to worry about running into a CSP snag again 😅)

client/src/components/JurisdictionAdmin/generateSheets.test.tsx
```diff
@@ -276,4 +309,238 @@ describe('generateSheets', () => {
       )
     })
   })
+
+  describe('downloadBatchTallySheets', () => {
+    // Snapshots generated by a Mac Ubuntu VM can differ ever so slightly from those generated
+    // by CircleCI in a way that isn't visually perceptible, so we tolerate a 0.1% diff
+    // TODO: Get to the root cause of the diffs
```
arsalansufi, line 316: Will open a GitHub issue for this if I can't figure it out shortly!

client/src/components/JurisdictionAdmin/generateSheets.test.tsx
```diff
@@ -276,4 +309,238 @@ describe('generateSheets', () => {
...
+        mockBatches,
+        constructContestChoices(6),
+        mockJurisdiction.name
+      )
+      await expect(Buffer.from(pdf)).toMatchPdfSnapshot({
+        tolerance: diffTolerance,
+      })
+    })
+
+    it('Adds page break after table if necessary - 2', async () => {
```
arsalansufi, line 447: I've covered every page break point reasonably possible through this sequence of `Adds page break after table if necessary - #` tests

arsalansufi: Realized I could consolidate them in a loop here: https://github.com/votingworks/arlo/pull/1482/commits/d5b1a0bd8af498770053ab50ef82f6096f385c28

client/src/components/JurisdictionAdmin/generateSheets.test.tsx
jonahkagan: One idea is to selectively remove pieces of content from the PDF to see if there's a specific part that causes the diff. But also the tiny diff tolerance seems fine

jonahkagan: One idea is to selectively remove pieces of content from the PDF to see if there's a specific part that causes the diff. But also the tiny diff tolerance seems fine

client/src/components/JurisdictionAdmin/generateSheets.ts
```diff
@@ -200,10 +202,354 @@ export const downloadAuditBoardCredentials = async (
...
+      text: `Jurisdiction: ${jurisdictionName}`,
+      wrapWidth: pageContentWidth,
+      x: pageMargin,
+      y,
+      bottomMargin: pBottomMargin,
+    })
+
+    y = renderTextWrapped({
+      doc,
+      text: `Audit Board: ${batch.auditBoard ? batch.auditBoard.name : ''}`,
```
jonahkagan, line 278: `batch.auditBoard` should always be defined at this point. I'm not sure if failing loudly or failing silently is better

jonahkagan (approved):
Nice job working through the jsPDF challenges

client/src/components/JurisdictionAdmin/generateSheets.test.tsx
arsalansufi: Definitely - I have a few courses of action here that I'm going to try before I fully shift back to L&A

client/src/components/JurisdictionAdmin/generateSheets.ts
arsalansufi: Hmm yeah, though I didn't know `batch.auditBoard` would always be defined here, I've encountered other cases like this and often feel on the fence about whether to fail loudly or silently. I can raise this as a topic for an eng discussion!

client/src/components/JurisdictionAdmin/generateSheets.test.tsx
arsalansufi: https://github.com/votingworks/arlo/issues/1490

Issue #1439, opened 2022-03-29
allow audit admin to upload files and launch on behalf of a jurisdiction


benadida:
Sometimes, a jurisdiction needs help from the audit admin. Today, the audit admin sends the files to VotingWorks support, and VotingWorks support uploads those files on behalf of the jurisdiction – with full logging that this was done by VotingWorks, of course.

For the sake of states being able to run their audits with less back-and-forth with VotingWorks support, we should allow audit admins to upload files and launch an audit on behalf of a jurisdiction – and log that it happened that way.

TBD: should this be full impersonation of a jurisdiction admin, or should it be limited to initial file and starting of the audit?

PR #1483, opened 2022-05-16
Allow audit admins to upload jurisdiction files

jonahkagan:
Task: #1439 

Gives audit admins the ability to upload jurisdiction files in the jurisdiction detail modal on the progress screen.

Details of the changes are in each commit.

Demo video (seemed to be too large for Github): https://drive.google.com/file/d/1ovUio7uelbb7Or82N23IQb6RZk8IzowB/view?usp=sharing

### DIFF
--- a/client/src/components/Atoms/FileUpload.tsx
+++ b/client/src/components/Atoms/FileUpload.tsx
@@ -1,0 +1,156 @@
+import React from 'react'
+import { FileInput, Button, Colors, ProgressBar } from '@blueprintjs/core'
+import { useForm } from 'react-hook-form'
+import styled from 'styled-components'
+import StatusTag from './StatusTag'
+import { IFileUpload } from '../useFileUpload'
+import AsyncButton from './AsyncButton'
+
+const ErrorP = styled.p`
+  margin-top: 10px;
+  color: ${Colors.RED3};
+`
+
+export interface IFileUploadProps extends IFileUpload {
+  acceptFileType: 'csv' | 'zip'
+  allowMultipleFiles?: boolean
+  disabled?: boolean
+}
+
+const FileUpload = ({
+  uploadedFile,
+  uploadFiles,
+  uploadProgress,
+  deleteFile,
+  downloadFileUrl,
+  acceptFileType,
+  allowMultipleFiles = false,
+  disabled = false,
+}: IFileUploadProps) => {
+  const { register, handleSubmit, formState, watch } = useForm<{
+    files: FileList
+  }>({ mode: 'onTouched' })
+
+  if (!uploadedFile.isSuccess) return null
+
+  const { file, processing } = uploadedFile.data
+
+  if (!uploadProgress && !file) {
+    const onUpload = async ({ files }: { files: FileList }) => {
+      try {
+        await uploadFiles(Array.from(files))
+      } catch (error) {
+        // Do nothing - toasting handled by queryClient
+      }
+    }
+
+    const files = watch('files')
+    const numFiles = files ? files.length : 0
+
+    return (
+      <form onSubmit={handleSubmit(onUpload)}>
+        <p>
+          <StatusTag>No file uploaded</StatusTag>
+        </p>
+        <p>
+          <FileInput
+            inputProps={{
+              accept: `.${acceptFileType}`,
+              name: 'files',
+              multiple: allowMultipleFiles,
+              ref: register(),
+            }}
+            hasSelection={numFiles > 0}
+            text={(() => {
+              if (numFiles === 0)
+                return allowMultipleFiles
+                  ? 'Select files...'
+                  : 'Select a file...'
+              if (numFiles === 1) return files[0].name
+              return `${numFiles} files selected`
+            })()}
+            disabled={disabled || formState.isSubmitting}
+          />
+        </p>
+        <p>
+          <Button
+            type="submit"
+            intent="primary"
+            loading={formState.isSubmitting}
+            disabled={disabled || numFiles === 0}
+          >
+            {allowMultipleFiles ? 'Upload Files' : 'Upload File'}
+          </Button>
+        </p>
+      </form>
+    )
+  }
+
+  if (uploadProgress) {
+    return (
+      <form>
+        <p>
+          <StatusTag intent="warning">Uploading</StatusTag>
+        </p>
+        <ProgressBar
+          key="uploading"
+          stripes={false}
+          intent="warning"
+          value={uploadProgress}
+        />
+      </form>
+    )
+  }
+
+  /* istanbul ignore next */
+  if (!(file && processing)) {
+    throw new Error('Invalid state')
+  }
+
+  if (!processing.completedAt) {
+    return (
+      <form>
+        <p>
+          <StatusTag intent="primary">Processing</StatusTag>
+        </p>
+        {processing.workTotal && (
+          <ProgressBar
+            key="processing"
+            stripes={false}
+            intent="primary"
+            value={processing.workProgress! / processing.workTotal}
+          />
+        )}
+      </form>
+    )
+  }
+
+  const { error } = processing
+  return (
+    <form>
+      <p>
+        {error ? (
+          <StatusTag intent="danger">Upload failed</StatusTag>
+        ) : (
+          <StatusTag intent="success">Uploaded</StatusTag>
+        )}
+        <a
+          href={downloadFileUrl}
+          target="_blank"
+          rel="noopener noreferrer"
+          style={{ marginLeft: '15px' }}
+        >
+          {file.name}
+        </a>
+      </p>
+      {error && <ErrorP>{error}</ErrorP>}
+      <p>
+        <AsyncButton disabled={disabled} onClick={deleteFile}>
+          Delete File
+        </AsyncButton>
+      </p>
+    </form>
+  )
+}
+
+export default FileUpload
--- a/client/src/components/testUtilities.tsx
+++ b/client/src/components/testUtilities.tsx
@@ -207,3 +207,10 @@ export const findAndCloseToast = async (expectedContent: string) => {
     expect(screen.queryByRole('alert')).not.toBeInTheDocument()
   )
 }
+
+// Enforces the value type for an object with values of type Mock, while
+// inferring the keys. This allows for good autocompletion of mocks in tests
+// while still enforcing their type.
+// Example: const mockNumbers = mockOfType<number>()({ 'one': 1, 'two': 2 })
+export const mockOfType = <T,>() => <Mock,>(mock: { [K in keyof Mock]: T }) =>
+  mock
--- a/client/src/components/testUtilities.tsx
+++ b/client/src/components/testUtilities.tsx
@@ -211,6 +211,6 @@ export const findAndCloseToast = async (expectedContent: string) => {
 // Enforces the value type for an object with values of type Mock, while
 // inferring the keys. This allows for good autocompletion of mocks in tests
 // while still enforcing their type.
-// Example: const mockNumbers = mockOfType<number>()({ 'one': 1, 'two': 2 })
-export const mockOfType = <T,>() => <Mock,>(mock: { [K in keyof Mock]: T }) =>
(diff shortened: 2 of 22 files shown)


client/src/components/useJurisdictions.ts
```diff
@@ -132,6 +132,10 @@ export const useJurisdictions = (electionId: string, refreshId?: string) => {
         `/api/election/${electionId}/jurisdiction`
       )
       return response && response.jurisdictions
-    }
+    },
+    // When a file input dialog closes, it triggers a window focus event,
+    // which causes a refetch by default, so we turn that off.
+    // https://github.com/tannerlinsley/react-query/issues/2960
+    { refetchOnWindowFocus: false }
```
arsalansufi, line 139: Nice find! I _think_ I ran into some similar unexpected fetches while working on batch tally sheets (perhaps explained by other refetch-triggering events)

client/src/components/useFileUpload.ts
```diff
@@ -0,0 +1,202 @@
...
+  const isProcessing = (fileInfo?: IFileInfo) =>
+    fileInfo && fileInfo.processing && !fileInfo.processing.completedAt
+  return useQuery<IFileInfo, ApiError>(key, () => fetchApi(url), {
+    // When a file input dialog closes, it triggers a window focus event,
+    // which causes a refetch by default, so we turn that off.
+    // https://github.com/tannerlinsley/react-query/issues/2960
+    refetchOnWindowFocus: false,
+    refetchInterval: fileInfo => (isProcessing(fileInfo) ? 1000 : false),
+    // Once a file is finished processing or is deleted, call onFileChange
+    // (but don't call it the very first time we load the file)
```
arsalansufi, line 40: Worth surfacing the fact that `onFileChange` isn't called the first time a file is loaded in a top-level comment for the hook, since this is unlike most `onChange` handlers (which don't distinguish `empty state --> populated state` from `populated state --> diff populated state`)

This also makes me wonder whether any consumers will ever need to trigger a change on the initial file upload and whether we should plan for that future

client/src/components/useFileUpload.ts
```diff
@@ -0,0 +1,202 @@
...
+          'batch-tallies',
+        ])
+        queryClient.invalidateQueries(['jurisdictions', jurisdictionId, 'cvrs'])
+        // We need to reload the jurisdictions list with new file upload status
+        queryClient.invalidateQueries([
+          'elections',
+          electionId,
+          'jurisdictions',
+        ])
+      },
```
arsalansufi, line 118: Related to the above comment, why is it that we only need to invalidate queries on subsequent file upload and not initial file upload? Is it because the data we invalidate doesn't exist until the first file is uploaded?

server/tests/api/test_ballot_manifest.py
```diff
@@ -129,13 +146,6 @@ def test_ballot_manifest_clear(
     )
     assert json.loads(rv.data) == {"file": None, "processing": None}
 
-    jurisdiction = Jurisdiction.query.get(jurisdiction_ids[0])
-    assert jurisdiction.manifest_num_batches is None
-    assert jurisdiction.manifest_num_ballots is None
-    assert jurisdiction.batches == []
-    assert jurisdiction.manifest_file_id is None
-    assert File.query.get(file_id) is None
```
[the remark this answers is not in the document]
jonahkagan: Yeah, I think it does give sufficient confident. And also I've been trying to stick more to testing the interface, not the internals.

client/src/components/testUtilities.tsx
```diff
@@ -207,3 +207,10 @@ export const findAndCloseToast = async (expectedContent: string) => {
     expect(screen.queryByRole('alert')).not.toBeInTheDocument()
   )
 }
+
+// Enforces the value type for an object with values of type Mock, while
+// inferring the keys. This allows for good autocompletion of mocks in tests
+// while still enforcing their type.
+// Example: const mockNumbers = mockOfType<number>()({ 'one': 1, 'two': 2 })
+export const mockOfType = <T,>() => <Mock,>(mock: { [K in keyof Mock]: T }) =>
```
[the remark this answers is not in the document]
jonahkagan: Good callout, will change

client/src/components/useFileUpload.ts
```diff
@@ -0,0 +1,202 @@
...
+  options: { onFileChange?: () => void; enabled?: boolean } = {}
+) => {
+  const isFirstFetch = useRef(true)
+  const isProcessing = (fileInfo?: IFileInfo) =>
+    fileInfo && fileInfo.processing && !fileInfo.processing.completedAt
+  return useQuery<IFileInfo, ApiError>(key, () => fetchApi(url), {
+    // When a file input dialog closes, it triggers a window focus event,
+    // which causes a refetch by default, so we turn that off.
+    // https://github.com/tannerlinsley/react-query/issues/2960
+    refetchOnWindowFocus: false,
```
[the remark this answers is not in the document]
jonahkagan: Yeah, I thought about that! I think refetching on window focus could be nice for users in general (imagining if they switch tabs and come back). That being said, it's definitely a nice-to-have and Arlo's been totally fine without it for now. What do you think? I could go either way.

client/src/components/useFileUpload.ts
jonahkagan: Yeah, this is the really tricky part of this whole thing. There are two things going on here:
1. When the page initially loads, there's a list of jurisdictions loaded (including file upload status for each jurisdiction). The modal used to be populated with data from that request. However, I didn't want to poll that endpoint during file uploads to get the status for a single file, since it can be pretty expensive (though maybe this was premature optimization...). It also seemed more straightforward for the file upload component+hook to be self-contained and only think about one file. So I needed some way to reload the jurisdiction list after a file upload completed. Since file uploads are only "complete" after background processing is finished, it's not as simple as just reloading once the initial upload request completes - we need to wait until processing finishes. So that's why we invalidate the jurisdictions request on file _load_, but not on the first load (since there's no chance a file has changed on the first load).
2. In some audit types, there are two jurisdiction files to be uploaded (e.g. manifest and CVRs for ballot comparison). CVRs depend on manifests, since we need to know the batch names from the manifest in order to validate the CVR. That means if a user changes the manifest, the CVR needs to be reprocessed. We made a product decision a while back to do that automatically on manifest changes, which is nice from a user perspective and tricky technically. When a manifest file finishes processing, the server will automatically start the reprocessing of the existing CVR file. That means the frontend needs to know to start polling for the CVR file status after the manifest finishes processing. This is the other reason we use the `onFileChange` handler to invalidate CVRs (and batch tallies) on manifest file load, but not on the first load.

Confusing stuff! I wish there was a way to make it simpler, and I'm open to any suggestions.

server/tests/batch_comparison/test_batch_tallies.py
```diff
@@ -160,6 +160,55 @@ def test_batch_tallies_replace(
...
+    jurisdiction = Jurisdiction.query.get(jurisdiction_ids[0])
+    assert jurisdiction.batch_tallies_file_id is None
+    assert File.query.get(file_id) is None
+    assert jurisdiction.batch_tallies is None
+
+    set_logged_in_user(client, UserType.AUDIT_ADMIN, DEFAULT_AA_EMAIL)
+    rv = client.get(
+        f"/api/election/{election_id}/jurisdiction/{jurisdiction_ids[0]}/batch-tallies/csv"
+    )
+    assert rv.status_code == 404
```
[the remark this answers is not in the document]
jonahkagan: Oh wait, actually JAs don't have access to the download link (it's just not part of the JA UI for file uploads currently), so that's why the test is this way. I think they could definitely have access in the future if we upgrade their UI to match the new AA UI, but that's for later

server/tests/api/test_ballot_manifest.py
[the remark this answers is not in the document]
arsalansufi: > I've been trying to stick more to testing the interface, not the internal

:100:

client/src/components/Atoms/FileUpload.tsx
```diff
@@ -0,0 +1,156 @@
...
+  downloadFileUrl,
+  acceptFileType,
+  allowMultipleFiles = false,
+  disabled = false,
+}: IFileUploadProps) => {
+  const { register, handleSubmit, formState, watch } = useForm<{
+    files: FileList
+  }>({ mode: 'onTouched' })
+
+  if (!uploadedFile.isSuccess) return null
```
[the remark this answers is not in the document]
arsalansufi: Got it, that makes sense!

Of course out of scope for this PR but I do like having explicit loading state indicators to avoid a jitter as items disappear and reappear

client/src/components/useFileUpload.ts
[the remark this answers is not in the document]
arsalansufi: Good point about refetching when users switch tabs and come back!

I also don't have a strong preference here but like the idea of making `refetchOnWindowFocus: false` the default to avoid confusion when writing tests involving interactions that trigger window focus (e.g. "Why is this fetch call happening twice when I think it should be happening once? Has my code negatively impacted client-side caching?")

If we do do this, I realize some (many?) existing tests may require updating so feel free to defer to a follow-up issue!

client/src/components/useFileUpload.ts
arsalansufi: Ah! Just meant noting in a function comment so someone doesn't have to read through the implementation of the hook to understand this behavior

client/src/components/useFileUpload.ts
arsalansufi: Wow, thank you for all this context!

I don't have any immediate thoughts on how to best simplify the larger architecture, but I do still think that it's a bit unexpected from an interface perspective for `onFileChange` to only run on subsequent file changes and not the first.

One alternative may be to call the option `onUploadSuccess` and expose a param indicating whether it's the initial upload, which consumers can then use

```ts
onUploadSuccess: (isInitialUpload?: boolean) => void
```

client/src/components/useFileUpload.ts
arsalansufi: (Though this may become irrelevant given my idea below)

arsalansufi (approved):
I know we're still hashing through some longer-term questions, but code looks good and my questions have been answered! LGTM!

client/src/components/useFileUpload.ts
jonahkagan: I meant `onFileChange` to mean "when a file is either successfully uploaded, uploaded but fails processing, or is deleted" - in other words, whenever the state of the uploaded file in the database changes. That's why it doesn't run on first load, because the uploaded file didn't change. I'm pretty confident that's the behavior we want. Would making that more clear (through name or comment) help?

client/src/components/useFileUpload.ts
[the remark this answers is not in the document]
jonahkagan: I like that reasoning. I'll try making it the default and see how it goes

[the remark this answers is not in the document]
jonahkagan: Ok it was pretty easy - just had to change one test

client/src/components/useFileUpload.ts
arsalansufi: Ah I see now! The source of my confusion was a misinterpretation of the phrase "first load." I was interpreting it as the first upload when it's actually referring to loading of a previously uploaded file (if any)

Tweaking the existing explanatory comment could be helpful.

If I'm still missing something (very possible haha), happy to hop on a quick chat!

client/src/components/useFileUpload.ts
```diff
@@ -22,6 +22,14 @@ export interface IFileUpload {
...
 }
 
+/**
+ * useUploadedFile loads the current uploaded file state from the server. If the
+ * file is processing, it will refetch the file status every second. It takes an
+ * optional argument options.onFileChange, which will be called whenever the
+ * uploaded file on the server changes (i.e. when a new file finishes processing
+ * or when a file is deleted). It is not called when the file state is first
+ * loaded from the server, nor when a file is processing.
+ */
```
arsalansufi, line 32: 💯

arsalansufi (approved):
Changes LGTM!

PR #1484, opened 2022-05-17
Fix audit board confirmation modal title wrapping

arsalansufi:
🤦

_Before and after_

before after

### DIFF
--- a/client/src/components/AuditBoard/Ballot.tsx
+++ b/client/src/components/AuditBoard/Ballot.tsx
@@ -84,7 +84,7 @@ const ConfirmationModalTitle = styled.span`
 const ConfirmationModalSubTitle = styled.span`
   font-size: 16px;
   font-weight: 400;
-  whitespace: normal; /* Allow long batch names to wrap */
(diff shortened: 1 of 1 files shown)


jonahkagan (approved):
I'm surprised that stylelint didn't catch this...

arsalansufi:
> I'm surprised that stylelint didn't catch this...

I was wondering the same! As I'm looking through our CircleCI config, I'm actually beginning to wonder whether we even run stylelint at all :mag:

arsalansufi:
Okay! Looks like stylelint isn't running

I ran `yarn stylelint --config .stylelintrc-css.js src` manually in the `client` directory and surfaced a number of lint errors. After adding [`property-no-unknown`](https://stylelint.io/user-guide/rules/list/property-no-unknown/) to the list of rules, I was able to surface this issue in particular:

```
src/components/AuditBoard/Ballot.tsx
 87:3  ✖  Unexpected unknown property "whitespace"   property-no-unknown
```

Opened an issue to run stylelint as part of Arlo's linting! https://github.com/votingworks/arlo/issues/1485

PR #1486, opened 2022-05-17
Remove redundant "Aggregated" from batch/ballot retrieval list buttons

arsalansufi:
All in the PR title! We concluded that the "Aggregated" in the copy felt redundant

-----

_Before and after_

before

after

-----

Made the same change to the ballot retrieval buttons for non-batch audits

```
$ git grep "Aggregated"
$
```


jonahkagan (approved):
Thanks for circling back to this

PR #1487, opened 2022-05-17
Hybrid: validate CVR votes aren't more than contest settings votes

jonahkagan:
If the audit admin enters total vote counts for a contest choice that are less than the number of votes for that choice in the CVRs, something is wrong!

### DIFF
--- a/server/api/sample_sizes.py
+++ b/server/api/sample_sizes.py
@@ -93,6 +93,12 @@ def validate_hybrid_manifests_and_cvrs(contest: Contest):
             f" times the number of votes allowed ({contest.votes_allowed})"
         )
 
+    if any(count.non_cvr < 0 for count in vote_counts.values()):
+        raise UserError(
+            f"For contest {contest.name}, the CVRs contain more votes"
+            " than were entered in the contest settings."
+        )
+
 
 def sample_size_options(election: Election) -> Dict[str, Dict[str, SampleSizeOption]]:
     if not election.contests:
--- a/server/api/sample_sizes.py
+++ b/server/api/sample_sizes.py
@@ -93,10 +93,20 @@ def validate_hybrid_manifests_and_cvrs(contest: Contest):
             f" times the number of votes allowed ({contest.votes_allowed})"
         )
 
-    if any(count.non_cvr < 0 for count in vote_counts.values()):
+    choices_by_id = {choice.id: choice for choice in contest.choices}
+    invalid_count = next(
+        (
+            (choices_by_id[choice_id], count)
+            for choice_id, count in vote_counts.items()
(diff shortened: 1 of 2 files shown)


server/api/sample_sizes.py
```diff
@@ -93,6 +93,12 @@ def validate_hybrid_manifests_and_cvrs(contest: Contest):
             f" times the number of votes allowed ({contest.votes_allowed})"
         )
 
+    if any(count.non_cvr < 0 for count in vote_counts.values()):
```
arsalansufi, line 96: How exactly does a `count.non_cvr < 0` mean that the CVRS contain more votes than were entered in the contest settings?

server/tests/hybrid/test_hybrid.py
```diff
@@ -720,6 +720,49 @@ def test_hybrid_manifest_validation_few_non_cvr_ballots(
...
+            "name": "Contest 1",
+            "isTargeted": True,
+            "choices": [
+                {"id": str(uuid.uuid4()), "name": "Choice 1-1", "numVotes": 13},
+                {"id": str(uuid.uuid4()), "name": "Choice 1-2", "numVotes": 10},
+            ],
+            "numWinners": 1,
+            "votesAllowed": 2,
+            "jurisdictionIds": jurisdiction_ids[:2],
+        },
```
arsalansufi, line 744: What about this setup triggers the error? For example, is it the fact that 13 is greater than 2 (votes allowed)?

arsalansufi (approved):
Some understanding questions but LGTM!

server/api/sample_sizes.py
jonahkagan: In hybrid audits, we get two pieces of data from users:
1. The CVRs
2. The total votes reported for each candidate
In order to figure out the vote counts for the non-CVR ballots, we subtract the CVR votes from the total votes. Then we have the vote counts for both the CVR ballots and the non-CVR ballots, and we can perform the hybrid audit on each segment (ballot comparison for CVR ballots, ballot polling for non-CVR ballots).

The error that happened in this case was that the non-CVR vote count was negative, which is invalid and caused an assertion failure in the audit math.

server/tests/hybrid/test_hybrid.py
jonahkagan: It's that 13 is less than the total votes for Choice 1-1 in the CVR. I can try to make that a bit more clear

server/api/sample_sizes.py
arsalansufi: Got it, thanks for explaining!

Negative non-CVR count --> more CVRs than total votes ✅

PR #1489, opened 2022-05-18
Pass options through to jsPDF mock

arsalansufi:
We've been seeing console errors during tests that cover our use of jspdf-autotable.

errors

Despite the above error, our PDF tables are fitting on the page and rendering fine.

This is because we aren't always passing options through to the jsPDF mock used by tests. So while we're [using pt units when creating batch tally sheets](https://github.com/votingworks/arlo/blob/2d90eb4dc394a2f77c7607ba59925a4a55f55843/client/src/components/JurisdictionAdmin/generateSheets.ts#L217), some tests are [using the default mm units](https://github.com/votingworks/arlo/blob/2d90eb4dc394a2f77c7607ba59925a4a55f55843/client/src/components/AuditAdmin/Progress/JurisdictionDetail.test.tsx#L38). Our [table cell min width of 100](https://github.com/votingworks/arlo/blob/2d90eb4dc394a2f77c7607ba59925a4a55f55843/client/src/components/JurisdictionAdmin/generateSheets.ts#L236) fits fine on the page in pt but not always in mm (1mm = 2.83pt).

This PR accordingly makes sure that we pass options through to the jsPDF mock. The above console errors are no more!

### DIFF
--- a/client/src/components/AuditAdmin/Progress/JurisdictionDetail.test.tsx
+++ b/client/src/components/AuditAdmin/Progress/JurisdictionDetail.test.tsx
@@ -35,11 +35,10 @@ jest.mock('axios')
 const mockSavePDF = jest.fn()
 jest.mock('jspdf', () => {
   const { jsPDF } = jest.requireActual('jspdf')
-  const mockjspdf = new jsPDF({ format: 'letter' })
-  // eslint-disable-next-line func-names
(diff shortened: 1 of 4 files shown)


jonahkagan (approved):
What a simple fix! Nice find

PR #1493, opened 2022-05-19
Fix AA uploading jurisdiction files form state bug

jonahkagan:
There was a small bug where after deleting a CVR file/reuploading a new
CVR file, the latest file state wasn't showing up in the detail modal, which
was causing the CVR file type dropdown to be incorrectly
enabled/disabled.

The root cause of this was that the jurisdiction progress screen was
storing the entire jurisdiction object in state for whichever
jurisdiction's detail modal was open. When the jurisdictions list was
refetched, that stale object was sticking around. I fixed that by
switching to only storing the jurisdiction id in state.

Then, I realized that it would be snappier to not even use that
jurisdiction list data in the modal, but to instead rely entirely on the
useCVRs response (which updates quicker). This is also cleaner and
better matches the data pattern we use for the other file uploads
(manifest, tallies). So I changed that as well.


arsalansufi (approved):
LGTM!

PR #1494, opened 2022-05-19
Always use react-query mutateAsync to ensure form spinners work correctly

jonahkagan:
react-query mutations have two main methods: `mutate`, which takes an `onSuccess` callback, and `mutateAsync`, which returns a Promise. I had experimented with using both in different places where I thought appropriate, but I realized that in order to play nicely with react-hook-form, we need to always use `mutateAsync`. The reason why is that we have a pattern of using `formState.isSubmitting` to show a spinner on the submit button while the form submission is in progress. `formState.isSubmitting` is controlled by `handleSubmit(onSubmit)`, which waits for the provided `onSubmit` function to complete. When using `mutate`, `onSubmit` returns immediately, and thus `formState.isSubmitting` is immediately false, and we don't get any spinner.


client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -129,15 +129,20 @@ const BatchResultsForm = ({
...
-      { onSuccess: closeForm }
-    )
+    try {
+      await recordBatchResults.mutateAsync({
+        batchId: batch.id,
+        resultTallySheets: [tallySheet],
+      })
+    } catch (e) {
+      // Do nothing - errors toasted by queryClient
+    }
```
jonahkagan, line 144: not a huge fan of this empty catch pattern but I don't have any better ideas

arsalansufi: Out of curiosity, what would happen if we just didn't wrap in a try-catch at all? It sounds like we expect `.mutateAsync` to handle all errors internally

arsalansufi (approved):
Nice find!

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
jonahkagan: `mutateAsync` will return a rejected promise (which throws an exception when `await`ed), which won't be caught and will crash the app.

I don't love having the empty try-catch, but haven't been able to think of a better solution. Open to ideas

PR #1506, opened 2022-06-29
Don't crash worker on caught task errors

jonahkagan:
Any error that occurs during a task that we can catch and save to the
database, we should send to Sentry but not crash the worker, since it's
safe to continue working. Any error that happens in the rest of the
worker outside of the task context can still crash as before.

### DIFF
--- a/server/worker/tasks.py
+++ b/server/worker/tasks.py
@@ -118,10 +118,10 @@ def run_task(task: BackgroundTask, db_session=db_session):
 
         if isinstance(error, UserError):
             logger.info(f"TASK_USER_ERROR {log_data}")
-
-        log_data["traceback"] = str(traceback.format_tb(error.__traceback__))
-        logger.error(f"TASK_ERROR {log_data}")
-        sentry_sdk.capture_exception(error)
+        else:
+            log_data["traceback"] = str(traceback.format_tb(error.__traceback__))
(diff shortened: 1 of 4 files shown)


server/worker/worker.py
```diff
@@ -4,6 +4,10 @@
 from server.worker.tasks import run_new_tasks
 from server.sentry import configure_sentry
 
+# We have to import all of the api so that the modules load and all of the task
+# handlers get registered as background_tasks.
+from server import api  # pylint: disable=unused-import
```
jonahkagan, line 9: This was happening implicitly before through the import of `server.sentry`, which accidentally was importing all the api through a series of imports. I just happened upon this and decided to make it explicit

server/worker/tasks.py
```diff
@@ -119,11 +118,10 @@ def run_task(task: BackgroundTask, db_session=db_session) -> bool:
 
         if isinstance(error, UserError):
             logger.info(f"TASK_USER_ERROR {log_data}")
-            return True
```
arsalansufi, line 122: I think we still want a `return` here since we don't want to log user errors to sentry. And we also don't want to double log user errors as both a `TASK_USER_ERROR` and `TASK_ERROR`

arsalansufi: No prob!

arsalansufi (approved):
LGTM!

PR #1529, opened 2022-06-30
Treat CSV decoding errors as user errors

arsalansufi:
While we do our best to validate files and detect their encoding before parsing them as CSVs with that detected encoding, unusual files, e.g. XLS files mislabeled with a .csv extension, can still result in decoding errors, as we've recently seen:

error

This PR updates these errors to be user errors instead of internal errors. We're confident in the completeness of our CSV parsing and expect nearly all of these issues to be resolved by user action.

We'll still have visibility into these errors, even once they're converted to user errors, in #arlo-activity.

I augmented the default CSV parsing error message to make it easier for us to provide support in this case:

error


server/util/csv_parse.py
```diff
@@ -60,8 +60,17 @@ class CSVColumnType(NamedTuple):
 def parse_csv(file: BinaryIO, columns: List[CSVColumnType]) -> CSVDictIterator:
     validate_not_empty(file)
     text_file = decode_csv(file)
-    validate_comma_delimited(text_file)
-    csv: CSVIterator = py_csv.reader(text_file, delimiter=",")
+    csv: CSVIterator
+    try:
+        validate_comma_delimited(text_file)
+        csv = py_csv.reader(text_file, delimiter=",")
```
jonahkagan, line 66: Does wrapping just these two lines actually catch errors that happen after `validate_comma_delimited`? My guess is no, since the reader provides a streaming API. I think a test case where the file is mostly valid and then has one undecodable character later on might check whether this works or not.

If it doesn't work, you might consider wrapping this whole function in another function to catch errors, or wrapping the entire body of the function.

arsalansufi: Great point - I hadn't considered the fact that the `py_csv.reader` API is a streaming API. Will do some testing and update as needed!

arsalansufi: I tried all sorts of things to create a file that starts off valid but ends invalid, including concatenating the contents of my current test file onto the end of a valid CSV. But I kept on finding that other checks were catching this, e.g. either chardet was detecting no encoding or the bad lines were simply treated as bad CSV rows (e.g. wrong number of cells in row 14).

After struggling to create the hypothetical situation of an undecodable character later on in the file, I decided to continue to rely on a single line read but with clearer intent in the `decode_csv` helper.

I think that this will catch the overwhelming majority of realistic cases, most notably the XLS file mislabeled with a .csv extension case, and is an improvement over what we have today!

jonahkagan: Cool, thanks for looking into that. I like the new approach!

jonahkagan (approved):

PR #1551, opened 2022-07-13
Implementing tests to investigate MACRO bug

umbernhard:


server/audit_math/sampler.py
```diff
@@ -150,6 +150,7 @@ def draw_ppeb_sample(
     sample_tuples: List[Tuple[Any, Tuple[Any, Any]]] = []
 
     for batch in sample:
+        print(batch)
```
arsalansufi, line 153: Probably a remnant from earlier debugging that can now be removed

server/tests/audit_math/test_sampler.py
```diff
@@ -182,3 +183,39 @@ def test_ballot_labels():
...
+                }
+            }
+
+    with open("server/tests/audit_math/audit_data/dekalb_2022_primary_manifest.csv") as csv:
+        for row in DictReader(csv):
+            batches[("Dekalb", row["Batch Name"])][name]["ballots"] = int(row["Number of Ballots"])
+
+    sample = sampler.draw_ppeb_sample(seed, contest, sample_size, 0, batches)
+
+    assert len(sample) == sample_size
```
arsalansufi, line 220: Since `test_dekalb_primary_2022` includes all of what's here in `test_correct_sample_size` and more, we could maybe just get rid of this test in favor of `test_dekalb_primary_2022` - I don't feel strongly about this!

arsalansufi (approved):
I have pretty limited context here but code LGTM!

server/tests/audit_math/test_sampler.py
umbernhard: yeah, you're probably right about that. I'll kill it!

server/audit_math/sampler.py
umbernhard: I need a precommit hook that just greps for `print` :joy:

umbernhard:
This wound up not being needed.

PR #1552, opened 2022-07-13
Fix sampled batch results query with multiple tally sheets

jonahkagan:
The "distinct" constraint was only selecting a single tally sheet for
each batch. Instead, we need to use "group by" to aggregate and sum
across batches.


umbernhard (approved):
lgtm, pending tests!

umbernhard (approved):

Issue #1505, opened 2022-06-28
Allow uploading of ScannedBallotInformation file with Hart CVRs


jonahkagan:
Some jurisdictions using Hart have a ballot ID that is actually printed on the ballot and want to use this ID to confirm they have the correct ballot when auditing. This ID is not found in the CVR XML file for each ballot, but is in a separate file ScannedBallotInformation.csv in the UniqueIdentifier field. Each row in the ScannedBallotInformation file also has the CvrGuid, so we can link it to a corresponding XML file.

We should allow jurisdictions to optionally upload this file in addition to the zip file of XML files. If the file is uploaded, then we should use UniqueIdentifier as our imprinted ID field. If not, we should continue with our default of using the CvrGuid field from the XML file.

PR #1558, opened 2022-07-25
Update Hart CVR parsing to handle scanned ballot information CSVs

arsalansufi:
_Commit-by-commit reviewing recommended_

# Overview

Issue link: https://github.com/votingworks/arlo/issues/1505

This PR updates Hart CVR parsing to handle scanned ballot information CSVs for Washington and potentially others down the road. Lots of additional context can be found in the linked issue!

As you can see in these screencaptures, providing a scanned ballot information CSV is optional. In the first upload, we provide both a CVRs ZIP and a scanned ballot information CSV. In the second upload, we only provide a CVRs ZIP. And in the third upload, we only provide a scanned ballot information CSV, which results in an error. Both jurisdiction managers and audit admins can upload scanned ballot information CSVs.

_Jurisdiction manager UI_

https://user-images.githubusercontent.com/12616928/180813935-76a1b681-f157-42ab-a031-830041886b7f.mov

_Audit admin UI_

https://user-images.githubusercontent.com/12616928/180884235-fc252b61-7ce7-41df-98cb-fcd41e0c49f2.mov

# Testing

- [x] Added automated tests for both client-side and server-side logic, including error cases
- [x] Tested success and error flows manually

### DIFF
--- a/client/src/components/Atoms/CSVForm.tsx
+++ b/client/src/components/Atoms/CSVForm.tsx
@@ -91,6 +91,10 @@ const CSVFile = ({
               {title && <H4>{title}</H4>}
               <FormSectionDescription>
                 {description}
+                {values.cvrFileType === CvrFileType.HART &&
+                  ' ' +
+                    'For Hart, you can also provide an optional scanned ballot information CSV. ' +
+                    'If provided, the unique identifiers in the CSV will be used as imprinted IDs.'}
                 {sampleFileLink && (
                   <>
                     <br />
@@ -127,15 +131,19 @@ const CSVFile = ({
               <>
                 <FormSection>
                   <FileInput
-                    inputProps={
-                      values.cvrFileType === CvrFileType.HART
-                        ? { accept: '.zip', name: 'zip' }
-                        : {
-                            accept: '.csv',
-                            name: 'csv',
-                            multiple: values.cvrFileType === CvrFileType.ESS,
-                          }
-                    }
+                    inputProps={{
+                      // While this component is named CSVFile, it can accept zip files in the case
+                      // of Hart CVRs
+                      // TODO: Consider renaming the component and its internals accordingly
+                      accept:
+                        values.cvrFileType === CvrFileType.HART
+                          ? '.zip,.csv'
+                          : '.csv',
+                      name: 'csv',
+                      multiple:
+                        values.cvrFileType === CvrFileType.ESS ||
+                        values.cvrFileType === CvrFileType.HART,
+                    }}
                     onInputChange={e => {
                       const { files } = e.currentTarget
                       setFieldValue(
@@ -145,7 +153,12 @@ const CSVFile = ({
                     }}
                     hasSelection={!!values.csv}
                     text={(() => {
-                      if (!values.csv) return 'Select a file...'
+                      if (!values.csv) {
+                        return values.cvrFileType === CvrFileType.ESS ||
+                          values.cvrFileType === CvrFileType.HART
+                          ? 'Select files...'
+                          : 'Select a file...'
+                      }
                       if (values.csv.length === 1) return values.csv[0].name
                       return `${values.csv.length} files selected`
                     })()}
--- a/server/api/cvrs.py
+++ b/server/api/cvrs.py
@@ -221,15 +221,24 @@ def column_value(
     row_number: int,
     header_indices: Dict[str, int],
     required: bool = True,
+    file_name: str = None,
 ):
     index = header_indices.get(header)
     if index is None:
         if required:
-            raise UserError(f"Missing required column {header}")
+            raise UserError(
+                f"Missing required column {header} in {file_name}."
+                if file_name is not None
+                else f"Missing required column {header}."
+            )
         return None
     value = row[index] if index < len(row) else None
     if required and (value is None or value == ""):
-        raise UserError(f"Missing required column {header} in row {row_number}.")
+        raise UserError(
+            f"Missing required column {header} in row {row_number} in {file_name}."
+            if file_name is not None
+            else f"Missing required column {header} in row {row_number}."
+        )
     return value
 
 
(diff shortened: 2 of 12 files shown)


server/api/cvrs.py
```diff
@@ -782,20 +783,93 @@ def join_ballots_to_interpretations(
...
+        scanned_ballot_information_file: BinaryIO,
+    ) -> Dict[str, str]:
+        validate_not_empty(scanned_ballot_information_file)
+        text_file = decode_csv(scanned_ballot_information_file)
+
+        # Determine whether the scanned ballot information CSV has a format version row and skip it
+        # if so
+        has_format_version_row = False
+        first_line = text_file.readline()
+        if first_line.startswith("#FormatVersion") or "," not in first_line:
```
arsalansufi, line 826: It's very possible that every scanned ballot information CSV has this format version row, but I didn't want to make that assumption

server/tests/ballot_comparison/test_cvrs.py
```diff
@@ -1744,7 +1780,148 @@ def test_hart_cvr_upload(
...
+        },
+        {
+            "scanned_ballot_information_file_contents": "",
+            "expected_processing_status": ProcessingStatus.ERRORED,
+            "expected_processing_error": "CSV cannot be empty.",
+            "expected_processing_work_progress": 0,
+        },
+    ]
+
+    for test_case in test_cases:
```
arsalansufi, line 1861: I really like using table-driven tests as a way to consolidate redundant test logic! But I'd also ideally use some kind of subtest construct to better delineate the subtests and corresponding subtest errors, snapshot names, etc.

I decided to punt on setting up proper subtests to get this fix out in time for Washington's audits but can make a ticket to revisit

client/src/components/Atoms/CSVForm.tsx
```diff
@@ -127,15 +131,19 @@ const CSVFile = ({
...
-                        : {
-                            accept: '.csv',
-                            name: 'csv',
-                            multiple: values.cvrFileType === CvrFileType.ESS,
-                          }
-                    }
+                    inputProps={{
+                      // While this component is named CSVFile, it can accept zip files in the case
+                      // of Hart CVRs
+                      // TODO: Consider renaming the component and its internals accordingly
```
jonahkagan, line 137: I forgot until I read this comment that I actually meant to deprecate this component. I built an improved file uploading component (that's currently only in use in the AA progress screen) called FileUpload that is intended to replace this on the JM screen. Made a task to track: https://github.com/votingworks/arlo/issues/1559

That makes me think we should probably also update the file uploader on the AA progress screen to also support this additional file type for Hart.

server/api/cvrs.py
```diff
@@ -782,20 +783,93 @@ def join_ballots_to_interpretations(
...
+            validate_comma_delimited(text_file)
+        reader = csv.reader(text_file, delimiter=",")
+
+        headers = next(reader)
+        if "CvrId" not in headers:
+            raise UserError("Scanned ballot information CSV missing CvrId column.")
+        if "UniqueIdentifier" not in headers:
+            raise UserError(
+                "Scanned ballot information CSV missing UniqueIdentifier column."
+            )
```
jonahkagan, line 843: Check out how the `get_header_indices` and `column_value` functions are used elsewhere in this file. They handle a lot of the edge cases and error handling needed for parsing out CSV values. That should make some of this code a bit simpler and more consistent with the rest of the CVR parsing code.

server/api/cvrs.py
```diff
@@ -915,11 +989,16 @@ def parse_cvr_ballots() -> Iterable[CvrBallot]:
                 )
 
             db_batch = batches_by_key.get(batch_number)
+            imprinted_id = (
+                cvr_guid_to_unique_identifier_mapping[cvr_guid]
+                if cvr_guid in cvr_guid_to_unique_identifier_mapping
+                else cvr_guid
+            )
```
jonahkagan, line 996: Should we throw an error if there's a CVR that's not in the ScannedBallotInformation file? Or do you think it's better to just fall back to the cvr guid?

server/api/cvrs.py
jonahkagan: I've generally been taking the approach that since all of these files are auto-generated, they will look exactly how the sample files look unless I see an actual counterexample. That way I don't complicate the code trying to account for the endless number of variations that might exist and just focus on the ones we actually encounter. So I think it's fine to assume the format version row exists. That being said, not a big deal either way

client/src/components/Atoms/CSVForm.tsx
```diff
@@ -145,7 +153,12 @@ const CSVFile = ({
                     }}
                     hasSelection={!!values.csv}
                     text={(() => {
-                      if (!values.csv) return 'Select a file...'
+                      if (!values.csv) {
+                        return values.cvrFileType === CvrFileType.ESS ||
+                          values.cvrFileType === CvrFileType.HART
+                          ? 'Select files...'
+                          : 'Select a file...'
+                      }
```
jonahkagan, line 161: nice! surprised I let this slide in the past

server/tests/ballot_comparison/test_cvrs.py
jonahkagan: I've never used subtests before in Python - would be curious to see what that looks like if you have any examples or docs.

server/api/cvrs.py
arsalansufi: Nice! Thanks for pointing these out - https://github.com/votingworks/arlo/pull/1558/commits/d599ae4b63474ef008250428746a5057b87ad84c

server/api/cvrs.py
arsalansufi: Really good to know as a principle that these files are generated and not hand modified! Updated the code to assume that the format version row exists: https://github.com/votingworks/arlo/pull/1558/commits/72d2bbf0cee680adf9d426afa2bc169f4cc26caf

server/api/cvrs.py
arsalansufi: I went back and forth on this but landed on falling back to the CVR GUID as the safer option and the option least likely to require last-minute code changes on our end when Washington actually runs their audit next week

We can easily change this given what we learn from Washington's actual files and audit experience

server/tests/ballot_comparison/test_cvrs.py
arsalansufi: These docs cover the gist! https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests.

This note captures one of the key benefits for looped tests like the ones I've written:

> Without using a subtest, execution would stop after the first failure, and the error would be less easy to diagnose because the value of `i` wouldn’t be displayed

jonahkagan: Sweet, I wonder if that works with `pytest`

client/src/components/Atoms/CSVForm.tsx
arsalansufi: Ooh great to know! The separation of concerns with the new FileUpload feels much cleaner

Gave audit admins the ability to upload scanned ballot information CSVs as well and added a corresponding screencapture to the PR description: https://github.com/votingworks/arlo/pull/1558/commits/55750e0d781155edbb86f48f62cb13a889087a04

server/tests/ballot_comparison/test_cvrs.py
arsalansufi: Good q! There does appear to be a plugin for `pytest` subtest support: https://pypi.org/project/pytest-subtests. Opened an exploratory ticket: https://github.com/votingworks/arlo/issues/1560

arsalansufi:
@jonahkagan Ready for a follow-up look!

server/api/cvrs.py
```diff
@@ -782,20 +793,98 @@ def join_ballots_to_interpretations(
...
+                file_name="scanned ballot information CSV",
+            )
+            unique_identifier = column_value(
+                row,
+                "UniqueIdentifier",
+                row_number,
+                header_indices,
+                file_name="scanned ballot information CSV",
+            )
+            if cvr_guid != "" and unique_identifier != "":
```
jonahkagan, line 871: empty strings are handled as errors by `column_value` unless you set `required = False`

jonahkagan, line 871: empty strings are handled as errors by `column_value` unless you set `required = False`

jonahkagan (approved):
One small thing to check out but looks great overall!

server/api/cvrs.py
arsalansufi: Ah good catch!

PR #1561, opened 2022-08-09
csv_parse: Replace undecodable chararacters

jonahkagan:
If we incorrectly detect the encoding of a CSV and only find out deep into the file parsing, we should simply replace the undecodable chararacters we find rather than crashing, since we already got so far it's probably just a few characters.

### DIFF
--- a/server/util/csv_parse.py
+++ b/server/util/csv_parse.py
@@ -104,19 +104,19 @@ def decode_csv(file: IO[bytes]) -> TextIO:
         if detector.done or i > 500:
             break
     detector.close()
-    if not detector.result["encoding"]:
-        raise CSVParseError(INVALID_CSV_ERROR)
     encoding = detector.result["encoding"]
+    if not encoding:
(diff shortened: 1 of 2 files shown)


server/util/csv_parse.py
```diff
@@ -104,29 +104,32 @@ def decode_csv(file: IO[bytes]) -> TextIO:
...
             )
         ) from error
-    text_file.seek(0)
 
-    return text_file
+    # Just to be safe, if we still got the encoding wrong and there are
+    # undecodable characters deeper in the file, turn on the "replace errors"
+    # option to replace any invalid chars with a question mark so we don't crash.
+    file.seek(0)
+    return io.TextIOWrapper(file, encoding=encoding, errors="replace", newline=None)
```
arsalansufi, line 132: Ah nice - `errors="replace"` feels like a great call!

arsalansufi (approved):

jonahkagan:
Cypress tests are broken due to a Chrome download error, but gonna merge anyway and we can fix that later

PR #1563, opened 2022-08-09
Verify that Clear Ballot CVR file has contest columns

arsalansufi:
# Overview

This PR addresses an alert for a Clear Ballot CVR file that actually turned out not to be a Clear Ballot CVR file at all. More specifically, the file had no contest columns (columns beginning with `"Choice_"`), resulting in this "find" operation raising a `StopIteration` exception:

```python
first_contest_column = next(
    i for i, header in enumerate(headers) if header.startswith("Choice_")
)
```

There's a broader bucket of errors here: uploading the wrong file. We typically catch these errors downstream and surface them as user errors. Once in a while, though, a wrong file triggers a non-user error, as happened here. It's not yet clear to me what a generalized fix for this would look like. We could as a matter of habit encode more assumptions about headers for each CVR file type upfront.

# Testing

- [x] Tested manually with the file that triggered PagerDuty
- [x] Added an automated test

### DIFF
--- a/server/api/cvrs.py
+++ b/server/api/cvrs.py
@@ -249,6 +249,12 @@ def parse_clearballot_cvrs(
     cvr_file = retrieve_file(jurisdiction.cvr_file.storage_path)
     cvrs = csv_reader_for_cvr(cvr_file)
     headers = next(cvrs)
+
+    if not any(header.startswith("Choice_") for header in headers):
+        raise UserError(
+            "CVR file should have at least one column beginning with 'Choice_'"
+        )
+
     first_contest_column = next(
         i for i, header in enumerate(headers) if header.startswith("Choice_")
     )
--- a/server/tests/ballot_comparison/test_cvrs.py
+++ b/server/tests/ballot_comparison/test_cvrs.py
@@ -1085,8 +1085,11 @@ CLEARBALLOT_CVRS = """RowNumber,BoxID,BoxPosition,BallotID,PrecinctID,BallotStyl
 14,BATCH2,6,2-2-6,p,bs,ps,TABULATOR2,s,r,,,1,0,1
 """
 
-CLEARBALLOT_CVRS_MISSING_CONTEST_COLUMNS = """RowNumber,BoxID,BoxPosition,BallotID,PrecinctID,BallotStyleID,PrecinctStyleName,ScanComputerName,Status,Remade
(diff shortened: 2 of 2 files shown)


server/api/cvrs.py
```diff
@@ -249,6 +249,12 @@ def parse_clearballot_cvrs(
...
     headers = next(cvrs)
+
+    if not any(header.startswith("Choice_") for header in headers):
+        raise UserError(
+            "CVR file should have at least one column beginning with 'Choice_'"
+        )
+
     first_contest_column = next(
         i for i, header in enumerate(headers) if header.startswith("Choice_")
     )
```
jonahkagan, line 260: `next` can take a default, so another option here might be:

```python
    first_contest_column = next(
        i for i, header in enumerate(headers) if header.startswith("Choice_"),
        None
    )
    if first_contest_column is None:
       raise ...
```
The way you have it is totally fine too, just offering

server/tests/ballot_comparison/test_cvrs.py
```diff
@@ -1085,6 +1085,10 @@ def test_cvr_reprocess_after_manifest_reupload(
 14,BATCH2,6,2-2-6,p,bs,ps,TABULATOR2,s,r,,,1,0,1
 """
 
+CLEARBALLOT_CVRS_MISSING_CONTEST_COLUMNS = """RowNumber,BoxID,BoxPosition,BallotID,PrecinctID,BallotStyleID,PrecinctStyleName,ScanComputerName,Status,Remade
```
jonahkagan, line 1088: This feels like a less realistic case. I doubt an actual ClearBallot CVR file would not have any contest columns. I think a more representative test would be something more like what we saw (a diff file completely). I'm thinking more about readers of the code wondering why we're testing this

jonahkagan (approved):
Left some comments, nothing blocking though

server/tests/ballot_comparison/test_cvrs.py
arsalansufi: Great point! Let me update this test to make it more realistic

server/api/cvrs.py
arsalansufi: Nice! I don't have a strong preference either way, so I'll just keep this as is

arsalansufi:
Temporarily disabling main branch checks so that I can merge - will re-enable right after (see https://github.com/votingworks/arlo/issues/1562)

PR #1564, opened 2022-08-11
Cobble together script to sort Hart XML CVRs into two groups

arsalansufi:
Creating this PR for documentation purposes!

From Ginny in Slack:

> OK - here's the fun new voting system CVR quirk of the day. This county scanned 6 batches in Hart's "Audit mode" to use in their other type of audit and then scanned everything else in another mode (let's call it Live b/c that's what we use).  This is why the batch numbers are duplicated, even though all of these ballots were tabulated (the audit mode is not some type of rescan of ballots already tabulated, it's just there for...  no one is exactly sure, none of the other WA Hart counties use that function).  They can pull a ScannedBallotInformation file for just those "audit mode" batches... is that enough info to help us generate a workable CVR file?

The answer to that last question was yes!

I cobbled together this script to sort Hart XML CVRs into two groups. All CVRs represented in the provided ScannedBallotInformation.csv were sorted into one directory, and all CVRs _not_ represented in that provided ScannedBallotInformation.csv were sorted into another directory.

```
python -m scripts.parse-xml-cvrs /path/to/cvrs unused-param
```

I then ran the original unmodified `parse-cml-cvrs` script, treating the two groups as different tabulators (though they were really just different modes on the same tabulator).

```
python -m scripts.parse-xml-cvrs --cvrs-exported-by-tabulator /path/to/sorted/cvrs ./dominion-format.csv
```

This was all necessary because, like Hart with multiple tabulators, there were duplicate batch names across ballots scanned in the two different modes. Unfortunately, unlike Hart with multiple tabulators, where CVRs can be exported by tabulator, we can't export CVRs by mode and can only generate a ScannedBallotInformation.csv by mode, as Ginny noted above.

More context in this Slack thread: https://votingworks.slack.com/archives/C01DUGG1D8E/p1660149845606469

### DIFF
--- a/scripts/parse-xml-cvrs.py
+++ b/scripts/parse-xml-cvrs.py
@@ -2,6 +2,7 @@
 import sys
 import os
 import csv
+import shutil
 from xml.etree import ElementTree
 from collections import defaultdict
 
@@ -55,6 +56,15 @@ def parse_cvr_file(file_path, use_directory_name_as_tabulator=False):
     return cvr
 
 
+def parse_scanned_ballot_information_csv(file_path):
+    unique_identifiers = set()
+    with open(file_path) as csv_file:
+        reader = csv.reader(csv_file)
+        for row in reader:
+            unique_identifiers.add(row[15])
+    return unique_identifiers
+
+
 if __name__ == "__main__":
     if not (
         len(sys.argv) == 3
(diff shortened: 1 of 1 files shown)


arsalansufi:
As noted in the PR description, just created this for documentation purposes and in case these script modifications ever come in handy agin

PR #1569, opened 2022-08-18
Use CircleCI browser tools to install Chrome for Cypress tests

arsalansufi:
Issue link: https://github.com/votingworks/arlo/issues/1562

This PR updates Arlo CI to use CircleCI browser tools to install Chrome for Cypress tests. Arlo CI currently installs Chrome manually. This approach recently broke, and the new approach should be more robust. The new approach also mirrors what we do in vxsuite.


jonahkagan (approved):
Awesome, so much easier!

Issue #1499, opened 2022-06-21
Implement flow for rolling back from round 2


jonahkagan:
Support Users should be able to:
- Undo the round start for any round of the audit
- Reopen rounds that have completed

We sometimes need to do this manually/using scripts when an audit gets stuck in a bad state due to auditing mistakes.

arsalansufi:
Updated plan after discussion in Slack:

- Allow support users to roll back an audit to a previous round (two things happening under the hood, undoing current round start and re-opening previous round)
- Allow audit admins to reopen audit boards

Full Slack thread for context: https://votingworks.slack.com/archives/CKCVA0F9S/p1660838320351479

PR #1571, opened 2022-08-23
Add support user endpoint for reopening current round

arsalansufi:
## Overview

Another step for https://github.com/votingworks/arlo/issues/1499, this PR adds a support user endpoint for reopening the current audit round. We currently only have [a script](https://github.com/votingworks/arlo/blob/main/scripts/reopen-round.py) for this.

## Testing

- [x] Added unit tests

### DIFF
--- a/server/api/support.py
+++ b/server/api/support.py
@@ -171,8 +171,11 @@ def get_election(election_id: str):
         auditType=election.audit_type,
         online=election.online,
         jurisdictions=[
-            dict(id=jurisdiction.id, name=jurisdiction.name,)
+            dict(id=jurisdiction.id, name=jurisdiction.name)
             for jurisdiction in election.jurisdictions
+        ],
+        rounds=[
+            dict(id=round.id, ended_at=round.ended_at) for round in election.rounds
         ],
         deletedAt=isoformat(election.deleted_at),
     )
--- a/server/api/support.py
+++ b/server/api/support.py
@@ -421,7 +421,7 @@ def log_in_as_jurisdiction_admin(email: str):
     return redirect("/")
 
 
-@api.route("/support/elections/<election_id>/rounds/current/reopen", methods=["PATCH"])
+@api.route("/support/elections/<election_id>/reopen-current-round", methods=["PATCH"])
 @restrict_access_support
 def reopen_current_round(election_id: str):
     election = get_or_404(Election, election_id)
--- a/server/api/support.py
+++ b/server/api/support.py
@@ -419,3 +419,24 @@ def log_in_as_jurisdiction_admin(email: str):
         session, UserType.JURISDICTION_ADMIN, email, from_support_user=True
     )
     return redirect("/")
+
+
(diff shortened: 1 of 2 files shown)


server/api/support.py
```diff
@@ -419,3 +419,24 @@ def log_in_as_jurisdiction_admin(email: str):
         session, UserType.JURISDICTION_ADMIN, email, from_support_user=True
     )
     return redirect("/")
+
+
+@api.route("/support/elections/<election_id>/rounds/current/reopen", methods=["PATCH"])
```
jonahkagan, line 424: For the support API, I've been trying to keep the paths as shallow as possible (instead of having nested resource paths with multiple IDs in them). It's an experiment to see if that ends up being cleaner than the long paths in the main Arlo APIs.

So to follow the pattern, I'd probably make this one:
```
PATCH /support/rounds/<round_id>/reopen
```
or
```
DELETE /support/rounds/<round_id>/ended-at
```
What do you think?

server/tests/api/test_support.py
```diff
@@ -770,3 +770,55 @@ def test_support_clear_offline_results_wrong_audit_type(
             }
         ]
     }
+
+
+def test_reopen_current_round(
+    client: FlaskClient, election_id: str, contest_ids: List[str], round_1_id: str,
+):
+    def is_round_completed(round_id: str) -> bool:
```
jonahkagan, line 778: When possible, I try to test Arlo using the API interface rather than looking into the db. Is there a way that the support API will show that the round has been reopened? If so, can that be used in the test? (The support user interface will probably need some way to know this anyway)

I'm fine with keeping these assertions in addition if it's important to make sure that all the relevant data was deleted (which in this case it might be).

jonahkagan (approved):

server/api/support.py
arsalansufi: Ah interesting! I chose to use election ID in the path here because we only ever want to reopen the current round. Rather than having the client identify and provide the current round, I figured it'd be easier to have the server look up the current round ID itself. On that note, I intentionally didn't include round ID in the path

jonahkagan: That makes sense to me. What do you think about shortening the path then to something like `/elections/<election_id>/reopen-round`? I think when I saw the multiple segments `/rounds/current/reopen` it suggested to me that we would be building a nested structure where `current` could be replaced by a round id for other round-based operations.

jonahkagan: Or maybe `/elections/<election_id>/reopen-current-round`

arsalansufi: Yes shortening the path sounds good to me!

server/tests/api/test_support.py
arsalansufi: Makes a lot of sense! As far as I could tell, there wasn't already a way to retrieve round data using the support user API, so I updated the get-election endpoint. And I used that endpoint/data where I could in the round-reopening tests

https://github.com/votingworks/arlo/pull/1571/commits/baf5113e9b6acffec6559f1180f3e01b083a10bd

jonahkagan (approved):

PR #1573, opened 2022-08-25
Give audit admins the ability to reopen audit boards

arsalansufi:

## Overview

PR number 4 for https://github.com/votingworks/arlo/issues/1499 (flow for rolling back from round 2)!

This PR gives audit admins the ability to reopen audit boards, an ability analogous to their ability to unfinalize results for batch audits. It also removes this ability from the support user UI since support users can now just auth as audit admins for this functionality.

_Audit admin reopening an audit board after a support user has reopened the round_

https://user-images.githubusercontent.com/12616928/186720275-aa56c436-8c75-476b-b1be-a749adcf3e51.mov

_Audit admin view of audit boards_

audit-admin

_Support user view of audit boards_

Note how the option to reopen has been removed. We've still kept the audit board names, however, to provide context for the "Clear audit boards" action.

support-user

## Testing

Tested all of the following manually and/or with new automated tests:

- [x] Support user can reopen an audit board, when the conditions are right
- [x] Audit admin can reopen an audit board, when the conditions are right
- [x] Support user doesn't see option to reopen when audit boards are offline
- [x] Audit admin doesn't see audit boards table at all when audit boards are offline


client/src/components/AuditAdmin/Progress/JurisdictionDetail.tsx
```diff
@@ -274,34 +319,49 @@ const RoundStatusSection = ({
...
-                  // state is way high up in the component hierarchy
-                  window.location.reload()
-              }}
+              onClick={() =>
+                unfinalizeBatchResults({
+                  electionId,
+                  jurisdictionId: jurisdiction.id,
+                  roundId: round.id,
+                })
+              }
```
arsalansufi, line 334: I noticed that we don't disable the existing "Unfinalize Results" button when the current round is completed. Clicking the button surfaces an error that provides a brief explanation. I followed that same pattern for the new "Reopen" buttons:

error

Open to changing this behavior, of course! We could add a suggestion to the error message to reach out to a support user to reopen the round. Or we could disable the button and add an explanatory tooltip

client/src/components/AuditAdmin/Progress/JurisdictionDetail.tsx
```diff
@@ -274,34 +319,49 @@ const RoundStatusSection = ({
...
             </AsyncButton>
           </div>
         )
-      else return <p>Data entry complete</p>
-    if (auditBoards.length === 0)
+      }
+      return (
+        <div>
+          <p>Data entry complete</p>
+          {auditSettings.online && (
```
arsalansufi, line 345: When audit boards are offline, an AA can't take any meaningful actions on audit boards, so I opted not to display the audit boards table in this case

This differs from support users, where we still display offline audit boards, just with no reopen option. Support users have other abilities, like clearing all audit boards, so displaying even offline audit boards to them makes more sense

client/src/components/AuditAdmin/Progress/AuditBoardsTable.tsx
```diff
@@ -0,0 +1,68 @@
...
+          >
+            Reopen
+          </Button>
+        )
+      },
+    })
+  }
+  return (
+    <>
+      <Table columns={columns} data={auditBoards} />
```
jonahkagan, line 62: The `Table` component was an experiment with using the react-table library. I think my takeaway after trying it out is that the fanciness is maybe more trouble than it's worth. I can't remember exactly what made it hard, but I just remember that I kept running into annoying little problems and found code using `Table` to be tricky to maintain. Since then I've been leaning towards using the basic styled `HTMLTable` instead.

I probably should have left a comment to that effect to deprecate `Table`.

It's not a big deal either way but just wanted to clue you in on my thinking and hear your thoughts.

client/src/components/AuditAdmin/Progress/JurisdictionDetail.tsx
```diff
@@ -188,26 +189,76 @@ const CvrsFileUpload = ({
...
+    await api(
+      `/election/${electionId}/jurisdiction/${jurisdictionId}/round/${roundId}/audit-board/${auditBoardId}/sign-off`,
+      { method: 'DELETE' }
+    )
+  )
+  if (success) {
+    // Reload the whole page instead of properly refreshing the state from the server since the
+    // state is so high up in the component hierarchy
+    // TODO: Use react-query instead
+    window.location.reload()
```
jonahkagan, line 259: nice, I like moving the data refresh logic into the api request function!

client/src/components/SupportTools/SupportTools.tsx
```diff
@@ -484,28 +469,24 @@ const Jurisdiction = ({ jurisdictionId }: { jurisdictionId: string }) => {
...
+                areAuditBoardsOnline={election.online}
+                auditBoards={auditBoards}
+                reopenAuditBoard={async auditBoard => {
+                  await reopenAuditBoard.mutateAsync({
+                    auditBoardId: auditBoard.id,
+                    jurisdictionId,
+                  })
+                  toast.success(`Reopened ${auditBoard.name}`)
+                }}
+              />
```
jonahkagan, line 489: What if we just remove this feature for support users completely? In general, the approach I've been taking is to only have support features for things that users can't do themselves. Support users can always log in as a user if they need to do something on the user's behalf. Leaves us with less code, which is easier to maintain.

client/src/components/AuditAdmin/Progress/JurisdictionDetail.tsx
jonahkagan: This pattern seems good enough for now.

I think adding instructions about contacting support is a bit tricky in Arlo because VotingWorks isn't the only host for Arlo - some states host it themselves. I've never quite known what to do so I have avoided putting specific instructions in the app.l

arsalansufi: Ah yes good point about some states hosting Arlo themselves - will leave as is

client/src/components/AuditAdmin/Progress/AuditBoardsTable.tsx
arsalansufi: Very helpful context! I did find the interface hard to understand, and I generally think that compositional approaches to tables (e.g. through multiple components, `table`, `th`, `tr`, `td`, etc.) are more intuitive

client/src/components/SupportTools/SupportTools.tsx
arsalansufi: Great call! Will remove the "Reopen" buttons and the corresponding support user API endpoints. I'll leave the list of audit boards since that's still potentially useful context for the support-user specific "Clear audit boards" button

client/src/components/AuditAdmin/Progress/AuditBoardsTable.tsx
arsalansufi: https://github.com/votingworks/arlo/pull/1573/commits/34770e246afe016ab576bbb02528c02bbb88e7e1

jonahkagan (approved):

Issue #1528, opened 2022-06-29
Switch dev server to vite


jonahkagan:
- [x] Upgrade to Node 16
- [x] Eject react-scripts
- [x] Switch to Vite
- [ ] Fix skipped tests

PR #1575, opened 2022-08-29
Upgrade to Node v16

jonahkagan:
Task: #1528 

Arlo was on Node v12. Vite requires Node v14+. Node 16 is the latest LTS version.

Required small upgrades to @testing-library/react and typescript in order to get past some errors. I tried to make the smallest possible version upgrades that would fix the errors for now, since upgrading TypeScript especially will be a much bigger task. These upgrades in turn caused some new lint errors, which were mostly auto-fixable.

### DIFF
--- a/client/src/components/AuditAdmin/timers.test.tsx
+++ b/client/src/components/AuditAdmin/timers.test.tsx
@@ -52,7 +52,7 @@ const loadEach = [
   aaApiCalls.getSettings(auditSettings.all),
 ]
 
-// TODO: Fix these tests after moving to Vite
+// TODO: Fix these tests after switching to Vite
 describe.skip('timers', () => {
   let clock = FakeTimers.install()
   beforeEach(() => {
--- a/client/src/components/AuditAdmin/timers.test.tsx
+++ b/client/src/components/AuditAdmin/timers.test.tsx
@@ -52,7 +52,8 @@ const loadEach = [
   aaApiCalls.getSettings(auditSettings.all),
 ]
 
-describe('timers', () => {
+// TODO: Fix these tests after moving to Vite
+describe.skip('timers', () => {
   let clock = FakeTimers.install()
   beforeEach(() => {
     clock = FakeTimers.install()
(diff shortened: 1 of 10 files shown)


client/src/components/AuditAdmin/timers.test.tsx
```diff
@@ -52,7 +52,8 @@ const loadEach = [
   aaApiCalls.getSettings(auditSettings.all),
 ]
 
-describe('timers', () => {
+// TODO: Fix these tests after moving to Vite
+describe.skip('timers', () => {
```
jonahkagan, line 56: I got some weird errors with the fake timers. I think I will likely need to upgrade Jest, but can't do that until after ejecting react-scripts (which controls Jest version).

client/src/components/useContestsBallotComparison.ts
```diff
@@ -38,7 +38,7 @@ const useContestsBallotComparison = (
   const [contests, setContests] = useState<IContest[] | null>(null)
 
   const updateContests = async (newContests: INewContest[]) => {
-    if (putContests(electionId, newContests)) {
+    if (await putContests(electionId, newContests)) {
```
jonahkagan, line 41: this was just a bug that the typescript upgrade detected

client/src/components/SupportTools/SupportTools.tsx
```diff
@@ -234,6 +234,7 @@ const Organization = ({ organizationId }: { organizationId: string }) => {
             <input
               type="text"
               name="name"
+              id="name"
```
jonahkagan, line 237: testing library upgrade caught that this input needed an `id` in order to be associated correctly with its label

client/src/components/AuditAdmin/timers.test.tsx
```diff
@@ -52,7 +52,8 @@ const loadEach = [
   aaApiCalls.getSettings(auditSettings.all),
 ]
 
-describe('timers', () => {
+// TODO: Fix these tests after switching to Vite
```
jonahkagan, line 55: I got some weird errors with the fake timers. I think I will likely need to upgrade Jest, but can't do that until after ejecting react-scripts (which controls Jest version).

arsalansufi (approved):
Like the bug fixes we got out of the upgrades! LGTM

eventualbuddha (commented):
LGTM

PR #1577, opened 2022-08-30
Use vite for dev server and build

jonahkagan:
Task: #1528 

Updates the Arlo client app to use the vite dev server and to use vite to build the production app. 

_Review by commits_


eventualbuddha:
Looking sane so far. Let me know if you have specific questions.

app.json
```diff
@@ -2,7 +2,7 @@
   "addons": ["heroku-postgresql:hobby-free"],
   "buildpacks": [
     { "url": "heroku/nodejs" },
-    { "url": "https://github.com/moneymeets/python-poetry-buildpack.git" },
+    { "url": "https://github.com/davegaeddert/python-poetry-buildpack.git" },
```
jonahkagan, line 5: covered in #1579, just left it in for testing with a review app in Heroku

client/src/index.tsx
```diff
@@ -7,12 +7,14 @@ import { Integrations } from '@sentry/tracing'
 import './index.css'
 import App from './App'
 
-Sentry.init({
-  dsn: (window as any)._arlo_sentry_dsn, // eslint-disable-line @typescript-eslint/no-explicit-any
-  environment: (window as any)._arlo_flask_env, // eslint-disable-line @typescript-eslint/no-explicit-any
-  integrations: [new Integrations.BrowserTracing()],
-  tracesSampleRate: 0.2,
-})
+if (process.env.NODE_ENV === 'production') {
```
jonahkagan, line 10: My understanding is that `process.env.NODE_ENV` gets replaced with either `"development"` or `"production"` depending on whether you're running the vite dev server or using a vite production build. @eventualbuddha does that sound right?

jonahkagan:
Whoops, still have to resolve some Cypress issues before this is ready for review.

client/src/index.tsx
eventualbuddha: Yes, but I can't remember whether that happens with vanilla `vite`. I think it does, because we're not specifically setting that `define` config in any of the vxsuite `vite` config.

client/index.html
```diff
@@ -0,0 +1,22 @@
...
+  <head>
+    <meta charset="utf-8" />
+    <link rel="shortcut icon" href="/favicon.ico" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta name="theme-color" content="#000000" />
+    <link rel="manifest" href="/manifest.json" />
+    <title>Arlo, by VotingWorks</title>
+    <script type="text/javascript">
+      // Inject environment variables from the Flask backend by rendering this
+      // HTML page as a template.
```
jonahkagan, line 12: I simplified this logic (see the old public/index.html for what it was before). In dev, these injections won't work because index.html is rendered by vite, not Flask, but that's fine because I changed the rest of the code to only use them in production.

arsalansufi (approved):
lgtm

client/src/index.tsx
```diff
@@ -7,12 +7,14 @@ import { Integrations } from '@sentry/tracing'
...
+  Sentry.init({
+    dsn: (window as any)._arlo_sentry_dsn, // eslint-disable-line @typescript-eslint/no-explicit-any
+    environment: (window as any)._arlo_flask_env, // eslint-disable-line @typescript-eslint/no-explicit-any
```
eventualbuddha, line 13: I know these were already this way before, but you can extend the `window` interface to include additional properties a la [`window.kiosk`](https://github.com/votingworks/vxsuite/blob/d40343e36e44292768642accd8ebdaeec5d04e4b/libs/%40types/kiosk-browser/kiosk-browser.d.ts#L263).

eventualbuddha (approved):

client/src/index.tsx
jonahkagan: Oh cool, it's about time I learn how to use declaration files

PR #1578, opened 2022-08-31
Update CircleCI config

jonahkagan:
Vite requires an updated version of Node, which meant I needed to change the base image in our CircleCI config in order for vite to run in CI. I ended up making some other upgrades while I was there:
- Use new generation of CircleCI convenience images
- Use orbs to install dependencies (they have built in caching so we don't have to do it manually)
- Stop making a production build during the client-side dependency
install step (I'm not sure why we were ever doing that in the first place)


eventualbuddha (commented):
LGTM

arsalansufi (approved):

PR #1580, opened 2022-09-01
Allow AAs to reopen audit boards before all have signed off

arsalansufi:
I recently realized that https://github.com/votingworks/arlo/pull/1573, which moved the ability to reopen audit boards from support users to audit admins, resulted in a loss of functionality, namely the ability to reopen audit boards _before_ all audit boards in a jurisdiction have signed off. This PR restores that ability!

audit-boards-table


jonahkagan (approved):
Nice catch!

PR #1581, opened 2022-09-01
Upgrade to Cypress v10

jonahkagan:
Task: #1528 

In order to debug a failing Cypress test during the migration to vite, I needed to upgrade Cypress to v10 in order to run it on my ARM machine.

_Review by commits_


client/cypress/plugins.ts
```diff
@@ -0,0 +1,67 @@
+/* eslint-disable @typescript-eslint/camelcase */
```
arsalansufi, line 1: Why exactly is this needed? I'm not seeing any non camelCase variables in this file

arsalansufi (approved):
LGTM!

client/cypress/plugins.ts
jonahkagan: It's the field names on L21-23

Issue #1549, opened 2022-07-13
Write-in choices for audit boards


jonahkagan:
## Problem
Currently, Arlo doesn't have any built-in support for write-ins. Instead, we recommend that audit admins add a "write-in" contest choice when setting up contests (see [docs](https://docs.voting.works/arlo/audit-administrator/ballot-manifest/target-contests)).

This has caused a problem in ballot comparison audits specifically, since a write-in may be adjudicated as valid (counting as a vote) or invalid (counting as an undervote). If the CVR has a write-in that was invalid but the audit board selects the "write-in" choice, it creates a discrepancy.

We need a way for audit boards to be able to easily specify if a write-in was valid or invalid so that we can correctly compare the audit board's interpretation to the CVR.

 ## Proposed solution

Add an "Invalid Write-In" choice to the audit board ballot auditing screen. When this choice is selected, it should count as a vote for no candidate. If auditing a vote-for-1 contest, this is the same as a blank vote. If auditing a vote-for-n contest, the invalid write-in should simply be ignored.

This logic should apply for ballot comparison, ballot polling, and hybrid audits.

The audit report should show that the audit board selected invalid write-in.

Bonus: only show the "Invalid Write-In" choice if the audit admin added a write-in choice for the contest (e.g. if there's a choice that fuzzy matches the string "write-in" among the contest choices)

![Image](https://user-images.githubusercontent.com/530106/186234691-66d27d3f-53c7-4555-8122-f58018f7aa6a.png)

umbernhard:
What @jonahkagan's proposing is probably the best way to handle it but that may not always be the way that corresponds to a jurisdiction's adjudication rules. I think we may need to dig into that a bit more. 

Question for @ginvdr about how to handle these: I think the rules in various jurisdictions may differ on whether the presence of an invalid write-in vote would nullify another vote. I believe this may be a sticking point in Michigan, specifically?

ginvdr:
In Michigan's case (and I think they are the one with the oddest rule here), the invalid write-in would actually award the vote to another candidate in the case of an overvote and I don't think they'd use this button but the button for the other candidate.  We may need to worry about it if they ever end up doing ballot comparison but I think that's a long way off.

arsalansufi:
Didn't tackle the "bonus" piece and opened a separate issue for that (https://github.com/votingworks/arlo/issues/1589) so we can close this one out!

PR #1583, opened 2022-09-02
Add new "Invalid Write-In" option to audit board UI

arsalansufi:

## Overview

Issue link: https://github.com/votingworks/arlo/issues/1549

This PR adds a new "Invalid Write-In" option to the audit board UI. Lots of great context on why we're adding this in the linked issue. Note that, for vote-for-n contests, it's possible to select "Invalid Write-In" alongside a valid selection.

https://user-images.githubusercontent.com/12616928/188223978-2e217956-3c7f-4c6a-81ee-db613149814d.mov

## Testing

- [x] Added automated tests
- [x] Manually tested audit board UI
- [x] Manually tested report generation


server/models.py
```diff
@@ -677,6 +677,11 @@ class BallotInterpretation(BaseModel):
     # recompute this in different places.
     is_overvote = Column(Boolean, nullable=False)
 
+    # If a ballot has an invalid write-in with no other selections, the corresponding interpretation
+    # will be BLANK. If a ballot for a vote-for-n contest has an invalid write-in alongside a valid
+    # selection, the corresponding interpretation will be VOTE.
+    has_invalid_write_in = Column(Boolean, nullable=False)
```
arsalansufi, line 683: Open to feedback here! The primary alternative is a new INVALID_WRITE_IN interpretation. This would work great for vote-for-1 contests but would break down when a voter has an invalid write-in alongside a valid selection for a vote-for-n contest. In that case, we'd want the interpretation to be VOTE but with some flag indicating that there was also an invalid write-in for report-generation purposes (i.e. what I've proposed in this PR)

arsalansufi: Relatedly, I've assumed that it's useful to be able to record a valid selection alongside an invalid write-in for a vote-for-n contest. That assumption may be wrong, and we may just want audit boards to ignore the invalid write-in and only pick the valid selection for this case. If that's the case, the INVALID_WRITE_IN interpretation is probably the desirable implementation.

jonahkagan: I think this makes sense given the behavior we want. The presence of an invalid write-in seems orthogonal to the rest of the interpretation.

One thought that I had looking at the rest of the code here - there's some complexity stemming from the `VOTE` vs `BLANK` interpretation cases. I wonder if things would get simpler if we represented the blank ballot case as a `VOTE` interpretation with no selected choices?

arsalansufi: Ah interesting - I could definitely see that simplifying things. Let me look into that 🔍

arsalansufi: Prepared a proof of concept! https://github.com/votingworks/arlo/commit/367ad6947bee7a25c117399affbfef3707e7cace

My conclusion: It didn't simplify the logic much, mostly just resulted in logic getting moved around. There is something kind of nice/intuitive about `Interpretation.VOTE` always representing a selection. So I think I'm going to keep things as is.

Will work on tests and then mark this as ready for review!

jonahkagan: Sounds good! Thanks for exploring that

arsalansufi:
@jonahkagan I've added tests, and this is now ready for review!

client/src/components/AuditBoard/BallotAudit.tsx
```diff
@@ -226,14 +229,37 @@ const BallotAuditContest = ({
...
+        newInterpretation = Interpretation.VOTE
+      } else if (interpretation.hasInvalidWriteIn) {
+        newInterpretation = Interpretation.BLANK
+      } else {
+        newInterpretation = null
+      }
       setInterpretation({
         ...interpretation,
-        interpretation: choiceIds.length > 0 ? Interpretation.VOTE : null,
+        interpretation: newInterpretation,
```
jonahkagan, line 262: This logic is getting a bit complicated. I wonder if it would be simpler to store in React state a direct representation of the checkboxes and then derive an interpretation from it.

jonahkagan (commented):
Nice thorough testing!

How confident are you about the actual impact on the audit risk measurement? I didn't see any tests that tested that specifically.

client/src/components/AuditBoard/BallotAudit.tsx
arsalansufi: Good call - let me see if I can tweak 🛠️

arsalansufi:
> How confident are you about the actual impact on the audit risk measurement? I didn't see any tests that tested that specifically.

Good q! I did some manual testing to verify that behavior was reasonable, e.g. selecting "Invalid Write-In" for most ballots required a second round, selecting the winning candidate and "Invalid Write-In" (both) for most ballots resulted in the audit completing, etc.

I admittedly also relied on my knowledge of the backend implementation, with no new interpretations and just a new field that the round ending logic ignores. But for completeness, I'll add some tests there, too.

client/src/components/AuditBoard/BallotAudit.tsx
```diff
@@ -1,7 +1,6 @@
-import React from 'react'
-import { Formik, FormikProps, Field } from 'formik'
+import React, { useState } from 'react'
```
arsalansufi, line 1: Before proceeding with my state cleanup, I decided to remove `Formik` from this form to simplify things 🧹

I recommend reviewing this commit using [`?w=1`](https://github.com/votingworks/arlo/pull/1583/commits/6a556ec5ccc9822f324eaacdca86f36d66290eed?w=1)

arsalansufi:
@jonahkagan Comments addressed and ready for another look!

server/tests/api/test_rounds.py
```diff
@@ -209,6 +209,56 @@ def test_rounds_complete_audit(
...
+    client: FlaskClient, election_id: str, contest_ids: List[str], round_1_id: str,
+):
+    run_audit_round_all_blanks(
+        round_1_id, contest_ids[0], contest_ids, invalid_write_in_ratio=1
+    )
+    rv = client.get(f"/api/election/{election_id}/round")
+    rounds = json.loads(rv.data)["rounds"]
+    assert len(rounds) == 1
+    assert_is_date(rounds[0]["endedAt"])
+    assert not rounds[0]["isAuditComplete"]
```
jonahkagan, line 259: Thanks for adding these. This is great.

Hate to do this to ya but I think we also need tests for ballot comparison. You might be able to modify an existing test or two to use invalid write-ins and just verify that the outcome didn't change.

client/src/components/AuditBoard/BallotAudit.tsx
jonahkagan: Hurrah! Formik is a pain in the butt. I started using react-hook-form instead and have found that to be a lot nicer to use.

client/src/components/AuditBoard/Ballot.tsx
```diff
@@ -103,6 +106,47 @@ interface IProps {
   ) => void
 }
 
+interface IRenderedInterpretationProps {
+  interpretation: IBallotInterpretationFormRepresentation
+  contest: IContest
+}
+
+const RenderedInterpretation: React.FC<IRenderedInterpretationProps> = ({
```
jonahkagan, line 114: Nit: `Rendered` is implicit since all components are rendered. Maybe just `Interpretation`?

client/src/components/AuditBoard/BallotAudit.tsx
```diff
@@ -193,68 +212,84 @@ const BallotAudit: React.FC<IProps> = ({
...
+          isBlankVoteChecked: true,
+          isContestNotOnBallotChecked: false,
+          isInvalidWriteInChecked: false,
+        })
       } else {
-        newInterpretation = null
+        setInterpretation({
+          ...interpretation,
+          isBlankVoteChecked: false,
+        })
```
jonahkagan, line 249: I think these two branches could be collapsed. Toggling Blank Vote should completely specify the state of the checkboxes.

```ts
setInterpretation({
   ...interpretation,
   choiceIds: [],
   isBlankVoteChecked: checked,
   isContestNotOnBallotChecked: false,
   isInvalidWriteInChecked: false,
})
```

Not sure if you had them separated out on purpose.

I think the same applies for Contest Not on Ballot below

client/src/components/AuditBoard/BallotAudit.tsx
```diff
@@ -193,68 +212,84 @@ const BallotAudit: React.FC<IProps> = ({
...
+          isBlankVoteChecked: false,
+          isContestNotOnBallotChecked: false,
+          isInvalidWriteInChecked: true,
+        })
       } else {
-        newInterpretation = null
+        setInterpretation({
+          ...interpretation,
+          isInvalidWriteInChecked: false,
+        })
```
jonahkagan, line 278: In a vote-for-1, shouldn't selecting `Invalid Write-In` also clear the selected choices?

jonahkagan: maybe:

```ts
setInterpretation({
   ...interpretation,
   choices: contest.votesAllowed === 1 ? [] : interpretation.choices,
   isContestNotOnBallotChecked: false,
   isBlankVoteChecked: false,
   isInvalidWriteInChecked: checked,
})
```

client/src/components/AuditBoard/ballotInterpretation.ts
```diff
@@ -0,0 +1,65 @@
+import { IBallotInterpretation, Interpretation } from '../../types'
+
+export const INVALID_WRITE_IN = 'INVALID_WRITE_IN'
+
+/**
+ * A modified representation of a ballot interpretation, optimized for the audit board form
+ */
+export interface IBallotInterpretationFormRepresentation {
```
jonahkagan, line 8: Nit: I think `IBallotInterpretationFormState` would be a slightly more conventional name for this

client/src/components/AuditBoard/ballotInterpretation.ts
```diff
@@ -0,0 +1,65 @@
...
+    interpretation = Interpretation.CONTEST_NOT_ON_BALLOT
+  }
+  return {
+    contestId,
+    interpretation,
+    choiceIds,
+    comment,
+    hasInvalidWriteIn: isInvalidWriteInChecked,
+  }
+}
```
jonahkagan, line 65: Nice, I think this approach is clearer. How does it feel to you?

client/src/components/AuditBoard/Ballot.tsx
arsalansufi: I'd initially named this `Interpretation`, but it conflicted with the imported `Interpretation` enum:

https://github.com/votingworks/arlo/blob/86f7deabc1fda6e561964ab73186aed589b318c9/client/src/components/AuditBoard/Ballot.tsx#L13

I can maybe rename to `InterpretationSummary`

client/src/components/AuditBoard/BallotAudit.tsx
arsalansufi: I'd initially collapsed the branches exactly as you've described but then separated them out since I felt that the separation made the logic easier to grok. But I don't feel strongly either way and can re-collapse!

client/src/components/AuditBoard/ballotInterpretation.ts
arsalansufi: Also agree that it feels much clearer!

client/src/components/AuditBoard/BallotAudit.tsx
arsalansufi: Ah actually @jonahkagan I'm now remembering why I didn't add logic like this. Even for a vote-for-1 today, we allow the audit board to select multiple choices since a voter may have overvoted. Likewise, I figured that a voter could specify both a choice and an invalid write-in, even for a vote-for-1. And the audit board UI should allow audit board members to transcribe exactly what they see on a ballot. Let me know if I'm missing something!

jonahkagan (approved):

Issue #1433, opened 2022-03-23
Support user: make list of audits more manageable


mcchilders:
The current list of orgs and audits in those orgs that the support user sees (and has to scroll through) is getting really long and unwieldy. We can solve this by:

- highlighting recent/current audits at the very top
- hiding older audits automatically / behind some collapsable 
- making audits collapsible by org

Let's discuss the best way to go here.

mattroe:
From sync discussion: proposed solution to filter for active audits with some limited audit status.

arsalansufi:
- [ ] Add "Log in as" buttons that log you in as an audit admin _and_ navigate to a specific audit
- [ ] Explore pinning
- [ ] Explore search bar

Relevant Slack thread: https://votingworks.slack.com/archives/CKCVA0F9S/p1661897867743139

arsalansufi:
Another piece of context: Until very recently, Ginny had been relying on the audit admin UI (as an audit admin in nearly every active org) but just recently switched to the support user UI. So we're now collecting real feedback from Ginny on support user pain points

benadida:
My sense on this one is that if we prioritize current audits over old ones, this might solve 90+% of the problem?

mattroe:
[Latest description of the problem](https://docs.google.com/document/d/1FjXSCCCq1EV8VFr6EXcjXXIvMoX3l7ixcWx3PYk7FKw/edit?usp=sharing) courtesy of @ginvdr

mattroe:
9/6 chat:
- what's the definition of an active audit?
- sort based on last updated

PR #1584, opened 2022-09-06
Add support user button for logging in to audit directly as arbitrary audit admin

arsalansufi:
## Overview

One of many support user improvements that we're considering for https://github.com/votingworks/arlo/issues/1433, this PR adds a support user button for logging in to an audit directly as an arbitrary audit admin. (Ginny has confirmed that picking any arbitrary audit admin in the relevant organization is fine and even desirable.)

Prior to this PR, there was no login button on the individual audit screen for support users, and support users still had to select an audit after logging in as an audit admin, making a number of support user flows clunky.

log-in-as

https://user-images.githubusercontent.com/12616928/188958673-b7483ce1-250a-49c8-a94f-dc619ccedc6b.mov

## Testing

- [x] Added automated tests
- [x] Tested manually

### DIFF
--- a/client/src/components/SupportTools/SupportTools.tsx
+++ b/client/src/components/SupportTools/SupportTools.tsx
@@ -393,16 +393,23 @@ const Audit = ({ electionId }: { electionId: string }) => {
 
   if (!election.isSuccess) return null
 
-  const { auditName, auditType, jurisdictions, rounds } = election.data
+  const { id, auditName, auditType, jurisdictions, rounds } = election.data
 
   return (
     <Column>
       <H2>{auditName}</H2>
-      <Tag large style={{ marginBottom: '15px' }}>
+      <div style={{ marginBottom: '10px' }}>
+        <AnchorButton href={`/api/support/elections/${id}/login`} icon="log-in">
+          Log in as audit admin
+        </AnchorButton>
+      </div>
+      <Tag large style={{ marginBottom: '10px' }}>
         {prettyAuditType(auditType)}
       </Tag>
-      <RoundsTable electionId={electionId} rounds={rounds} />
-      <H3 style={{ marginTop: '10px' }}>Jurisdictions</H3>
+      <div style={{ marginBottom: '10px' }}>
+        <RoundsTable electionId={electionId} rounds={rounds} />
+      </div>
+      <H3>Jurisdictions</H3>
       <ButtonList>
         {jurisdictions.map(jurisdiction => (
           <LinkButton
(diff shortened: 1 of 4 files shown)


server/api/support.py
```diff
@@ -407,6 +407,24 @@ def log_in_as_jurisdiction_admin(email: str):
     return redirect("/")
 
 
+@api.route(
+    "/support/elections/<election_id>/login", methods=["GET"],
+)
+@restrict_access_support
```
arsalansufi, line 413: The other possible API was to augment the existing `/support/audit-admins/<email>/login` endpoint to take an `election_id` query param. But in that world, the client would still have to provide the email of a specific audit admin to log in as. Having the server look up an email on its own makes the endpoint much more reusable across client contexts.

client/src/components/SupportTools/SupportTools.tsx
```diff
@@ -393,16 +393,23 @@ const Audit = ({ electionId }: { electionId: string }) => {
...
   return (
     <Column>
       <H2>{auditName}</H2>
-      <Tag large style={{ marginBottom: '15px' }}>
+      <div style={{ marginBottom: '10px' }}>
+        <AnchorButton href={`/api/support/elections/${id}/login`} icon="log-in">
+          Log in as audit admin
+        </AnchorButton>
+      </div>
+      <Tag large style={{ marginBottom: '10px' }}>
```
jonahkagan, line 406: I feel like I want the audit type tag to stay next to the audit name to make it clear that it's a property of that audit.

What do you think about putting them on the same line like so?
Screen Shot 2022-09-07 at 11 00 46 AM

client/src/components/SupportTools/SupportTools.tsx
```diff
@@ -393,16 +393,23 @@ const Audit = ({ electionId }: { electionId: string }) => {
 
   if (!election.isSuccess) return null
 
-  const { auditName, auditType, jurisdictions, rounds } = election.data
+  const { id, auditName, auditType, jurisdictions, rounds } = election.data
 
   return (
     <Column>
       <H2>{auditName}</H2>
```
jonahkagan, line 400: While I was here, I noticed that the audit name wraps with the column, which doesn't look great for longer audit names. Maybe the `H2` could be outside of the `Column`?

client/src/components/SupportTools/SupportTools.tsx
jonahkagan: (I also added `intent="primary"` to the log in button, just an idea)

server/api/support.py
jonahkagan: Makes sense to me!

jonahkagan (approved):

client/src/components/SupportTools/SupportTools.tsx
arsalansufi: Nice - I like both those changes!

client/src/components/SupportTools/SupportTools.tsx
arsalansufi: Good catch - will tweak 🛠️

Issue #1440, opened 2022-03-29
Audit planner


benadida:
## Problem

Right now, in order to help states plan their audits, VotingWorks support runs audit scenarios by creating hypothetical audits and seeing where the numbers land. One of the main questions they're trying to answer is which audit type to use (since different audit types will require different amounts of work).

There are sample size calculators outside of Arlo that can be used for ballot polling and ballot comparison, but not for batch comparison.

## Proposed Solution

Create an audit planner tool to compare different audit types without having to go through the whole flow to create an audit.

## Requirements
- Takes as input candidate vote totals, total ballots cast, and number of winners 
- Estimate sample sizes for ballot polling, ballot comparison, and batch comparison
- Allow trying out different risk limits
- Only needs to estimate sample size for one contest at a time
- Sample size estimates should be shareable alongside the input data
- Doesn't require logging into Arlo (so it can be used broadly)

Sample size to generate:
- Ballot polling: 90% probability of completion (see audit_math/bravo.py)
- Ballot comparison: standard (see audit_math/supersimple.py)
- Batch comparison: estimate given by this formula: `- ln(risk limit) / diluted margin + 2`, where `risk limit` is a decimal and `diluted margin = (winner - loser) / total ballots cast` for the smallest `winner - loser` margin among all the candidate pairs.

Note that the unit for batch comparison is batches, not ballots.

ginvdr:
For this calculator, total ballots, number of winners, and candidate vote totals will be useful. It would be very helpful if it's built so that information is entered and a user can then see sample sizes for all audit types without re-entry.

jonahkagan:
### Mockups
![IMG_6854](https://user-images.githubusercontent.com/530106/186546387-140768d6-4405-42e3-97ef-1598855117f2.jpg)
![IMG_6857](https://user-images.githubusercontent.com/530106/186546401-d089a44e-14bb-42e6-a2d0-9441695e9aeb.jpg)

arsalansufi:
Follow-ups:
- [ ] Add input for number of votes allowed

arsalansufi:
Now that I've opened issues for follow-ups (above), I'm going to mark this base issue as done! We can prioritize the above as we see fit

PR #1585, opened 2022-09-06
Set up scaffolding for public Arlo pages

arsalansufi:
A first step for https://github.com/votingworks/arlo/issues/1440 (the new public audit planner), this PR sets up scaffolding for public Arlo pages.

Once we deploy this, it will be possible to navigate to https://arlo.voting.works/planner and see the incomplete audit planner. I don't believe that this is a big deal but correct me if I'm wrong - happy to introduce some sort of gating logic for prod!

https://user-images.githubusercontent.com/12616928/188966061-892f14c0-5544-4ab3-88c2-6154faa51a8b.mov

### DIFF
--- a/client/src/components/PublicPages/PublicPages.tsx
+++ b/client/src/components/PublicPages/PublicPages.tsx
@@ -1,0 +1,22 @@
+import React from 'react'
+import { Redirect, Route, Switch } from 'react-router-dom'
+
+import AuditPlanner from './AuditPlanner'
+import { Wrapper } from '../Atoms/Wrapper'
+
+const PublicPages: React.FC = () => {
+  return (
+    <Wrapper>
+      <Switch>
+        <Route exact path="/public/audit-planner">
+          <AuditPlanner />
+        </Route>
+        <Route>
+          <Redirect to="/" />
+        </Route>
+      </Switch>
+    </Wrapper>
+  )
+}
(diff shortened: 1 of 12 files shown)


client/src/components/PublicPages/PublicPages.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { Redirect, Route, Switch } from 'react-router-dom'
+
+import AuditPlanner from './AuditPlanner'
+import { Wrapper } from '../Atoms/Wrapper'
+
+const PublicPages: React.FC = () => {
+  return (
+    <Wrapper>
+      <Switch>
+        <Route exact path="/public/audit-planner">
```
arsalansufi, line 11: I'm just now noticing in the mocks a `/plan?...` URL

plan

More than happy to tweak the routing structure if we don't want `/public/` in the URL!

jonahkagan: Since this is going to be a URL that's typed out and passed around a lot, I want to optimize for that case. That's the motivation behind `arlo.voting.works/plan`. Though now I'm wondering if `arlo.voting.works/planner` might be a bit more clear. Either way, yeah, let's remove `/public/`.

jonahkagan (approved):

PR #1587, opened 2022-09-07
Use correct port for support login links in development

jonahkagan:
To ensure that login links always redirect to the frontend dev server at port 3000 (rather than backend dev server at 3001), we use the same fix we used for auth redirects when migrating to Vite.


arsalansufi (approved):
Thanks for investigating and fixing this!

Issue #1565, opened 2022-08-11
Batch Inventory flow


mattroe:
## Context
- Jxs with short certification timelines and limited reconciliation processes may not have the logistical capacity to create ballot manifests quickly and accurately from scratch.
- These jxs do have source data (CVRs) that can be used to generate preliminary precinct/batch totals to reconcile with other ballot accounting data to identify discrepancies and then more rapidly create ballot manifest files.
- Relatedly, jxs performing batch comparison audits have requested totaling batches from CVR files to expedite the process of audit preparation.

## Flow
![batch-inventory](https://user-images.githubusercontent.com/530106/188956262-e9ad4f4c-8c3a-4201-bb05-a4fe9fa0e72e.jpg)

1. Jx manager logs into Arlo
2. Jx manager uploads CVR and tabulator mapping file exported from voting system
3. Jx manager prints out worksheet ([sample worksheet](https://docs.google.com/spreadsheets/d/1GdnUIj-hUsfrW3A1HPJQdwZGtMbIVpaID9hug7lYs8A/edit#gid=0))
4. Jx manager inventories batches using worksheet
5. Jx manager signs off that batches have been inventoried
6. Jx manager downloads ballot manifest and candidate totals by batch files

PR #1590, opened 2022-09-12
Batch Inventory flow

jonahkagan:
Task: #1565 

_Review by commits_

Adds an experimental batch inventory flow accessible from the jurisdiction manager audit setup screen. Since this is a prototype, I tried to separate it as much as possible from the rest of Arlo by:
- Adding one new db table for all the data related to this flow
- Creating a minimal UI to link to the batch inventory flow from the jurisdiction manager screen (which will need to be feature flagged in some way)
- Making a separate API prefix `/elections/:id/jurisdictions/:id/batch-inventory` for all the new API endpoints

Some other notable pieces of this:
- I prototyped what components for a wizard might look like (though I didn't fully abstract them yet) - this might be a useful pattern to use in the AA audit setup flow, which is a bit messy currently
- I was really torn on whether to have the file upload interaction consist of picking two files and then uploading them together or having two separate file uploaders. I ultimately opted for the latter since it leveraged existing components and patterns, but I think there are less opportunities for things to go wrong if you upload both files together, since they are interdependent. In the future I'd like to experiment with a file upload interaction where you can select multiple files using multiple inputs and then upload them together.

https://user-images.githubusercontent.com/530106/191138173-e3d1e86b-eabb-48b1-8f51-ca0bd09e5791.mov


jonahkagan:
@arsalansufi this is ready for an initial review

server/api/batch_inventory.py
```diff
@@ -0,0 +1,466 @@
...
+    storage_path = store_file(
+        request.files["cvr"].stream,
+        f"audits/{election.id}/jurisdictions/{jurisdiction.id}/"
+        + timestamp_filename("batch-inventory-cvrs", "csv"),
+    )
+
+    batch_inventory_data = BatchInventoryData.query.get(jurisdiction.id)
+    if not batch_inventory_data:
+        batch_inventory_data = BatchInventoryData(jurisdiction_id=jurisdiction.id)
+        db_session.add(batch_inventory_data)
```
jonahkagan, line 180: One of the downsides to having a separate table is that we have to do some gymnastics to create the BatchInventoryData object only when it's needed. I thought it was worthwhile, but open to thoughts on this

client/src/components/JurisdictionAdmin/BatchInventory.tsx
```diff
@@ -0,0 +1,483 @@
...
+  right?: React.ReactElement
+}> = ({ left, right }) => (
+  <div
+    style={{
+      display: 'flex',
+      padding: '20px',
+      backgroundColor: Colors.LIGHT_GRAY2,
+      justifyContent: 'space-between',
+    }}
+  >
```
jonahkagan, line 125: @arsalansufi I'm curious if you have thoughts about using styled-components vs inline styles. I just did whatever was quickest for prototyping, but would like to decide on a strategy and clean up accordingly

arsalansufi: Yes good q! In my work on the audit planner, I've been trying to consistently use styled components instead of inline styles. There are limitations to inline styles, e.g. lack of support of `:hover`, `:active`, `:focus` selectors, lack of support for media queries, less separation of style and structure. And while these limitations don't pose problems for simple cases, styled components are guaranteed to work for all cases. So if we are to recommend one over the other as a general rule, I'd pick styled components

Of course, I do get that it can be cumbersome when you just need to add a one-off margin to a button, for example

client/src/components/JurisdictionAdmin/BatchInventory.tsx
```diff
@@ -1,11 +1,245 @@
...
+    {right ?? <div />}
+  </div>
+)
+
+const UploadElectionResultsStep: React.FC<{ nextStep: () => void }> = ({
+  nextStep,
+}) => {
+  return (
+    <StepContainer>
+      <StepPanel>
```
arsalansufi: Love the wizard concept! I like that the components are composable and leave state management to the consumer

A problem with the first iteration of the wizard component at Clever was the fact that it tried to bake state management into the component. The component interface grew really complex over time as it tried to support an ever growing number of state management cases. The second iteration of that component looked more like what you've begun hashing out in this PR

client/.stylelintrc.js
```diff
@@ -1,15 +1,15 @@
...
+  processors: ['stylelint-processor-styled-components'],
+  extends: [
+    'stylelint-config-palantir',
+    'stylelint-config-prettier',
+    'stylelint-config-styled-components',
+  ],
+  rules: {
+    'selector-max-id': 1,
+    'selector-max-universal': 1,
+    'order/properties-order': null,
```
arsalansufi, line 13: 👋 🗑️

server/models.py
```diff
@@ -926,6 +926,38 @@ class ActivityLogRecord(Base):
...
+    )
+    tabulator_status_file = relationship(
+        "File",
+        foreign_keys=[tabulator_status_file_id],
+        single_parent=True,
+        cascade="all, delete-orphan",
+    )
+
+    election_results = Column(JSON)
+    tabulator_id_to_name = Column(JSON)
```
arsalansufi, line 955: I'm always a little bit hesitant to use JSON fields instead of more structured fields. But I take it we're doing this to get something up and running quickly

server/api/batch_inventory.py
arsalansufi: I see below that `upload_batch_inventory_tabulator_status` does not have a case for `if not batch_inventory_data`. Is the assumption that the client will always call `upload_batch_inventory_cvr` first?

Wondering if it's safer to abstract this `if not batch_inventory_data` check to a helper that both call or a decorator that both use

arsalansufi (commented):
Still working my way through review but wanted to get this first batch of comments over to you!

server/models.py
jonahkagan: Yeah, I imagine this will likely change in the future, so sticking to quick and dirty for now.

server/api/batch_inventory.py
jonahkagan: Yeah, the frontend requires the files be uploaded in order. I agree that ideally the backend wouldn't assume that without enforcing it. Maybe I'll just add a quick line of validation to enforce that.

client/src/components/JurisdictionAdmin/BatchInventory.tsx
jonahkagan: Ok, that sounds good. I'll try moving in that direction.

client/src/components/JurisdictionAdmin/BatchInventory.tsx
jonahkagan: That's a great lesson!

One pattern I've been experimenting with is using hooks for state management and stateless components for UI (e.g. the `useConfirm` hook and `Confirm` component). I was inspired by react-hook-form, which only does form state management and plugs into whatever UI you want. So we could potentially also try building a `useSteps` hook to manage state if we wanted.

arsalansufi: Ooh that's an interesting possibility!

arsalansufi (commented):
The approach of keeping everything as separate as possible makes a lot of sense to me! No big picture concerns from me. Left one UX-related observation in Slack and can review again after edits given feedback from Georgia

jonahkagan:
@arsalansufi this is ready for review. Main changes: I lightened up the design a bit, added feature flagging, and added tests.

client/src/components/JurisdictionAdmin/BatchInventory.tsx
```diff
@@ -0,0 +1,502 @@
...
+    deleteFile: () => deleteFile.mutateAsync(),
+    // TODO make this actually work
+    downloadFileUrl: `${url}/file`,
+  }
+}
+
+interface ISignOffQueries {
+  status: UseQueryResult<{ signedOffAt: string | null }>
+  signOff: UseMutationResult<unknown, unknown, void, unknown>
+  undoSignOff: UseMutationResult<unknown, unknown, void, unknown>
```
arsalansufi, line 90: Very minor, but the first param to these `UseMutationResult` generics could be `void` instead of `unknown`, since the mutation functions (`() => fetchApi(url, { method: 'POST' })` and `() => fetchApi(url, { method: 'DELETE' })`) both return `Promise<void>`:

```ts
signOff: UseMutationResult<void, unknown, void, unknown>
undoSignOff: UseMutationResult<void, unknown, void, unknown>
```

client/src/components/JurisdictionAdmin/BatchInventory.tsx
```diff
@@ -0,0 +1,502 @@
...
+    onSuccess: () => queryClient.invalidateQueries(key),
+  })
+  return {
+    status,
+    signOff,
+    undoSignOff,
+  }
+}
+
+const StepContainer = styled(Card).attrs({ elevation: 1 })`
```
arsalansufi, line 116: TIL `.attrs`!

server/api/batch_inventory.py
arsalansufi: Looping back to this comment: Were you still planning to add that validation to `upload_batch_inventory_tabulator_status`?

client/src/components/JurisdictionAdmin/BatchInventory.tsx
```diff
@@ -0,0 +1,502 @@
...
+
+  if (
+    !(
+      cvrUpload.uploadedFile.isSuccess &&
+      tabulatorStatusUpload.uploadedFile.isSuccess &&
+      signOffQueries.status.isSuccess
+    )
+  ) {
+    return null
+  }
```
arsalansufi, line 493: Having a little bit of trouble understanding this condition - what state is it capturing? More specifically, under what circumstances can these `.isSuccess` booleans be false?

client/src/components/JurisdictionAdmin/JurisdictionAdminView.tsx
```diff
@@ -29,6 +39,13 @@ const JurisdictionAdminView: React.FC = () => {
     jurisdictionId: string
   }>()
 
+  const auth = useAuthDataContext()
+  assert(auth?.user?.type === 'jurisdiction_admin')
+  const jurisdiction = auth.user.jurisdictions.find(
+    j => j.id === jurisdictionId
+  )
+  const { organizationId } = jurisdiction?.election || {}
```
arsalansufi, line 47: What do you think of packaging all this feature flagging logic in some kind of `useJurisdictionAdminFeatureFlag` hook that calls `useAuthDataContext` and extracts `organizationId` internally? E.g.

```ts
const isBatchInventoryEnabled = useIsJurisdictionAdminFeatureFlagEnabled(
  'BATCH_INVENTORY',
  jurisdictionId,
)
```

(Of course, there are lots of permutations to the above)

arsalansufi (commented):
Looks great!

server/api/batch_inventory.py
jonahkagan: Oops, forgot that, thanks! 58228a3f

client/src/components/JurisdictionAdmin/JurisdictionAdminView.tsx
jonahkagan: Sure, I made a more specific hook to just this feature flag for now since it's the only one we have: cf37105b

client/src/components/JurisdictionAdmin/BatchInventory.tsx
jonahkagan: Nice, thanks! cf37105b

arsalansufi (approved):
LGTM!

client/src/components/JurisdictionAdmin/JurisdictionAdminView.tsx
```diff
@@ -38,13 +29,7 @@ const JurisdictionAdminView: React.FC = () => {
...
     jurisdictionId: string
   }>()
-
-  const auth = useAuthDataContext()
-  assert(auth?.user?.type === 'jurisdiction_admin')
-  const jurisdiction = auth.user.jurisdictions.find(
-    j => j.id === jurisdictionId
-  )
-  const { organizationId } = jurisdiction?.election || {}
+  const enableBatchInventory = useBatchInventoryFeatureFlag(jurisdictionId)
```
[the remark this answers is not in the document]
jonahkagan: Good call!

PR #1591, opened 2022-09-13
Add security warning to audit board credentials PDF

jonahkagan:
Adds a callout with a security warning at the top of each page of the audit board credentials PDF. Also uses the more recent formatting approach to lay out the PDF, since all of the line y positions were gonna have to change anyway. I tweaked some of the copy to be simpler as well.

Before
Audit Board Credentials - Los Angeles - Jonah Test (36)

After
Audit Board Credentials - Los Angeles - Jonah Test (35)


client/src/components/JurisdictionAdmin/generateSheets.ts
```diff
@@ -150,63 +166,135 @@ export const downloadAuditBoardCredentials = async (
...
-      auditBoardsWithoutBallots.push(board.name)
-    }
+    if (i > 0) doc.addPage('letter')
+
+    let y = pageMargin
+    const x = pageMargin
+    const wrapWidth = pageContentWidth
+    doc.setFont('Helvetica', 'normal').setFontSize(defaultFontSize)
+
+    doc.setFillColor('#F8EDE2') // Light yellow from Blueprint
```
arsalansufi, line 190: Was gonna suggest importing this color directly from blueprint, but I couldn't seem to find it 🤷

```ts
import { Colors } from '@blueprintjs/core'
```

colors

arsalansufi (approved):
LGTM!

client/src/components/JurisdictionAdmin/generateSheets.ts
jonahkagan: Oh yeah it's a tint I got from the web inspector. But I'll try those golds and see if any work

jonahkagan: Yeah seems like they don't offer any light tints because they use opacity to lighten them in their CSS, which we can't do with jsPDF

arsalansufi: Got it! Thanks for looking into that

PR #1593, opened 2022-09-13
Upgrade typescript-eslint to v3.0.1

jonahkagan:
This seems to be the lowest version that officially supports TS 3.9, which we upgraded to in https://github.com/votingworks/arlo/pull/1575. There was a warning about incompatibility and I was starting to see some linting issues.

Had to fix a number of lint errors as a result (mainly just adding return types to module exports).


client/src/components/useFileUpload.ts
```diff
@@ -87,6 +89,7 @@ export const useUploadFiles = (key: string[], url: string) => {
   }
 }
 
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
```
arsalansufi, line 92: Given how often we have to disable this rule and how much we rely on type inference, did you consider disabling this rule globally? Or are you thinking we'll gradually add explicit types to all these React queries?

arsalansufi (approved):
LGTM!

client/src/components/useFileUpload.ts
arsalansufi: Hmm I also don't see a strong benefit to this rule and am having a hard time thinking of errors that this rule would catch that wouldn't be caught by TS type checking in general

Given how frequently we'd have to override it, I'm okay with disabling it globally!

jonahkagan: @eventualbuddha do you know of any reason we should keep this lint rule enabled?

eventualbuddha: Personally, I'd leave this enabled and add type annotations. For me, this is primarily about documentation and improving type error locality. It's possible that it may also improve performance of type checking, but this is merely speculation on my part though you can imagine how it might be true for long chains of function calls.

What do I mean by error locality? Check out this [example](https://www.typescriptlang.org/play?#code/N4KAZgrgdgxgLgSwPZQAQHMCmcDKSC22AFglOgBQCUqoqdqATthA2rfRwIYBcqAjABoQHDgCNeAJiEcAviDkgYKAM5waqTgNSitMVDNQBeDNjyE4JMlQDc8kPdCRYiFCdwFipdAHUEFgErMrAAqAJ4ADpgAglBQSHCcLlBUvMAavFAQ+KKYDNbaGVk5eagwhdm5+jTC9ExwLGw1Iun80s0FqFJN+naKKmppmtq6VcZY7uaWPn5EgfUhEdGx8YnIyZS2MkA). In the first case, we have an error at the call site because TS infers that the return type doesn't have a `c` property:
image

In the second we have an error in the function itself and not at the call site because we've declared our intention to have a `c` property:
image

I don't feel _that_ strongly about this rule in Arlo, so take this as no more than a nudge toward keeping it. I'd push back more strongly in VxSuite mostly because I work in that repo more 😉 .

jonahkagan: Thanks @eventualbuddha, helpful to hear this perspective. I do see the benefits in terms of documentation and in helping to catch mistakes inside a function.

I'm leaning towards leaving it enabled for now and continuing to disable it for react-query related stuff as needed since I've already done the necessary work. We can keep an eye on it and revisit later!

eventualbuddha: I'm admittedly pretty new to react-query, but I haven't found typing them to be too annoying. [Here's an example from vxsuite](https://github.com/votingworks/vxsuite/blob/259b9f2e48bb4b11e914d70a51b12017a7a597ae/frontends/election-manager/src/hooks/use_write_ins_query.ts#L13):

```ts
export type WriteInsQuery = UseQueryResult<Admin.WriteInRecord[]>;

/**
 * Returns a query for all write-ins matching the given criteria.
 */
export function useWriteInsQuery({
  contestId,
  status,
  enabled = true,
}: Props = {}): WriteInsQuery {
```

PR #1595, opened 2022-09-15
Redesign FileUpload component to be more visually stable

jonahkagan:
The existing FileUpload component jumped around a lot because it changed layout during the file uploading process. This redesign uses a consistent layout across states. Now, the only time the component height changes is when showing an error message (since messages can be very long, there's no good way to avoid it).

Video of the new design

https://user-images.githubusercontent.com/530106/190516929-df168ae2-cf18-4d71-84e8-9c0b59d1fe90.mov

A sampling of states
Screen Shot 2022-09-15 at 10 50 42 AM


client/src/components/AuditAdmin/AuditAdminView.tsx
```diff
@@ -141,7 +141,7 @@ const AuditAdminView: React.FC = () => {
               <Progress
                 jurisdictions={jurisdictions}
                 auditSettings={auditSettings}
-                round={rounds[rounds.length - 1]}
+                round={rounds.length > 0 ? rounds[rounds.length - 1] : null}
```
jonahkagan, line 144: This was an unrelated issue I found where `round` was getting set to `undefined` instead of `null` before the audit started. The code that consumed it had compensated by using `!!round` as opposed to `round !== null`, but I wanted to tighten it up, since they types say that it's `null`.

arsalansufi: Nice catch!

client/src/components/Atoms/FileUpload.tsx
```diff
@@ -1,154 +1,190 @@
...
+const buttonAndTagWidth = '115px' // Wide enough for the longest text
+
+const FileStatusTag = styled(StatusTag)`
+  width: ${buttonAndTagWidth};
+  text-align: center;
+`
+
+const FileStatusTagWithProgress = styled(StatusTagWithProgress)`
+  width: ${buttonAndTagWidth};
+  text-align: center;
```
arsalansufi, line 28: Minor: This `text-align: center;` could be applied universally within `StatusTag`, since I imagine almost any case that wants to increase the width will want the text centered

client/src/components/Atoms/StatusTag.tsx
```diff
@@ -1,9 +1,36 @@
...
+`
+
+export const StatusTagWithProgress: React.FC<{
+  progress: number
+} & Omit<ITagProps, 'minimal'>> = ({ progress, children, ...props }) => (
+  <StatusTagWithProgressBar {...props}>
+    {children}
+    <ProgressBar stripes={false} value={progress} intent={props.intent} />
+  </StatusTagWithProgressBar>
+)
```
arsalansufi, line 34: I imagine in most cases where we'll use a `StatusTagWithProgress`, we'll swap it out with a plain `StatusTag` once the relevant work is done and progress has hit 100%. Rather than exporting two components that consumers have to toggle, would it make sense to export a single `StatusTag` component with a `progress: number | undefined` prop that renders no progress bar when `progress` is `undefined`?

client/src/components/Atoms/FileUpload.tsx
```diff
@@ -1,154 +1,190 @@
...
 `
 
 export interface IFileUploadProps extends IFileUpload {
+  title: string
   acceptFileTypes: ('csv' | 'zip')[]
   allowMultipleFiles?: boolean
-  disabled?: boolean
+  uploadDisabled?: boolean
+  deleteDisabled?: boolean
```
arsalansufi, line 36: Nice! So customizable 🧐

arsalansufi (approved):
Some minor thoughts but otherwise LGTM! 🎨

client/src/components/Atoms/FileUpload.tsx
arsalansufi: Though on second thought I also see the benefit of applying it here when we actually need it since its purpose may not be clear if added to the original base-width `StatusTag`

client/src/components/Atoms/StatusTag.tsx
jonahkagan: I briefly explored that. I can't remember exactly why I abandoned it and decided to stick with separate components. I'll give it another quick look.

client/src/components/Atoms/FileUpload.tsx
jonahkagan: I think I agree with both perspectives haha. I'll probably just leave it unless you have a strong leaning

client/src/components/Atoms/StatusTag.tsx
jonahkagan: Ok it was a bit hairy but I pulled it off: 8e215f59

PR #1596, opened 2022-09-15
Polish audit planner election results card

arsalansufi:
## Overview

Some simple UI polish that I decided to move into its own PR since it's unrelated to the other PR I'm working on to complete the second half of the audit planner UI! Specifically decreases the size of the input labels in the election results card, moves the candidate remove buttons such that the empty space on the right of the card is no more, and tweaks the uneditable state.

Reviewing with [`?w=1`](https://github.com/votingworks/arlo/pull/1596/files?w=1) recommended

https://user-images.githubusercontent.com/12616928/190547570-2999902e-cc41-4a27-b373-49c843adc7e2.mov

Regarding the smaller input labels, I did look ahead to the "Audit Plan" card and verified that the smaller labels will look okay more generally.

## Testing

- [x] Updated automated tests
- [x] Tested manually

### DIFF
--- a/client/src/components/PublicPages/AuditPlanner/ElectionResultsCard.tsx
+++ b/client/src/components/PublicPages/AuditPlanner/ElectionResultsCard.tsx
@@ -5,6 +5,7 @@ import {
   Button,
   Card,
   Classes,
+  Colors,
   FormGroup,
   H2,
   HTMLTable,
@@ -36,7 +37,6 @@ const CandidatesTable = styled(HTMLTable)`
   table-layout: fixed;
 
   &.bp3-html-table th {
-    font-size: 14px;
     padding: 8px;
   }
 
@@ -67,7 +67,7 @@ const CandidatesTable = styled(HTMLTable)`
     width: 100%;
   }
   &.bp3-html-table .bp3-input[readonly] {
-    background: transparent;
+    background: ${Colors.LIGHT_GRAY5};
     box-shadow: none;
   }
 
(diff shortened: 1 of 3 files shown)


client/src/components/PublicPages/AuditPlanner/ElectionResultsCard.tsx
```diff
@@ -189,52 +192,52 @@ const ElectionResultsCard: React.FC<IProps> = ({
...
+                            ) {
+                              return 'At least 1 candidate must have greater than 0 votes'
+                            }
+                            return true
+                          },
+                          valueAsNumber: true,
+                        })}
+                        type="number"
+                      />
+                      {editable && (
```
jonahkagan, line 228: I like this new arrangement!

I feel like we're still losing the structure in readonly mode because we lose the shape of the input fields. What do you think of something like this?

Screen Shot 2022-09-15 at 6 13 13 PM

This is a fill of `LIGHT_GRAY5`. We could also experiment with a stroke. I also played with removing the intent from the remove buttons to differentiate their state.

client/src/components/PublicPages/AuditPlanner/ElectionResultsCard.tsx
```diff
@@ -36,7 +36,7 @@ const CandidatesTable = styled(HTMLTable)`
   table-layout: fixed;
 
   &.bp3-html-table th {
-    font-size: 18px;
+    font-size: 14px;
```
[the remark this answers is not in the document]
arsalansufi: Good catch!

client/src/components/PublicPages/AuditPlanner/ElectionResultsCard.tsx
arsalansufi: Nice, I like the screenshot! Incorporated ✅

jonahkagan (approved):

PR #1600, opened 2022-09-28
Implement audit planner backend

arsalansufi:
_Reviewing by commit recommended_

## Overview

Issue link: https://github.com/votingworks/arlo/issues/1440

This PR implements the audit planner backend! Specifically the public sample size computation endpoint, stubbed in previous PRs. There are still a few interesting design/computation edge cases to discuss, which we can do in the context of this PR.

https://user-images.githubusercontent.com/12616928/192893246-243405b7-8b29-4694-bff3-bee91b513e7a.mov

## Testing

- [x] Added automated tests
- [x] Tested manually


server/api/public.py
```diff
@@ -42,15 +47,106 @@
...
+    if total_ballots_cast < candidate_votes_sum:
+        raise Conflict("Total ballots cast cannot be less than sum of candidate votes")
+
+    contest = Contest(
+        choices=[
+            ContestChoice(
+                id=f"candidate-{i}",
+                num_votes=candidate["votes"]
+                + 0.01  # Safeguard against math errors involving 0s
+                if candidate["votes"] == 0
```
arsalansufi, line 95: Happy to discuss this `+ 0.01`! The primary edge case that this is protecting against is the case that one candidate has 0 votes and the number of total ballots cast is higher than the winning candidate's votes, e.g.

```
Helga Hippo: 10 votes
Bobby Bear: 0 votes
Total ballots cast: 11 votes <-- Important that this is greater than Helga Hippo's votes
```

The ballot polling computation crashes with inputs like this because `s_w` is 1 so `z_l` is negative infinity.

```python
        else:
            z_w = (2 * s_w).ln()
            z_l = (2 - 2 * s_w).ln()
    
            T = Decimal(min(get_test_statistics(contest.margins, sample_results).values()))
    
            weighted_alpha = (Decimal(1.0) / alpha) / T
            return int(
                (
>                   (weighted_alpha.ln() + (z_w / Decimal(2))) / (p_w * z_w + p_l * z_l)
                ).quantize(Decimal(1), rounding=ROUND_CEILING)
            )
E           decimal.InvalidOperation: [<class 'decimal.InvalidOperation'>]

server/audit_math/bravo.py:73: InvalidOperation
```

When the number of total ballots cast exactly matches the winning candidate's votes, there's no issue because the landslide short-circuit kicks in:

https://github.com/votingworks/arlo/blob/e4b637ee5f96ed2cce7d266ec4f4792ee9434e65/server/audit_math/bravo.py#L358-L364

While I could change the ballot polling logic itself, I've been approaching this task conservatively and trying to avoid updates to the audit math directly.

server/api/public.py
```diff
@@ -42,15 +47,106 @@
...
+        contest
+    )
+    return contest_for_sampler
+
+
+def is_full_hand_tally_required(
+    contest: sampler_contest.Contest, risk_limit_percentage: int
+) -> bool:
+    is_tie = contest.diluted_margin == 0
+    return is_tie or risk_limit_percentage == 0
```
arsalansufi, line 114: Both these situations (specifically `risk_limit_percentage == 0`) can trigger zero-related errors (e.g. ZeroDivisionErrors), so I special-case and short-circuit

server/api/public.py
```diff
@@ -42,15 +47,106 @@
...
+    contest: sampler_contest.Contest, risk_limit_percentage: int
+) -> int:
+    if is_full_hand_tally_required(contest, risk_limit_percentage):
+        return contest.ballots
+
+    sample_size = int(
+        math.ceil(-1 * math.log(risk_limit_percentage / 100) / contest.diluted_margin)
+        + 2
+    )
+    return min(sample_size, contest.ballots)
```
arsalansufi, line 152: Interesting question here: What should the max possible number of batches for a batch comparison audit be? For audit methods where the unit is ballots, it's clear that the max should be all ballots. For batches, we don't actually know how many total batches there are, and I'm not sure what the average number of ballots per batch is.

There's also a general design question here: What should the UI display when we want to recommend a full hand tally?
- The total number of ballots/batches? (What this PR currently does)
- "All ballots" / "All batches"?
- "Full hand tally"?

client/src/components/PublicPages/AuditPlanner/AuditPlanner.test.tsx
```diff
@@ -251,16 +251,16 @@ test('Entering election results - validation and submit', async () => {
     await areExpectedErrorMessagesDisplayed({
       displayed: ['Required'],
     })
-    userEvent.type(numberOfWinnersInput, '3')
+    userEvent.type(numberOfWinnersInput, '2')
     await areExpectedErrorMessagesDisplayed({
-      displayed: ['Cannot be greater than number of candidates'],
+      displayed: ['Must be less than number of candidates'],
```
umbernhard, line 256: Do we have plans to support vote-for-n contests? This will have to change for that, fwiw.

server/api/public.py
umbernhard: Could you add a comment explaining that?

server/api/public.py
umbernhard: I think I like "Full hand tally" the most, but would love @ginvdr's thoughts.

server/api/public.py
umbernhard: this is actually a really interesting corner case, nice find! We should probably update the Arlo backend to handle it (tho I defer to @jonahkagan)

umbernhard (approved):
This is great. There are a few nitpicks here and there, but overall it's looking really good.

client/src/components/PublicPages/AuditPlanner/AuditPlanner.test.tsx
arsalansufi: Good q! So we can currently support vote-for-n contests, e.g. 3 candidates, 2 winners (in the UI, there's an option to add candidates - you aren't restricted to 2).

I only changed the logic to prevent the same number of winners as candidates, e.g. 3 candidates, 3 winners, to avoid hitting this error in audit math:

https://github.com/votingworks/arlo/blob/e4b637ee5f96ed2cce7d266ec4f4792ee9434e65/server/audit_math/bravo.py#L356

umbernhard: ah this is my bad for forgetting how things work! I don't mean n-winners, I mean vote-for-n (i.e. there may only be one winner, but you can vote for 3 candidates or whatever). I _think_ it should just work as long as we aren't checking that (number of total votes cast) <= (number of ballots) (basically you have to do a multiplicity thing where you actually check (number of votes cast) <= n*(number of ballots), where n is the total number of votes.

arsalansufi: Ah I see, thanks for clarifying! This is really interesting. We actually do check that the total number of votes is less than the total ballots cast:

error

To perform the proper multiplied check, we'd need to add an input to the audit planner for number of votes allowed. I know @jonahkagan's original mocks didn't include an input for that, likely to keep things simple. Any thoughts on revisiting the inclusion/exclusion of that input, Jonah?

server/api/public.py
```diff
@@ -42,15 +47,108 @@
...
+                num_votes=candidate["votes"]
+                + 0.01  # Safeguard against math errors involving 0s
+                if candidate["votes"] == 0
+                else candidate["votes"],
+            )
+            for i, candidate in enumerate(candidates)
+        ],
+        num_winners=num_winners,
+        total_ballots_cast=total_ballots_cast,
+        votes_allowed=num_winners,
```
arsalansufi, line 102: Related to the discussion about number of votes allowed (https://github.com/votingworks/arlo/pull/1600#discussion_r982681006), because we don't expose an input for this, I'm just setting number of votes allowed to number of winners.

I've done a quick spot check to verify that the audit math doesn't actually need number of votes allowed.

arsalansufi: Dropped a question in Slack for greater visibility: https://votingworks.slack.com/archives/CKCVA0F9S/p1664387582424249

client/src/components/PublicPages/AuditPlanner/AuditPlanner.test.tsx
arsalansufi: Related to this, because we don't expose an input for number of votes allowed, I'm just setting number of votes allowed to number of winners here:

https://github.com/votingworks/arlo/blob/c6774fd7ca8a3f6fad2cae56e5749048b87e6991/server/api/public.py#L102

I've done a quick spot check to verify that sample size calculations don't actually need number of votes allowed.

arsalansufi: Opened a discussion thread in Slack since there are a few moving pieces here: https://votingworks.slack.com/archives/CKCVA0F9S/p1664387582424249

server/api/public.py
```diff
@@ -42,15 +47,106 @@
...
+        raise Conflict("Number of winners must be less than number of candidates")
+
+    candidate_votes_sum = sum(candidate["votes"] for candidate in candidates)
+    if total_ballots_cast < candidate_votes_sum:
+        raise Conflict("Total ballots cast cannot be less than sum of candidate votes")
+
+    contest = Contest(
+        choices=[
+            ContestChoice(
+                id=f"candidate-{i}",
```
jonahkagan, line 92: Should we require unique candidate names and use `candidate["name"]` here?

server/api/public.py
jonahkagan: First off, I think the audit planner needs to have a way to handle computation errors, because as we've learned, it's very hard to catch every audit math edge case ahead of time. @arsalansufi I think you've already got this in place, but just checking.

Given that we've caught this one (nice job @arsalansufi!), I think we should try to fix it, but as a separate bug fix to the actual math. I don't think the audit planner should defend against it separately, because if it's a problem in the planner it's a problem in real audits as well.

server/api/public.py
jonahkagan: I think the audit math should handle this (same argument as above - this could impact real audits)

server/api/public.py
jonahkagan: I lean towards "Full hand tally" as well

server/api/public.py
```diff
@@ -42,15 +47,106 @@
...
+        "batchComparison": compute_batch_comparison_sample_size,
     }
-    for risk_limit_percentage in list(range(0, 21)):
-        sample_sizes["ballotComparison"][str(risk_limit_percentage)] = 0
-        sample_sizes["ballotPolling"][str(risk_limit_percentage)] = 0
-        sample_sizes["batchComparison"][str(risk_limit_percentage)] = 0
+    for audit_type in audit_types:
+        for risk_limit_percentage in list(range(0, 21)):
+            sample_size_function = sample_size_functions[audit_type]
+            sample_sizes[audit_type][str(risk_limit_percentage)] = sample_size_function(
```
jonahkagan, line 62: Why strings for risk limit percentages? I don't have a strong opinion, just would have expected numbers

client/src/components/PublicPages/AuditPlanner/AuditPlanner.test.tsx
arsalansufi: Short-term: Decided to remove the remove requirement that `total ballots cast >= total number of votes` https://github.com/votingworks/arlo/pull/1600/commits/42a9b20d7257a6d451f0cefc6255316497a30222

Long-term: We'll likely want to add an input for the number of votes allowed. Took note of this in https://github.com/votingworks/arlo/issues/1440#issuecomment-1261352006. Will look into this after this PR!

server/api/public.py
arsalansufi: > I think the audit planner needs to have a way to handle computation errors

Confirming that the UI does handle unexpected errors!

image

> I think we should try to fix it, but as a separate bug fix to the actual math. I don't think the audit planner should defend against it separately, because if it's a problem in the planner it's a problem in real audits as well.

This line of reasoning makes perfect sense to me. To avoid scope creep on this PR, I've opened a follow-up bug task. I'm leaving the audit-planner-specific fix in place for now since, as a public page, I do think it's more likely to receive funky inputs and create PagerDuty noise for us. The follow-up ticket notes that we should remove the audit-planner-specific fix once the root fix is in place.

https://github.com/votingworks/arlo/issues/1601

server/api/public.py
arsalansufi: Good point! The ballot polling and ballot comparison audit math was actually already handling the tie case. I pushed the logic for the zero risk limit, though, into those functions. (Makes sense that they weren't covering those since you can't select a risk limit of 0 in our standard UI.)

https://github.com/votingworks/arlo/pull/1600/commits/a741eb5c784f3662e27d790a718abb3c3a0bcbf5

server/api/public.py
arsalansufi: Yeah good q! Even had I used numbers here, when calling `jsonify`, the numerical keys would've automatically been converted to string keys anyway (since JSON keys have to be strings).

So I decided to be explicit in my conversion to strings instead of implicit. The latter could result in other engineers digging to figure out what would happen to the numerical keys

jonahkagan: Ah, makes sense

server/api/public.py
arsalansufi: Voilà! https://github.com/votingworks/arlo/pull/1600/commits/7e1c9503bcc1613d5fc0e94a5c1a91760272b770

full-hand-tally

server/api/public.py
arsalansufi: I thought about this, but didn't see a huge benefit. I could (maybe) see a world where someone accidentally repeats candidate names in the original form, entering votes in like batches, e.g.

```
Helga Hippo | 10
Helga Hippo | 22
...
```

So this would prevent that. It's an easy enough change, so I'll make it.

arsalansufi: https://github.com/votingworks/arlo/pull/1600/commits/dc8030f03a5e188a156af600c9abc47534f8ee2f

jonahkagan (approved):
Looks great!

Issue #1605, opened 2022-09-30
Limit size of numbers provided to audit planner


arsalansufi:
There's a whole bucket of errors surfaced by suuuuper large numbers, e.g. 1e30.

https://sentry.io/organizations/votingworks/issues/3632355750/?project=5504628&referrer=slack

Let's set some reasonable (larger than the population of the world lol) limits to make it harder for users to crash the underlying sample size math.

PR #1607, opened 2022-09-30
Limit size of numbers provided to audit planner

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/arlo/issues/1605

There's a natural human impulse when provided a calculation tool to enter the largest numbers possible 😆

lol

As much as I feel this impulse too, these super large numbers can crash our underlying sample size math. Rather than playing whack-a-mole with all the places that large numbers can break our sample size math, this PR limits the size of numbers provided to the audit planner. I've settled on a limit of `1e15` (1 quadrillion), which should, as noted in the code:

> Leave enough buffer to support an election of galactic scale while making it hard for users to crash the sample size math by holding down the 0 key :D

limit

## Testing

- [x] Added tests for client-side logic
- [x] Added tests for server-side logic

### DIFF
--- a/client/src/components/PublicPages/AuditPlanner/AuditPlanner.test.tsx
+++ b/client/src/components/PublicPages/AuditPlanner/AuditPlanner.test.tsx
@@ -254,10 +254,16 @@ test('Entering election results - validation and submit', async () => {
       notDisplayed: ['At least 1 candidate must have greater than 0 votes'],
     })
     userEvent.clear(candidate0VotesInput)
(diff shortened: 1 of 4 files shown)


benadida (approved):
I can live with galactic scale.

eventualbuddha:
I guess we better hope for low turnout in the next [Asimov galactic election](https://en.wikipedia.org/wiki/Galactic_Empire_(Asimov)). 500 quintillion residents!

PR #1608, opened 2022-09-30
Make audit planner mobile friendly and add utils for responsive design

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/arlo/issues/1614

This PR makes the audit planner mobile friendly in the most basic way possible (still leaving plenty of room for improvement but at least ensuring that the page isn't broken on mobile).

It also adds utils for responsive design that'll hopefully come in handy with upcoming work to make the audit board screens mobile friendly / responsive! I'm excited about these! [`window.matchMedia`](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia), which the utils use under the hood, is magic 🪄 and basically brings the power of CSS media queries to JS.

### Before and after on iPhone SE width

before after

### Responsiveness in action

https://user-images.githubusercontent.com/12616928/193356028-7a6b61fd-c452-45ed-92bf-7c81ea0bef9e.mov

## Testing

- [x] Tested manually (see above)

### DIFF
--- a/client/src/components/PublicPages/AuditPlanner/AuditPlanCard.tsx
+++ b/client/src/components/PublicPages/AuditPlanner/AuditPlanCard.tsx
@@ -6,7 +6,10 @@ import SampleSize from './SampleSize'
 import SegmentedControl from '../../Atoms/SegmentedControl'
 import { AuditType } from '../../useAuditSettings'
 import { IElectionResults } from './electionResults'
+import { useCssBreakpoints } from '../../../utils/responsiveness'
 import { useSampleSizes } from './sampleSizes'
+
+const HIDDEN_LABEL_CLASS_NAME = 'hidden-label'
 
 interface IContainerProps {
   disabled?: boolean
@@ -17,9 +20,10 @@ const Container = styled(Card)<IContainerProps>`
     margin: auto;
     margin-top: 24px;
     margin-bottom: 72px;
+    max-width: 640px;
     opacity: ${props => (props.disabled ? '0.5' : '1')}
     padding: 0;
-    width: 640px;
+    width: 100%;
   }
 `
 
@@ -48,6 +52,15 @@ const SubHeading = styled(H3)`
   }
 `
 
+const RiskLimitSlider = styled(Slider)`
+  &.bp3-slider .${HIDDEN_LABEL_CLASS_NAME} {
+    color: transparent;
+  }
+  &.bp3-slider .bp3-slider-handle .${HIDDEN_LABEL_CLASS_NAME} {
+    color: white;
+  }
+`
+
 const SampleSizeSection = styled.div`
   background: #f3f8ff; // A custom tint of Blueprint v4 @blue5
   border-bottom-left-radius: 3px; // Match Blueprint card
@@ -71,6 +84,7 @@ const AuditPlanCard: React.FC<IProps> = ({ disabled, electionResults }) => {
     }
   }, [])
 
+  const { isMobileWidth } = useCssBreakpoints()
   const [selectedAuditType, setSelectedAuditType] = useState<
     Exclude<AuditType, 'HYBRID'>
   >('BALLOT_POLLING')
@@ -105,23 +119,24 @@ const AuditPlanCard: React.FC<IProps> = ({ disabled, electionResults }) => {
               { label: 'Batch Comparison', value: 'BATCH_COMPARISON' },
             ]}
             value={selectedAuditType}
+            vertical={isMobileWidth}
           />
         </Section>
 
         <Section>
           <SubHeading>Risk Limit</SubHeading>
-          <Slider
+          <RiskLimitSlider
             disabled={disabled}
             labelRenderer={value => (
               <span
                 // A hack to display the listed values on both the slider axis and the slider label,
                 // and all other values on only the slider label. More recent versions of the
                 // Blueprint slider actually support this differentiation
-                style={{
-                  color: [0, 5, 10, 15, 20].includes(value)
-                    ? undefined
-                    : 'white',
-                }}
+                className={
+                  ![0, 5, 10, 15, 20].includes(value)
+                    ? HIDDEN_LABEL_CLASS_NAME
+                    : undefined
+                }
               >
                 {value}%
               </span>
--- a/client/src/components/PublicPages/AuditPlanner/ElectionResultsCard.tsx
+++ b/client/src/components/PublicPages/AuditPlanner/ElectionResultsCard.tsx
@@ -20,6 +20,7 @@ import {
   ICandidateFormState,
   IElectionResultsFormState,
 } from './electionResults'
+import { useCssBreakpoints } from '../../../utils/responsiveness'
 
 const HIDDEN_INPUT_CLASS_NAME = 'hidden-input'
 
@@ -224,7 +225,7 @@ const ElectionResultsCard: React.FC<IProps> = ({
   planAudit,
 }) => {
   const { confirm, confirmProps } = useConfirm()
-
+  const { isMobileWidth, isTabletWidth, isDesktopWidth } = useCssBreakpoints()
   const {
     control,
     formState,
@@ -382,11 +383,12 @@ const ElectionResultsCard: React.FC<IProps> = ({
             <tr>
               <td>
                 <Button
+                  aria-label={isMobileWidth ? 'Add Candidate' : undefined}
                   disabled={!editable}
                   icon="plus"
                   onClick={() => addCandidate(constructNewCandidate())}
                 >
-                  Add Candidate
+                  {(isTabletWidth || isDesktopWidth) && 'Add Candidate'}
                 </Button>
               </td>
               <td />
--- a/client/src/utils/responsiveness.ts
+++ b/client/src/utils/responsiveness.ts
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react'
+import { useEffect, useState } from 'react'
 
 /**
  * Given a CSS media query, returns whether the media query is matched, e.g.
@@ -8,7 +8,7 @@ import { useState, useEffect } from 'react'
  * Listens for updates, e.g. in response to window resizing.
  */
 export function useMediaQuery(query: string): boolean {
-  const [isMatched, setIsMatched] = useState(false)
+  const [isMatched, setIsMatched] = useState(window.matchMedia(query).matches)
 
   useEffect(() => {
     const matchMedia = window.matchMedia(query)
--- a/client/src/utils/responsiveness.ts
+++ b/client/src/utils/responsiveness.ts
@@ -1,0 +1,54 @@
+import { useState, useEffect } from 'react'
+
+/**
+ * Given a CSS media query, returns whether the media query is matched, e.g.
+ *
+ * const isDesktopWidth = useMediaQuery('(min-width: 64em)')
+ *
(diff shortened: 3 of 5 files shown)


client/src/components/PublicPages/AuditPlanner/ElectionResultsCard.tsx
```diff
@@ -382,11 +383,12 @@ const ElectionResultsCard: React.FC<IProps> = ({
             <tr>
               <td>
                 <Button
+                  aria-label={isMobileWidth ? 'Add Candidate' : undefined}
                   disabled={!editable}
                   icon="plus"
                   onClick={() => addCandidate(constructNewCandidate())}
                 >
-                  Add Candidate
+                  {(isTabletWidth || isDesktopWidth) && 'Add Candidate'}
```
arsalansufi, line 391: As noted in the PR description, there's still room for improvement, e.g. preventing form elements from hopping around when error strings exceed two lines, labels wrapping on really small screens, etc. But this PR covers the basics

client/src/utils/responsiveness.ts
```diff
@@ -0,0 +1,51 @@
...
+    return () => {
+      matchMedia.removeEventListener('change', handleChange)
+    }
+  }, [query])
+
+  return isMatched
+}
+
+export const BREAKPOINT_MOBILE = '40em'
+export const BREAKPOINT_DESKTOP = '64em'
```
arsalansufi, line 29: Oh the art of picking breakpoints. I consulted a few articles (there are hundreds) and ultimately took inspiration from Clever's components library:

[`@breakpointS: 40em;`](https://github.com/Clever/components/blob/454320efdc04e432e1aaa0f38656ba702fa77084/src/less/layout.less#L6-L10)
[`@breakpointL: 64em;`](https://github.com/Clever/components/blob/454320efdc04e432e1aaa0f38656ba702fa77084/src/less/layout.less#L26-L30)

client/src/utils/responsiveness.ts
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * const isMobileWidth = useMediaQuery("(max-width: 48em)")
+ *
+ * Listens for updates, e.g. in response to window resizing.
+ */
+export function useMediaQuery(query: string): boolean {
+  const [isMatched, setIsMatched] = useState(false)
+
+  useEffect(() => {
+    const matchMedia = window.matchMedia(query)
```
arsalansufi, line 14: CSS media queries can take you far, but sometimes you need to control a prop/styles in a third-party library or tweak something more complex that can't be accomplished by hiding/unhiding elements.

In the past, when I've inevitably come across this need to access the screen width in JS, I've used `window.innerWidth` plus a debounced window resize listener. This time around, I discovered [`window.matchMedia`](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia), which as noted in the PR description, is magic 🪄 and basically brings the power of CSS media queries to JS!

Relevant article: https://webdevetc.com/blog/matchmedia-events-for-window-resizes

client/src/utils/responsiveness.ts
arsalansufi, line 29: Oh the art of picking breakpoints. I consulted a few articles (there are hundreds) and ultimately took inspiration from Clever's design system!

[`@breakpointS: 40em;`](https://github.com/Clever/components/blob/454320efdc04e432e1aaa0f38656ba702fa77084/src/less/layout.less#L6-L10)
[`@breakpointL: 64em;`](https://github.com/Clever/components/blob/454320efdc04e432e1aaa0f38656ba702fa77084/src/less/layout.less#L26-L30)

client/src/components/PublicPages/AuditPlanner/ElectionResultsCard.tsx
arsalansufi, line 391: As noted in the PR description, there's still room for improvement, e.g. preventing form elements from hopping around when error strings exceed two lines, labels wrapping on really small screens, etc. But this PR covers the basics

client/src/components/PublicPages/AuditPlanner/AuditPlanCard.tsx
```diff
@@ -114,18 +125,18 @@ const AuditPlanCard: React.FC<IProps> = ({ disabled, electionResults }) => {
...
-                style={{
-                  color: [0, 5, 10, 15, 20].includes(value)
-                    ? undefined
-                    : 'transparent',
-                }}
+                className={
+                  ![0, 5, 10, 15, 20].includes(value)
+                    ? HIDDEN_LABEL_CLASS_NAME
+                    : undefined
+                }
```
arsalansufi, line 139: Here's the issue I'm fixing here (note the streaks of white across the axis labels):

issue

arsalansufi, line 139: Here's the issue I'm fixing here (note the streaks of white across the axis labels):

issue

client/src/utils/responsiveness.ts
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * const isDesktopWidth = useMediaQuery('(min-width: 64em)')
+ *
+ * Listens for updates, e.g. in response to window resizing.
+ */
+export function useMediaQuery(query: string): boolean {
+  const [isMatched, setIsMatched] = useState(false)
+
+  useEffect(() => {
+    const matchMedia = window.matchMedia(query)
```
arsalansufi, line 14: CSS media queries can take you far, but sometimes you need to control a prop/styles in a third-party library or tweak something more complex that can't be accomplished by hiding/unhiding elements.

In the past, when I've inevitably come across this need to access the screen width in JS, I've used `window.innerWidth` plus a debounced window resize listener. This time around, I discovered [`window.matchMedia`](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia), which as noted in the PR description, is magic 🪄 and basically brings the power of CSS media queries to JS! It's supported in all modern browsers (and non-modern browsers, too, like IE lol).

Relevant article: https://webdevetc.com/blog/matchmedia-events-for-window-resizes

client/src/utils/responsiveness.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }, [query])
+
+  return isMatched
+}
+
+/** A breakpoint for mobile vs. tablet */
+export const BREAKPOINT_M = '40em'
+/** A breakpoint for tablet vs. desktop */
+export const BREAKPOINT_L = '64em'
```
arsalansufi, line 31: Oh the art of picking breakpoints. I consulted a few articles (there are hundreds). Here's one that I took inspiration from: https://polypane.app/blog/the-breakpoints-we-tested-in-2021-and-the-ones-to-test-in-2022

client/src/components/PublicPages/AuditPlanner/AuditPlanCard.tsx
```diff
@@ -105,23 +119,24 @@ const AuditPlanCard: React.FC<IProps> = ({ disabled, electionResults }) => {
...
-                style={{
-                  color: [0, 5, 10, 15, 20].includes(value)
-                    ? undefined
-                    : 'white',
-                }}
+                className={
+                  ![0, 5, 10, 15, 20].includes(value)
+                    ? HIDDEN_LABEL_CLASS_NAME
+                    : undefined
+                }
```
arsalansufi, line 139: Here's the issue that I'm fixing here (note the streaks of white across the axis labels):

issue

client/src/components/PublicPages/AuditPlanner/ElectionResultsCard.tsx
```diff
@@ -382,11 +383,12 @@ const ElectionResultsCard: React.FC<IProps> = ({
...
                 </Button>
```
arsalansufi, line 392: As noted in the PR description, there's still room for improvement. Remaining issues include:

- Form elements hopping around when error strings exceed two lines
- Labels wrapping on really small screens
- The "Clear" confirmation modal not being full screen

But this PR covers the basics

jonahkagan (approved):
Looks great! It's pretty cool how little you needed to change to make this work on mobile. I think we got lucky. Next time we should remember to plan for mobile from the get-go - lesson learned.

client/src/utils/responsiveness.ts
arsalansufi: Yes good q! This article walks through the reasoning for using `em` over `rem` or `px`: https://zellwk.com/blog/media-query-units/ (some of it involves the quirks/limitations of Safari in particular)

client/src/utils/responsiveness.ts
```diff
@@ -0,0 +1,54 @@
...
+ * const isDesktopWidth = useMediaQuery('(min-width: 64em)')
+ *
+ * Listens for updates, e.g. in response to window resizing.
+ */
+export function useMediaQuery(query: string): boolean {
+  const [isMatched, setIsMatched] = useState(false)
+
+  useEffect(() => {
+    const matchMedia = window.matchMedia(query)
+    setIsMatched(matchMedia.matches)
```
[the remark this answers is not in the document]
arsalansufi: To the first q, this is an unlikely scenario but say you have the following code:

```ts
const useAlternateLayout = useMediaQuery(useLowerThreshold
  ? '(min-width: 400px)'
  : '(min-width: 800px)'
)
```

If `useLowerThreshold` were toggled from `false` to `true`, the query provided to `useMediaQuery` would change. And if the "initial" value were set outside the `useEffect`, the hook wouldn't consider the updated query until the screen was (unrelatedly) resized.

We could additionally set the initial value in the `useState` invocation, but we'd still need the initial value in the `useEffect` for the case outlined above.

To the second q, agreed! Will update

client/src/components/PublicPages/AuditPlanner/ElectionResultsCard.tsx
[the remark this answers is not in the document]
arsalansufi: Ah yeah, so when you keep the text in the button, it affects the layout of the table and makes the second column smaller than the first, causing the "Total Ballots Cast" label to wrap while "Number of Winners" label doesn't. That then throws off the alignment of the inputs. Fixable by other means for sure! But removing the button text felt like a simple and reasonable fix

issue

client/src/utils/responsiveness.ts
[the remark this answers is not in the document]
jonahkagan: Gotcha, that makes sense!

client/src/components/PublicPages/AuditPlanner/ElectionResultsCard.tsx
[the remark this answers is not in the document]
jonahkagan: Gotcha. Makes me think that having this whole form be in one table isn't the right solution. But I think your workaround is good for now

client/src/utils/responsiveness.ts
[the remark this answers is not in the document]
arsalansufi: Actually thought about the second suggestion some more and settled on something a bit different to ensure that the proper values are set before the DOM is painted.

The operations performed inside the effect are asynchronous (updating state values, setting up listeners), so changing from `useEffect` to `useLayoutEffect` wouldn't make a difference. I instead updated the `useMediaQuery` hook to set its initial state value via a call to `window.matchMedia` since the first values returned by the `useMediaQuery` hook do affect the first paint.

https://github.com/votingworks/arlo/pull/1608/commits/f798919b7878c5509b2945ff05ba81bbf9da93bc

PR #1618, opened 2022-10-06
Add placeholder candidate names to audit planner

arsalansufi:
## Overview

This PR adds placeholder candidate names (Candidate 1, Candidate 2, etc.) to the audit planner so that users don't have to provide them if they don't want to.

https://user-images.githubusercontent.com/12616928/194336861-e123e140-3df7-404d-8f18-59c5aaf6391f.mov

## Testing

- [x] Updated automated tests
- [x] Tested manually


client/src/components/PublicPages/AuditPlanner/ElectionResultsCard.tsx
```diff
@@ -294,11 +294,12 @@ const ElectionResultsCard: React.FC<IProps> = ({
                     intent={errors.candidates?.[i]?.name && 'danger'}
                   >
                     <input
-                      aria-label={`Candidate ${i} Name`}
+                      aria-label={`Candidate ${i + 1} Name`}
                       className={classnames(
                         Classes.INPUT,
                         errors.candidates?.[i]?.name && Classes.INTENT_DANGER
                       )}
+                      defaultValue={`Candidate ${i + 1}`}
```
jonahkagan, line 302: Is there an easy way to auto-select the existing text on first focus so that it's easy to replace the existing candidate name?

jonahkagan (approved):

client/src/components/PublicPages/AuditPlanner/ElectionResultsCard.tsx
arsalansufi: I don't believe `autoFocus` will do the trick since it doesn't affect text selection. Also, the linter nudges away from it since it has negative accessibility implications:

lint

I found a change that works (https://github.com/votingworks/arlo/pull/1618/commits/a2e6c3c8eb02bb0accae467c52e526ce6ad4bd06), but it's unfortunately triggering `act` wrapping warnings in tests 🤔:

```
Warning: An update to ElectionResultsCard inside a test was not wrapped in act(...).
    
    When testing, code that causes React state updates should be wrapped into act(...):
```

arsalansufi: Okay! After some digging, I've found that using `watch` or `formState.touched` from `useForm` causes `act` warnings to surface.

For `formState.touched`, literally just the act of destructuring `touched` from `formState`, even if you don't use it anywhere, has this effect, e.g.

```ts
const { formState } = useForm(...)
const { touched } = formState
```

mind-blown

Turns out we've been getting these same warnings for the `BatchRoundDataEntry` form:

batch-round-data-entry

Others seem to have encountered this as well: https://github.com/react-hook-form/react-hook-form/discussions/3108.

I'm going to open a separate issue to resolve these errors for all affected Arlo forms 🛠️

arsalansufi: Also in case it wasn't clear, despite the above, I'm keeping my fix in place for the auto-selection on first focus. I updated the screen capture in the PR description to show this behavior

Issue #1601, opened 2022-09-28
BRAVO ballot polling sample size calculation can throw InvalidOperation error when candidate has 0 votes


arsalansufi:
When one candidate has 0 votes and the number of total ballots cast is higher than the winning candidate's votes, e.g.

```
Helga Hippo: 10 votes
Bobby Bear: 0 votes
Total ballots cast: 11 votes <-- Important that this is greater than Helga Hippo's votes
```

The ballot polling computation crashes. This seems to be because `s_w` is 1 so `z_l` is negative infinity.

```python
        else:
            z_w = (2 * s_w).ln()
            z_l = (2 - 2 * s_w).ln()
    
            T = Decimal(min(get_test_statistics(contest.margins, sample_results).values()))
    
            weighted_alpha = (Decimal(1.0) / alpha) / T
            return int(
                (
>                   (weighted_alpha.ln() + (z_w / Decimal(2))) / (p_w * z_w + p_l * z_l)
                ).quantize(Decimal(1), rounding=ROUND_CEILING)
            )
E           decimal.InvalidOperation: [<class 'decimal.InvalidOperation'>]

server/audit_math/bravo.py:73: InvalidOperation
```

When the number of total ballots cast exactly matches the winning candidate's votes, there's no issue because the landslide short-circuit kicks in:

https://github.com/votingworks/arlo/blob/e4b637ee5f96ed2cce7d266ec4f4792ee9434e65/server/audit_math/bravo.py#L358-L364

The public sample size calculation endpoint contains a one-off patch to avoid this error:

https://github.com/votingworks/arlo/blob/42a9b20d7257a6d451f0cefc6255316497a30222/server/api/public.py#L89-L90

Once we implement the root fix in the ballot polling sample size logic, we should remove that one-off patch.

Relevant comment thread: https://github.com/votingworks/arlo/pull/1600#discussion_r982579908

Issue #1503, opened 2022-06-23
BRAVO: pick sample size for correct candidate pair


jonahkagan:
Currently, our BRAVO math computes sample sizes only for the candidate pair with the smallest margin. However, when our BRAVO math computes p-values, it computes them for all candidate pairs and then returns the largest p-value. In some cases, the audited sample may have a p-value that meets the risk limit for the candidate pair with the smallest margin but _not_ for another candidate pair. In these cases, the sample size for the next round will not compute correctly, since the candidate pair with the smallest margin has met the risk limit.

To fix this, instead of picking the candidate pair with the smallest margin to compute sample sizes for, we should compute sample sizes for all candidate pairs and pick the largest one. In the first round, this should correspond to picking the candidate pair with the smallest margin. In later rounds, this should correspond to picking the candidate pair with the least "proof" of correct results from the auditing so far.

PR #1619, opened 2022-10-06
Update BRAVO sample size calculation to consider all candidate pairs

arsalansufi:
_Reviewing by commit recommended_

## Overview

Issue link(s): https://github.com/votingworks/arlo/issues/1503, https://github.com/votingworks/arlo/issues/1601

This PR updates BRAVO sample size calculation to consider all candidate pairs, not just the pair with the closest margin in the reported election results, since information from a prior audit round can indicate that other pairs need closer consideration.

It also generally cleans up the BRAVO sample size calculation logic, addressing other quirks (e.g.  https://github.com/votingworks/arlo/issues/1601) along the way.

## Testing

- [x] Used existing tests to confirm that there haven't been any unexpected regressions
- [x] Added a new test to cover changes in expected behavior
- [x] Used the audit planner to test a number of cases manually


server/audit_math/bravo.py
```diff
@@ -34,45 +35,59 @@ def get_expected_sample_sizes(
...
+                continue
+            s_w = p_w / (p_w + p_l)
+            z_w = (2 * s_w).ln()
+            z_l = (2 - 2 * s_w).ln()
+
+            possible_sample_size = math.ceil(
                 (weighted_alpha.ln() + (z_w / Decimal(2))) / (p_w * z_w + p_l * z_l)
-            ).quantize(Decimal(1), rounding=ROUND_CEILING)
-        )
+            )
```
arsalansufi, line 68: Subtle tweak here. We were previously using `.quantize(Decimal(1), rounding=ROUND_CEILING)` to round up to the nearest integer. I now just use `math.ceil` because 1) it's simpler to understand and 2) `.quantize` sometimes throws invalid operation errors due to precision issues ([reference](https://stackoverflow.com/questions/31795109/decimal-quantize-raises-invalidoperation))

arsalansufi:
@jonahkagan I don't think you necessarily need to review this but added you just in case you were curious!

server/audit_math/bravo.py
```diff
@@ -517,3 +512,25 @@ def compute_risk(
...
+def is_tie(contest: Contest) -> bool:
+    winners = contest.margins["winners"]
+    losers = contest.margins["losers"]
+
+    if not losers:
+        return False
+
+    smallest_p_w = min([winner_stats["p_w"] for winner_stats in winners.values()])
+    largest_p_l = max([loser_stats["p_l"] for loser_stats in losers.values()])
+    return bool(smallest_p_w == largest_p_l)
```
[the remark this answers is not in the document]
umbernhard: don't you just love types in python? :joy:

server/audit_math/bravo.py
```diff
@@ -517,3 +512,25 @@ def compute_risk(
...
+
+
+def is_landslide(contest: Contest) -> bool:
+    losers = contest.margins["losers"]
+
+    if not losers:  # pragma: no cover
+        return False
+
+    largest_p_l = max([loser_stats["p_l"] for loser_stats in losers.values()])
+    return bool(largest_p_l == 0)
```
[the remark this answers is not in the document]
umbernhard: Huh, this is a funky case. I guess if it's vote for two, but I think in that case "total ballots" is equal to "total votes cast", so I believe this case should be precluded by the frontend. But nevertheless I'm not sad about being robust to it everywhere!

server/audit_math/ballot_polling.py
```diff
@@ -3,17 +3,11 @@
...
 from .sampler_contest import Contest
 from . import bravo, minerva
-
-
-class SampleSizeOption(TypedDict):
-    type: Optional[Literal["ASN"]]
-    size: int
-    prob: Optional[float]
-
+from .ballot_polling_types import SampleSizeOption
```
umbernhard, line 10: is there a reason this got split into it's own file? (Not complaining, just curious if it's a stylistic choice or a type-checker satisfying choice)

server/audit_math/bravo.py
```diff
@@ -7,20 +7,21 @@
...
 
-def get_expected_sample_sizes(
-    alpha: Decimal, contest: Contest, sample_results: Dict[str, int]
-) -> int:
+def get_expected_sample_size(
+    alpha: Decimal, contest: Contest, cumulative_sample_results: Dict[str, int]
+) -> SampleSizeOption:
     """
-    Returns the expected sample size for a BRAVO audit of <contest>
+    Returns the expected sample size or ASN (average sample number) for a BRAVO audit of <contest>
```
umbernhard, line 24: Ah this is a linguistic ambiguity, technically "expected sample size" and "ASN" mean the same thing here.

server/audit_math/bravo.py
umbernhard: I think this may have to change back. We were using the Decimal class specifically because of it's improved precision. This discussion is **old** so it might be worth discussing again, but here's the PR that introduced the Decimal module: https://github.com/votingworks/arlo/pull/829

server/audit_math/bravo.py
```diff
@@ -34,45 +35,59 @@ def get_expected_sample_sizes(
...
+            s_w = p_w / (p_w + p_l)
+            z_w = (2 * s_w).ln()
+            z_l = (2 - 2 * s_w).ln()
+
+            possible_sample_size = math.ceil(
                 (weighted_alpha.ln() + (z_w / Decimal(2))) / (p_w * z_w + p_l * z_l)
-            ).quantize(Decimal(1), rounding=ROUND_CEILING)
-        )
+            )
+            if possible_sample_size > sample_size:
```
umbernhard, line 69: would it make more sense to accumulate the possible sample sizes into a dict mapping to the winner-loser pair they correspond to, so we don't have to do the `expected_prob` calculation a bunch of times? It is a relatively expensive computation.

server/audit_math/bravo.py
```diff
@@ -357,97 +371,78 @@ def get_sample_size(
...
-
-    if asn <= 0:
-        raise ValueError("Sample indicates the audit is over!")
+    samples: Dict[str, SampleSizeOption] = {}
 
-    p_w = Decimal("inf")
-    p_l = Decimal(0)
-    best_loser = ""
-    worse_winner = ""
+    samples["asn"] = get_expected_sample_size(alpha, contest, cumulative_sample)
```
umbernhard, line 409: Am I wrong that this can throw an error that isn't handled? Should it be handled here, or are you relying on something further up the call stack handling it?

umbernhard: Ah, I guess technically ties and landslides are precluded by the earlier checks, so maybe we really should just crash if a ValueError gets thrown here....

server/audit_math/bravo.py
umbernhard: Hmm, thinking about this some more, maybe this is okay. Since Decimal was being used to handle an integer value, I doubt we'll get into a situation where that's an issue...all the floats are still being treated as Decimals, so I think this is fine.

umbernhard (approved):
Looks good, most of my comments are nitpicks. 

My only question: you introduced throwing errors in some functions for ties and landslides, but not in others. I don't really see any error handling in the BRAVO module, so I'm trying to understand what was gained there. It feels like we may have stopped midway?

server/audit_math/bravo.py
jonahkagan: Could also just move probability of completion calculation to after the loop since we only need it at the end.

server/audit_math/ballot_polling.py
arsalansufi: Good q! The motivation was avoiding a circular dependency. If I'd kept `SampleSizeOption` in `server/audit_math/ballot_polling.py`, `server/audit_math/bravo.py` would have to import from `server/audit_math/ballot_polling.py`, but `server/audit_math/ballot_polling.py` already imports from `server/audit_math/bravo.py` 🔁

server/api/public.py
```diff
@@ -133,7 +127,10 @@ def compute_ballot_polling_sample_size(
         risk_limit_percentage, contest, None, None
     )
     sample_size: int = sample_size_options.get(
-        "0.9", sample_size_options.get("asn", sample_size_options.get("all-ballots")),
+        "0.9",
+        sample_size_options.get(
+            "asn", sample_size_options.get("all-ballots", {"size": contest.ballots})
+        ),
     )["size"]
```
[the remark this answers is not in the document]
jonahkagan: Looking at this now, I'm realizing it might be confusing to users to return the ASN as a fallback when there's no 0.9 key.

jonahkagan (approved):

server/audit_math/bravo.py
arsalansufi: Reworded to make that clear: https://github.com/votingworks/arlo/pull/1619/commits/ceda9db32a577fbc5d8796220a3c7ed685d08fe5

server/audit_math/bravo.py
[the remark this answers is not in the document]
arsalansufi: > so I believe this case should be precluded by the frontend. But nevertheless I'm not sad about being robust to it everywhere!

At least in the audit planner case, this change has made a difference!

_Before / After_

before after

> also nitpick p_w is a proportion, not a probability

Helpful distinction! Updated the GitHub comment accordingly

server/audit_math/bravo.py
arsalansufi: Yup my thinking exactly. I see the importance of using `Decimal` for the intermediate calculations, but for this last step, where we're rounding to an integer, it's not as relevant

server/audit_math/bravo.py
arsalansufi: Good to know that the calculation is relatively expensive! I'd wrongly assumed that it wasn't and kept it in the loop to keep the code simpler. Moved it out of the loop in https://github.com/votingworks/arlo/pull/1619/commits/9b6275658ccfdaa0acee40fc8e29ba7060814358

server/audit_math/bravo.py
arsalansufi: Yup exactly! My thinking was that `get_expected_sample_size` should make its limitations clear, and it's on consumers to perform relevant checks before using it

server/api/public.py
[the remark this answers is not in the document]
arsalansufi: Good observation! In practice, the only time we should have to use the ASN is when we hit this case:

https://github.com/votingworks/arlo/blob/d81016ee63b72ec6ec57da2696b219fb35c0de27/server/audit_math/bravo.py#L413-L416

At the bare minimum, let me change the order of precedence in this chain of fallbacks for clarity, making `all-ballots` higher priority than `asn` (though there aren't any cases where both `all-ballots` and `asn` will be specified)

Especially after this PR, I'm not sure under what other circumstances we'll not have an `0.9` key

Another option would be to return `0.7`, `0.8`, and `0.9` keys in addition to `asn` for the landslide case. And then remove the `asn` fallback in the audit planner endpoint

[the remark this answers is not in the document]
arsalansufi: I'm gonna move forward with this PR but can revisit the above in a follow-up since it's fairly separate from this PR anyway!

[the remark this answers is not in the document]
jonahkagan: For that landslide case, what would get returned for the '0.9' key if we didn't short-circuit? I'm wondering if that short-circuit logic to return `1` as the sample size should actually be part of the ASN calculating function, not the whole sample size calculation.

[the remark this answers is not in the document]
arsalansufi: Mm great q - when you pass a landslide and remove the short circuit (or more specifically move the short circuit to the ASN calculating function, which will err if you don't short circuit), you still get the same output:

```
{
  "asn": {
    "type": "ASN",
    "size": 1,
    "prob": 1,
  },
}
```

You get a sample size of 0 for all of the `0.7`, `0.8`, and `0.9` quants, so they aren't added to the response at all:

https://github.com/votingworks/arlo/blob/0b014891f5e72d3ff88d643f0ea3f99a5de77fe3/server/audit_math/bravo.py#L450-L455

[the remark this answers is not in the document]
jonahkagan: Thanks for looking into that. I think we could pursue one of the options you outlined above to make things more consistent/clear, but it doesn't feel high priority.

PR #1626, opened 2022-10-11
Restore original Blueprint heading color

jonahkagan:
There's no reason to override it, and the original (darker) color looks better, as we learned when building the audit planner.

Tested by poking around the app and all looks well!

### DIFF
--- a/client/src/components/Atoms/Steps.tsx
+++ b/client/src/components/Atoms/Steps.tsx
@@ -70,7 +70,7 @@ const StepListItemCircle = styled.div<{
 export const StepListItem = styled(({ current, ...props }) => (
   <H5 {...props} />
 ))<{ current: boolean }>`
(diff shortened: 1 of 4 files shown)


client/src/components/Atoms/Steps.tsx
```diff
@@ -70,7 +70,7 @@ const StepListItemCircle = styled.div<{
 export const StepListItem = styled(({ current, ...props }) => (
   <H5 {...props} />
 ))<{ current: boolean }>`
-  color: ${props => (props.current ? Colors.DARK_GRAY5 : Colors.GRAY3)};
+  color: ${props => (props.current ? Colors.DARK_GRAY1 : Colors.GRAY3)};
```
jonahkagan, line 73: Confirmed that `DARK_GRAY1` is the standard text color using the inspector

Issue #1531, opened 2022-06-30
Batch comparison tally entry accounts


jonahkagan:
Currently, in batch audits, jurisdiction managers are responsible for batch result data entry. They create audit boards, but the audit boards don't have any way to log in. We should give them the ability to log in in order to distribute the actual data entry work in large batch audits.

**Solution**
Add a new tally entry role for batch audits. JMs can create tally entry accounts that can log in and enter/edit results for any batches.

Tally entry accounts initiate log in with a login link (similar to existing audit board login), but there is just one link for the whole jurisdiction. The JM must enter a code shown on the tally entry account screen in order to confirm the login.

**Implementation details**
- Create a new user type
- JM management flow API endpoints
  - POST turn on (creates log in link)
  - GET login link
  - GET login requests
  - POST confirm login (enter code)
  - DELETE reject login
- Login flow API endpoints
  - POST start login
  - GET login status (and code)

**Implementation plan**
- [x] Remove audit boards from batch audits (1)
- [x] Add tally entry accounts (backend) (2)
- [x] Tally entry accounts - login + tally entry (frontend) (2)
- [x] JM batch audit workflow (frontend) (2)
- [x] Fancier single-digit-at-a-time login code input
- [x] Improve user-facing login errors
- [ ] Refactor and clean up Header components
- [ ] Replace AuthDataContext with useCurrentUser
- [x] Ensure that support tools respects new tally entry accounts (e.g. don't display audit boards)
- [x] Figure out what the jurisdiction progress detail modal should show
- [x] Auto-select the appropriate step based on progress

ginvdr:
This may add complexity but I think jurisdiction manager's in *really* large audits will also want a way to review the data entry from each audit board.  Which somewhat ties into

mattroe:
Potential solution discussed: https://votingworks.slack.com/archives/C01F2CH62TS/p1661189447768099

PR #1627, opened 2022-10-12
Tally entry accounts (backend)

jonahkagan:
Task: #1531

Backend API for tally entry user login for batch comparison audits.

Includes several new endpoints:
- `GET /tallyentry/<passphrase>` - The login link that a JM shares with tally entry users. This is the login entry point.
- `POST /auth/tallyentry/code` - Endpoint to for a tally entry user to request a confirmation code, submitting member names/party affiliations.
- `GET /auth/tallyentry/election/<election_id>/jurisdiction/<jurisdiction_id>` - Endpoint for a JM to check for login requests from tally entry users
- `POST /auth/tallyentry/election/<election_id>/jurisdiction/<jurisdiction_id>/confirm` - Endpoint for a JM to submit a login confirmation code for a tally entry user to log them in

As well as an update to `/api/me` to include tally entry users (most importantly, their login confirmation code and a timestamp indicating when their login is confirmed).

### DIFF
--- a/server/auth/lib.py
+++ b/server/auth/lib.py
@@ -20,10 +20,15 @@ class UserType(str, enum.Enum):
     # Audit boards are represented by the AuditBoard record and use
     # AuditBoard.id as their login key. In the real world, a member of an audit
     # board will log in on behalf of the audit board by navigating to
-    # /audit-board/<passphrase>, initiating the session. We consider the whole
+    # /auditboard/<passphrase>, initiating the session. We consider the whole
     # audit board to be one "logged in user," as opposed to trying to know
     # which specific audit board member logged in.
     AUDIT_BOARD = "audit_board"
+    # Tally entry users are represented with a TallyEntryUser record and use
+    # TallyEntryUser.id as their login key. A tally entry user (or users, if
+    # there are multiple people working together), will log by navigating
+    # /tallyentry/<passphrase>.
+    TALLY_ENTRY = "tally_entry"
 
 
 _SUPPORT_USER = "_support_user"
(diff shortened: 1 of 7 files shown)


benadida (commented):
overall lgtm

server/tests/test_auth.py
```diff
@@ -494,6 +513,384 @@ def test_audit_board_log_in(
         ) < timedelta(seconds=1)
 
 
+def test_tally_entry_login(
```
jonahkagan, line 516: I recommend reading through this test for an overview of the flow.

server/auth/routes.py
```diff
@@ -317,6 +340,142 @@ def auditboard_passphrase(passphrase: str):
...
+    db_session.commit()
+
+    # We set the tally entry user in the session even though they haven't fully
+    # logged in yet. This allows the user to hit /api/me to retrieve
+    # jurisdiction name and login code to show on the login screens. The
+    # restrict_access decorator ensures that they can't do anything else until
+    # their login code is confirmed by the JA.
+    set_loggedin_user(session, UserType.TALLY_ENTRY, tally_entry_user.id)
+
+    return redirect("/tally-entry")
```
jonahkagan, line 365: This is the frontend route where tally entry content will live. I know it's confusingly close to the login link (`/tally-entry` vs `/tallyentry/<passphrase>`), but I thought this would be an easy way to avoid collisions with frontend routing/proxying. Though I didn't actually try out using overlapping URLs, so it might not be an issue. Dunno if anyone has experience with this.

server/auth/lib.py
```diff
@@ -135,14 +140,25 @@ def check_access(
...
+        assert user_type == UserType.TALLY_ENTRY
+        user = TallyEntryUser.query.filter_by(id=user_key).one()
+        if user.login_confirmed_at is None:
+            raise Unauthorized(
+                "Your jurisdiction manager must confirm your login code."
+            )
+        if jurisdiction and user.jurisdiction_id != jurisdiction.id:
+            raise Forbidden(
+                description=f"User does not have access to jurisdiction {jurisdiction.id}"
+            )
```
jonahkagan, line 160: We don't use `restrict_access` with tally entry users in this PR, but we implement and test it here, since it's a relevant part of the auth scheme. It will be used in coming PRs to authorize batch auditing endpoints.

jonahkagan:
Ok this is now fully tested and ready for review

server/auth/routes.py
benadida: one trend I've seen is to have all APIs behind `/api/` to differentiate URLs that should show up in the browser bar from ones that are just JSON. But this seems fine with comments.

benadida (approved):

server/auth/routes.py
jonahkagan: Yeah, both of these routes are user-facing, but one has to hit the server while one hits the front-end routing, so I don't think the `/api/` approach will help here.

PR #1631, opened 2022-10-13
Speed up pre-commit checks

jonahkagan:
_Review by commits_

(╯°□°）╯︵ ┻━┻

A variety of tweaks found in various internet sources. Brings average pre-commit hook time down from >1m to <10s on my machine.

### DIFF
--- a/client/tsconfig.json
+++ b/client/tsconfig.json
@@ -12,9 +12,13 @@
     "moduleResolution": "node",
     "resolveJsonModule": true,
     "isolatedModules": true,
-    "noEmit": true,
+    "incremental": true,
+    "declaration": true,
+    "emitDeclarationOnly": true,
+    "outDir": "dist",
     "types": ["node", "@types/jest", "@testing-library/jest-dom"],
     "jsx": "preserve"
   },
-  "include": ["src"]
+  "include": ["src"],
+  "exclude": ["**/node_modules", "**/.*/"]
 }
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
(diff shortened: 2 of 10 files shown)


client/tsconfig.json
```diff
@@ -12,9 +12,13 @@
...
+    "incremental": true,
+    "declaration": true,
+    "emitDeclarationOnly": true,
+    "outDir": "dist",
     "types": ["node", "@types/jest", "@testing-library/jest-dom"],
     "jsx": "preserve"
   },
-  "include": ["src"]
+  "include": ["src"],
+  "exclude": ["**/node_modules", "**/.*/"]
```
jonahkagan, line 23: Yes, we were type checking all of node_modules before 🤦

arsalansufi: 🙈

arsalansufi (approved):
LGTM!

Side note: I actually didn't even notice that we had pre-commit hooks set up in Arlo. They don't seem to be running for me 🤔

jonahkagan:
@arsalansufi woah! Maybe we don't need them if you've been getting by fine without

arsalansufi:
Yeah good point! Auto-formatting basically covers this need for me

client/tsconfig.json
eventualbuddha: What, really? I'd expect TypeScript to know not to do this.

jonahkagan: Well, I'm not 100% sure actually. I think I remember adding that exclude line and rerunning the lint-staged command and seeing a big speedup. I got the tip from this doc: https://github.com/microsoft/TypeScript/wiki/Performance#misconfigured-include-and-exclude.

But reading closer in this other section of the doc, it says node_modules is excluded by default: https://github.com/microsoft/TypeScript/wiki/Performance#specifying-files. So maybe I was wrong or not diligent in only changing one setting at a time.

PR #1632, opened 2022-10-13
Tally entry users - tally entry screen

jonahkagan:
_Review by commits_

- Uses the existing `BatchRoundDataEntry` component to allow tally entry users to enter tallies. Hides the ability to copy or finalize. I didn't put a ton of effort into making this nice (from both a design perspective and a code perspective) since it will be replaced for #1623.
- Also updates the permissions on the necessary API endpoints. Since a tally entry user can edit all of the batches that a JM can edit, I just reused the same endpoints. 

Screen Shot 2022-10-13 at 4 24 39 PM

### DIFF
--- a/client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
+++ b/client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
@@ -434,10 +434,9 @@ const BatchRoundDataEntry: React.FC<IBatchRoundDataEntryProps> = ({
   return (
     <div>
       <div>
-        <p>
-          When you have examined all of the ballots assigned to you, enter the
-          number of votes recorded for each candidate/choice from the audited
-          ballots in each batch.
+        <p className="bp3-text-large">
+          For each batch, enter the number of votes tallied for each
+          candidate/choice.
         </p>
         {resultsFinalizedAt && (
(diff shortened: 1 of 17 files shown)


jonahkagan:
@arsalansufi again, just looking for an initial review. There are so many parts that I want to make sure they all fit together before writing tests and finalizing anything.

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -434,10 +434,9 @@ const BatchRoundDataEntry: React.FC<IBatchRoundDataEntryProps> = ({
   return (
     <div>
       <div>
-        <p>
-          When you have examined all of the ballots assigned to you, enter the
-          number of votes recorded for each candidate/choice from the audited
-          ballots in each batch.
+        <p className="bp3-text-large">
```
arsalansufi, line 437: Same comment here as in the previous PR re using `Classes.TEXT_LARGE`

arsalansufi (commented):
Thanks for the digestible PRs! Draft looks good to me!

jonahkagan:
@arsalansufi This is ready for review now

arsalansufi (approved):
🚢

PR #1637, opened 2022-10-18
Remove all state logic from Steps component

jonahkagan:
I realized that the approach of marking one `StepListItem` with a `current` prop was brittle - if the `StepListItem` wasn't a direct child of `StepList`, there was no way to access the `current` prop in `StepList`.

Also, upon using the components more, I think they will be easier to reason about if they are purely graphical and completely controlled.

### DIFF
--- a/client/src/components/Atoms/Steps.tsx
+++ b/client/src/components/Atoms/Steps.tsx
@@ -1,7 +1,6 @@
 import React from 'react'
 import styled from 'styled-components'
 import { Card, Colors, H5, Icon } from '@blueprintjs/core'
-import { assert } from '../utilities'
 
 /**
  * A set of components to display a multi-step process. Consists of a card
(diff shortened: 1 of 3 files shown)


client/src/components/Atoms/Steps.tsx
```diff
@@ -143,3 +149,17 @@ export const StepActions: React.FC<{
     {right || <div />}
   </StepActionsRow>
 )
+
+// Utility function to compute step state for the most common step list pattern:
+// - a single current step
+// - all previous steps are complete
+// - all subsequent steps are incomplete
```
arsalansufi, line 156: Nice! I like this compromise - leaving state management up to consumers but providing utilities for common cases. Similar in spirit to composable components

arsalansufi (approved):

PR #1638, opened 2022-10-19
Jurisdiction manager 3-step flow for batch comparison audits

jonahkagan:
_Review by commits_

Task: #1531 

Adds a 3-step flow for jurisdiction managers during a round in a batch comparison audit:
1. Prepare Batches (download retrieval list and tally sheets)
2. Set Up Tally Entry Accounts (share login link, confirm login codes)
3. Enter Tallies (review/edit/finalize audited batch tallies)

https://user-images.githubusercontent.com/530106/196826785-ceef108c-6701-4f9e-b98f-17918ff57942.mov


client/src/components/JurisdictionAdmin/BatchRoundSteps/EnterTalliesStep.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { StepPanel, StepActions } from '../../Atoms/Steps'
+import LinkButton from '../../Atoms/LinkButton'
+
+interface IEnterTalliesStepProps {
+  previousStepUrl: string
+  jurisdiction: IJurisdiction
+  round: IRound
+}
+
+const EnterTalliesStep: React.FC<IEnterTalliesStepProps> = ({
```
jonahkagan, line 18: @arsalansufi I thought you might want to see roughly what this component looks like so far since it will be rendering the list/detail batch auditing view.

jonahkagan:
@arsalansufi the only thing I still need to do is update the Cypress tests, but I figured I'd put it up for review now in case you wanted to get a head start on reviewing while I work on that.

jonahkagan:
@arsalansufi actually, I also realized I need to clean up all the inline CSS still!

client/src/components/Atoms/Layout.tsx
```diff
@@ -0,0 +1,21 @@
...
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+`
+
+export const ButtonRow = styled.div`
+  display: flex;
+  flex-direction: row;
+  gap: 10px;
+`
```
jonahkagan, line 21: Curious to hear your thoughts about this approach. I was trying to abstract some common cases while also still maintaining some specificity (vs just making a `Flex` component that takes all the flex props). Not sure what the best approach is

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton.tsx
```diff
@@ -23,15 +24,16 @@ const DownloadBatchTallySheetsButton = ({
...
-  const { batches } = batchesQuery.data
-  // Batch comparison audits only support a single contest
-  const contest = contests[0]
-
   const onClick = async () => {
+    // Wait for the batches/contests to load in case they haven't yet.
+    while (!batchesQuery.isSuccess || contests === null) {
+      // eslint-disable-next-line no-await-in-loop
+      await sleep(100)
+    }
```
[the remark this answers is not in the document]
jonahkagan: That's a fair concern. Maybe I should make it wait for `batchesQuery.isSettled`, which will account for both request success and failure? I could also add a timeout.

I want to push for the best UX if possible - an unexplained disabled button feels a bit confusing. We could also just load the data on button press instead of in advance. What do you think? I know we discussed this a bit once when you originally implemented this button.

client/src/components/Atoms/Layout.tsx
jonahkagan: Yeah, I also went through the same debates. Here's my thinking of a way to approach this:
- Make an opinionated component based on the cases we think are most common
- Use inline styles to override when necessary
- If we see a common override, create a new component or add a prop

client/src/components/JurisdictionAdmin/DownloadBatchTallySheetsButton.tsx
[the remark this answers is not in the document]
arsalansufi: Agreed that the disabled button isn't the best UX!

Waiting for `batchesQuery.isSettled` (and the relevant analog for `contests`) sounds great! Basically just want to make sure that we break out of the loop even under error circumstances

My gut tells me that a timeout isn't may not be worth the complexity but don't feel strongly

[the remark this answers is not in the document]
jonahkagan: Ok turns out `isSettled` is only available in a newer version of react-query, but I was able to still handle the error case (ebd801cf). There isn't an equivalent error case for contests, so I converted that to use react-query (fb0e2725).

client/src/components/Atoms/Layout.tsx
jonahkagan: Updated to `Row` and `Column` elements here: [b8d4b0e](https://github.com/votingworks/arlo/pull/1638/commits/b8d4b0e65c14fb718f9b09d161686ddb0f714f30)

jonahkagan:
@arsalansufi Ready for re-review!

client/src/components/Atoms/Layout.tsx
```diff
@@ -1,21 +1,52 @@
-import styled from 'styled-components'
+import styled, { css } from 'styled-components'
 
-export const CenterRow = styled.div`
+interface BoxProps {
+  justify?: 'start' | 'center' | 'end' | 'space-between' | 'space-around'
+  align?: 'start' | 'center' | 'end' | 'stretch' | 'baseline'
```
arsalansufi, line 5: Might want to name these props to directly match the CSS properties, e.g. `justifyContent` and `alignItems`, to avoid ambiguity with the many other CSS properties that start with `justify-` and `align-`, e.g. `justify-self` and `align-self`

arsalansufi (approved):
One last comment then LGTM!

client/src/components/Atoms/Layout.tsx
jonahkagan: Ok, will do

PR #1643, opened 2022-10-21
Fancy login code input

jonahkagan:
Adds a custom input component for login codes that shows one text input per digit.

https://user-images.githubusercontent.com/530106/197087212-2c84f58c-ae59-47a4-8cc8-26c2abf34dc8.mov

### DIFF
--- a/client/src/components/Atoms/CodeInput.tsx
+++ b/client/src/components/Atoms/CodeInput.tsx
@@ -1,0 +1,114 @@
+import React, { useEffect } from 'react'
+import styled from 'styled-components'
+import { Classes } from '@blueprintjs/core'
+import { replaceAtIndex, range } from '../../utils/array'
+import { Row } from './Layout'
+import { assert } from '../utilities'
+
+const DigitInput = styled.input.attrs({
+  className: Classes.INPUT,
+  type: 'text',
+})`
+  text-align: center;
+`
+
+// We only support a small subset of the HTMLInputElement interface to start,
+// but this can be expanded as needed.
+type InputProps = Pick<
+  React.InputHTMLAttributes<HTMLInputElement>,
+  'name' | 'id'
+>
+
+interface ICodeInputProps extends InputProps {
+  length: number
+  value: string
+  onChange: (value: string) => void
+}
+
+/**
+ * CodeInput takes the same props as a controlled text input, but
+ * actually renders and coordinates multiple individual digit inputs.
+ *
+ * Example:
+ *  const [value, setValue] = useState('')
+ *  <CodeInput length={3} value={value} onChange={setValue} />
+ */
+const CodeInput: React.FC<ICodeInputProps> = ({
+  length,
+  value = '',
+  onChange,
+  ...htmlInputProps
+}) => {
+  assert(/^\d*$/.test(value), 'CodeInput value must be a string of digits')
+
+  const digitInputRefs = range(0, length - 1).map(() =>
+    React.createRef<HTMLInputElement>()
+  )
+
+  const focusDigitInput = (index: number) => {
+    const digitInputRef = digitInputRefs[index]
+    if (digitInputRef?.current) {
+      digitInputRef.current.focus()
+    }
+  }
+
+  const moveFocusRight = (index: number) => {
+    if (index < length - 1) {
+      focusDigitInput(index + 1)
+    }
+  }
+
+  const moveFocusLeft = (index: number) => {
+    if (index > 0) {
+      focusDigitInput(index - 1)
+    }
+  }
+
+  const onDigitKeyDown = (index: number, key: string) => {
+    const firstEmptyDigitIndex = value.length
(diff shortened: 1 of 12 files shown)


client/src/components/Atoms/CodeInput.tsx
```diff
@@ -0,0 +1,114 @@
...
+ *  const [value, setValue] = useState('')
+ *  <CodeInput length={3} value={value} onChange={setValue} />
+ */
+const CodeInput: React.FC<ICodeInputProps> = ({
+  length,
+  value = '',
+  onChange,
+  ...htmlInputProps
+}) => {
+  assert(/^\d*$/.test(value), 'CodeInput value must be a string of digits')
```
arsalansufi, line 42: This isn't possible on the version of TS that we're on but just thought I'd note that, on the latest version of TS, you can use [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) to enforce this at compile time:

```ts
type NumericalString = `${number}`;
```

Kinda mind blowing!

whoa-1

whoa-2

client/src/components/Atoms/CodeInput.tsx
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }
+
+  // Whenever we have no digits entered, focus the first digit input
+  // E.g. on mount, after backspacing, or after form reset
+  useEffect(() => {
+    if (value === '') {
+      focusDigitInput(0)
+    }
+  })
```
arsalansufi, line 92: I believe, reading the comment, that this is intentional, but just sanity checking that we're intentionally not including a dependency array here and running this hook on every render

client/src/components/JurisdictionAdmin/BatchRoundSteps/TallyEntryAccountsStep.tsx
```diff
@@ -142,6 +145,14 @@ const TurnOnTallyEntryAccountsPrompt: React.FC<ITurnOnTallyEntryAccountsPromptPr
   )
 }
 
+const CodeInput = styled(CodeInputAtom)`
+  input {
+    width: 70px;
+    font-size: 50px;
+    height: 80px;
+  }
```
arsalansufi, line 153: Instead of specifying this style here, we could bake these styles into `CodeInput`, toggle-able via an optional size prop, e.g. `size="large"` or just `large`

arsalansufi (approved):

client/src/components/Atoms/CodeInput.tsx
jonahkagan: That is pretty wild! Who comes up with this stuff??

client/src/components/Atoms/CodeInput.tsx
jonahkagan: Good question. I have been trying out never including a dependency array for `useEffect` hooks to make sure that I'm not relying on that performance optimization for correctness.

Do you think there's a performance reason to include one here? I could think of a few cases where there's a rerender and `value` stays `''` where we could avoid refocusing, but it seems likely negligible to me.

client/src/components/JurisdictionAdmin/BatchRoundSteps/TallyEntryAccountsStep.tsx
jonahkagan: I thought about that. I don't have a strong opinion, but thought it might be nice to see what size was desired in other use cases before building any size styling into the component.

jonahkagan: (I'm specifically thinking of the jurisdiction admin login screen as the next potential use case)

arsalansufi: Seems reasonable to me!

arsalansufi: I think I tend to err on the side of baking logic into components to encourage consistency, even if it's not yet clear what the consistent standard is that we want to settle on

client/src/components/Atoms/CodeInput.tsx
arsalansufi: > I have been trying out never including a dependency array for useEffect hooks to make sure that I'm not relying on that performance optimization for correctness

Very interesting!

Agreed that the performance concern here is negligible

client/src/components/JurisdictionAdmin/BatchRoundSteps/TallyEntryAccountsStep.tsx
jonahkagan: That makes sense as an approach to maintain visual consistency especially. Though on the other hand, it can make it harder to modify later without breaking consumers. I think I'll leave this one for now, but will noodle on that more.

PR #1644, opened 2022-10-24
Redesign batch audit data entry to use list-and-detail view

arsalansufi:
# Overview

Issue link: https://github.com/votingworks/arlo/issues/1623

This PR redesigns batch audit data entry to use a list-and-detail view that we think will be more manageable than the current "one giant table" view, especially for large batch audits like the upcoming one in Georgia!

I'm still writing automated tests but wanted to proactively get this up for review for UI feedback.

https://user-images.githubusercontent.com/12616928/197563548-1bb7659a-f3c5-441f-9fd6-32e31d4b9581.mov

# Testing

- [x] Manual testing
- [ ] Still writing automated tests


client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -1,606 +1,644 @@
-/* eslint-disable jsx-a11y/no-autofocus */
-import React, { useState } from 'react'
+import classnames from 'classnames'
```
arsalansufi, line 1: The changes to this file are so substantial it's probably easier to review the new code in isolation instead of the diff, hence the delete commit followed by a new file commit: https://github.com/votingworks/arlo/pull/1644/commits/5d812b985823e55dde6ae1370e32a684feb14f12

arsalansufi, line 1: The changes to this file are so substantial it's probably easier to review the new code in isolation instead of the diff, hence the delete commit followed by a new file commit (https://github.com/votingworks/arlo/pull/1644/commits/952bc6424f27e5cb01fd9b2ad795cb7bc30935ed)

arsalansufi:
@jonahkagan I'm still writing automated tests and also know that I'll need to account for merge conflicts once https://github.com/votingworks/arlo/pull/1638 is merged (though I tried to get ahead of some, e.g. the removal of the `showFinalizeAndCopyButtons` prop and those buttons from `BatchRoundDataEntry`). But wanted to proactively get some eyes on this, especially the UX!

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -1,606 +1,644 @@
...
+                    }
+                    saveResults={saveResults}
+                    sheets={sheets}
+                  />
+                }
+              >
+                Vote Totals
+              </Tab>
+              {sheets.length > 1 &&
+                sheets.map((sheet, i) => (
```
arsalansufi, line 397: Once edge case that I'm still thinking through is the case that the number of sheets is so large that the tab bar overflows - looking into scrolling and wrapping

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,645 @@
...
+  })
+
+  // Reset state whenever a new batch is selected
+  useEffect(() => {
+    reset({ resultTallySheets: initialResultTallySheets })
+    setSelectedTabId(VOTE_TOTALS_TAB_ID)
+    setIsEditing(false)
+  }, [selectedBatchId])
+
+  // Auto-select first search match
```
jonahkagan, line 178: This felt slightly slow to me when I watched the video. Is it possible to make it a bit more responsive?

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
jonahkagan: Yes, I was thinking that could be an issue, though likely a rare case. My gut says scrolling will be better than wrapping.

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,645 @@
...
+  const contests = useContestsJurisdictionAdmin(electionId, jurisdictionId)
+  const recordBatchResults = useRecordBatchResults(
+    electionId,
+    jurisdictionId,
+    roundId
+  )
+  const { confirm, confirmProps } = useConfirm()
+
+  const [searchQuery, setSearchQuery] = useState('')
+  const [selectedBatchId, setSelectedBatchId] = useState<IBatch['id']>()
```
jonahkagan, line 130: I wonder if we should auto-select the first batch. I think it might work a bit better with the feeling of "when I filter the batches, the top one in the list is selected"

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,645 @@
...
+              }}
+              // Defaults to false but noting explicitly for clarity. Rendering all tabs at all
+              // times is important for react-hook-form's state management
+              renderActiveTabPanelOnly={false}
+              selectedTabId={selectedTabId}
+            >
+              <Tab
+                id={VOTE_TOTALS_TAB_ID}
+                key={VOTE_TOTALS_TAB_ID}
+                panel={
```
jonahkagan, line 372: How is the spacing between the tabs and the results table determined? Looking at the video, the spacing looked a bit too big, but then when I ran it locally, it looked fine. I could swear the spacing is different...
Screen Shot 2022-10-24 at 10 17 48 AM

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,645 @@
...
+
+        {isEditing || formState.isSubmitting ? (
+          <Button
+            disabled={isRenamingSheet}
+            icon="tick"
+            intent="primary"
+            loading={formState.isSubmitting}
+            onClick={saveResults}
+          >
+            Save Results
```
jonahkagan, line 629: We're standardizing on Tally/Tallies instead of Result/Results for user-facing copy. There are a few places that should be changed

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,645 @@
...
+          </Button>
+        ) : (
+          <Button
+            disabled={areResultsFinalized}
+            icon="edit"
+            onClick={enableEditing}
+          >
+            Edit Results
+          </Button>
```
jonahkagan, line 638: Might be worth doing a pass using keyboard navigation for power-users doing repeated tally entry. An idea off the top of my head:
- After searching, hit enter to focus edit button
- Hit enter to edit, which focuses first input
- Tab through inputs
- Tab to save button, hit enter, which focuses search bar again

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,645 @@
...
+        description:
+          'You have unsaved changes. ' +
+          'Are you sure you want to leave this batch without saving changes?',
+        yesButtonLabel: 'Yes, Leave Batch without Saving',
+        onYesClick: () => {
+          setSelectedBatchId(batchId)
+        },
+        noButtonLabel: 'Return to Batch',
+      })
+      return
```
jonahkagan, line 213: Love that you caught this case! It made me realize that we might need a clearer way to discard changes during editing mode. I think we had that in the old UI. Maybe a secondary button labeled "Discard Changes" next to the save button would do it.

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,645 @@
...
+  // ---------- Handlers (START) ----------
+
+  const selectBatch = (batchId: string) => {
+    if (formState.isDirty) {
+      confirm({
+        title: 'Unsaved Changes',
+        description:
+          'You have unsaved changes. ' +
+          'Are you sure you want to leave this batch without saving changes?',
+        yesButtonLabel: 'Yes, Leave Batch without Saving',
```
jonahkagan, line 207: I think we can standardize this copy with the idea of discarding changes that I mentioned below. Maybe just label these "Discard Changes" and "Cancel"

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,645 @@
...
+        {!selectedBatch ? (
+          <p>Select a batch to enter audit results.</p>
+        ) : (
+          <>
+            <H4>{selectedBatch.name}</H4>
+            <Tabs
+              id={selectedBatch.name}
+              onChange={(newTabId: string) => {
+                cancelSheetRename()
+                setSelectedTabId(newTabId)
```
jonahkagan, line 362: Is it possible to remove the blue focus outline that happens when you click on a tab or on a batch list item?

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,645 @@
...
+                )}
+              </td>
+            </tr>
+          ))}
+        </tbody>
+      </BatchResultTallySheetTable>
+
+      <Actions>
+        {!isTotalsSheet && (
+          <SheetActions>
```
jonahkagan, line 552: I think the solution of having all the sheet actions at the bottom is workable. My only quibbles are that (1) it's a bit cluttered - there are a lot of controls shown (2) the sheet actions are far away from the sheet name (especially the rename action).

Two other ideas I have:
- Add a dropdown menu in the current tab name that has two options: Rename Sheet and Delete Sheet. The menu would be triggered by clicking a small icon button in the sheet name (similar styling to a browser tab delete button, but a diff icon - maybe a downward triangle). The rename option could either make the sheet name editable in place or show a modal.
- Add a dropdown menu to the "Add a sheet" button with similar functionality

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,645 @@
...
+    <>
+      <BatchResultTallySheetTable>
+        <thead>
+          <tr>
+            <th>Choice</th>
+            <th>Votes</th>
+          </tr>
+        </thead>
+        <tbody>
+          {contest.choices.map(choice => (
```
jonahkagan, line 498: What happens when there are a large number of candidates? I've started to establish a pattern where the step panel is fixed height and scrolling happens internally, so I'm wondering how this component scrolls (if at all)

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,645 @@
...
+              {isEditing && (
+                <Button
+                  disabled={isRenamingSheet}
+                  icon="add"
+                  minimal
+                  onClick={addSheet}
+                >
+                  {sheets.length > 1
+                    ? 'Add sheet'
+                    : 'Use multiple tally sheets'}
```
jonahkagan, line 439: Even though I like the way these sentence case button labels look, I think maybe we should title case them for consistency with the rest of the UI. If I could go back, I would make them all sentence case...

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,645 @@
...
+      // Reset the form's isDirty value back to false
+      reset({ resultTallySheets })
+    } catch {
+      // Errors are automatically toasted by the queryClient
+      setIsEditing(true)
+    }
+  }
+
+  const onInvalidSubmit: SubmitErrorHandler<IBatchRoundDataEntryFormState> = errors => {
+    // Open the first sheet with a validation error
```
jonahkagan, line 248: This is a really nice touch!

jonahkagan (commented):
Awesome work! I just reviewed from a UX perspective to start. Quite a lot of tricky corner cases to handle here

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -1,544 +1,644 @@
...
+                >
+                  {sheets.length > 1
+                    ? 'Add sheet'
+                    : 'Use multiple tally sheets'}
+                </Button>
+              )}
+            </Tabs>
+          </>
+        )}
+      </Detail>
```
jonahkagan, line 445: I forgot one important feature when handing this off to you: We need to show who last updated the batch tally (JM or tally entry user)! Let's discuss where to add this in

arsalansufi: Ah yes I noticed this in the WIP branch you handed over! Happy to discuss

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -1,544 +1,659 @@
...
     name: 'resultTallySheets',
   })
 
-  const onSubmit = async ({
+  // Reset state whenever a new batch is selected
+  useEffect(() => {
+    reset({ resultTallySheets: initialResultTallySheets })
+    setSelectedTabId(VOTE_TOTALS_TAB_ID)
+    setIsEditing(false)
+  }, [selectedBatchId])
```
jonahkagan, line 172: I think you might be able to remove this effect if you made a subcomponent for a single batch (the entire contents of the detail view, including the whole form).

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
```diff
@@ -1,544 +1,659 @@
...
+  // Auto-select first search match
+  useEffect(() => {
+    if (
+      debouncedSearchQuery &&
+      filteredBatches.length > 0 &&
+      !formState.isDirty
+    ) {
+      setSelectedBatchId(filteredBatches[0].id)
+    }
+  }, [debouncedSearchQuery, filteredBatches, formState, setSelectedBatchId])
```
jonahkagan, line 190: I wonder if we even need the `selectedBatchId` state anymore. Could it just be derived as the first item in `filteredBatches`, since we always select the first item in the list now?

jonahkagan (commented):
I know you're still in the process of changing things, but I started looking at the code just to get a sense of it and wanted to give some high-level structural feedback in case it's helpful as you keep working.

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
arsalansufi: I've set up the most rudimentary horizontal scrolling when the number of sheets is large but have taken note in https://github.com/votingworks/arlo/issues/1655 (a catch-all issue for follow-ups) that there's room for improvement, e.g. horizontal scrolling of just the tab bar.

jm

te

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
arsalansufi: Ooh I think the root here wasn't the debounce time (250ms) but inconsistent usage of the debounced search query. While the batch selection was being debounced, the batch list filtering was not, resulting in a flicker. I now consistently use the debounced search query: https://github.com/votingworks/arlo/pull/1644/commits/40473164c4164ca8af506f39dd8d6d7e847274cf

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
arsalansufi: Good call: https://github.com/votingworks/arlo/pull/1644/commits/28bfee21a2ed884fdbe78b2be76fa80467bcbbf1

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
arsalansufi: Done!

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
arsalansufi: This was a really interesting one! We need the blue focus outline (or some indication in general) for accessible tab navigation, but we don't need/want it for mouse navigation. I initially looked at using the `:focus-visible` selector, which allows you to distinguish focus by keyboard vs. focus by mouse.

But then discovered that this is a common enough wish that Blueprint provides [a utility](https://blueprintjs.com/docs/versions/3/#core/accessibility.focus-management) for it! We'll now get the desired behavior globally.

https://github.com/votingworks/arlo/pull/1644/commits/4aaf474beb1ba2a02764e5983b20b9ad2cd56449

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
arsalansufi: As discussed in person, went with the second approach ✅

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
arsalansufi: I once again have the most rudimentary scrolling in place, both for long lists of choices and long lists of batches. But know that there's room for improvement, e.g. via sticky elements, like a sticky search bar, sticky tab bar, and maybe even sticky actions. Taking note of this in https://github.com/votingworks/arlo/issues/1655

_Long list of batches_

https://user-images.githubusercontent.com/12616928/198332310-e2289af3-f5ee-409d-b91f-06c7028ffe1e.mov

_Long list of choices_

https://user-images.githubusercontent.com/12616928/198332305-82b5b09b-0e75-4ca6-9658-1c84242d4b6a.mov

client/src/components/JurisdictionAdmin/BatchRoundDataEntry.tsx
arsalansufi: Ah yes good q - we do still need it for the case that someone explicitly clicks on a batch from the list, which of course won't always be the first batch

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,600 @@
...
+    ) {
+      selectTabWithSheetIndex(sheetFields.length - 1)
+      setAutoSelectLastTabOnceNumSheetsIs(undefined)
+    }
+  }, [
+    autoSelectLastTabOnceNumSheetsIs,
+    selectTabWithSheetIndex,
+    setAutoSelectLastTabOnceNumSheetsIs,
+    sheetFields,
+  ])
```
arsalansufi, line 173: Not my favorite part of this code 😅

I know `autoSelectLastTabOnceNumSheetsIs` is a convoluted name. The challenge here is that we use `useFieldArray` sheet field IDs as tab IDs. And when a new sheet is created, we don't have the new sheet field ID right away.

In earlier iterations of this code, I tried used other fields that we know upon sheet creation as tab IDs, e.g. sheet indexes and sheet names, but both of those created other bigger headaches, like tabs not updating properly

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,600 @@
...
+    try {
+      await saveBatchResults(resultTallySheets)
+    } catch {
+      // Errors are automatically toasted by the queryClient
+      setIsEditing(true)
+      return
+    }
+    selectVoteTotalsTab()
+    // Reset the form's isDirty value back to false
+    reset({ resultTallySheets })
```
arsalansufi, line 279: react-hook-form (a bit unintuitively) doesn't reset `isDirty` to false after successful submission

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,600 @@
...
+    const sheetNames = sheetFields.map((_, i) =>
+      i === currentSheetIndex
+        ? draftSheetName
+        : watch(`resultTallySheets[${i}].name`)
+    )
+    if (new Set(sheetNames).size < sheetNames.length) {
+      return 'Must be unique'
+    }
+    return ''
+  })()
```
arsalansufi, line 325: As you've probably started to notice, sheet name changes are handled outside of react-hook-form up until the point a new sheet name is confirmed. At that point, we call `setValue` to plug back into react-hook-form

Lots of reasons for this, the key one being that we can't always render the sheet name input (even as a hidden input) when it's tucked away inside the popover. And as you'll also see elsewhere in this PR, for react-hook-form to work ideally, inputs should always be present and not unmounting and remounting

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchRoundDataEntry.test.tsx
```diff
@@ -0,0 +1,458 @@
...
+      <BatchRoundDataEntry
+        electionId="1"
+        jurisdictionId="1"
+        roundId="round-1"
+      />
+      <ToastContainer />
+    </QueryClientProvider>
+  )
+
+describe('Batch comparison data entry', () => {
```
arsalansufi, line 69: I've essentially adapted the existing tests to work for the new UI. But there are lots of new interactions that I want to test more deeply. Took note of this in https://github.com/votingworks/arlo/issues/1655

arsalansufi:
@jonahkagan This is ready for follow-up review! I've started tracking follow ups in https://github.com/votingworks/arlo/issues/1655.

I posted a new screencapture in the PR description, though it doesn't cover all UI flows. Working on a better one but didn't want to block review on that!

arsalansufi:
After some discussion, we've decided to take this in a slightly different direction. I found it easier to open a new PR than to edit this one! https://github.com/votingworks/arlo/pull/1661

As noted in the new PR's description:

> This second attempt differs from the first attempt in that you now edit and save changes to one sheet at a time. In the first attempt, you accumulated a large number of changes across sheets and saved them all at once. The new approach has benefits from both product and technical perspectives.

PR #1647, opened 2022-10-24
QueryClient test helper

jonahkagan:
We have been using the same queryClient instance from App.tsx in all tests. This has been mostly fine, but sometimes there can be weird interactions between tests since they are using the same query cache. I started to run into problems with this (honestly surprised it didn't become an issue sooner), so decided to change to creating a new QueryClient per test case.


arsalansufi (approved):
:100:

PR #1648, opened 2022-10-24
Set query client stale time to 1s

jonahkagan:
See comment in the diff for explanation.

### DIFF
--- a/client/src/components/HomeScreen.tsx
+++ b/client/src/components/HomeScreen.tsx
@@ -409,18 +409,21 @@ const WideField = styled(FormField)`
 
 const CreateAudit = ({ organizations }: { organizations: IOrganization[] }) => {
   const history = useHistory()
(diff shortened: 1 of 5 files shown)


client/src/components/HomeScreen.tsx
```diff
@@ -409,18 +409,21 @@ const WideField = styled(FormField)`
...
+  >(
+    (newAudit: INewAudit) =>
+      fetchApi('/api/election', {
+        method: 'POST',
+        body: JSON.stringify(newAudit),
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      }),
+    { onSuccess: () => queryClient.invalidateQueries('orgs') }
```
jonahkagan, line 426: This code relied on the existing refetching behavior instead of explicitly invalidating the query cache.

arsalansufi (approved):
Nice!

PR #1651, opened 2022-10-26
Redesign Jurisdiction Manager status bar

jonahkagan:
Closes: #1624 

_Review by commits_

Redesigns the jurisdiction manager status bar to a more minimal, composable component.

Design goals:
- Provide context (which jurisdiction? which audit?) without taking up a ton of screen real estate/visual weight - the name of the audit isn't actually the most important thing on the screen, so it shouldn't be the largest
- Remove status messages. If we need to tell the user a message (e.g. you're done with your file uploads, wait for the audit to start) we should do that where they will notice it. Note that for this PR, I just removed those messages because I don't think people even notice them. We should follow up to add them in more appropriate places.
- Show batch auditing progress (there's no other way to track it in the batch interface). We could potentially use the same pattern for ballot audits, but that's out of scope for this PR (since there's already progress tracking in place)
- Minimally tweak existing screen content to work well with the new status bar. Though more comprehensive redesigns are tempting, the goal here is just to get the status bar in place.

Technical goals:
- Make the StatusBar a container to support composition
- Render the StatusBar deeper in the component hierarchy. Currently, the existing StatusBox is rendered at the top of the tree, so it needs to have its own parallel logic about the status of the audit. By rendering it deeper in the tree, it can leverage the same logic that decides what content to show on the screen.

### Audit setup - before & after
Screen Shot 2022-10-26 at 1 13 25 PM

Screen Shot 2022-10-26 at 12 52 58 PM

### Ballot audit - Create audit boards - before & after
Screen Shot 2022-10-26 at 2 33 08 PM

Screen Shot 2022-10-26 at 2 32 55 PM

### Ballot audit - offline tally entry - before & after
Screen Shot 2022-10-26 at 1 12 49 PM

Screen Shot 2022-10-26 at 1 02 49 PM

### Ballot audit - online tally entry - before & after
Screen Shot 2022-10-26 at 1 13 07 PM

Screen Shot 2022-10-26 at 12 53 16 PM

### Ballot audit - full hand tally entry (deprecated) - before & after
Screen Shot 2022-10-26 at 1 14 01 PM

Screen Shot 2022-10-26 at 12 52 17 PM

### Batch audit - tally entry - before & after 
Screen Shot 2022-10-26 at 1 13 13 PM

Screen Shot 2022-10-26 at 12 53 09 PM

### No ballots sampled - before & after
Screen Shot 2022-10-26 at 3 14 39 PM

Screen Shot 2022-10-26 at 3 13 22 PM

### Audit complete - before & after
Screen Shot 2022-10-26 at 3 03 51 PM

Screen Shot 2022-10-26 at 3 06 39 PM


jonahkagan:
@arsalansufi I still need to update tests, but this is ready for initial review otherwise.

jonahkagan:
Ok tests are updated!

client/src/components/Atoms/StatusBar.tsx
```diff
@@ -0,0 +1,51 @@
...
+  IFlexboxProps,
+} from './Layout'
+
+/**
+ * A container component that shows an audit status bar. Uses space-between to
+ * separate left/right children.
+ */
+export const StatusBar = styled(Row).attrs({
+  justifyContent: 'space-between',
+  className: Classes.TEXT_LARGE,
```
arsalansufi, line 17: I've found that the attributes provided to `attrs` aren't type checked. I haven't had a chance to dig into a fix but thought I'd flag 🎏

client/src/components/Atoms/StatusBar.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react'
+import styled from 'styled-components'
+import { Divider, Classes } from '@blueprintjs/core'
+import {
+  Row,
+  // IFlexboxProps is needed by TS in order to export StatusBar
+  // eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
```
arsalansufi, line 7: Interesting!

client/src/components/Atoms/CSVForm.tsx
```diff
@@ -12,12 +12,16 @@ import {
...
-import FormSection, { FormSectionDescription } from './Form/FormSection'
+import { FormSectionDescription } from './Form/FormSection'
 import { ErrorLabel, SuccessLabel } from './Form/_helpers'
 import { sum } from '../../utils/number'
 import FormButton from './Form/FormButton'
 import AsyncButton from './AsyncButton'
 
+// CSVFile is deprecated in favor of FileUpload
+
+const FormSection = styled.div``
```
arsalansufi, line 23: Can probably just drop `FormSection` in favor of plain divs since we aren't applying any styles (unless I'm missing something or you intend to add styles soon)

arsalansufi (approved):
LGTM! 🎨

client/src/components/Atoms/StatusBar.tsx
jonahkagan: Oh, dang, that's annoying! Thanks for the heads up

client/src/components/Atoms/CSVForm.tsx
jonahkagan: Here I was actually _removing_ the styles that are built into the existing `FormSection` component, but I didn't want to impact other forms that were relying on it. I'm hoping to replace the last instances of the `CSVForm` component soon, so all this code can be removed

jonahkagan: Switched to plain divs

PR #1652, opened 2022-10-26
Combine hybrid results

umbernhard:
This is a (for now) ugly script that runs the SUITE p-value computation on two independent audits (one ballot comparison and one ballot polling) to support California's two-step auditing procedure. This is a first step towards more broadly supporting stratified audits. 

I have tested it on a few hand-crafted test cases, but one thing I need to do before I merge is test it on some actual SUITE data to make sure it behaves the way I expect it to.


scripts/combine-hybrid-results.py
```diff
@@ -0,0 +1,119 @@
...
+    return suite.BallotComparisonStratum(
+        contest.total_ballots_cast, vote_counts, misstatements, sample_size,
+    )
+
+def ballot_polling_stratum(audit: Election, remap):
+    contest = audit.contests[0]
+    vote_counts = {choice.id: choice.num_votes for choice in contest.choices}
+
+    remapped_vote_counts = {remap[id1]: vote_counts[id1] for id1 in vote_counts}
+    suite_contest = sampler_contest.from_db_contest(contest)
```
jonahkagan, line 29: I think this is unused and can be removed

scripts/combine-hybrid-results.py
```diff
@@ -0,0 +1,119 @@
...
+    )
+    return suite.BallotComparisonStratum(
+        contest.total_ballots_cast, vote_counts, misstatements, sample_size,
+    )
+
+def ballot_polling_stratum(audit: Election, remap):
+    contest = audit.contests[0]
+    vote_counts = {choice.id: choice.num_votes for choice in contest.choices}
+
+    remapped_vote_counts = {remap[id1]: vote_counts[id1] for id1 in vote_counts}
```
jonahkagan, line 28: could probably just do the remapping the first time you define `vote_counts`

scripts/combine-hybrid-results.py
```diff
@@ -0,0 +1,119 @@
...
+    suite_contest = sampler_contest.from_db_contest(contest)
+
+    sample_results = contest_results_by_round(contest)
+    remapped_sample_results = {}
+
+    for rd in sample_results:
+        remapped_sample_results[rd] = {}
+        for id1 in sample_results[rd]:
+            remapped_sample_results[rd][remap[id1]] = sample_results[rd][id1]
```
jonahkagan, line 37: If you want to get fancy with dictionary comprehensions...
```suggestion
    sample_results = { 
      round_id: {
        remap[choice_id]: result
        for choice_id, result in round_results.items()
      }
      for round_id, round_results in contest_results_by_round(contest).items()
    }
```

jonahkagan (approved):
Looks great!

scripts/combine-hybrid-results.py
umbernhard: ah I knew there was a way to do this I just couldn't figure it out.

PR #1653, opened 2022-10-27
Button to reject tally entry login request

jonahkagan:
It can get a little annoying to have no way to clean up unused login requests. Maybe this won't be an issue in real audits, but I figured it's always nice to have a way to say no.

https://user-images.githubusercontent.com/530106/198161974-5d14ca35-18f1-4bb2-afff-32cd2373f7cb.mov


arsalansufi (approved):
Nice! I caught myself wanting this feature while testing yesterday

PR #1654, opened 2022-10-27
Improve wording of error message when CVR batch doesn't match manifest

jonahkagan:
The arranging of field names and values in the sentence was a bit confusing.


arsalansufi (approved):
:100:

PR #1658, opened 2022-10-27
Restore main python-poetry-buildpack

jonahkagan:
We were temporarily using a fork to work around a poetry deprecation: https://github.com/votingworks/arlo/pull/1579. I confirmed that the main buildpack now works using a review app.


umbernhard (approved):
Woohoo!

Issue #1655, opened 2022-10-27
Batch audit data entry improvements


arsalansufi:

arsalansufi:
Display who last updated a batch (JM or tally entry users) is documented in its own issue: https://github.com/votingworks/arlo/issues/1662

PR #1661, opened 2022-10-31
Redesign batch audit data entry to use list-and-detail view, take 2

arsalansufi:
# Overview

Issue link: https://github.com/votingworks/arlo/issues/1623

A second attempt at redesigning batch audit data entry to use a list-and-detail view that we think will be more manageable than the current "one giant table" view, especially for large batch audits like the upcoming one in Georgia!

This second attempt differs from the first attempt (https://github.com/votingworks/arlo/pull/1644) in that you now edit and save changes to one sheet at a time. In the first attempt, you accumulated a large number of changes across sheets and saved them all at once. The new approach has benefits from both product and technical perspectives.

_Tally entry user view, full demo_

https://user-images.githubusercontent.com/12616928/199528164-2b751410-5f41-4e14-b4e1-b30114ca7b37.mov

_JM view, scrolling_

https://user-images.githubusercontent.com/12616928/199530201-f0ad454a-285a-4a2b-8fcf-fe9195476cf3.mov

_Tally entry user view, tallies finalized_

tally-entry-user-view-finalized

_JM view_

jm-view

_JM view, tallies finalized_

jm-view-finalized

# Testing

- [x] Tested the jurisdiction manager UI manually
- [x] Tested the tally entry user UI manually
- [x] Tested finalizing tallies manually
- [x] Updated tests (though I want to add more, noted in https://github.com/votingworks/arlo/issues/1655)


client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,472 @@
...
+    setIsEditing(false)
+  }
+
+  return (
+    <>
+      {isEditing && sheets.length > 1 ? (
+        // A special tab bar with a sheet name form input replacing the selected tab, rendered here
+        // instead of in the parent component so that we can create a separate react-hook-form form
+        // per sheet (sharing the form across sheets requires diligent resetting)
+        <Tabs id={batch.name} selectedTabId={selectedTabId}>
```
arsalansufi, line 312: An interesting tradeoff here. I could've either gone with one react-hook-form form shared across sheets, which would've allowed for one tab bar, or two tab bars, which allows for a separate react-hook-form form per sheet (making state management bugs due to incorrect resetting less likely).

The latter felt more important to prioritize. One remaining negative side effect of the two tab bars is that, after you finish editing a sheet (and we return to a different tab bar under the hood), we reanimate the selection of the already selected tab:

199039307-625a4aed-84c6-4a97-b911-a35698608564

Still thinking through workarounds, one of which is just disabling tab bar animation altogether by setting the `animate` prop to false. Already tried toggling the `animate` prop intelligently to just prevent the initial animation and not subsequent ones, but that interfered with the selected tab indicator in other ways.

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,472 @@
...
+                className={classnames(
+                  Classes.INPUT,
+                  errors.name && Classes.INTENT_DANGER
+                )}
+                key={tab.id}
+                name="name"
+                readOnly={isSubmitting}
+                ref={register({
+                  required: true,
+                })}
```
arsalansufi, line 326: I'm no longer checking that the sheet name is unique client-side because there just isn't enough space for "Must be unique" error text. And making the input outline red without any explanatory text (as we do for this and other inputs when empty) feels like it wouldn't provide enough information to the user.

error

arsalansufi (commented):
Note that I'm tracking follow-ups in https://github.com/votingworks/arlo/issues/1655

The most important is displaying who last updated a batch (JM or tally entry users), which I know we want for upcoming audits. I just didn't want to block this PR and initial QA on that and other changes documented in that issue.

client/src/components/Atoms/ListAndDetail.tsx
```diff
@@ -0,0 +1,119 @@
...
+ *    <Detail>
+ *      {detailsForSelectedListItem}
+ *    </Detail>
+ *  </ListAndDetail>
+ */
+
+export const ListAndDetail = styled(Card)`
+  display: grid;
+  grid-template-columns: 1fr 3fr;
+  overflow-y: auto;
```
jonahkagan, line 24: I think the list and the detail panels should each scroll separately.

client/src/components/Atoms/ListAndDetail.tsx
```diff
@@ -0,0 +1,119 @@
...
+ *    </List>
+ *    <Detail>
+ *      {detailsForSelectedListItem}
+ *    </Detail>
+ *  </ListAndDetail>
+ */
+
+export const ListAndDetail = styled(Card)`
+  display: grid;
+  grid-template-columns: 1fr 3fr;
```
jonahkagan, line 23: I think to maintain visual consistency, the change in content width of the detail view shouldn't change the layout of the list/detail panels. I'm specifically thinking of the case of adding many tally sheet tabs:

https://user-images.githubusercontent.com/530106/199078316-70544926-613b-4701-9d47-ab25222839ab.mov

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,472 @@
...
+                          errors.results?.[choice.id] && Classes.INTENT_DANGER
+                        )}
+                        name={`results[${choice.id}]`}
+                        readOnly={isSubmitting}
+                        ref={register({
+                          min: 0,
+                          required: true,
+                          valueAsNumber: true,
+                        })}
+                        type="number"
```
jonahkagan, line 383: I feel like the increment/decrement buttons that show up on the right side of the input are probably not so useful here. They also aren't aligned correctly. Not a huge deal, but if they're easy to remove, that could be nice
Screen Shot 2022-10-31 at 11 11 18 AM

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,472 @@
...
+            if (selectedTabId === VOTE_TOTALS_TAB_ID) {
+              return <span>Select a sheet to make edits.</span>
+            }
+            if (isEditing) {
+              return (
+                <>
+                  <Button
+                    disabled={isSubmitting}
+                    icon="delete"
+                    intent="danger"
```
jonahkagan, line 411: I think this can have neutral intent - it's only discarding a small amount of data, and it's not actually deleting something that's already been saved. The danger intent makes it a bit too attention-grabby on the screen.

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,160 @@
...
+          }}
+        >
+          {filteredBatches.map(batch => (
+            <ListItem
+              key={batch.id}
+              onClick={() => selectBatch(batch.id)}
+              selected={batch.id === selectedBatchId}
+            >
+              {batch.name}
+            </ListItem>
```
jonahkagan, line 129: I just realized there's a task that's quite difficult in this new view: finding a batch with no tallies entered (e.g. if you want to finalize tallies, but you missed one batch). I'm not sure that we need to fix it now, but adding a simple tick mark icon to the right end of the list item might be an easy fix.

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,472 @@
...
+                          valueAsNumber: true,
+                        })}
+                        type="number"
+                        // Visually hide this input instead of completely unmounting it since
+                        // rendering no input at all interferes with react-hook-form's state
+                        // management, specifically its isDirty tracking
+                        style={{ display: !isEditing ? 'none' : undefined }}
+                      />
+                      {!isEditing && (
+                        <span>{selectedSheet?.results[choice.id] || 0}</span>
```
jonahkagan, line 390: Add `toLocaleString()` to make sure large numbers format with commas

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,472 @@
...
+                </td>
+              </tr>
+            ))}
+          </tbody>
+        </BatchResultTallySheetTable>
+
+        <BatchResultTallySheetButtonRow>
+          {(() => {
+            if (selectedTabId === VOTE_TOTALS_TAB_ID) {
+              return <span>Select a sheet to make edits.</span>
```
jonahkagan, line 403: I like the way you handled this case! Small copy tweak:

```suggestion
              return <span>Select a sheet to edit tallies</span>
```

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,472 @@
...
+          ))}
+          <Tabs.Expander />
+          {sheets.length > 1 && (
+            <Button
+              icon="add"
+              disabled={areResultsFinalized}
+              minimal
+              onClick={addSheet}
+            >
+              Add Sheet
```
jonahkagan, line 215: It might be nice to go straight into edit mode when adding a sheet. Is that easy? Bonus: auto focus the tally sheet name input as well.

If this is possible, maybe we could have empty inputs instead of default to 0 tallies.

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,472 @@
...
+              )
+            }
+            return (
+              <ButtonGroup>
+                <Button
+                  disabled={areResultsFinalized}
+                  icon="edit"
+                  onClick={() => setIsEditing(true)}
+                >
+                  {sheets.length === 1 ? 'Edit Tallies' : 'Edit Sheet'}
```
jonahkagan, line 435: On edit, can we auto focus the first tally input?

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
jonahkagan: Also, I noticed that if I click add sheet, click edit sheet, and then don't fill it out and navigate to another batch, the sheet is still saved with 0 for each tally. This makes sense in the current world, but if we automatically went into edit mode, maybe adding a sheet doesn't actually save the sheet until I enter tallies? The new sheet could be considered an unsaved change (requiring discard)

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,160 @@
...
+        </Callout>
+      )}
+      <ListAndDetail>
+        <List
+          search={{
+            placeholder: 'Search batches...',
+            setQuery: setSearchQuery,
+          }}
+        >
+          {filteredBatches.map(batch => (
```
jonahkagan, line 122: Could use an empty state when there are no matching batches. Maybe just a simple line of text: "No batches found"

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,472 @@
...
+      updatedSheets[Math.min(currentSheetIndex, updatedSheets.length - 1)].id
+    )
+    await saveBatchResults(updatedSheets.map(sheetWithoutId))
+  }
+
+  // ---------- Handlers (end) ----------
+
+  return (
+    <Detail>
+      <H4>{batch.name}</H4>
```
jonahkagan, line 190: Biggest nit of all time: can we baseline align the batch name with the text in the search field?

Screen Shot 2022-10-31 at 11 36 57 AM

client/src/components/Atoms/ListAndDetail.tsx
```diff
@@ -0,0 +1,119 @@
...
+    </ListContainer>
+  )
+}
+
+// ---------- ListItem ----------
+
+const ListItemContainer = styled.li<{ selected?: boolean }>`
+  border-bottom: 1px solid ${Colors.LIGHT_GRAY4};
+
+  button.${Classes.BUTTON} {
```
jonahkagan, line 79: Could we change the background color of the button on hover to match the border defined above? I was noticing the slight mismatch of grays creating a bit of a weird look.

jonahkagan (approved):
First review pass is only design feedback. Will do another pass right after this for technical feedback. Not sure if it's useful to separate, but figured I'd try it out.

Overall, this looks really really good! It feels much more straightforward to use! Great work!

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,472 @@
...
+  const currentSheetIndex = sheets.findIndex(
+    sheet => sheet.id === selectedTabId
+  )
+
+  // ---------- Handlers (start) ----------
+
+  const updateCurrentSheet = async (updatedSheet: IBatchResultTallySheet) => {
+    const updatedSheets = sheets.map((sheet, i) =>
+      i === currentSheetIndex ? { ...updatedSheet, id: sheet.id } : sheet
+    )
```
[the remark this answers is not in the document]
jonahkagan: Though I'm looking at the implementation of `replaceAtIndex` and this approach actually seems better haha

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,472 @@
...
+  })
+  const { errors, formState, handleSubmit, register, reset } = formMethods
+  // Important gotcha! You have to access properties on the formState to subscribe to it:
+  // https://github.com/react-hook-form/react-hook-form/issues/9002
+  const { isSubmitting, isDirty } = formState
+
+  // Communicate up to the parent whether or not there are unsaved changes
+  useEffect(() => {
+    setAreChangesUnsaved(isDirty)
+  }, [isDirty, setAreChangesUnsaved])
```
jonahkagan, line 287: Is there any way to move this into an event handler? It's [not recommended](https://beta.reactjs.org/learn/you-might-not-need-an-effect#passing-data-to-the-parent) to use an effect for this. I know that might not be possible due to how react-hook-form works

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,472 @@
...
+  selectedTabId,
+  setAreChangesUnsaved,
+  setIsEditing,
+  sheets,
+  updateSheet,
+}) => {
+  const tabs = tabsFromSheets(sheets)
+  const isTotalsSheet = selectedTabId === VOTE_TOTALS_TAB_ID
+  const selectedSheet = isTotalsSheet
+    ? undefined
```
jonahkagan, line 273: I've generally had a standard in Arlo of using `null` for intentionally "none" values rather than `undefined`. But I know that's confusing since we have the opposite standard in VxSuite...

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
jonahkagan: Oof this is tough! I agree with your prioritization, though I do love the animation. Did you try lifting up the react-hook-form and see bugs, or was that just a preventative choice? Maybe we leave it as is, disable animation for now and try to find a workaround as a follow up?

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
jonahkagan: Great thinking, I think this makes a lot of sense. We can tweak the server-side error message to be a bit more friendly to this case.

jonahkagan (commented):
First pass at a technical review. I have to run out for a bit so I didn't make it through everything, but just focused on the main parts.

client/src/components/Atoms/ListAndDetail.tsx
arsalansufi: https://github.com/votingworks/arlo/pull/1661/commits/0f91c7c368267c8d1dfe0760d7a6f5f310acaac0

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,472 @@
...
+                        ref={register({
+                          min: 0,
+                          required: true,
+                          valueAsNumber: true,
+                        })}
+                        type="number"
+                        // Visually hide this input instead of completely unmounting it since
+                        // rendering no input at all interferes with react-hook-form's state
+                        // management, specifically its isDirty tracking
+                        style={{ display: !isEditing ? 'none' : undefined }}
```
jonahkagan, line 387: This seemed weird to me, so I did a little investigation. My understanding is that `react-hook-form` determines `isDirty` by comparing the `defaultValues` to the current state of the form inputs. When we remove the inputs from the DOM, react-hook-form automatically unregisters them, so the current state of the form inputs no longer matches the default values. It looks like there's an option for `useForm` called [`shouldUnregister` ](https://legacy.react-hook-form.com/v6/api#useForm)that you can set to `false` to change this behavior so that the inputs don't get unregistered when removed from the DOM. I played around with it a bit and it seemed to solve the problem!

I'm definitely losing a bit of confidence in this whole uncontrolled inputs approach that react-hook-form takes. I know it's supposed to improve performance, but it's definitely got some tricky bits. I think this would be trivial in a controlled input context.

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,472 @@
...
+                  onClick={() => setIsEditing(true)}
+                >
+                  {sheets.length === 1 ? 'Edit Tallies' : 'Edit Sheet'}
+                </Button>
+                <Popover
+                  content={
+                    <Menu>
+                      {sheets.length === 1 ? (
+                        <MenuItem
+                          icon="applications"
```
jonahkagan, line 442: Cute icon choice!

client/src/components/TallyEntryUser/TallyEntryScreen.test.tsx
```diff
@@ -21,16 +21,13 @@ const renderScreen = () =>
...
     ]
     await withMockFetch(expectedCalls, async () => {
       renderScreen()
       await screen.findByRole('heading', { name: 'Enter Tallies' })
-      screen.getByText(
-        'For each batch, enter the number of votes tallied for each candidate/choice.'
-      )
-      screen.getByText('Batch One')
+      expect(screen.getAllByText('Batch One')).toHaveLength(2)
       // TODO fill in comprehensive tests once the tally entry interface is finalized
```
jonahkagan, line 31: can remove this comment and point to the new unit tests you wrote - they are quite comprehensive!

jonahkagan (commented):
Ok, finished up my review!

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
arsalansufi: I originally did have the react-hook-form up a level, and while I eventually worked through the bugs, the resulting code was complex and felt prone to bugs down the road, which started to defeat one of the key purposes of switching from https://github.com/votingworks/arlo/pull/1644 to this

That said, I came up with a way to toggle animation on and off that works! https://github.com/votingworks/arlo/pull/1661/commits/c844e4946da0f211188156018f722191a173ca76

I also really like the animation 😄

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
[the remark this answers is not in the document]
arsalansufi: I contemplated updating the implementation of `replaceAtIndex` to match the approach used here, but I see that the current implementation of `replaceAtIndex` allows updating indexes that aren't yet populated in the array, e.g. `replaceAtIndex([], 0, 'a')` ➡️ `['a']`, and I further see that `CodeInput` relies on that behavior. Gonna leave things as is accordingly

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
arsalansufi: Tweaked! https://github.com/votingworks/arlo/pull/1661/commits/84d792100272ef25c10241e95dc1ac4049b1ed04

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchRoundDataEntry.tsx
arsalansufi: Voila! https://github.com/votingworks/arlo/pull/1661/commits/a25fd1e3f0a221670c839a6e9602e26a636f2aa5

checks

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchRoundDataEntry.tsx
arsalansufi: https://github.com/votingworks/arlo/pull/1661/commits/61c981128507343571a70dc6058c16554606a12c

no-batches-found

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
arsalansufi: https://github.com/votingworks/arlo/pull/1661/commits/c3b9372fc957c60c575a7659161c096934e42804

arsalansufi: Decided to vertically middle align the batch name and search bar since it was a bit easier to do so precisely:

alignment

https://github.com/votingworks/arlo/pull/1661/commits/3d03b4a0df4040acdd77bdcd9d1d1d04e5d9ff67

client/src/components/Atoms/ListAndDetail.tsx
arsalansufi: https://github.com/votingworks/arlo/pull/1661/commits/7da851cd9e729e82afe9c00af758e647cd094fe9

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
arsalansufi: I started with this direction when first working on this PR, but one thing that gave me pause was the fact that the "Discard Changes" button could now do one of two things, either discard edits while leaving the sheet intact (for an existing sheet) or discard the whole sheet (for a new sheet).

Maybe that's okay if we change the copy for these two cases, e.g. "Discard Sheet" vs. "Discard Changes"

jonahkagan: Hmm, interesting point.

Imagining we leave the copy as "Discard Changes" - if you were a user and you made a new sheet and then clicked "Discard Changes," what would you expect to happen to the new sheet? I think I would expect the new sheet to get discarded completely. So maybe it won't be that confusing in practice? I worry that changing the copy might add some cognitive overhead where you have to read the button more closely.

arsalansufi: Thinking some more, this makes sense! "Discard Changes" is, in both cases, just discarding everything since the last save. In one case, that's edits to an existing sheet. In another, that's the creation of a new sheet plus edits to it, if any.

https://github.com/votingworks/arlo/pull/1661/commits/3137729a933ae05003476dd5e27b70f7269381c2

^ Required some code complexity to distinguish those two cases, but that's expected.

-----

I'm still thinking through how to best switch from 0 to empty defaults in vote choice inputs. This is straightforward for all cases but one: when you haven't yet entered any tallies and click "Use Multiple Tally Sheets". We automatically take you to Sheet 2, but Sheet 1 still needs to be filled out, too. And our new model of editing/saving one sheet at a time isn't conducive to navigating errors across multiple sheets. (I'm remembering now that this case was another reason why I diverged from the "drop you off in edit mode after creating a sheet" UI that we used in the first PR.)

The simplest approach here is probably just to fill in Sheet 1 with 0s that you can then edit.

arsalansufi: > The simplest approach here is probably just to fill in Sheet 1 with 0s that you can then edit after saving Sheet 2.

^ Ended up going with this: https://github.com/votingworks/arlo/pull/1661/commits/929c746776da909c62c66ca2a5910048bedd8567

arsalansufi: In terms of auto-focusing the name input, will discuss that in the other comment about auto-focus

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
arsalansufi: Handled this as well as auto-focusing name for newly created sheets in https://github.com/votingworks/arlo/pull/1661/commits/744c02c4c289e7879648e2e3233dd70e41c654ad!

The latest version of react-hook-form (v7) actually provides a [`setFocus` function](https://react-hook-form.com/api/useform/setfocus), but we're on a version that doesn't include this function (v6). Either way, `autoFocus` worked well (which is React's custom polyfill of the underlying HTML `autofocus` that applies on React component mount - [reference](https://blog.maisie.ink/react-ref-autofocus/#how-does-the-autofocus-prop-work))

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,160 @@
...
+  width: 100%;
+`
+
+interface IProps {
+  electionId: string
+  jurisdictionId: string
+  roundId: string
+}
+
+const BatchRoundDataEntry: React.FC<IProps> = ({
```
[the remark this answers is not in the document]
arsalansufi: Rename makes sense as does the move to the `BatchRoundSteps` directory! https://github.com/votingworks/arlo/pull/1661/commits/f2c6aa8e540a483d4cd30582619dab286e5e8da6

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
arsalansufi: So one funky finding here: After specifying `autoFocus` attributes (https://github.com/votingworks/arlo/pull/1661/commits/744c02c4c289e7879648e2e3233dd70e41c654ad), `formState.isDirty` is once again getting set to `true` when it shouldn't be, specifically for sheets with no tallies entered at the moment I click on another batch.

Removing the `autoFocus` attributes fixes this. 

https://user-images.githubusercontent.com/12616928/199392806-f750bcf8-81fa-4921-819d-7d4f6f421896.mov

To simplify the code, I'm just not going to use `formState.isDirty` and instead open the unsaved changes prompt any time you try to leave a batch while in edit mode.

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
arsalansufi: I couldn't think of a way to do so without bringing the react-hook-form form up a few levels. But for separate reasons (https://github.com/votingworks/arlo/pull/1661#discussion_r1011136794), I actually decided to stop using `formState.isDirty` altogether. I'm now just checking whether or not we're in edit mode: https://github.com/votingworks/arlo/pull/1661/commits/a814bee95694eab86dcb41bf9861ea37e2e0feb3

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
arsalansufi: I'm okay with prioritizing in-repo consistency rather than cross-repo, for now! https://github.com/votingworks/arlo/pull/1661/commits/87a25e572e818ac82cd3d1296997381eab1e2194

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,161 @@
...
+        onYesClick: () => setSelectedBatchId(batchId),
+        noButtonLabel: 'Cancel',
+      })
+      return
+    }
+    setSelectedBatchId(batchId)
+  }
+
+  return (
+    <Container>
```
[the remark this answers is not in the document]
arsalansufi: I like that! Taken care of in https://github.com/votingworks/arlo/pull/1661/commits/fa7b7e0558e9b8dd186ea0e350bd38c6dee074d4

inside-stepper

inside-stepper-finalized

client/src/components/Atoms/ListAndDetail.tsx
```diff
@@ -0,0 +1,119 @@
...
+  ${props =>
+    props.withTopBorder && `border-top: 1px solid ${Colors.LIGHT_GRAY2};`}
+  list-style: none;
+  margin: 0;
+  padding: 0;
+  padding-bottom: 40px;
+`
+
+interface IListProps {
+  search?: {
```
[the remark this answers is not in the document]
arsalansufi: Yes good q! I've been thinking of composable components vs. traditional props as a spectrum - you can take things as far as you'd like in either direction. Composable components provide more flexibility (you can use the provided composable component, e.g. `search={<ListSearch />}`, or your own, e.g. `search={<MyCustomSearch />}`), while traditional props can be simpler / more convenient (one less import, one less puzzle piece to put together).

In this case, I decided that there was little to be gained by making search composable. But am of course open to changing that down the road should the need arise for more customizable search elements.

My opinion isn't strong for this specific case, and I could convince myself of either direction

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,160 @@
...
+    batch.name.toLowerCase().includes(debouncedSearchQuery.toLowerCase())
+  )
+  const selectedBatch = batches.find(batch => batch.id === selectedBatchId)
+
+  // Auto-select first batch on initial load
+  useEffect(() => {
+    if (!selectedBatchId && filteredBatches.length > 0) {
+      setSelectedBatchId(filteredBatches[0].id)
+    }
+  }, [filteredBatches, selectedBatchId, setSelectedBatchId])
```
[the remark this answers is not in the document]
arsalansufi: > I'll offer suggestions to that effect (pun intended)

😆

The `useEffect` is gone! https://github.com/votingworks/arlo/pull/1661/commits/a3251e63d87bf9d027fd8c4f86ab77d4e15ae7ab

Decided to move data fetching into a wrapper component since I wasn't a huge fan of `selectedBatchId` possibly being `null` while `selectedBatch` wasn't.

In general, I do like using wrapper components for async data fetching. They can cover loading states, error states, and other preemptive return cases. And the components that they wrap and pass data to can then focus on core logic, knowing that data is defined.

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,160 @@
...
+      !areChangesUnsaved
+    ) {
+      setSelectedBatchId(filteredBatches[0].id)
+    }
+  }, [
+    debouncedSearchQuery,
+    filteredBatches,
+    areChangesUnsaved,
+    setSelectedBatchId,
+  ])
```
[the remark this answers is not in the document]
arsalansufi: I played around with removing this effect, but the resulting code felt more complex and prone to bugs than the effect (which I think is fairly easy to reason about for an effect), so I kept things as is.

Trying to use the search query onChange for everything is a bit tricky in this case because we have one data item that needs to be updated immediately (the search query) and another that needs to be debounced (the list of filtered batches / selected batch ID). Also, because we need the list of filtered batches when we render, not just the selected batch ID, I was having to either duplicate the list of filtered batches outside the search query onChange and inside it, or put it in state, both non-ideal.

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,472 @@
...
+
+  // ---------- Handlers (start) ----------
+
+  const updateCurrentSheet = async (updatedSheet: IBatchResultTallySheet) => {
+    const updatedSheets = sheets.map((sheet, i) =>
+      i === currentSheetIndex ? { ...updatedSheet, id: sheet.id } : sheet
+    )
+
+    // Update client-side state first for immediate UI feedback
+    setSheets(updatedSheets)
```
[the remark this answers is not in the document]
arsalansufi: Yes good q - this comment got me thinking about a few things.

I wasn't seeing flickering but an expected lag in UI feedback (noticeable if you throttle your network speed in the Chrome dev tools) for certain actions, namely sheet additions and removals (the former is not relevant any more since we aren't saving new sheets until tallies are entered and saved).

I'm realizing now that the sheet update action doesn't need this optimistic client-side state update. We're displaying values from react-hook-form state (not the `sheets` state value) and disabling other UI elements until the API call completes. So the optimistic `setSheets` call doesn't actually do anything, and we're also providing clear user feedback.

The sheet removal action, however, does benefit from an optimistic client-side state update. If we wait for the API call to complete without this update, we'd want to introduce some kind of loading state to indicate that a sheet removal is in progress and block other UI actions in the interim. In this case, the immediate feedback of the sheet going away (while we continue processing the API update behind the scenes) feels like a better UX.

Reading the article that you linked did remind me of the need to revert optimistic state updates in the case of failure, hopefully a very rare case, but important for consistency / minimal user confusion.

Relevant commits: https://github.com/votingworks/arlo/pull/1661/commits/af7a4798c644c648f5c257a06862d259b325c62b, https://github.com/votingworks/arlo/pull/1661/commits/56ef3405a9ce6e3dc5c7419e515f629a97d03c47

While I'm sure I could push some of this logic into the React query layer, I'm gonna punt on that for now

arsalansufi:
@jonahkagan This is ready for a follow-up look!

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchRoundDataEntry.tsx
jonahkagan: That looks great!

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
jonahkagan: Hmm, yes, that's a tricky case. I think the solution you chose works fine, but I do have another couple ideas for consideration:
- When switching to multiple tally sheets with no tally entered, drop the user into edit mode on Sheet 1 instead of Sheet 2 (that might help maintain the mental model of moving left-to-right to fill out the sheets)
- Don't auto-create a Sheet 2 at all. Switching to multiple tally sheets just enables multi-sheet mode, and then the second sheet has to be added by clicking add sheet. This might also have the benefit in the case where votes were already entered of showing the user immediately that their existing votes got transferred to Sheet 1. This would require allowing one sheet at a time to exist. So we'd also need a way to exit "multi-tally-sheet-mode" and collapse back to just showing vote totals (maybe a button in the Edit Sheets dropdown menu that's only enabled when there's only one sheet?)

Curious to hear if you think either of these directions is worth pursuing now

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
jonahkagan: Great!

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
jonahkagan: That works for me!

client/src/components/Atoms/ListAndDetail.tsx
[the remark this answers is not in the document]
jonahkagan: Thanks for sharing your thinking. I agree there's not really any flexibility needed here.

As I've been playing with more "logic-less" components that just do layout/styling, I've been enjoying how composability lets the markup show the layout structure a bit more clearly. E.g. in this case:

```ts
<List>
  <ListSearch />
  <ListItems>
     <ListItem />
     <ListItem />
     <ListItem />
  </ListItems>
</List>
```
Here I can see that there's a search bar on top of the list items. Or I could easily omit it or move it to the bottom (though as mentioned, we don't need that here).

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchRoundDataEntry.tsx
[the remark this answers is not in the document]
jonahkagan: Yeah, I like the wrapper approach too! I wish there were an easy way to do that inside a component - I think Suspense could be the ticket (if we ever make the upgrade to React 18)

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
arsalansufi: > When switching to multiple tally sheets with no tally entered, drop the user into edit mode on Sheet 1 instead of Sheet 2 (that might help maintain the mental model of moving left-to-right to fill out the sheets)

I like this solution for now! I think that it also shares this benefit that you mention for the second solution:

> This might also have the benefit in the case where votes were already entered of showing the user immediately that their existing votes got transferred to Sheet 1.

The second solution is more involved and will require a backend data model update if we'd like to persist whether or not you're in multi-sheet mode, so if we were to pursue it, I'd want to later. I also think that the distinction between one sheet in single-sheet mode and one sheet in multi-sheet mode _may_ create some user confusion. I'd want to hear from us internally / real users externally whether the current model is causing confusion before investing time in this change that may or may not minimize confusion.

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchRoundDataEntry.tsx
```diff
@@ -0,0 +1,160 @@
...
+            >
+              {batch.name}
+            </ListItem>
+          ))}
+        </List>
+
+        {!selectedBatch ? (
+          <Detail>
+            <p>Select a batch to enter tallies.</p>
+          </Detail>
```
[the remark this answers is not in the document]
jonahkagan: Great!

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
```diff
@@ -0,0 +1,472 @@
...
+  justifyContent: 'end',
+})`
+  min-height: 30px;
+`
+
+const VOTE_TOTALS_TAB_ID = 'vote-totals'
+
+interface IBatchResultTallySheetWithId extends IBatchResultTallySheet {
+  id: string
+}
```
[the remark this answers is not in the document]
jonahkagan: Ah that makes sense. Maybe a comment explaining the need for ids could be helpful for posterity.

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
jonahkagan: Incredible sidestep! Can't believe we never thought of this option before. It actually feels better as a user to know that I can never accidentally exit edit mode. And the code is much simpler now 👏

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchRoundDataEntry.tsx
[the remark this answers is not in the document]
jonahkagan: Ooh the full-bleed callout is particularly svelte!

jonahkagan (approved):

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
arsalansufi: https://github.com/votingworks/arlo/pull/1661/commits/b853017fc6e710f95ed7007a5e7046c59862633c

https://user-images.githubusercontent.com/12616928/199578535-244e1b97-266e-4aaf-a589-ef812fea96f2.mov

client/src/components/Atoms/ListAndDetail.tsx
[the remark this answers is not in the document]
arsalansufi: Yes, great point in favor of composable components!

A side note: I've often encountered the need to create props for some of the composable elements, instead of specifying them all as children, e.g.

```
<List search={<ListSearch />}>
  <ListItems>
    <ListItem />
    <ListItem />
  </ListItem>
</List>
```

This can be helpful if, for example, you need to know whether a search element has been provided or not, and you need to toggle styles accordingly. Or if you have to render an element somewhere that isn't purely before or after the other elements

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
arsalansufi: :tada: Feels so obvious in retrospect!

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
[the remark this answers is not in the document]
arsalansufi: :100: https://github.com/votingworks/arlo/pull/1661/commits/db43b9892b151ce40a7f0caaae18998837d83815

client/src/components/JurisdictionAdmin/BatchRoundSteps/BatchRoundTallyEntry/BatchDetail.tsx
```diff
@@ -0,0 +1,572 @@
...
+  }
+
+  const onValidSubmit: SubmitHandler<IBatchResultTallySheet> = async sheet => {
+    try {
+      await updateSheet(sheet)
+    } catch {
+      // Errors are automatically toasted by the queryClient
+      return
+    }
+    // Reset the form's isDirty value back to false
```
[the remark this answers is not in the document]
arsalansufi: Good catch! The `reset` call is no longer needed either now

client/src/components/Atoms/ListAndDetail.tsx
```diff
@@ -0,0 +1,152 @@
...
+ *      <ListItem selected>Item 2</ListItem>
+ *    </List>
+ *    <Detail>
+ *      {detailsForSelectedListItem}
+ *    </Detail>
+ *  </ListAndDetail>
+ */
+
+export const ListAndDetail = styled(({ fullBleed, ...props }) => (
+  <Card {...props} />
```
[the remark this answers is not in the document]
arsalansufi: Great call! https://github.com/votingworks/arlo/pull/1661/commits/6948dfceed69ce6345650c41620c59d01a077cad

client/src/components/JurisdictionAdmin/BatchRoundDataEntry/BatchDetails.tsx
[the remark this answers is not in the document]
arsalansufi: Updated ✅

Required a few new state values and switching the `Popover` from uncontrolled usage to controlled: https://github.com/votingworks/arlo/pull/1661/commits/ea96177f10e58b3a1b71b23ef47f77b2e7167adf

https://user-images.githubusercontent.com/12616928/199595160-2b30b72d-7c46-4900-8ab5-dcc284f20a4d.mov

PR #1667, opened 2022-11-02
Manual test fixture kit

jonahkagan:
Closes: #1660 

_Review by commits_

Adds a new directory `fixtures` with a script to generate fake election data. Documentation to explain it further is available in `fixtures/README.md`.

There are still some important cases not covered by this initial draft of the kit (e.g. multi-winner/vote-for-n contests, a contest that triggers a full hand tally) but they will be easy to add as follow ups.


fixtures/test/test_generate_election.py
```diff
@@ -0,0 +1,166 @@
...
+                    batch_tallies, cvrs, tallies[contest["name"]]
+                )
+                snapshot.assert_match(batch_tallies)
+
+
+# We don't snapshot the results of generating fixtures for this election, since
+# we may want to change the spec for other reasons. However, we can still check
+# that the resulting data is valid (using property assertions). It's important
+# to test with a slightly larger election because there are edge cases that only
+# show up at a slightly larger scale.
```
arsalansufi, line 141: Sanity check: This comment mentions testing with a "slightly larger" election, and this test is called "test_small_election" - is this basically a TODO to add a test for a larger election?

arsalansufi (approved):
This is great!! Excited to build more fixtures on top of this base

fixtures/test/test_generate_election.py
jonahkagan: Haha I see how that is confusing. `small-election.spec.json` is one of the real fixtures. It's larger than `simple_election`, but still on the small side. I'll update this comment to clarify

PR #1672, opened 2022-11-03
Added handling for if Decimal is NaN

umbernhard:
Occasionally, in our sample size probability computation, some of the internal variables can become NaN (due to the way the stats math works, if the numbers are too big/small). This makes sure that we gracefully handle that case.

### DIFF
--- a/server/tests/audit_math/test_bravo.py
+++ b/server/tests/audit_math/test_bravo.py
@@ -395,6 +395,10 @@ def test_compute_risk(contests):
             ("Biden", "Overvote"): 0.0,
             ("Biden", "Undervote/Blank"): 0.0,
         },
+        "test_caroline": {
+            ("Erin George", "Alex Hall"): 0.13,
+            ("Erin George", "Write In"): 0.13,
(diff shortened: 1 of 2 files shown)


server/tests/audit_math/test_bravo.py
```diff
@@ -395,6 +395,10 @@ def test_compute_risk(contests):
             ("Biden", "Overvote"): 0.0,
             ("Biden", "Undervote/Blank"): 0.0,
         },
+        "test_caroline": {
```
arsalansufi, line 398: As much as I love calling this the Caro test 😆, a name that describes the situation might be more helpful down the road

arsalansufi, line 398: As much as I love calling this the Caro test 😆, a name that describes the situation might be more helpful down the road

umbernhard: :sweat_smile: fair point!

arsalansufi (approved):
LGTM!

Issue #1662, opened 2022-10-31
Track and show which user last edited a batch


jonahkagan:
This should be shown in two places:
- The tally entry interface (for both JMs and tally entry users)
- The audit report

Note that the last editor might be a JM or tally entry user.

PR #1673, opened 2022-11-04
Track who last edited a batch and include in batch tally entry UI and reports

arsalansufi:
# Overview

Issue link: https://github.com/votingworks/arlo/issues/1662

This PR updates the batch tally entry UI to display who last edited a batch, whether it was a jurisdiction admin, tally entry user, or support user. Especially important with the introduction of tally entry users since more than one user can now enter batch tallies! The PR also adds who last edited a batch to batch audit reports.

ja

tally-entry-user

https://user-images.githubusercontent.com/12616928/200065796-7d783fd2-bc80-4210-be9c-e9c78faabc28.mov

# Testing

- [x] Added automated tests
- [x] Tested batch tally entry UI manually
- [x] Tested batch audit report generation manually

-----

_Updated screenshot (bolding removed, colon added)_

updated


server/models.py
```diff
@@ -404,6 +404,10 @@ class Batch(BaseModel):
         order_by="BatchResultTallySheet.created_at",
     )
 
+    last_edited_by_user_type = Column(String(200))
+    last_edited_by_user_key = Column(String(200))
+    last_edited_by_support_user_email = Column(String(200))
```
arsalansufi, line 409: I considered a number of possible schema, including foreign keys into the users / tally entry users tables, but ultimately settled on these three session fields as a simple way to identify users of various types

server/api/batches.py
```diff
@@ -82,9 +82,27 @@ def serialize_batch(batch: Batch) -> JSONDict:
             }
             for tally_sheet in batch.result_tally_sheets
         ],
+        "lastEditedBy": construct_batch_last_edited_by_string(batch),
```
arsalansufi, line 85: Decided to return a single processed value to the client instead of the three underlying fields (`last_edited_by_user_type`, `last_edited_by_user_key`, and `last_edited_by_support_user_email`), keeping the complexities of user identification encapsulated on the server.

I considered storing the single processed value in the DB instead of the three separate fields, but that would make it harder to change format down the road as well as to cross reference source-of-truth records.

server/api/batches.py
```diff
@@ -82,9 +82,27 @@ def serialize_batch(batch: Batch) -> JSONDict:
...
+def construct_batch_last_edited_by_string(batch: Batch) -> Optional[str]:
+    if batch.last_edited_by_support_user_email:
+        return batch.last_edited_by_support_user_email
+    if batch.last_edited_by_user_type == UserType.JURISDICTION_ADMIN:
+        return batch.last_edited_by_user_key  # Email
+    if batch.last_edited_by_user_type == UserType.TALLY_ENTRY:
+        tally_entry_user = TallyEntryUser.query.filter_by(
+            id=batch.last_edited_by_user_key
+        ).one_or_none()
+        if tally_entry_user is not None:
```
arsalansufi, line 98: One edge case that I've been thinking through is how to handle the case that the relevant tally entry user no longer exists, say because they were deleted.

This concern doesn't exist for deleted jurisdiction admins because we'll always have an email in `last_edited_by_user_key`.

client/cypress/support/commands.js
```diff
@@ -57,7 +57,9 @@ Cypress.Commands.add('logout', email => {
   cy.intercept('/auth/logout').as('logout')
   cy.intercept('/api/me').as('me')
   if (email) {
-    cy.findByText(email).click()
+    cy.findAllByText(email)
+      .eq(0)
```
arsalansufi, line 61: Had to add this because there can now be two elements on the page with the user email, the account button and the "Last edited by..." copy.

I tried `cy.findByRole('button', { name: email })`, but for whatever reason, that didn't work

server/models.py
arsalansufi, line 409: I considered a number of possible schema, including foreign keys into the users / tally entry users tables, but ultimately settled on these three session fields as a simple way to identify users of various types

server/api/batches.py
arsalansufi, line 85: Decided to return a single processed value to the client instead of the three underlying fields (`last_edited_by_user_type`, `last_edited_by_user_key`, and `last_edited_by_support_user_email`), keeping the complexities of user identification encapsulated on the server.

I considered storing the single processed value in the DB instead of the three separate fields, but that would make it harder to change format down the road as well as to cross reference source-of-truth records.

server/api/batches.py
arsalansufi, line 98: One edge case that I've been thinking through is how to handle the case that the relevant tally entry user no longer exists, say because they were deleted.

This concern doesn't exist for deleted jurisdiction admins because we'll always have an email in `last_edited_by_user_key`.

client/cypress/support/commands.js
arsalansufi, line 61: Had to add this because there can now be two elements on the page with the user email, the account button and the "Last edited by..." copy.

I tried `cy.findByRole('button', { name: email })`, but for whatever reason, that didn't work

client/src/components/JurisdictionAdmin/BatchRoundSteps/BatchRoundTallyEntry/BatchDetail.tsx
```diff
@@ -539,6 +546,12 @@ const BatchResultTallySheet: React.FC<IBatchResultTallySheetProps> = ({
         </BatchResultTallySheetTable>
 
         <BatchResultTallySheetButtonRow>
+          {batch.lastEditedBy && !isEditing && (
+            <LastEditedByText>
+              Last edited by <strong>{batch.lastEditedBy}</strong>
```
jonahkagan, line 551: I like the use of gray here to lower this in the visual hierarchy. Suggest using `Classes.TEXT_MUTED` for consistency with other gray text. The bold kind of cancels it out though, so I might remove the bold.

Also maybe adding a colon could make it a bit more readable (especially in the tally entry user case): e.g. `Last edited by: Shrek, Donkey`

server/models.py
benadida: Yeah you could unify all users in one table, but that's a very big change.

I'm a tad irked by the fact that `user_key` really refers to different fields in different tables depending on `user_type`. But I don't have anything better to propose, so I'm okay with it. Maybe document a little bit more what `user_key` is here?

arsalansufi: > One more general solution might be to model all users in one table with extra tables for data specific to each user type. That might be out of scope for this task though.

Modeling all users in one table definitely crossed my mind but of course felt out of scope

> That being said, maybe we can ask if this solution could be easily migrated to that one. I think the answer is yes, because we are storing enough info to reference the existing user record.

:100: Yeah, this was definitely a priority, being able to identify the exact user. That's why I avoided a simple string model, e.g. `last_edited_by: "Alice, Bob"`

> I'm a tad irked by the fact that `user_key` really refers to different fields in different tables depending on `user_type`. But I don't have anything better to propose, so I'm okay with it. Maybe document a little bit more what `user_key` is here?

Will add a comment! For reference, I picked these names to match the names of the fields that we store on the session:

https://github.com/votingworks/arlo/blob/551d8c4e223672ca1abfb502dba1127c3c45c5ef/server/auth/auth_helpers.py#L51-L54

I'll note one last alternative that I considered that would've avoided the vague names: separate fields for the different user types, e.g. `last_edited_by_jurisdiction_admin_email`, `last_edited_by_tally_entry_user_id`, etc. A downside to this model was that we'd have to enforce that only one of the fields was filled in at a time with application code.

server/api/batches.py
arsalansufi: Agree that the former is preferable to the latter for the exact reason you mention! Will add that assertion + comment

client/cypress/support/commands.js
arsalansufi: Ah yes, good catch!

client/src/components/JurisdictionAdmin/BatchRoundSteps/BatchRoundTallyEntry/BatchDetail.tsx
arsalansufi: ✅

Added an updated screenshot to the end of the PR description

server/api/reports.py
```diff
@@ -713,6 +714,7 @@ def sampled_batch_rows(election: Election, jurisdiction: Jurisdiction = None):
             "Audited?",
             "Audit Results",
             "Reported Results",
+            "Last Edited By",
```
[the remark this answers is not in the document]
arsalansufi: I admittedly had no idea what discrepancy was 🙈

Now that I've taken a closer look and understand what it is, 100% agree that it should be next to "Audit Results" and "Reported Results"

server/tests/helpers.py
```diff
@@ -279,14 +279,22 @@ def run_audit_round_all_blanks(
...
 
 
+def scrub_uuid(string: str) -> str:
+    return re.sub(UUID_REGEX, "UUID", string)
+
+
 def assert_match_report(report_bytes: bytes, snapshot):
     report = report_bytes.decode("utf-8")
-    snapshot.assert_match(scrub_datetime(report))
+    snapshot.assert_match(scrub_uuid(scrub_datetime(report)))
```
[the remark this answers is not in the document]
arsalansufi: Good point! Made the UUID scrubbing more targeted: https://github.com/votingworks/arlo/pull/1673/commits/cf11cf953d62ad4da95f3587b6ba23d9d2be9315

arsalansufi:
@jonahkagan I've addressed all comments!

That said, there's one last thing that I'm thinking through - how performant this logic will be when the number of batches is super large and all have been edited by tally entry users. In that case, we'll loop and make a separate query for every batch to get the tally entry user names, which could be slow (and generate a lot of DB load, though maybe not too much since the queries wouldn't be running in parallel as far as I can tell).

I'm looking into tweaking the logic to use either a SQL JOIN or SQL IN (e.g. `SELECT * WHERE tally_entry_user.id IN list_of_tally_entry_user_ids`)

jonahkagan:
Good thought. Worth trying it out to see. That might also be an argument for modeling this with separate fields for each user type, since that would allow you to join in the original query.

server/models.py
```diff
@@ -404,7 +405,24 @@ class Batch(BaseModel):
...
 
-    __table_args__ = (UniqueConstraint("jurisdiction_id", "tabulator", "name"),)
+    # Only one of the following should be populated, enforced via a check constraint defined below
+    last_edited_by_support_user_email = Column(String(200))
+    last_edited_by_jurisdiction_admin_email = Column(String(200))
+    last_edited_by_tally_entry_user_id = Column(
+        String(200), ForeignKey("tally_entry_user.id", ondelete="set null")
+    )
+
+    last_edited_by_tally_entry_user = relationship("TallyEntryUser")
```
arsalansufi, line 415: My understanding of SQL Alchemy relationships is very rudimentary. It's not clear to me what other params I'd want to specify here (e.g. I see that every other relationship specifies `back_populates`)

server/models.py
```diff
@@ -404,7 +405,24 @@ class Batch(BaseModel):
...
+
+    __table_args__ = (
+        UniqueConstraint("jurisdiction_id", "tabulator", "name"),
+        CheckConstraint(
+            "(cast(last_edited_by_support_user_email is not null as int) +"
+            " cast(last_edited_by_jurisdiction_admin_email is not null as int) +"
+            " cast(last_edited_by_tally_entry_user_id is not null as int)) <= 1",
+            "only_one_of_last_edited_by_fields_is_specified_check",
+        ),
```
arsalansufi, line 424: Addressed one of my main concerns about this alternative schema: That we have to somehow make sure that only one of the three `last_edited_by` fields is specified.

This constraint will now guarantee that! Manually verified that it works

server/models.py
```diff
@@ -404,7 +405,24 @@ class Batch(BaseModel):
...
+    )
```
arsalansufi, line 425: Also not totally clear to me is whether I need to tweak the batches query at all:

https://github.com/votingworks/arlo/blob/4fd1051dfefdb2a98cc9d277a189dc8938ab49ae/server/api/batches.py#L100-L112

We're joining with the tally entry users table automatically, and everything is working as far as I can tell. But I'm wondering whether I need to specify any `.options`

arsalansufi (commented):
> That might also be an argument for modeling this with separate fields for each user type, since that would allow you to join in the original query.

Yup! This convinced me to switch over

A few points that I'd like to sanity check with you @jonahkagan since I'm ramping up on SQL Alchemy for the first time, but otherwise, everything is working!

server/models.py
arsalansufi: Update: Did end up switching things up, mainly to avoid an inefficient query fanout: https://github.com/votingworks/arlo/pull/1673#issuecomment-1310887267

> A downside to this model was that we'd have to enforce that only one of the fields was filled in at a time with application code

Found a way to do this without application code and with a SQL constraint instead! https://github.com/votingworks/arlo/pull/1673#discussion_r1019718841

server/models.py
jonahkagan: This looks fine!

`back_populates` is only needed if you have a corresponding relationship defined in the other direction as well. E.g. in this case if you wanted a relationship like `TallyEntryUser.batch_tally_sheets_that_i_last_edited`

arsalansufi: (I left `back_populates` empty because I don't envision any cross-editing needs, e.g. I don't think we should be modifying the tally entry user through `last_edited_by_tally_entry_user`. But I might not be thinking about this in the right way)

server/models.py
```diff
@@ -404,14 +405,24 @@ class Batch(BaseModel):
...
-    # Match our session fields
-    last_edited_by_user_type = Column(String(200))
-    last_edited_by_user_key = Column(
-        String(200)
-    )  # An email for jurisdiction admins, a user ID for tally entry users
+    # Only one of the following should be populated, enforced via a check constraint defined below
     last_edited_by_support_user_email = Column(String(200))
+    last_edited_by_jurisdiction_admin_email = Column(String(200))
+    last_edited_by_tally_entry_user_id = Column(
+        String(200), ForeignKey("tally_entry_user.id", ondelete="set null")
```
jonahkagan, line 412: Did we want to say you can't delete a tally entry user if they've edited batches? If so, I think removing this `ondelete` will enforce that at the db level

server/models.py
jonahkagan: Wow cool!

server/models.py
jonahkagan: The way relationships work in SQLAlchemy is that when you access a relationship field on a db object, it fires off a select query for that related object. E.g. accessing `tally_sheet.last_edited_by_tally_entry_user` will trigger a select for `tally_sheet.last_edited_by_tally_entry_user_id` from the `tally_entry_user` table.

In order to join those in during the original query that loads the tally sheets, you do need to specify it with `joinedload`. So in this case, we should have this:

```py
.options(
  joinedload(Batch.result_tally_sheets).joinedload( 
    BatchResultTallySheet.results,
    BatchResultTallySheet.last_edited_by_tally_entry_user
  )
)
```

server/models.py
```diff
@@ -404,14 +405,24 @@ class Batch(BaseModel):
...
     )
 
-    # Match our session fields
-    last_edited_by_user_type = Column(String(200))
-    last_edited_by_user_key = Column(
-        String(200)
-    )  # An email for jurisdiction admins, a user ID for tally entry users
+    # Only one of the following should be populated, enforced via a check constraint defined below
     last_edited_by_support_user_email = Column(String(200))
+    last_edited_by_jurisdiction_admin_email = Column(String(200))
```
jonahkagan, line 410: I wonder if this should be a foreign key to the `User` table now

server/models.py
arsalansufi: Got it - thanks for clarifying!

server/models.py
arsalansufi: Assuming a foreign key on user ID*, I went back and forth on this. Because the session contains user email, and that's ultimately what we want to display, it initially felt roundabout to retrieve an ID given email at write time and convert back from ID to email at read time.

That said, consistency in how we handle JAs and tally entry users would be nice! And a foreign key on user ID is probably what we'll want if/when we have a universal users table. I ultimately did decide to switch for these reasons.

https://github.com/votingworks/arlo/pull/1673/commits/2b75cbba2c82e93b4a7b8d868ffa0f6319c5c0d9

\*I know we could also set up a foreign key on user email since it's unique, but user ID felt ideal for the other reasons outlined above

server/models.py
arsalansufi: Got it - very helpful! Added `joinedload`s to make sure we use JOINs and not multiple SELECTs (defeating the purpose of this schema switch). (Looked a little bit different than the above since last edited tracking happens at a batch level and not a sheet level.)

https://github.com/votingworks/arlo/pull/1673/commits/7e21986c408a678866ef6712dc80b4f06f021275

client/src/components/JurisdictionAdmin/BatchRoundSteps/BatchRoundTallyEntry/BatchRoundTallyEntry.test.tsx
```diff
@@ -162,6 +164,7 @@ describe('Batch comparison data entry', () => {
       row2 = within(rows[2]).getAllByRole('cell')
       expect(row1[1]).toHaveTextContent('1')
       expect(row2[1]).toHaveTextContent('2')
+      screen.getByText(hasTextAcrossElements('Last edited by: ja@example.com'))
```
jonahkagan, line 167: I think this `hasTextAcrossElements` is no longer necessary

server/api/batches.py
```diff
@@ -82,9 +82,33 @@ def serialize_batch(batch: Batch) -> JSONDict:
...
 
+def construct_batch_last_edited_by_string(batch: Batch) -> Optional[str]:
+    if batch.last_edited_by_support_user_email:
+        return batch.last_edited_by_support_user_email
+    if batch.last_edited_by_user_id:
+        user = batch.last_edited_by_user
+        assert (
+            user is not None
+        ), f"Unable to find user with ID {batch.last_edited_by_user_id}"
+        return user.email
```
jonahkagan, line 97: I think you can simplify this without hurting performance due to the joinedload
```suggestion
    if batch.last_edited_by_user:
        return batch.last_edited_by_user.email
```

jonahkagan (approved):
Nice work! Quite a winding path to arrive at this solution, but it seems solid

client/src/components/JurisdictionAdmin/BatchRoundSteps/BatchRoundTallyEntry/BatchRoundTallyEntry.test.tsx
arsalansufi: Good catch!

PR #1688, opened 2022-11-09
BRAVO: Fix division-by-zero error

### LINKED ISSUE #1678: Investigate division by zero error
https://sentry.io/organizations/votingworks/issues/3722074231/?project=5504628

From @mattroe:
> I was testing offline tally entry and just entered 0s for all but one candidate.

jonahkagan:
Fixes #1678 

Here's the line that had the divide-by-zero error: https://github.com/votingworks/arlo/blob/015c7fba37457933de77894e74628f2cd8d4314d/server/audit_math/bravo.py#L530

### DIFF
--- a/server/audit_math/bravo.py
+++ b/server/audit_math/bravo.py
@@ -7,6 +7,7 @@ Note that this library works for one contest at a time, as if each contest being
 targeted is being audited completely independently.
 """
 import math
+import decimal
 from decimal import Decimal
 from collections import defaultdict
 import logging
@@ -15,6 +16,9 @@ from scipy import stats
 
 from .sampler_contest import Contest
 from .ballot_polling_types import SampleSizeOption
+
+# Treat n/0 as Infinity instead of raising an error
+decimal.getcontext().traps[decimal.DivisionByZero] = False
 
 
 def get_expected_sample_size(
--- a/server/audit_math/bravo.py
+++ b/server/audit_math/bravo.py
@@ -7,7 +7,6 @@ Note that this library works for one contest at a time, as if each contest being
 targeted is being audited completely independently.
 """
 import math
-import decimal
 from decimal import Decimal
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


server/audit_math/bravo.py
```diff
@@ -16,6 +17,9 @@
 from .sampler_contest import Contest
 from .ballot_polling_types import SampleSizeOption
 
+# Treat n/0 as Infinity instead of raising an error
+decimal.getcontext().traps[decimal.DivisionByZero] = False
```
jonahkagan, line 21: @umbernhard curious what you think about this approach vs handling the specific divide-by-zero case. I generally would lean towards failing loudly, but it seems more useful to have division be well-defined across the integers for our math. If you like this approach, we could enable it for other audit methods as well.

arsalansufi (approved):
This approach seems reasonable to me!

jonahkagan:
@umbernhard ok, in that case I think I will just fix the specific case, since I'm not 100% confident that changing this setting won't impact other audit methods (not totally sure how the setting state is shared b/w python modules, and I don't want to find out)

Re: making things less clear, how would you differentiate "bad data" from this example case? In other words, what's "bad data" vs just unfavorable audit results?

umbernhard:
@jonahkagan I'm mostly thinking about when new people start playing with Arlo and enter in election info that isn't very representative of "real" elections. Like if you enter a really small contest (like 10 total votes), or a tie, it's conceivable to me that you could hit a corner case where something goes to 0 and the math throws an error. Instead of getting a red banner that says error, escalating to a full hand count may not be as obvious an indicator to folks that the data they entered wasn't really the type that Arlo is supposed to handle. Idk, I may be making mountains out of mole hills.

jonahkagan:
Gotcha, thanks for explaining. That makes me think that we should add guardrails for those types of cases, but that might be a separate task.

PR #1693, opened 2022-11-09
Convert AuditAdminView to react-query

jonahkagan:
Task: #1684 

_Review by commits_

My goal was to optimize the audit progress screen, which was requesting jurisdiction status 3 times on load. The refreshing logic that controls this is quite gnarly, so I needed to convert all of the data loading in AuditAdminView to react-query in order to unravel it.


client/src/components/AuditAdmin/useRoundsAuditAdmin.ts
```diff
@@ -65,57 +43,59 @@ export const drawSampleError = (rounds: IRound[]): string | null =>
...
-): [
-  IRound[] | null,
-  (sampleSizes: ISampleSizes) => Promise<boolean>,
-  () => Promise<boolean>
-] => {
-  const [rounds, setRounds] = useState<IRound[] | null>(null)
+  options?: UseQueryOptions<IRound[], ApiError, IRound[], string[]>
+): UseQueryResult<IRound[], ApiError> =>
+  useQuery(
+    ['elections', electionId, 'rounds'],
```
arsalansufi, line 51: I don't feel strongly about this, but could move the construction of this key to a function so that it can be reused in the invalidation calls below, too

I could see someone accidentally updating this key while forgetting to update the keys below

arsalansufi (approved):
LGTM!

client/src/components/AuditAdmin/useRoundsAuditAdmin.ts
jonahkagan: Good point. I'll do that. I'd like to figure out an approach that can work for all our queries, since they all have keys...

client/src/components/AuditAdmin/Setup/Contests/ContestForm.tsx
```diff
@@ -381,26 +360,21 @@ const ContestForm: React.FC<IProps> = ({
               )}
             />
           </FormWrapper>
-          {nextStage.state === 'processing' ? (
-            <Spinner />
-          ) : (
```
arsalansufi, line 378: Is some other loading indicator replacing this and other spinners removed in this commit? Or is a loading indicator no longer necessary? (My simplistic search for "Spinner" in this commit didn't surface any replacements)

arsalansufi (approved):
Thanks for breaking this refactor up into digestible commits! Code looks good

In terms of whether to move forward with this even though audit season has already begun, I'm fine with moving forward and trust your judgment!

client/src/components/AuditAdmin/Setup/Contests/ContestForm.tsx
jonahkagan: I couldn't figure out what this spinner was actually needed for. The submit button has a spinner using the `loading` property, so that seemed sufficient to me

jonahkagan:
Ok tests are done and this is ready for a new review. The first unreviewed commit is [76997d0](https://github.com/votingworks/arlo/pull/1693/commits/76997d0e4d94ce77de535149caabbb4323c09314)

carolinemodic:
A little stressed about how large this PR ended up being but at a high level it all makes sense to me. I didn't review every commit in super detail but the general idea seems reasonable.

jonahkagan:
Yeah it got huge. Maybe a record number of commits for me

PR #1696, opened 2022-11-14
Improve process for installing/running arlo locally

combscCode:
Made 4 changes to get the make commands working in devcontainers using the ubuntu-20.04 image.

* update apt before searching for packages
* add the -y flag to all `apt install` invocations. Previously only one invocation had the flag, just wanted to make it consistent and easy to use.
* start the postgresql service before trying to call it. Uses the service module, I *believe* this should work even if you use systemctl to manage services but I'd appreciate someone double checking me on that one.
* Check for whether AWS env vars are set before assuming you can write to an s3 bucket in cypress tests.


combscCode:
hmmm need to sort out the cypress tests...

Makefile
```diff
@@ -2,14 +2,16 @@
...
+	sudo apt install -y python3.8 python3.8-venv libpython3.8-dev libpq-dev graphicsmagick
 	# Install node: https://github.com/nodesource/distributions/blob/master/README.md#deb		
 	curl -fsSL https://deb.nodesource.com/setup_16.x | sudo -E bash -
 	sudo apt-get install -y nodejs
 	# Install poetry: https://python-poetry.org/docs/master/#osx--linux--bashonwindows-install-instructions 
 	curl -sSL https://install.python-poetry.org | python3.8 -	
 	sudo npm install -g yarn
-	sudo apt install postgresql
+	sudo apt install -y postgresql
+	sudo service postgresql start
```
umbernhard, line 14: This is a good nudge, I forgot we hadn't been starting the psql service. My only gripe: I would recommend replacing `service` with `systemctl start postgresql`, `service` is a much older way of doing things and may become deprecated in the future.

umbernhard (approved):
Looks good to me, other than the one nudge about `service`!

umbernhard:
This looks good to me, but I'd like @jonahkagan to take a look especially at the vscode stuff (I don't use vscode...)

Makefile
```diff
@@ -29,6 +31,12 @@ resetdb:
 
 dev-environment: deps initdevdb install-development resetdb
 
+install-cypress:
```
jonahkagan, line 34: What's the motivation behind this new target? I think cypress is already installed as a dep in package.json

jonahkagan (changes requested):
Thanks @combscCode for this PR. I have a few suggestions for scoping it down.

Makefile
combscCode: Hmmm the ubuntu-20 image I'm using doesn't come with some of the cypress dependencies out of the box, so even though yarn is installing cypress it's not enough. I saw in the README that y'all recommend running this in Ubuntu-20, so I figured it'd be good to add a makefile step for installing cypress deps.

Happy to keep this in my fork though, I probably should just be using an ubuntu image that comes with the cypress dependencies anyways.

jonahkagan: Ah gotcha. I can't remember if the Ubuntu image I'm using came with the deps or not. I'll keep an eye out for this next time I install Ubuntu and if there are missing deps, will add them to the Makefile. Thanks for pointing this out.

jonahkagan (approved):

PR #1701, opened 2022-11-15
Batch inventory: improve error messages for wrong tabulator status file type

jonahkagan:
There are a few different version of the tabulator status file that are exported by the EMS. Here, we add a few checks to detect the wrong version that Arlo doesn't know how to parse (an Excel version and an HTML version).


arsalansufi (approved):
LGTM!

Issue #1568, opened 2022-08-18
Error when selecting District of Columbia as state


ginvdr:
To repro:
1. Create an audit
2. On the audit settings screen, select "District of Columbia" from the "State" dropdown
3. Go to the Audit Progress screen

An error will occur:
```
Error: State topology not found: District Of Columbia
```

combscCode:
👋 Taking a look at this one.

PR #1702, opened 2022-11-15
Fix dc audit

combscCode:
Related to: https://github.com/votingworks/arlo/issues/1568

This should fix a bug that was originating in `/workspaces/arlo/client/src/components/AuditAdmin/Progress/ProgressMap.tsx`. Basically, there's a string comparison looking for a matching state name and DC should be `District of Columbia`, not `District Of Columbia` with a capitalized "of".

Another option to fix this would be to change the string comparison to be case insensitive. But given that "District of Columbia" is the correct way, I figured this made more sense.

### DIFF
--- a/client/src/components/AuditAdmin/Setup/Settings/states.ts
+++ b/client/src/components/AuditAdmin/Setup/Settings/states.ts
@@ -7,7 +7,7 @@ export const states: { [abbreviation: string]: string } = {
   CO: 'Colorado',
   CT: 'Connecticut',
   DE: 'Delaware',
-  DC: 'District Of Columbia',
+  DC: 'District of Columbia',
   FL: 'Florida',
   GA: 'Georgia',
   HI: 'Hawaii',


client/run-cypress-tests.sh
```diff
@@ -4,7 +4,10 @@ export ARLO_SMTP_HOST=localhost
 export ARLO_SMTP_PORT=1025
 export ARLO_SMTP_USERNAME=cypress-smtp-username
 export ARLO_SMTP_PASSWORD=cypress-smtp-password
-export ARLO_FILE_UPLOAD_STORAGE_PATH=s3://arlo-file-uploads-dev
+if [[ -n $AWS_ACCESS_KEY_ID ]] && [[ -n $AWS_SECRET_ACCESS_KEY ]]
+then
+    export ARLO_FILE_UPLOAD_STORAGE_PATH=s3://arlo-file-uploads-dev
+fi
```
jonahkagan, line 10: This change seems unrelated to this PR. Can you remove it? Feel free to make a separate PR for it if you'd like

jonahkagan (changes requested):
Thanks for catching this @combscCode!

client/run-cypress-tests.sh
combscCode: Ah yes, unfortunately this change is necessary for me to pass the cypress tests since my build pipeline account doesn't have any aws access key env vars set (I assume this is why the build fails for me but not for others working on the project?).

I've got another PR over [here](https://github.com/votingworks/arlo/pull/1696) that includes these changes, once it gets merged in I'll fix this one : )

jonahkagan: Oh interesting. Ok, sounds good to wait for the other PR!

jonahkagan (approved):

PR #1709, opened 2022-11-16
Make total row detection more rigorous (Clay County fix)

arsalansufi:
# Overview

Our total row detection logic is flagging a false positive in Clay County. We currently raise an error if any numeric column's last value looks like a total. After this PR, we'll only raise an error if _all_ numeric columns' last value looks like a total. Other forms of total row detection, like checking for the word "total", are still in place and unaffected by this PR.

# Testing

- [x] Updated unit tests
- [x] Manually verified that Clay County's file will now be accepted

### DIFF
--- a/server/tests/util/test_csv_parse.py
+++ b/server/tests/util/test_csv_parse.py
@@ -382,6 +382,28 @@ def test_parse_csv_total_row():
         == "It looks like the last row in the CSV might be a total row. Please remove this row from the CSV."
     )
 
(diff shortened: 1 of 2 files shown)


arsalansufi:
Update: We've come up with a simple workaround for Clay County - just have them reorder the rows in their CSV. That said, this change is probably still worth rolling out to avoid other false positives, unless we feel that it makes the total row checking too strict

jonahkagan (approved):
Great idea!

PR #1711, opened 2022-11-17
Improve data fetching/redirect on audit launch

jonahkagan:
There were two issues going on:
1. After launching the audit, the review screen would immediately redirect to the progress screen. This redirect could occur before the rounds query updated, which would lead to a flash of the rounds screen before showing the drawing sample message.
2. When drawing the sample completed, the jurisdictions query didn't reload, meaning the progress screen wouldn't show the sampled ballots for each jurisdiction until the user refreshed.

We fix both of those issues by moving the logic for redirecting/reloading into the onSuccess callback of the rounds query.

### DIFF
--- a/client/src/components/AuditAdmin/AuditAdminView.tsx
+++ b/client/src/components/AuditAdmin/AuditAdminView.tsx
@@ -1,5 +1,5 @@
-import React from 'react'
-import { useParams, Redirect } from 'react-router-dom'
+import React, { useRef } from 'react'
+import { useParams, Redirect, useHistory } from 'react-router-dom'
 import { Spinner, H3, Intent } from '@blueprintjs/core'
 import { useQueryClient } from 'react-query'
 import {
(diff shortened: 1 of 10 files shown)


client/src/components/AuditAdmin/AuditAdminView.tsx
```diff
@@ -29,17 +30,34 @@ interface IParams {
...
+      if (
+        lastFetchedRounds.current &&
+        !isDrawSampleComplete(lastFetchedRounds.current) &&
+        isDrawSampleComplete(rounds)
+      ) {
+        queryClient.invalidateQueries(jurisdictionsQueryKey(electionId))
+        history.push(`/election/${electionId}/progress`)
+      }
+      lastFetchedRounds.current = rounds
+    },
```
jonahkagan, line 53: This is a common problem I'm starting to see - how to trigger a frontend update when an "event" occurs on the server side. There's not really great support for this use case in the data fetching/state management libraries that I've seen. I'm curious to come up with a good approach that we can codify.

arsalansufi (approved):
LGTM! The Cypress batch comparison test seems to be failing with some regularity - I take it this might fix that?

jonahkagan:
I have seen that flaky test. I don't think this will help, but it's on my todo list to look into tomorrow

Issue #1674, opened 2022-11-04
Support uploading files larger than 5GB


arsalansufi:

jonahkagan:
This likely won't be an issue until we encounter an election with a similar number of ballots cast as OC's 2020 general election, which had 1.5M ballots.

jonahkagan:
Some exploratory work on uploading directly to S3: https://github.com/votingworks/arlo/tree/presigned-s3-upload

jonahkagan:
@carolinemodic putting on your radar this related task: https://github.com/votingworks/arlo/issues/1616. We've also seen timeouts on CVR file downloads. I don't think it's high priority but if it's easy to fix at the same time, might be worth considering.

PR #1713, opened 2022-11-17
Chunked CVR file uploads

jonahkagan:
Task: https://github.com/votingworks/arlo/issues/1674

So far, just a proof-of-concept for the backend.


arsalansufi (commented):
This approach seems reasonable to me! I'm curious how the consumers are going to chunk the files

chunk-test.py
```diff
@@ -0,0 +1,91 @@
...
+random.shuffle(chunks)
+
+chunked_upload_id = str(uuid.uuid4())
+
+
+# Overall approach:
+# For each file we want to upload:
+# 1. Upload each chunk to a separate file
+# 2. When we're done uploading all the chunks, concatenate the chunks into a single file
+# 3. Open the concatenated file as a file stream
```
eventualbuddha, line 26: Combine these so we don't have to duplicate the file data?

jonahkagan: I tried this at first but it got a little weird. There's no built in way to do it and so you have to get kind of fancy/hacky to meet the BinaryIO interface that the consuming code expects. I think it could be a cool follow up once we get it working

chunk-test.py
```diff
@@ -0,0 +1,91 @@
...
+    # Store each chunk in a separate file
+    # /tmp/<chunked_upload_id>/<file_name>/<chunk_number>
+    file_dir_path = os.path.join(
+        chunked_upload_dir_path(chunked_upload_id), chunk["fileName"]
+    )
+    os.makedirs(file_dir_path, exist_ok=True)
+    chunk_path = os.path.join(file_dir_path, str(chunk["chunkNumber"]))
+    print("Uploading", chunk_path)
+    with open(chunk_path, "wb") as chunk_file:
+        chunk_file.write(chunk["chunkContents"])
```
eventualbuddha, line 47: How big do we think these contents are going to be? Could/should we stream this rather than loading the whole thing in memory?

chunk-test.py
```diff
@@ -0,0 +1,91 @@
...
+    files = {}
+    for file_dir in os.scandir(upload_dir_path):
+        # Sort the chunks by name *as numbers*
+        chunk_file_names = sorted(os.listdir(file_dir.path), key=int)
+        chunk_file_paths = [
+            os.path.join(file_dir.path, chunk_file_name)
+            for chunk_file_name in chunk_file_names
+        ]
+        concatenated_file_path = os.path.join(file_dir.path, "concatenated")
+        concatenate_files(chunk_file_paths, concatenated_file_path)
```
eventualbuddha, line 71: Rather than writing the concatenated file to disk, create a `BinaryIO` that reads from the files in sequence? Effectively taking the `stdout` of `cat file1 file2 …`.

chunk-test.py
jonahkagan: I'm still trying to figure out what a good chunk size is. Maybe somewhere around 1-5MB? Streaming it should be pretty easy if needed

chunk-test.py
jonahkagan: Same response as here:
https://github.com/votingworks/arlo/pull/1713#discussion_r1025818840

PR #1715, opened 2022-11-19
Improve minerva landslide

ccombs-sqsp:
Looking at possible improvements to make to the minerva module when I found a TODO to handle landslide elections better. I tried temporarily changing the losing candidates' votes from 0 -> 1 while the athena module does its math thing, to avoid a margin of 1 which would throw a `ValueError`.

This PR is essentially replacing a heuristic (previously, just suggest pulling 4 ballots in a landslide) with a better heuristic (handles alpha values other than 10 properly, lower alpha results in more ballots being pulled). Previously if you submitted a landslide with a risk < 10, the algorithm would not return the right number of ballots to pull and the risk limiting property would technically be violated.

This PR contains:
- Improvements to install/test files, I have a PR open for that already.
- Landslide detection (if any candidates have 0 votes) and a fix by temporarily adding 1 vote to losing candidates.
- Warning generation if contest is so small that changing votes from 0 -> 1 might affect the size of the audit
  - This doesn't seem too concerning to me? Generally the size of the contest would have to be in the small dozens, in which case the difference of pulling 5 vs 8 vs 11 ballots doesn't seem too taxing (from my perspective).
- More tests for minerva to run.
  - Test landslides with different risk alphas
  - Test that contests submitted to the landslide function are not modified by the landslide function once it returns

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,14 +2,16 @@
 PATH := $(PATH):$(HOME)/.local/bin
 
 deps:
-	sudo apt install python3.8 python3.8-venv libpython3.8-dev libpq-dev graphicsmagick
+	sudo apt update
+	sudo apt install -y python3.8 python3.8-venv libpython3.8-dev libpq-dev graphicsmagick
 	# Install node: https://github.com/nodesource/distributions/blob/master/README.md#deb		
 	curl -fsSL https://deb.nodesource.com/setup_16.x | sudo -E bash -
 	sudo apt-get install -y nodejs
 	# Install poetry: https://python-poetry.org/docs/master/#osx--linux--bashonwindows-install-instructions 
 	curl -sSL https://install.python-poetry.org | python3.8 -	
 	sudo npm install -g yarn
-	sudo apt install postgresql
+	sudo apt install -y postgresql
+	sudo service postgresql start
 
 # this should only be used for development
 initdevdb:
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,8 @@ format-server:
 lint-server:
 	poetry run pylint server scripts fixtures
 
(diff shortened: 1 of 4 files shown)


combscCode:
This PR fixes issue #1728

umbernhard:
Hey Chris, thanks for submitting this. One nit: could you rebase this PR onto your other PR? Otherwise it's hard to tell what I should review here versus there, and then once that one gets merged this one will automatically rebase onto main, so it's less work for everybody.

umbernhard (commented):
I'm not totally sure of the utility of this change. In general, I think we're okay with hard failing if someone enters nonsensical data, which a landslide would be (since they almost never happen in the real world, except in uncontested contests, which no one should be auditing in the first place). I can be convinced otherwise, but I don't know that we want to support additional code that's not adding that much more value.

combscCode:
I think I buried the lede a bit with this PR, the most important bug this fixes is the case where there's a tight race between A and B, but candidate C gets no votes. Currently, Minerva will recommend just pulling 4 ballots rather than just focusing on the margin between A and B. (any candidate getting 0 votes causes Athena to throw an exception, which is handled by suggesting pulling 4 ballots)

Admittedly, this wasn't the use case I was thinking of when I first made this PR so it *is* probably too much code relative to the problem that it fixes. I made this code trying to fix the landslide issue and only discovered after-the-fact that it also fixes races with poorly performing third-party candidates! LMK if you think a fix for this issue is worth implementing, I can open another branch for that if we think it's worthwhile.

umbernhard:
Aha, that second use case is definitely a good catch! Yeah, file another PR if you can, but maybe open a ticket for it first. That's definitely a bug and shouldn't happen!

PR #1716, opened 2022-11-21
Attempt to fix flaky batch comparison Cypress test

jonahkagan:
I couldn't reproduce the flakiness, so I'm not sure if this will fix it. But worth a shot. These assertions (which fail in the flake) aren't necessary, so I thought removing them might help.


client/cypress/end-to-end/batch-comparison.cy.js
```diff
@@ -143,25 +143,16 @@ describe('Batch Comparison', () => {
                   })
                 })
               cy.findByRole('button', { name: 'Confirm' }).click()
+              cy.logout(jurisdictionAdmin)
```
jonahkagan, line 146: this is the actual fix

arsalansufi: Ooh great find!

arsalansufi (approved):
Thanks for looking into this!

PR #1723, opened 2022-11-21
Tweak batch tally entry styles

jonahkagan:
- Fix overflow handling so the list of batches scrolls, but the table of candidates sets the overall height (I noticed that before, the height of the list of batches was setting the overall height)
- Add some borders

Screen Shot 2022-11-21 at 1 50 10 PM

### DIFF
--- a/client/src/components/Atoms/ListAndDetail.tsx
+++ b/client/src/components/Atoms/ListAndDetail.tsx
@@ -28,15 +28,17 @@ export const ListAndDetail = styled.div`
   display: grid;
   grid-template-columns: 240px 1fr;
   height: 100%;
-  overflow-y: auto;
   width: 100%;
 `
 
 // ---------- List ----------
 
 const ListContainer = styled.div`
+  display: flex;
+  flex-direction: column;
   border-right: 1px solid ${Colors.LIGHT_GRAY2};
-  overflow-y: auto;
+  height: 0; /* Remove from the height calculation of the parent */
+  min-height: 100%; /* Take up the full height of the parent (based on the detail panel) */
 `
 
 const ListSearch = styled.div`
@@ -49,6 +51,7 @@ const ListItems = styled.ul`
   margin: 0;
   padding: 0;
   padding-bottom: 40px;
+  overflow-y: auto; /* Scroll the items if there are too many to fit */
 `
 
 interface IListProps {
@@ -140,6 +143,5 @@ export const ListSearchNoResults = styled.div`
 // ---------- Detail ----------
 
 export const Detail = styled.div`
-  overflow-y: auto;
   padding: 16px;
 `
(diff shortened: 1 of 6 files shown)


client/src/components/Atoms/ListAndDetail.tsx
```diff
@@ -28,15 +28,17 @@ export const ListAndDetail = styled.div`
...
 
 // ---------- List ----------
 
 const ListContainer = styled.div`
+  display: flex;
+  flex-direction: column;
   border-right: 1px solid ${Colors.LIGHT_GRAY2};
-  overflow-y: auto;
+  height: 0; /* Remove from the height calculation of the parent */
+  min-height: 100%; /* Take up the full height of the parent (based on the detail panel) */
```
jonahkagan, line 41: I found this trick here: https://stackoverflow.com/questions/68216918/dynamic-css-grid-with-max-height-relevent-to-another-columns-content. Gave me a new level of understanding of CSS!

client/src/components/Atoms/ListAndDetail.tsx
```diff
@@ -49,6 +51,7 @@ const ListItems = styled.ul`
   margin: 0;
   padding: 0;
   padding-bottom: 40px;
```
jonahkagan, line 53: @arsalansufi do you remember what this padding is for? I considered removing it as well

arsalansufi: Yes good q! I added this because I thought the last list item's bottom border looked a bit off when it touched the bottom section of the wizard:

bottom-border

I think the same idea holds with the border added to the bottom section of the wizard - we'd now have a double border:

double-border

I couldn't just remove the border on the last list item because we need that border when the list isn't tall enough to reach the end of the wizard. A negative 1px margin may do the trick, but the bottom padding also felt like an okay solution to me.

client/src/components/JurisdictionAdmin/BatchRoundSteps/EnterTalliesStep.tsx
```diff
@@ -12,9 +12,12 @@ import { Column, Row } from '../../Atoms/Layout'
 
 const Panel = styled(StepPanel)`
   padding: 0;
+  border-top: 1px solid ${Colors.LIGHT_GRAY2}; /* Match ListAndDetail */
+  border-bottom: 1px solid ${Colors.LIGHT_GRAY2};
```
arsalansufi, line 16: Thanks for getting to this! I know I said I would 😅, just had it low on my list of priorities

arsalansufi (approved):

client/src/components/Atoms/ListAndDetail.tsx
jonahkagan: Ok that was my guess! I'll leave the padding

PR #1724, opened 2022-11-21
Fix broken timers tests

jonahkagan:
These stopped working during the Vite upgrade so I skipped them to unblock myself.

When trying to fix them, I realized only one test actually relied on the fake clock (which is different from relying on fake timers - the logic looks at Date.now). So I moved the two other tests back into regular test files and then fixed that one test.

### DIFF
--- a/client/src/components/AuditAdmin/Setup/Review/Review.test.tsx
+++ b/client/src/components/AuditAdmin/Setup/Review/Review.test.tsx
@@ -20,6 +20,7 @@ import {
   jurisdictionMocks,
   contestMocks,
   auditSettingsMocks,
+  aaApiCalls,
 } from '../../../_mocks'
 
 const apiCalls = {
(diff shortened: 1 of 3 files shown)


client/src/components/AuditAdmin/Setup/Review/Review.test.tsx
```diff
@@ -1024,4 +1025,28 @@ describe('Audit Setup > Review & Launch', () => {
...
+    await withMockFetch(expectedCalls, async () => {
+      renderView()
+      await screen.findByRole('heading', { name: 'Sample Size' })
+      screen.getByText('Loading sample size options...')
+      jest.advanceTimersByTime(1000)
+      await screen.findByText(
+        'Choose the initial sample size for each contest you would like to use for Round 1 of the audit from the options below.'
+      )
+    })
+    jest.useRealTimers()
```
arsalansufi, line 1050: I don't feel strongly about this but we could use a `beforeEach` to reset the timers before every test

arsalansufi (approved):
LGTM!

client/src/components/AuditAdmin/Setup/Review/Review.test.tsx
jonahkagan: That's what I would normally do, but I figured since it's just one test in this file I'd try out putting it specifically in that test

PR #1725, opened 2022-11-21
Use react-query for useSampleSizes

jonahkagan:
I made this change while working on debugging the timers tests (https://github.com/votingworks/arlo/pull/1724) because I thought there was something weird going on with the polling. Turned out that wasn't the issue, but since this was already done, might as well move forward with it.

### DIFF
--- a/client/src/components/utilities.ts
+++ b/client/src/components/utilities.ts
@@ -75,32 +75,6 @@ export const downloadFile = (fileBlob: Blob, fileName?: string): void => {
   document.body.removeChild(a)
 }
 
-// Deprecated - use react-query's refetch interval
-export const poll = (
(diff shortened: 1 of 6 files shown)


client/src/components/utilities.ts
```diff
@@ -75,32 +75,6 @@ export const downloadFile = (fileBlob: Blob, fileName?: string): void => {
...
-        setTimeout(p, interval)
-      } else {
-        errback(new Error(`Timed out`))
-      }
-    } catch (err) {
-      errback(err)
-    }
-  })()
-}
-
```
arsalansufi, line 103: 🙌 👋

arsalansufi (approved):

PR #1726, opened 2022-11-22
Hart CVR parsing: Mitigate "Too many open files" errors

arsalansufi:
# Overview

In testing with Orange County's close-to-ready CVR files, I found myself hitting "Too many open files" errors locally. My local ulimit matches the ulimit on our prod servers (`ulimit -n` --> 1,024 * 1,024 = 1,048,576). Rather than finding a way to bump the ulimit even higher, which may not even be possible in prod (unless a higher tier of Heroku Dyno comes with a higher limit), I've modified the Hart CVR parsing code to open CVR files one at a time instead of all at once.

I also uncovered and fixed a fun bug involving a misuse of `.strip` in the process!

# Testing

- [x] Verified that existing tests still pass
- [x] Successfully uploaded OC's close-to-ready CVR files locally*

*By running this code in conjunction with https://github.com/votingworks/arlo/pull/1712 and making some manual tweaks to account for some CVRs not having a `BatchNumber` (which won't be the case in OC's final files)

### DIFF
--- a/server/api/cvrs.py
+++ b/server/api/cvrs.py
@@ -1033,6 +1033,7 @@ def parse_hart_cvrs(
 
     def parse_cvr_ballots() -> Iterable[CvrBallot]:
         for (cvr_zip_file_name, cvr_file_name), cvr_file_path in cvr_file_paths.items():
+            cvr_zip_file_name_without_extension = cvr_zip_file_name[:-4]
             cvr_file = open(cvr_file_path, "rb")
             cvr_xml = ET.parse(cvr_file)
             cvr_file.close()
@@ -1041,7 +1042,7 @@ def parse_hart_cvrs(
             batch_sequence = find(cvr_xml, "BatchSequence").text
 
             db_batch = batches_by_key.get(
-                (cvr_zip_file_name.strip(".zip"), batch_number)
+                (cvr_zip_file_name_without_extension, batch_number)
                 if use_cvr_zip_file_names_as_tabulator_names
                 else batch_number
             )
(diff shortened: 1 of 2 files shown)


server/api/cvrs.py
```diff
@@ -1261,6 +1297,7 @@ def parse_cvrs():
             session,
         )
         session.commit()
+        clean_up_file_system()
```
arsalansufi, line 1302: Manually inspected my `/tmp` directory to verify that this cleanup did in fact work as expected

server/api/cvrs.py
```diff
@@ -1041,7 +1042,7 @@ def parse_cvr_ballots() -> Iterable[CvrBallot]:
             batch_sequence = find(cvr_xml, "BatchSequence").text
 
             db_batch = batches_by_key.get(
-                (cvr_zip_file_name.strip(".zip"), batch_number)
+                (cvr_zip_file_name_without_extension, batch_number)
```
arsalansufi, line 1045: `.strip(".zip")` actually tells Python to remove any consecutive sequence (order doesn't matter) of `.`, `z`, `i`, or `p` from the start and end of the string. So `"in-person.zip".strip(".zip")` actually returns `"n-person"` 🤦. Definitely a fun find 😅

server/api/cvrs.py
```diff
@@ -1081,6 +1109,16 @@ def process_cvr_file(
 ):
     jurisdiction = Jurisdiction.query.get(jurisdiction_id)
 
+    temporary_directories_to_be_removed = []
+
+    def flag_temporary_directory_for_removal(path: str):
+        temporary_directories_to_be_removed.append(path)
```
jonahkagan, line 1115: This control flow around the temp dir feels a bit zig-zaggy. What if we created the temp dir here and passed it down into the functions that need it, then cleaned it up at the end?

arsalansufi: @jonahkagan Updated in https://github.com/votingworks/arlo/pull/1726/commits/32574634564b1ccdb8f37b540970e8abf9927f88 and re-ran my manual test to verify that 1) I can still successfully upload OC's CVRs and 2) temp directory creation and cleanup is working as expected under the hood

jonahkagan (approved):
Looks great!

PR #1729, opened 2022-11-23
Revert "Use react-query for useSampleSizes"

arsalansufi:
Proactively opening a revert commit for https://github.com/votingworks/arlo/pull/1725 since I believe it's causing issues with the UI continuing to indicate that sample sizes are being calculated even though the worker has finished calculating. Interestingly, the sample size does show up once in a while if I refresh over and over, so there may be a race condition.

Relevant Slack thread: https://votingworks.slack.com/archives/C013V6CT9H9/p1669163895545919


jonahkagan:
Should be fixed in https://github.com/votingworks/arlo/pull/1730

PR #1730, opened 2022-11-23
Fix sample size loading in status box

jonahkagan:
We weren't refetching the sample size response, so it could be in a loading state forever, which would cause the button to break.

Also added a guard in the button to prevent premature clicks from triggering an error.

### DIFF
--- a/client/src/components/Atoms/StatusBox.tsx
+++ b/client/src/components/Atoms/StatusBox.tsx
@@ -3,6 +3,7 @@ import { useParams } from 'react-router-dom'
 import styled from 'styled-components'
 import { Callout, H3, H4, Button } from '@blueprintjs/core'
 import { Formik } from 'formik'
+import { toast } from 'react-toastify'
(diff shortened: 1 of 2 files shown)


client/src/components/Atoms/StatusBox.tsx
```diff
@@ -329,7 +333,13 @@ const AuditAdminAnotherRoundStatusBox = ({
...
       })()}
       buttonLabel={`Start Round ${roundNum + 1}`}
-      onButtonClick={() => startNextRound(sampleSizes!)}
+      onButtonClick={async () => {
+        if (!sampleSizes) {
+          toast.info('Sample sizes are still loading')
+          return false
+        }
+        return startNextRound(sampleSizes)
+      }}
```
arsalansufi, line 342: Nice, thanks for updating this as well!

arsalansufi (approved):
LGTM!

PR #1732, opened 2022-11-23
adding a switch to reset the database on heroku for vuln scanning

umbernhard:


umbernhard:
We're accomplishing this via a Heroku plugin, so this is not needed.

PR #1737, opened 2022-11-30
Creates a seed db and script so that probely doesn't kill our database

umbernhard:


scripts/seed-probely-db.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+BACKUP=${1:-latest.dump}
+python -m scripts.resetdb --skip-db-creation
+pg_restore --verbose --clean --no-acl --no-owner -U arlo -d $DATABASE_URL $BACKUP
```
jonahkagan, line 4: I think we could replace the contents of restore-db-backup.sh with this script if we just add this line in:
```bash
DATABASE_URL=${DATABASE_URL:-postgresql://arlo:arlo@localhost:5432/arlo}
```

Then we don't need two separate scripts!

scripts/seed-probely-db.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+BACKUP=${1:-latest.dump}
```
jonahkagan, line 2: What if we put the probely db dump in `fixtures` with other fake data? (My motivation is simply avoiding clogging up the top level of the repo, which is not a big deal)

jonahkagan (approved):

scripts/seed-probely-db.sh
umbernhard: I can try to get that working! I'll probably make this and the above change in a follow up PR.

scripts/seed-probely-db.sh
jonahkagan: `dropdb` and `createdb` are Postgres tools. `resetdb` looks like it's calling similar functions but using Python libraries. There may be some differences, but if it works I wouldn't worry about it too much.

PR #1738, opened 2022-12-02
Fix contests before passing to athena module to avoid 0 or 1 margins.

combscCode:
This PR fixes the issue described in 1728 and is a simplification of the closed PR 1715.

Currently, minerva calculates the margins between all the candidates that are fed to it. Athena will throw a valueError when a margin is 0 or 1. This means that in a tight race between A and B where candidate C registered but got no votes, athena will throw a ValueError and Minerva will suggest pulling only 4 ballots.

This PR adds a function that ensures that 0 vote candidates are removed before being passed to Athena. In the case where only one candidate gets a vote, it puts back one 0 vote candidate so minerva doesn't suggest a sample size of 0.

### DIFF
--- a/server/audit_math/minerva.py
+++ b/server/audit_math/minerva.py
@@ -37,6 +37,39 @@ def make_arlo_contest(tally, num_winners=1, votes_allowed=1):
     )
 
 
+def fix_third_party_landslides(contest: Contest) -> Contest:
+    """Take an arlo contest and return a contest with all 0-vote candidates removed.
+
+    Use this function to clean contests before running minerva math on them. Athena
+    will throw a ValueError if it finds a margin between candidates of 0 or 1.
+    """
+
+    name = contest.name
+    ballots = contest.ballots
+    num_winners = contest.num_winners
+    votes_allowed = contest.votes_allowed
+    votes = {}
+    last_loser_candidate = None
+    for candidate, vote_total in contest.candidates.items():
+        if vote_total > 0:
+            votes[candidate] = vote_total
+        else:
+            last_loser_candidate = candidate
+    # If only one candidate received votes, make sure to add
+    # one of the losing parties back into the race.
+    if len(votes.keys()) == 1 and last_loser_candidate is not None:
+        votes[last_loser_candidate] = 0
+    return Contest(
+        name,
+        {
+            "ballots": ballots,
+            "numWinners": num_winners,
+            "votesAllowed": votes_allowed,
+            **votes,
+        },
+    )
+
+
 def make_sample_results(
     contest: Contest, votes_per_round: List[List]
 ) -> Dict[str, Dict[str, int]]:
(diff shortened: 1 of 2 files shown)


server/tests/audit_math/test_minerva.py
```diff
@@ -88,6 +88,24 @@ def test_get_sample_size_landslide():
     }
 
 
+def test_get_sample_size_landslide_third_party():
+    c = minerva.make_arlo_contest({"a": 1000, "b": 900, "c": 0})
+    res = minerva.get_sample_size(10, c, None, None)
```
umbernhard, line 93: Nitpick: it'd be nice to use a variable name here instead of just "10" for the risk limit. (If you wanted to drop a `RISK_LIMIT = 10` in global scope and just use that variable here that would be great!)

server/audit_math/minerva.py
```diff
@@ -37,6 +37,39 @@ def make_arlo_contest(tally, num_winners=1, votes_allowed=1):
...
+
+    Use this function to clean contests before running minerva math on them. Athena
+    will throw a ValueError if it finds a margin between candidates of 0 or 1.
+    """
+
+    name = contest.name
+    ballots = contest.ballots
+    num_winners = contest.num_winners
+    votes_allowed = contest.votes_allowed
+    votes = {}
```
umbernhard, line 51: I'm a little surprised the typechecker doesn't complain about these assignments

umbernhard (commented):
I'm still thinking through if this makes sense as a fix. I don't love the idea of throwing away candidates, since we do technically need to be checking that the winner did beat every loser. It's possible the right way to fix this is to actually fix the Minerva math code to handle it correctly...but that may be a bit out of scope. @ccombs-sqsp what do you think?

combscCode:
@umbernhard One simple alternative might be adding one vote to candidates with zero votes so that Athena doesn't throw the error. That would also (unfortunately) modify the results slightly, making the users pull maybe a couple more ballots than they really should. Otherwise if we wanted to fix the math, I believe we'd have to update the `athena` module [here](https://github.com/filipzz/athena/blob/e22995a3f8f61d96990ee78f6c2c11157dff0d0c/code/athena/athena.py#L206).

Totally understand why it'd be bad practice to be manually tweaking the vote counts of the contest before we feed it to Athena. Happy to start trying to modify Athena but 1. I'm not sure how easy it'll be to get a PR merged in over there and 2. I'm certainly *less* comfortable modifying the math in that repo. OFC I can also just drop the fix for now if there's bigger fish to fry!

server/audit_math/minerva.py
combscCode: I'm not sure if I know what you're referring to, is it that I could just pass these values directly to the constructor? I must admit that I wrote this section before I had coffee so my mind was a bit sluggish 😅

umbernhard: Nothing significant to my idle musings!

combscCode:
Closing this to unclutter my github PR page! feel free to re-open if there's interest in merging in these changes.

PR #1745, opened 2022-12-09
Update extra batch selection logic to select at least 2% of batches in each jurisdiction

arsalansufi:
# Overview

The Peach State has requested this slight modification to the extra batch selection logic that we implemented for their general audit, for their upcoming runoff audit.

We'll now not only be guaranteeing that every county has to audit at least 2 batches but also that every county has to audit at least 2% of their ballots. This shouldn't change the number of batches that small counties have to audit but will definitely increase the number of batches that large counties have to audit, creating for an even more compelling public narrative.

As part of this PR, I also tweaked the logic that determines whether or not the audit is complete since the notion of a round 2 is irrelevant for these extra batch, non-RLA audits.

From a process perspective for the runoff audit, we're going to have Georgia select a custom sample size of 0. This will make it such that batch selection is entirely determined by the extra batch selection logic and not RLA batch selection logic.

# Testing

- [x] Updated automated tests
- [x] Manually tested that a custom sample size of 0 poses no problems
- [x] Manually tested that large counties will in fact have to audit more batches
- [x] Manually tested audit completion

### DIFF
--- a/server/api/rounds.py
+++ b/server/api/rounds.py
@@ -74,6 +74,10 @@ def active_targeted_contests(election: Election) -> List[Contest]:
 def is_audit_complete(round: Round):
     if not round.ended_at:
         return None
+    # Don't use risk measurements to determine completion for audits with extra batches and always
+    # complete these audits after one round
+    if is_enabled_sample_extra_batches_by_counting_group(round.election):
+        return True
     targeted_round_contests = (
         RoundContest.query.filter_by(round_id=round.id)
         .join(Contest)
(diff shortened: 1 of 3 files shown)


server/api/rounds.py
```diff
@@ -74,6 +74,10 @@ def active_targeted_contests(election: Election) -> List[Contest]:
 def is_audit_complete(round: Round):
     if not round.ended_at:
         return None
+    # Don't use risk measurements to determine completion for audits with extra batches and always
+    # complete these audits after one round
+    if is_enabled_sample_extra_batches_by_counting_group(round.election):
+        return True
```
umbernhard, line 80: have we looked at what happens when we manually reopen a round with this feature? I assume it will be overridden just fine?

server/api/rounds.py
```diff
@@ -991,6 +1005,41 @@ def sample_batches(
...
+            def compute_percentage_of_jurisdiction_ballots_selected(
+                selected_batch_ids, num_jurisdiction_ballots
+            ):
+                num_jurisdiction_ballots_selected = sum(
+                    [
+                        batch_id_to_num_ballots[batch_id]
+                        for batch_id in selected_batch_ids
+                    ]
+                )
+                return num_jurisdiction_ballots_selected / num_jurisdiction_ballots
```
umbernhard, line 1021: what is the data type here? We don't run the risk of an integer division forcing this to 0 do we?

arsalansufi: Good q! In Python, if you want to use integer division, you have to explicitly do so with a double `//`, so we're good! But I'll do some double checking to be sure

server/api/rounds.py
arsalansufi: Did a manual test just to be sure, and yup, reopening still works fine!

Reopening clears `round.ended_at`:

https://github.com/votingworks/arlo/blob/53d57c6d61e5b37708793f129f469de4fbaa8636/server/api/support.py#L462

And that's still the first thing we check in `is_audit_complete`:

https://github.com/votingworks/arlo/blob/2d95d9e4e5717d6af99da0e3f6a4d7c23b535f82/server/api/rounds.py#L74-L76

server/api/rounds.py
arsalansufi:
```
$ python3
Python 3.8.10 (default, Jun 22 2022, 20:18:18) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 1 / 3
0.3333333333333333
>>> int(1) / int(3)
0.3333333333333333
>>>
```

^ Another sanity check

PR #1794, opened 2023-01-27
Move sessions to the backend, stored in the database.

benadida:


benadida:
I'm not yet sure how I want to test this code, so thinking about that. Would love any comments on the approach, the specifics, etc.

server/migrations/versions/973694cdbf49_add_a_sessions_table_as_we_move_.py
```diff
@@ -0,0 +1,31 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table('session',
+                    sa.Column('created_at', server.models.UTCDateTime(), nullable=False),
+                    sa.Column('updated_at', server.models.UTCDateTime(), nullable=False),
+                    sa.Column('id', sa.String(length=200), nullable=False),
+                    sa.Column('data', sa.JSON(), nullable=True),
```
arsalansufi, line 25: What exactly is this `data` object going to contain? I imagine at the very least a user type and user key:

https://github.com/votingworks/arlo/blob/6b5f18ee4782934cd48ac2e803e828375a833abc/server/auth/auth_helpers.py#L53-L54

If that understanding is correct, explicit fields for those values instead of a generic data object might make sense.

I've seen loosely defined sessions get overloaded with non-session data and misused as caches. While I'm confident that we wouldn't mistakenly do that, doesn't hurt to prevent that

arsalansufi (commented):
One thought!

server/migrations/versions/973694cdbf49_add_a_sessions_table_as_we_move_.py
benadida: I agree with the risk you mention, but I think the pain of explicit database fields for sessions makes any added state management super painful. Even today, we have a number of other fields that go into the session, like the CSRF token, the OAuth session token for Auth0, the support-user separate auth layer. So we'd need to define at least 8-10 fields, and I suspect we'd still end up with an "other" JSON bucket in case we add another library that depends on Flask sessions in its own way.

umbernhard (approved):
Looks good, pending failing tests!

server/migrations/versions/973694cdbf49_add_a_sessions_table_as_we_move_.py
arsalansufi: Got it! With that context, I agree that keeping things flexible makes sense

arsalansufi (approved):

server/tests/api/test_elections.py
```diff
@@ -385,11 +385,14 @@ def test_list_organizations_not_authorized(
     jurisdiction_ids: List[str],  # pylint: disable=unused-argument
 ):
     aa_user = User.query.filter_by(email=DEFAULT_AA_EMAIL).one()
+    db_session.expunge(aa_user)  # pylint: disable=no-member
```
benadida, line 388: If someone can enlighten me as to how I can get mypy to learn that `scoped_session` does indeed have an `expunge` member, that would be lovely.

server/tests/test_background_tasks.py
```diff
@@ -278,7 +278,7 @@ def db_error():
             "startedAt": assert_is_date,
             "completedAt": assert_is_date,
             "error": asserts_startswith(
-                '(psycopg2.errors.NotNullViolation) null value in column "audit_name" violates not-null constraint'
+                '(psycopg2.errors.NotNullViolation) null value in column "audit_name"'
```
benadida, line 281: The error message I was getting in my run was more explicit, talking about the table on which this error occurred. My interpretation is that this is a diff version of Postgres situation, and might as well make the test resilient to different versions of Postgres.

server/tests/batch_comparison/snapshots/snap_test_batch_inventory.py
```diff
@@ -90,3 +90,45 @@
 Tabulator 2 - BATCH1,2,1\r
 Tabulator 2 - BATCH2,2,0\r
 """
+
```
benadida, line 93: I'm not sure why this was added here. I'll see if I can remove it.

server/tests/test_app.py
```diff
@@ -12,6 +12,11 @@ def test_index(client: FlaskClient):
     assert b"Arlo, by VotingWorks" in rv.data
 
 
+def test_logo(client: FlaskClient):
```
benadida, line 15: needed to increase coverage and my files were already fully covered, so thought this was a useful added test to add one more line of coverage.

server/websession.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def save_session(self, app, session, response):
+        if not session.modified:
+            return
+
+        # we're about to update the database, and we want to cancel any pending and uncommitted updates.
+        # because we use scoped_session, which maintains a thread-local sqlalchemy session,
+        # calling remove here tidies everything up, including potentially rolling back an uncommitted txn,
+        # and on next db_session call, we'll be using a new thread-local session.
+        db_session.remove()
```
jonahkagan, line 50: What's an example of a pending/uncommitted update that would be happening when `save_session` is called? Just want to get an intuition for why this is useful

server/websession.py
```diff
@@ -0,0 +1,97 @@
...
+
+        saved_session = WebSession.query.filter_by(id=session.sid).first()
+
+        if saved_session:
+            saved_session.data = dict(session)
+        else:
+            new_session = WebSession(id=session.sid, data=dict(session))
+            db_session.add(new_session)
```
jonahkagan, line 58: Any worries about race conditions here? Do we need an atomic upsert? I haven't thought about it deeply but figured you might have

jonahkagan (commented):
Thanks for tackling this! I'm glad it turned out to be fairly straightforward (aside from the testing issues)

server/websession.py
benadida: great question! I did not think through this issue.

Thinking through it a bit, I don't think this is a problem. This only happens if a session has been deleted by the cleanup process (or if someone is messing with their cookie). If there's a race condition, it would be because two requests are happening at once... and whichever one completes second will overwrite the cookie from the first, and the first allocated session will be garbage collected a few minutes later.

So, good callout, but I think this is safe. Almost as important I don't know how we can do anything better, because we want to allocate a fresh session anyways.

server/worker/worker.py
```diff
@@ -12,6 +13,7 @@
     configure_sentry()
     while True:
         run_new_tasks()
+        cleanup_sessions()
```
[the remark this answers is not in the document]
benadida: I think the cleanup has to be a background task, since it's based on timeouts, and we can't guarantee there will be a chance to clean it up via normal web requests, because the user may simply not show up again.

It's not a huge deal if this job fails for a few hours, because there are other safeguards preventing the sessions from staying valid beyond their expiration date. This job here is mostly to garbage collect the database periodically.

server/websession.py
jonahkagan: Ah ok so it's specific to when flask actually calls these methods. That's quite subtle. Could you add a bit of context around that in the comment here explaining it?

Also, just wondering if we're sure we want to run `db_session.commit()` in `save_session`. Would it break anything to make the updating of the web session part of the same transaction as the rest of the endpoint handler?

server/websession.py
jonahkagan: That reasoning makes sense to me!

server/websession.py
benadida: Is there a different place we would call `db_session.commit()` ? My understanding is if we don't call it here, then the handler from `app.py` will roll back, thus preventing an update. I believe this approach follows the pattern we currently use, where the specific URL handler is responsible for committing its own data, and so is this session handler.

But, open to other thoughts!

server/worker/worker.py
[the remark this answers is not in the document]
jonahkagan: Gotcha, that's good context. I agree a background task is appropriate. I have a hygenic impulse to make it a separate process from the background worker and run it periodically (e.g. using [Heroku Scheduler](https://devcenter.heroku.com/articles/scheduler)), but I can live with piggybacking on the worker process for now.

server/websession.py
jonahkagan: We could change the `app.teardown_context` handler to be a bit smarter and only rollback on error, otherwise commit (see https://flask.palletsprojects.com/en/1.1.x/api/#flask.Flask.teardown_appcontext). That would ensure that every request ends in a transaction commit (we could even use that to avoid having to manually commit in all of our endpoint handlers!)

I think we should think through what behavior we want - i.e. ideally, when should the session update be committed to the database? How coupled should it be to the success of the request?

benadida: My immediate reaction to that idea is negative, but I would be happy to discuss it.

Either way, though, I think there's a good case to be made for the session database commit to be independent of the business logic commit. You might want to save an error message in the sessions state, for example. In fact I tried to figure out how to make the session store's database connection a *different* one, under separate transaction control, but `scoped_session` appears to be too smart and seems to prevent me from doing this, or at least I couldn't figure it out. But if I had my way, I'd make the session's database connection independent of the URL handler's.

jonahkagan: I buy that it should be independent of the business logic, since the request happened whether or not it succeeded.

jonahkagan (approved):

Issue #1775, opened 2022-12-24
Support user: create a "download audit report" button that's clickable at any point during the audit


carolinemodic:
Issue automatically created from slack feedback reported by bernhard 
https://votingworks.slack.com/archives/C01F2CH62TS/p1669063695531989?thread_ts=1669063695.531989&cid=C01F2CH62TS

umbernhard:
@ccombs-sqsp if you want a slightly less-involved UI/download thing to warm up to #1520, maybe try this one?

combscCode:
oh btw if you want to ping me, ping me at @combscCode! Not a big deal either way, but the other one is used mostly for work.

jonahkagan:
I think this was solved by adding the discrepancy report feature @arsalansufi. I'm gonna close it but let me know if you think otherwise

arsalansufi:
Agreed!

PR #1795, opened 2023-02-07
Combsc/download audit button/1775

combscCode:
## What
Addresses #1775, creates a `Download Audit Button` on the support user page whenever an audit is running.

## Details
Most of the files changed are just screenshots being updated because I modified the `AsyncButton` component which is used in a lot of places.
- Modified `AsyncButton` to take an optional icon
- Added a new `AuditContainer` component to handle rendering instead of reusing `Column`
- Added a `Download Audit Report` button that will only render if the audit has a non-zero number of rounds.

## Alternatives Considered
Previously, the `Column` component was used to display audit information for the support user. To the best of my understanding, this component seems designed to be used with two side-by-side Columns, since that's how some of the support user pages are laid out. In order to accomplish this, the `Column` class is given a 50% width and a right-padding.

To me, it seems that reusing this component in the audit screen is a mistake. Limiting it to 50% of the screen forced the buttons to either shrink or wrap around to two lines. One solution I considered was simply turning it into a `<div>` which would let the audit table display on the entire screen. This was a bit jarring given that it was previously limited to 50%, so I decided to make a new component that limits it to 60%. Happy to turn it back into a simple `<div>` but I think the 60% looks pretty close to how it used to look and allows the buttons to display nicely.


combscCode:
Converting to draft, don't think I understand what's going on with the props and icon ordering.

combscCode:
Verified that icons still render as expected for other uses of the button. Ready to get reviewed!

combscCode:
Closing this to unclutter my github PR page! feel free to re-open if there's interest in merging in these changes.

Issue #1750, opened 2022-12-24
Support user: improve Support user flow for navigating between AA and JM sessions


carolinemodic:
support tools feedback - navigating logging in back and forth between an audit admin and a jurisdiction manager is a lot of clicks - either using the back button or starting over by going to the org screen.  Am I missing something?

Issue automatically created from slack feedback reported by ginny 
https://votingworks.slack.com/archives/C01F2CH62TS/p1664371529164859?thread_ts=1664371529.164859&cid=C01F2CH62TS

ginvdr:
Resolved with last round of support user improvements.

PR #1822, opened 2023-09-13
Support Tools: Improved navigation

jonahkagan:
Fixes: #1433, #1750 

Improves navigation in Support Tools by adding:
- A list of active audits on the home screen (where an "active" audit has had some activity in the past 2 weeks)
- A "Log in" button for each jurisdiction on the audit screen
- Breadcrumb navigation for audit/jurisdiction screens

_Review by commit_

https://github.com/votingworks/arlo/assets/530106/b3933283-c57a-41f2-aef4-82faaa6cb590

### DIFF
--- a/client/src/components/SupportTools/Breadcrumbs.tsx
+++ b/client/src/components/SupportTools/Breadcrumbs.tsx
@@ -1,0 +1,23 @@
+import React from 'react'
+
+interface BreadcrumbsProps {
+  children: React.ReactNode
+}
+
+const Breadcrumbs: React.FC<BreadcrumbsProps> = ({ children }) => (
+  <div className="bp3-text-large">
+    {React.Children.toArray(children).flatMap((child, index) =>
+      index === 0
+        ? [child]
+        : [
+            <span key={`separator-${index}`} style={{ fontSize: '1em' }}>
+              {' '}
+              &raquo;{' '}
+            </span>,
+            child,
+          ]
+    )}
+  </div>
+)
+
+export default Breadcrumbs
--- a/client/src/components/SupportTools/List.tsx
+++ b/client/src/components/SupportTools/List.tsx
@@ -1,0 +1,49 @@
+import React from 'react'
+import { useHistory } from 'react-router-dom'
+import styled from 'styled-components'
+import { Colors } from '@blueprintjs/core'
+
+export const List = styled.ul`
+  list-style: none;
+  padding: 0;
+  margin: 0;
+`
+
+const ItemContainer = styled.li`
+  border-top: 1px solid ${Colors.LIGHT_GRAY4};
+  &:last-child {
+    border-bottom: 1px solid ${Colors.LIGHT_GRAY4};
+  }
+
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  gap: 10px;
+  padding: 10px;
+  cursor: pointer;
+  color: ${Colors.BLUE2};
+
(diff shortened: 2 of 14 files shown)


server/api/support.py
```diff
@@ -66,6 +68,43 @@ def auth0_create_audit_admin(email: str) -> Optional[str]:
...
+        Election.query.filter(
+            Election.id.in_(
+                ActivityLogRecord.query.filter(
+                    ActivityLogRecord.timestamp
+                    > datetime.now(timezone.utc) - timedelta(days=14)
+                )
+                .with_entities(
+                    ActivityLogRecord.info["base"]["election_id"].as_string()
+                )
+                .subquery()
```
arsalansufi, line 84: Nice! I was curious to see how you're identifying "active" audits. This seems like a solid method, conceptually.

I'll be curious to see if it slows down when there's lots of activity, and retrieving 14 days worth of records potentially becomes a slow operation. Do we have any sense of this from past years, how expensive it might be to retrieve that much activity?

client/src/components/SupportTools/Breadcrumbs.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+
+interface BreadcrumbsProps {
+  children: React.ReactNode
+}
+
+const Breadcrumbs: React.FC<BreadcrumbsProps> = ({ children }) => (
```
arsalansufi, line 7: 🍞

arsalansufi (approved):
LGTM! 🎨 🧭

client/src/components/SupportTools/List.tsx
```diff
@@ -0,0 +1,49 @@
...
+interface ILinkItemProps {
+  children: React.ReactNode
+  to: string
+  style?: React.CSSProperties
+}
+
+export const LinkItem: React.FC<ILinkItemProps> = ({ to, children, style }) => {
+  const history = useHistory()
+  return (
+    <ItemContainer role="link" onClick={() => history.push(to)} style={style}>
```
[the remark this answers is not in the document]
jonahkagan: I'm not sure. I only used an `li` instead of an anchor because I needed to nest another anchor tag inside it (for the jurisdiction log in button), which apparently is invalid HTML. I'll see if the `tabindex` is needed

server/api/support.py
jonahkagan: That's a good question, though since it never gets returned back from the DB server but is just used as a subquery, I'm hoping it's not a huge issue. Still worth a sanity check, which I can do locally

client/src/components/SupportTools/List.tsx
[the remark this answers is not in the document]
arsalansufi: Ah yes, I've run into the no anchor/button tags inside anchor/button tags limitation before, too. I've previously worked around this by having the two anchors/buttons be siblings, with one styled to sit on top of the other. Using an `li` for the first, tagged appropriately, seems fine, too!

server/api/support.py
jonahkagan: Ok, I faked the current date being the height of audit season last year (mid-December) and the request took <50ms. I think this is looking ok!

client/src/components/SupportTools/List.tsx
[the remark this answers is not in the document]
jonahkagan: Added `tabindex` - good call!

PR #1824, opened 2023-09-19
Sample preview

### LINKED ISSUE #1773: AA: provide jx-sample info before launching audit
Summary from [Arlo Product Improvements](https://docs.google.com/document/d/16wlZblYvuDUKPUF0rvHPO1YhMP_Z4zdsLqF2tisGHJs/edit#heading=h.ln9exyh772gu):

> Georgia’s November audit failed to launch, due to some jurisdiction managers uploading batch totals files that were in a different format than we expected. This type of issue has been a frequent occurrence across the lifespan of Arlo, with customers like Pennsylvania, Georgia, and Michigan running into errors that are only uncovered by launching the audit. In short, we can’t plan for weird customer input, so we should give ourselves the opportunity to detect it without making it public first. We should have a “pre-launch” phase, that runs the logic that launches an audit (parsing all the JM files, building the full dataset needed for sampling, and drawing the sample) but without notifying JMs until an official launch is executed. 

Issue automatically created from slack feedback reported by ginny 
https://votingworks.slack.com/archives/C01F2CH62TS/p1668708000948369

jonahkagan:
Closes #1773 

Adds a button to preview the sample to the Review & Launch screen, allowing audit admins to see how many ballots/batches would be sampled for each jurisdiction. This helps catch any errors in audit setup that would lead to an incorrect sample.

Also, I took the opportunity to make some small design improvements to the setup flow and the review screen in particular.

_Review by commit_

Basic demo of the feature

https://github.com/votingworks/arlo/assets/530106/405fcf63-4875-4eea-9cc1-7562422a72b1

Handling long jurisdiction lists
Screenshot 2023-09-14 at 10 45 48 AM

Walkthrough of the whole setup flow with design tweaks

https://github.com/votingworks/arlo/assets/530106/25ffee50-5795-4b24-80ed-fb39782c6031


client/src/components/AuditAdmin/Setup/Contests/ContestForm.tsx
```diff
@@ -292,7 +294,15 @@ const ContestForm: React.FC<IProps> = ({
         setFieldValue,
         isSubmitting,
       }: FormikProps<{ contests: IContestValues[] }>) => (
-        <form data-testid="form-one">
+        <form
+          data-testid="form-one"
+          style={{ width: '100%' }}
+          onSubmit={e => {
+            e.preventDefault()
+            if (isOpportunisticFormClean(touched, values)) goToNextStage()
```
arsalansufi, line 302: I found it a bit unintuitive that `isOpportunisticFormClean` could be called even when not dealing with an opportunistic form. I think I'd prefer the check for whether the form is for target vs. opportunistic contests to be separated from the check for whether the form is clean, e.g.

```
const isOpportunistic = !isTargeted;
if (isOpportunistic && isFormClean(touched, values)) goToNextStage();
```

Separately, I'm sure I'm missing something simple but why can't we call `goToNextStage()` if the form is clean but for target contests?

arsalansufi (approved):
The sample preview and other UI tweaks look great!

client/src/components/AuditAdmin/Setup/Contests/ContestForm.tsx
jonahkagan: This was existing code that I just moved around. I'm pretty sure it's buggy. The whole architecture for the setup stepper needs to be redone. It's too brittle for me to even risk editing it.

arsalansufi: Ah missed that it was existing code! Def fine to leave as is then

PR #1827, opened 2023-09-20
Fix bug in reporting when contest not on ballot

jonahkagan:
We used to express when a contest wasn't on a ballot by not recording an interpretation for that contest. At some point we changed to having an explicit interpretation type, but forgot to update one piece of code
that handled this case. This function translates our interpretation into CVR format so that it can be easily compared to the actual CVR (e.g. in audit math or in reporting). The bug was that we were treating it like a
blank ballot, which is not too far off, so it likely slipped by unnoticed.

### DIFF
--- a/server/api/rounds.py
+++ b/server/api/rounds.py
@@ -440,15 +440,34 @@ def sampled_ballot_interpretations_to_cvrs(
                 ),
                 None,
             )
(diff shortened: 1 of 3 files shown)


arsalansufi (approved):
Nice catch!

PR #1828, opened 2023-09-20
Improve clarity of discrepancy columns in audit report

jonahkagan:
Instead of only showing vote deltas when there are both audited and reported votes, show any mismatch between the audited result and the reported result. This makes this column the one-stop shop for any audit discrepancy (instead of having to also look in the discrepancy column, which most users find confusing).

Also renames the "Vote Delta" to "Change in Results" and "Discrepancy" to "Change in Margin" to try to make those values a bit clearer.

_Review by commit_

### DIFF
--- a/server/api/reports.py
+++ b/server/api/reports.py
@@ -255,15 +255,19 @@ def pretty_choice_votes(
 
 
 def batch_vote_deltas(
-    reported_results: Dict[str, int], audited_results: Dict[str, int],
-) -> Dict[str, int]:
+    reported_results: Dict[str, int], audited_results: Optional[Dict[str, int]],
+) -> Union[str, Dict[str, int]]:
+    if audited_results is None:
+        return "Batch not audited"
     return {
         choice_id: reported_results[choice_id] - audited_results[choice_id]
         for choice_id in reported_results.keys()
     }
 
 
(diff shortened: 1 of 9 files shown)


server/api/reports.py
```diff
@@ -697,8 +697,8 @@ def sampled_ballot_rows(election: Election, jurisdiction: Jurisdiction = None):
             result_columns.append(f"Audit Result: {contest.name}")
             if show_cvrs:
                 result_columns.append(f"CVR Result: {contest.name}")
-                result_columns.append(f"Vote Delta: {contest.name}")
-                result_columns.append(f"Discrepancy: {contest.name}")
+                result_columns.append(f"Change in Results: {contest.name}")
+                result_columns.append(f"Change in Margin: {contest.name}")
```
arsalansufi, line 701: I like the new column names!

Likely a relic of an earlier iteration, but the PR description says that "Vote Delta" was changed to "Tally Discrepancies" when it looks like it was actually changed to "Change in Results" - might want to update the PR description accordingly (I like "Change in Results")

arsalansufi (approved):

server/api/reports.py
jonahkagan: Yes, thank you! I remember writing that and thinking it might become outdated lol

PR #1829, opened 2023-09-21
100% coverage for server

jonahkagan:
We've been so close to this for a while, so it didn't take long to clean up the loose ends. My general approach was:
- Add tests for cases we care about
- Turn some error cases into assertions, since we don't actually expect those cases to happen
- Ignore code that's very hard to test or is not so important (e.g. unlikely invalid cases in the deprecated full hand tally flow)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -8,7 +8,7 @@ omit =
     fixtures/*
 
 [report]
-fail_under = 99.6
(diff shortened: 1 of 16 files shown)


arsalansufi (approved):
Thanks for doing this! 💯

PR #1830, opened 2023-09-26
Change jest coverage threshold to use absolute number

jonahkagan:


arsalansufi (approved):
Woohoo - excited to try this new approach out!

Issue #1813, opened 2023-08-17
Discrepancy tracking


mattroe:
[Discrepancy tracking brief](https://docs.google.com/document/d/1mtEOX-oWQyocOmLjFoPLAPJjbcRPEg_aIAXAFsoOeXY/edit?usp=sharing)

mattroe:
@ginvdr to define the process of how the support team investigates discrepancies today step-by-step and insights from the user groups

mattroe:
- Priority of subproblems: surfacing info > tracking comms/resolution; audit board discrepancies > file upload
- Examples of discrepancy tracking: https://docs.google.com/spreadsheets/d/1IT-AJrBm09qH3v8JL-m-V6iWZvITgvEr_V-YfAiuGso/edit#gid=867755159 & https://docs.google.com/spreadsheets/d/1R5pO_dmzCOMp-33Y1DuAIMrbDGhicIVG20RitC0O6ZM/edit#gid=0
- Seems important to have an active resolution flow
- Should we prototype this out of the system? Is the live update actually the root issue?
- Ginny pulls this report every 30m for some active large audits
- General consensus it is worth tackling the spreadsheet before building out an interface to test the problem
- Agreement to show human and RLA math discrepancies

jonahkagan:
Initial scope: A spreadsheet download that helps support users track audit board discrepancies in a more streamlined way.

jonahkagan:
New scope:
- An indicator of which jurisdictions have discrepancies in the progress screen
- A button for audit admins to download the discrepancy report

PR #1831, opened 2023-09-27
Discrepancy indicators

jonahkagan:
Task: #1813 

_Review by commit_

Adds a Discrepancies column to the audit admin Audit Progress screen for ballot comparison and batch comparison audits. This column shows a count of the number of discrepancies in the audit results after a jurisdiction finalizes their results. It's based on the same data that goes in the "Change in Results" audit report column (aka vote deltas) rather than the "Change in Margins" column (which is the internal audit math discrepancy value) since that data has been designed to flag any discrepancy between audit results and reported results, rather than just the ones that will impact the outcome of the statistical audit.

Next PR will include access to a discrepancy report.

To ensure this change will not slow down the jurisdictions endpoint (which is already one of our slowest), I tested this on some large elections. I found that the loading of CVRs was the bottleneck, and optimized it accordingly. The endpoint is still not the fastest, but now the impact of adding this new logic to calculate discrepancies has no noticeable impact on the overall time (at least not in my testing).

Screenshot 2023-09-27 at 10 07 27 AM

Screenshot 2023-09-27 at 9 55 52 AM

### DIFF
--- a/server/api/discrepancies.py
+++ b/server/api/discrepancies.py
@@ -1,10 +1,10 @@
 from typing import Dict, Optional, Union
 
-from ..audit_math import supersimple
+from ..audit_math import supersimple, macro
 from ..models import *  # pylint: disable=wildcard-import,unused-wildcard-import
 
 
-# { contest_choice_id: vote delta }
+# { choice_id: vote delta }
 ContestVoteDeltas = Dict[str, int]
 
 
@@ -45,11 +45,12 @@ def ballot_vote_deltas(
 
 
 def batch_vote_deltas(
-    reported_results: Dict[str, int], audited_results: Dict[str, int],
+    reported_results: macro.ChoiceVotes, audited_results: macro.ChoiceVotes
 ) -> Optional[ContestVoteDeltas]:
     deltas = {
         choice_id: reported_results[choice_id] - audited_results[choice_id]
         for choice_id in reported_results.keys()
+        if choice_id != "ballots"
     }
 
     if all(delta == 0 for delta in deltas.values()):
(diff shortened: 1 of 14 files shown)


server/api/discrepancies.py
```diff
@@ -45,11 +45,12 @@ def ballot_vote_deltas(
 
 
 def batch_vote_deltas(
-    reported_results: Dict[str, int], audited_results: Dict[str, int],
+    reported_results: macro.ChoiceVotes, audited_results: macro.ChoiceVotes
 ) -> Optional[ContestVoteDeltas]:
     deltas = {
         choice_id: reported_results[choice_id] - audited_results[choice_id]
         for choice_id in reported_results.keys()
+        if choice_id != "ballots"
```
arsalansufi, line 53: For my understanding, what is this check for?

arsalansufi (approved):
Looks great and thanks for proactively testing perf!

server/api/discrepancies.py
jonahkagan: The macro module defined a data format for batch tallies that was a single dictionary where the keys were all the contest choice ids plus one special `"ballots"` key for the total number of ballots. Ideally, it would have been structurally separated (e.g. `{ "choices": {...votes by choice id...}, "ballots": number }`). Since it's flattened, this just works around that special key. It wasn't an issue in the previous PR because the code that consumed the vote deltas was iterating over a separate list of contest choices defined in the contest (implicitly ignoring `"ballots"`), whereas now there's some code that iterates over the keys of vote deltas directly.

arsalansufi: Got it, thanks for clarifying

Issue #1541, opened 2022-07-13
Calculate p-value as a background task


jonahkagan:
When the last audit board signs off, the request spins for a long time. (This is because we’re calculating the risk measurement and finalizing the audit.) Since the audit board doesn’t actually need to know that all of that stuff completed, it could happen as a background task.

jonahkagan:
Fixed by #1833

PR #1833, opened 2023-09-28
Finish round button

### LINKED ISSUE #1826: Finish round button
Instead of ending the round automatically when the last jurisdiction finishes auditing, let the audit admin press a button to end the round.

There should be some guards to make sure they don't end the round before reviewing discrepancies.

jonahkagan:
Closes #1826, https://github.com/votingworks/arlo/issues/1541

_Review by commits_

Instead of ending the round automatically when the last jurisdiction finishes auditing, lets the audit admin press a button to end the round.

https://github.com/votingworks/arlo/assets/530106/4dd146b1-7d5d-47b7-9be3-b8b976f79839


server/api/rounds.py
```diff
@@ -519,6 +616,31 @@ def create_round(election: Election):
     return jsonify({"status": "ok"})
 
 
+@api.route("/election/<election_id>/round/<round_id>/finish", methods=["POST"])
```
arsalansufi, line 619: Is it ever the case that we need to finish a round other than the most current? Wondering if this endpoint really requires a `round_id` param

arsalansufi (approved):
One q, otherwise LGTM!

server/api/rounds.py
jonahkagan: Good q. No, we don't need to finish other rounds.

The convention in Arlo has been to overspecify params in the API urls and then verify them in the endpoint (e.g. check out the pattern in `undo_round_start`). So this endpoint should validate that the given round id is the current round (which I'm now realizing is missing here).

I'm feeling unsure whether it's better to change to the style you suggest (which I agree is simpler), or to stick to the more REST-y convention of specifying explicitly the resource you want to modify.

arsalansufi: Got it - I could go either way on this one. I do like consistency, but also think that it's okay to diverge from this convention if it isn't doing much for us

jonahkagan: Decided to update to `/election/<election_id>/rounds/current/finish` to still have a bit of that REST-y vibe will still getting the benefits of the simplification. And updated the undo round start endpoint to match. Thanks for the suggestion!

arsalansufi: Nice, I like that compromise!

PR #1835, opened 2023-10-02
Audit progress: Persist filter/sort in URL

### LINKED ISSUE #1513: Audit progress: persist filter and sort in URL
Sometimes audit admins and/or support users want to focus on tracking the progress of a particular jurisdiction (or a few particular jurisdictions). For audits with many jurisdictions, it can be difficult to continually sort through the table to find those jurisdictions. We could help address this with a feature to pin jurisdictions to the top.

jonahkagan:
Closes #1513 

In order to allow audit admins to track a specific jurisdiction or a few jurisdictions over time, save the filter/sort state of the audit progress table in the URL. This will allow the user to refresh the page for updates (or even bookmark/share it) while keeping the focus on the desired jurisdictions.

https://github.com/votingworks/arlo/assets/530106/cf977281-0726-4ad2-9804-ab72df361388

### DIFF
--- a/client/src/components/Atoms/Table.tsx
+++ b/client/src/components/Atoms/Table.tsx
@@ -1,5 +1,5 @@
-import React from 'react'
-import { useTable, useSortBy, Column, Row } from 'react-table'
+import React, { useEffect } from 'react'
+import { useTable, useSortBy, Column, Row, SortingRule } from 'react-table'
 import styled from 'styled-components'
 import { Icon, HTMLTable, Button } from '@blueprintjs/core'
 import { downloadFile } from '../utilities'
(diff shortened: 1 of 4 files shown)


client/src/components/Atoms/Table.tsx
```diff
@@ -91,22 +93,30 @@ export const Table = <T extends object>({
...
       columns: React.useMemo(() => columns, [columns]),
       autoResetSortBy: false,
+      initialState: initialSortBy && { sortBy: initialSortBy },
     },
     useSortBy
   )
 
+  useEffect(() => {
+    if (onSortByChange) onSortByChange(sortBy)
+  }, [sortBy, onSortByChange])
```
jonahkagan, line 118: This seems to be the recommended way to listen for state changes: https://github.com/TanStack/table/blob/v7/docs/src/pages/docs/faq.md#how-can-i-use-the-table-state-to-fetch-new-data

arsalansufi (approved):

PR #1840, opened 2023-11-06
Remove sandbox orgs from "sample extra batches" feature flag

jonahkagan:


arsalansufi (approved):
Thanks for looking into this!

PR #1869, opened 2024-01-24
Update Hart CVR parsing to support a Washington county with two servers

arsalansufi:
## Overview

This PR updates Hart CVR parsing to support a Washington county with two servers. More specifically, the PR:

1. Supports uploading multiple scanned ballot information CSVs
2. Uses Workstation values from scanned ballot information CSVs as tabulator names, if need be

I arrived at these two updates as a solution after taking a closer look at the test files that the Washington county uploaded last year.

Note that we briefly considered removing the existing mechanism for providing tabulator names (uploading multiple CVR ZIP files where the ZIP file names are tabulator names) in favor of the new mechanism. But have decided to play it safe and keep both tabulator-name-providing mechanisms around for now. More on that discussion here: https://votingworks.slack.com/archives/CKCVA0F9S/p1705682547647309

## Testing

- [x] Updated automated tests
- [x] Manually tested with the Washington county's files


server/api/cvrs.py
```diff
@@ -857,21 +857,25 @@ def parse_hart_cvrs(
...
-    Separate from the ZIP file(s), an optional scanned ballot information CSV can be provided. If
-    provided, the `UniqueIdentifier`s in it will be used as imprinted IDs. Otherwise, `CvrGuid`s
-    will be used.
+    Separate from the ZIP files, optional scanned ballot information CSVs can be provided. If
+    provided, the "Workstation" values in them will be used as tabulator names, and the
+    "UniqueIdentifier" values in them will be used as imprinted IDs. Otherwise, "CvrGuid" values
+    will be used as imprinted IDs.
+
+    If both multiple ZIP files are provided and scanned ballot information CSVs are provided, the
+    ZIP file names will take precedence over the "Workstation" values as tabulator names.
```
arsalansufi, line 869: I had to pick one method to take precedence and landed on higher precedence for the multiple ZIP files, since ZIP file names are easier to update than scanned ballot information CSVs

server/api/cvrs.py
```diff
@@ -1355,23 +1377,18 @@ def validate_cvr_upload(
...
-        for file in files:
-            if file.mimetype in ["application/zip", "application/x-zip-compressed"]:
-                num_zip_files += 1
-            elif does_file_have_csv_mimetype(file):
-                num_csv_files += 1
-        if num_zip_files == 0:
-            raise BadRequest("Please submit a ZIP file export.")
-        if not (
-            (num_zip_files == len(files))
-            or (num_zip_files == len(files) - 1 and num_csv_files == 1)
```
arsalansufi, line 1370: This validation was previously enforcing that there be only one scanned ballot information CSV, a constraint that this PR removes

server/tests/ballot_comparison/test_cvrs.py
```diff
@@ -2169,7 +2149,169 @@ class TestCase(TypedDict):
...
+            },
+        )
+
+        if test_case["expected_processing_status"] == ProcessingStatus.PROCESSED:
+            cvr_ballots = (
+                CvrBallot.query.join(Batch)
+                .filter_by(jurisdiction_id=jurisdiction_ids[0])
+                .order_by(CvrBallot.imprinted_id)
+                .all()
+            )
```
arsalansufi, line 2295: I know we'd rather use the API layer for these checks, instead of retrieving records from the DB directly. Just mimicking the convention used by other tests in this file

server/api/cvrs.py
```diff
@@ -1050,29 +1082,14 @@ def parse_interpretations(cvr_xml: ET.ElementTree):
...
         return ",".join(interpretations)
 
-    # When only one CVR zip file is provided, we use just the batch name as the unique identifier
-    # for a batch and expect batch names to be unique across tabulators.
-    # Otherwise, when more than one CVR zip file is provided, we use the zip file names as
-    # tabulator names, and batch names need not be unique.
-    # For the former, check that the batch names are actually unique before proceeding.
     duplicate_batch_name = find_first_duplicate(
         batch.name for batch in jurisdiction.batches
     )
```
jonahkagan, line 1087: Given how this variable is now used, might be clearer to do something like:
```py
use_tabulator_in_batch_key = find_first_duplicate(...) is not None
```

jonahkagan (approved):
Looks great!

server/api/cvrs.py
arsalansufi: Nice, I like that! https://github.com/votingworks/arlo/pull/1869/commits/869d91c6725ad821898b37f5aae74be8b3224d15

Issue #1780, opened 2022-12-24
Improve CVR data validation for candidate names across jurisdictions


carolinemodic:
We need a way to prevent CVR-based audits from being set up incorrectly if jurisdictions have differing candidate names in their CVRs for the same contest (e.g. diff spelling or case).

See thread below for an example of this going wrong: https://votingworks.slack.com/archives/C013V6CT9H9/p1671038373066389

Issue automatically created from slack feedback reported by jonah 
https://votingworks.slack.com/archives/C01F2CH62TS/p1671045929426709

mattroe:
Tagging as end of January to support first ballot comparison audits (NV)

jonahkagan:
Some past context: https://votingworks.slack.com/archives/CKCVA0F9S/p1640040419002000

PR #1870, opened 2024-01-29
Validate CVR choice names across jurisdictions in ballot comparison audits

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/arlo/issues/1780

Nevada ran into an issue with inconsistent CVR choice names during its pilot audit last year. One jurisdiction's CVR file used different casing for choice names than the other jurisdictions' CVR files. The choices were wrongly treated as distinct, all the way through to the audit board UI. You can see the inconsistency here:

<table>

</table>

This PR validates CVR choice names across jurisdictions in ballot comparison audits and prevents audit admins from starting the audit if such an inconsistency is detected:

<table>

</table>

Note that we're not auto-fixing anything in this PR. Just notifying audit admins and letting them update their CVR files as needed. We've thus far only seen this issue with Dominion CVRs, which are easy to update - just a header row needs updating.

In terms of our logic for detecting inconsistencies, we use the jurisdiction with the most CVR choices (picking arbitrarily if there's a tie) as the source of truth. We then verify that the CVR choices in all other jurisdictions are found in the source-of-truth list. We recognize that there are edge cases here, e.g., what if jurisdiction 1 has choices A, B, and C, and jurisdiction 2 has choices C and D (where D is a valid choice). While we don't realistically expect to encounter this edge case, the workaround would be to add choice D to the CVR files for jurisdiction 1.

## Testing

- [x] Updated automated tests
- [x] Manually tested using Nevada's pilot audit data
- [ ] Manually tested using other ballot comparison audits


client/.eslintrc.js
```diff
@@ -81,12 +81,12 @@ module.exports = {
         devDependencies: true,
       },
     ],
-    'no-unused-vars': [1, { argsIgnorePattern: '^_' }],
-    '@typescript-eslint/no-unused-vars': [1, { argsIgnorePattern: '^_' }],
+    'no-unused-vars': [2, { argsIgnorePattern: '^_' }],
+    '@typescript-eslint/no-unused-vars': [2, { argsIgnorePattern: '^_' }],
```
arsalansufi, line 85: 1 = warn, 2 = error - warnings don't cause CI to fail

server/api/contests.py
```diff
@@ -104,6 +110,21 @@ def serialize_contest(contest: Contest) -> JSONDict:
         "jurisdictionIds": [j.id for j in contest.jurisdictions],
     }
 
+    if contest.election.audit_type == AuditType.BALLOT_COMPARISON:
+        cvr_choice_names_by_jurisdiction = {}
+        for jurisdiction in contest.jurisdictions:
+            metadata = cvrs.cvr_contests_metadata(jurisdiction)
```
arsalansufi, line 116: This helper applies contest name standardizations

client/src/components/AuditAdmin/Setup/Review/ContestChoiceNameConsistencyError.tsx
jonahkagan, line 1: In general I've tried to implement as much of the business logic as possible on the backend. Any specific reason you chose to compute this on the frontend? I don't know if I have a strong reason it should happen on the backend in this particular case other than convention

arsalansufi: Yeah good question! I was trying to avoid having the backend entirely construct the error message, since user-facing messaging feels like a frontend concern. But you're right that that's not the current convention in Arlo.

I could alternatively have the backend provide the `IContestChoiceNameConsistencyError` objects that the frontend constructs (or something similar) so that business logic is better consolidated on the backend but the frontend can still handle display / messaging. Will try that out

arsalansufi: Updated! https://github.com/votingworks/arlo/pull/1870/commits/66ac9c43d88b0ceed120974ade41bff13fe87b8a + https://github.com/votingworks/arlo/pull/1870/commits/6f48daaf5824ecfe053d7c0c2206610a149da920

jonahkagan (approved):

PR #1871, opened 2024-01-29
Fix Heroku builds by locking python-poetry-buildpack version

arsalansufi:
## Overview

Heroku builds recently started failing. On the left is a successful build, and on the right is a failed build:

heroku

Recent updates to the [python-poetry-buildpack](https://github.com/moneymeets/python-poetry-buildpack) seem to be responsible. Simplest fix for now is to lock to a specific version of the buildpack [by specifying a commit hash](https://devcenter.heroku.com/articles/buildpacks#buildpack-references). Others have also used this technique to bypass issues with the buildpack, [as noted in this comment](https://github.com/moneymeets/python-poetry-buildpack/issues/67#issuecomment-1810362297). Through trial-and-error, I've found the latest commit that still works for us.

Note that I'm not sure how the `app.json` file and Heroku console interact. It's possible that, even after merging this, I'll have to manually update the buildpacks section of the console.

## Testing

Manually updated the buildpacks section of the console for staging and deployed successfully

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,9 @@ deps:
 	# Install node: https://github.com/nodesource/distributions/blob/master/README.md#deb		
 	curl -fsSL https://deb.nodesource.com/setup_16.x | sudo -E bash -
 	sudo apt-get install -y nodejs
-	# Install poetry: https://python-poetry.org/docs/master/#osx--linux--bashonwindows-install-instructions 
-	curl -sSL https://install.python-poetry.org | POETRY_VERSION=1.1.13 python3.8 -
+	# Install poetry: https://python-poetry.org/docs/master/#osx--linux--bashonwindows-install-instructions
+	# Keep the local dev POETRY_VERSION in sync with the Heroku config var
+	curl -sSL https://install.python-poetry.org | POETRY_VERSION=1.1.15 python3.8 -
 	sudo npm install -g yarn
 	sudo apt install -y postgresql
 	sudo systemctl start postgresql
--- a/app.json
+++ b/app.json
@@ -2,7 +2,9 @@
   "addons": ["heroku-postgresql:hobby-free"],
   "buildpacks": [
     { "url": "heroku/nodejs" },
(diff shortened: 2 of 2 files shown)


jonahkagan (approved):
Seems fine for now.

Did you try updating to a more recent Poetry version?

Is it worth reporting an issue with the buildpack repo?

arsalansufi:
@jonahkagan I timeboxed an attempt to upgrade Poetry and couldn't quite get things working. I updated the `POETRY_VERSION` Heroku config var to the latest (1.7.1) and ran into the following error when building in Heroku:

```
Dependency walk failed at consistent-sampler (*) @ git+https://github.com/votingworks/consistent_sampler.git
```

I updated my local Poetry and tried re-running `poetry install` in the hopes that that would modify the `poetry.lock` or `pyproject.toml` but didn't see any changes.

While I was at it, I did notice a discrepancy between the Poetry version installed for local dev and the one installed by Heroku. I accordingly updated local dev to match Heroku: https://github.com/votingworks/arlo/pull/1871/commits/aa76581c0940f65769dcb9f11081ac2844e6cd52.

Re reporting an issue with the buildpack repo, that makes sense to me. Can do.

Will also open a Vx issue for updating Poetry and removing the stopgap fix introduced by this PR, which will likely be rendered unnecessary by updating Poetry.

arsalansufi:
Actually, on second thought, I decided not to open an issue with the buildpack repo. The issue makes sense to me. The buildpack recently added a `poetry self add` command ([link](https://github.com/moneymeets/python-poetry-buildpack/blob/1fbf02409a3f25153870cf878da44dccb2101af3/bin/compile#L63)), which isn't supported by our outdated version of Poetry. There was even [a recent correction](https://github.com/moneymeets/python-poetry-buildpack/pull/68) to that line to support older versions of Poetry, but our version is particularly outdated. I think that upgrading Poetry is the right fix.

jonahkagan:
Thanks for diving deeper. That all makes sense to me

Issue #1809, opened 2023-06-28
CVR upload: support NV ES&S


mattroe:
[Problem summary](https://docs.google.com/document/d/15rVW8OPNjnvegeITgZg2o_QY6NbCRrzcsGjx0IfjpEw/edit?usp=sharing)

mattroe:
@ginvdr link to example files

mattroe:
next step: @ginvdr determine if the pipe-delimiting solution is viable

could we get info on the step by step process of the flow for exporting?

mattroe:
call notes: https://votingworks.slack.com/archives/C013V6CT9H9/p1694030207355219

arsalansufi:
Decided for now that https://github.com/votingworks/arlo/pull/1873 poses more risk than it's worth.

The current plan is now as follows: If Nevada runs into CVR upload errors, we'll ask them to try the Excel workaround. If that doesn’t work, we'll use our script workaround.

arsalansufi:
Okay looks like the Excel workaround will work! Exporting from ES&S as an Excel file and then saving that Excel file as a CSV results in a properly quoted CSV for Lander (the county that had trouble last time around). FWIW they've only shared a CVR file with us and not a ballot information file, but we have every reason to believe that the workaround should work for the ballot information file, too. Closing this out accordingly

PR #1873, opened 2024-01-31
Fix ES&S ballots file quoting

arsalansufi:
## Overview

The first of two PRs for https://github.com/votingworks/arlo/issues/1809.

ES&S ballots files (part of an ES&S CVR export) sometimes contain unquoted values, notably in the two "Ballot Exception" columns, e.g.,

```
...,Overvote, Undervote,...
```

which should read

```
...,"Overvote, Undervote",...
```

We currently use [a script](https://github.com/votingworks/arlo/blob/49b0298df14e688a066a4b46d922b5ea9200d5e1/scripts/fix-ess-cvr-quotes.py#L25-L36) to fix these files when we encounter them. This PR simply productionizes that fix.

In my next PR, I'll productionize the second part of the referenced script, which performs a similar fix for the other ES&S CVR export files.

## Testing

- [x] Updated automated tests
- [ ] Manually tested with a real problem ES&S ballots file

### DIFF
--- a/server/api/cvrs.py
+++ b/server/api/cvrs.py
@@ -498,6 +498,24 @@ def parse_dominion_cvrs(
     return contests_metadata, parse_cvr_rows()
 
 
+def fix_ess_ballots_file_quoting(line: str) -> str:
+    """
+    ESS ballots files sometimes contain unquoted values, notably in the two "Ballot Exception"
+    columns, e.g.,
+    ...,Overvote, Undervote,...
+    which should read
+    ...,"Overvote, Undervote",...
+    This function detects and fixes this issue at the individual line level.
+    """
+    unquoted_value_regex = re.compile(r",([a-zA-Z ]+(?:, [a-zA-Z ]+)+),")
+    properly_quoted_line = re.sub(unquoted_value_regex, r',"\1",', line)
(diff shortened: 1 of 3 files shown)


server/api/cvrs.py
```diff
@@ -573,7 +591,10 @@ def parse_ballots_file(
         ballots_file: TextIO,
     ) -> Iterator[Tuple[str, CvrBallot]]:  # (CVR number, ballot)
         validate_comma_delimited(ballots_file)
-        ballots_csv = csv.reader(ballots_file, delimiter=",")
+        ballots_file_cleaned = (
+            fix_ess_ballots_file_quoting(line) for line in ballots_file
+        )
+        ballots_csv = csv.reader(ballots_file_cleaned, delimiter=",")
```
arsalansufi, line 597: I do want to be mindful of false positives. Might be helpful to log/record somewhere the updates that were made, if any. Can incorporate that into my next PR

arsalansufi:
Decided for now that this change poses more risk than it's worth.

The current plan is now as follows: If Nevada runs into CVR upload errors, we'll ask them to try the Excel workaround. If that doesn’t work, we'll use our script workaround.

Issue #1849, opened 2023-11-29
Discrepancy tracking: present negative numbers when there are fewer ballots in the manifest than expected


carolinemodic:
Presenting a negative and positive number provides more context on the type of discrepancy and can result in an accurate total.

Issue automatically created from slack feedback reported by ginny 
https://votingworks.slack.com/archives/C01F2CH62TS/p1700060181665519?thread_ts=1700060181.665519&cid=C01F2CH62TS

PR #1876, opened 2024-02-06
Use negative numbers when num ballots in manifest fewer than expected

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/arlo/issues/1849

This one proved to be even easier than increasing batch audit sample size by 1 :smile:

_Before_

No distinction between more or fewer ballots than expected; confusing total `Difference From Expected Ballots`

<table>
before
</table>

_After_

Distinction between more or fewer ballots than expected; no total `Difference From Expected Ballots`

<table>
after
</table>

## Testing

- [x] Updated automated tests
- [x] Tested manually

### DIFF
--- a/client/src/components/AuditAdmin/Progress/Progress.test.tsx
+++ b/client/src/components/AuditAdmin/Progress/Progress.test.tsx
@@ -177,7 +177,6 @@ describe('Progress screen', () => {
       const footers = within(rows[5]).getAllByRole('cell')
       expect(footers[2]).toHaveTextContent('6,351')
       expect(footers[3]).toHaveTextContent('4,254')
-      expect(footers[4]).toHaveTextContent('10')
     })
   })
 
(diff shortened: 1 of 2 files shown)


client/src/components/AuditAdmin/Progress/Progress.test.tsx
```diff
@@ -177,7 +177,7 @@ describe('Progress screen', () => {
       const footers = within(rows[5]).getAllByRole('cell')
       expect(footers[2]).toHaveTextContent('6,351')
       expect(footers[3]).toHaveTextContent('4,254')
-      expect(footers[4]).toHaveTextContent('30')
```
jonahkagan, line 180: I think the reason I didn't do this originally is because it made the total confusing. I thought the total should reflect the magnitude of the discrepancies in aggregate.

arsalansufi: Yeah the total is particularly tricky. I see arguments in both directions re what it should be. Fwiw @ginvdr requested this change and noted that she thought the total prior to this change was incorrect:

<table>
slack
<table>

There's an argument to be made for not including a total for the `Difference From Expected Ballots` column at all

arsalansufi: Cool, went ahead and removed: https://github.com/votingworks/arlo/pull/1876/commits/1ea15dd57bb1b8adf837da549ea04cf9d8228340

And updated the PR description accordingly

jonahkagan (approved):

Issue #1815, opened 2023-08-17
Batch: support audits of multiple contests


mattroe:

mattroe:
NV would most likely the first candidate for this feature in 02/24

mattroe:
Most likely to be used by Michigan on 3/15; no batch audit until then

PR #1877, opened 2024-02-09
Add contest ID to SampledBatchDraw model

arsalansufi:
## Overview

This PR adds contest ID to the `SampledBatchDraw` model in preparation for multi-contest batch comparison audits. This should be the only data model update necessary.

## Testing

Ran the migration locally on the latest prod data and verified that it worked as expected.

### DIFF
--- a/server/migrations/versions/c012fa6b13a9_sampledbatchdraw_contest_id.py
+++ b/server/migrations/versions/c012fa6b13a9_sampledbatchdraw_contest_id.py
@@ -1,0 +1,57 @@
+# pylint: disable=invalid-name
+"""SampledBatchDraw.contest_id
+
+Revision ID: c012fa6b13a9
+Revises: 83bc53b14021
+Create Date: 2024-02-09 15:38:22.522331+00:00
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "c012fa6b13a9"
+down_revision = "83bc53b14021"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # Add SampledBatchDraw.contest_id
+    op.add_column("sampled_batch_draw", sa.Column("contest_id", sa.String(length=200)))
+    op.create_foreign_key(
+        op.f("sampled_batch_draw_contest_id_fkey"),
+        "sampled_batch_draw",
+        "contest",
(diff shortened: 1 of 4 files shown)


server/migrations/versions/c012fa6b13a9_sampledbatchdraw_contest_id.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = "c012fa6b13a9"
+down_revision = "83bc53b14021"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
arsalansufi, line 20: Modeled this migration after the [very similar migration for adding a contest ID to the `SampledBallotDraw` model](https://github.com/votingworks/arlo/blob/73d5140148604687d2de4de40d5056d41f99a178/server/migrations/versions/5b7164faeda5_sampledballotdraw_contest_id.py)

server/migrations/versions/c012fa6b13a9_sampledbatchdraw_contest_id.py
```diff
@@ -0,0 +1,57 @@
...
+    # Populate the field for existing SampledBatchDraws
+    op.execute(
+        """
+        UPDATE sampled_batch_draw
+        SET contest_id = contest.id
+        FROM contest
+        JOIN election ON election.id = contest.election_id
+        JOIN round ON election.id = round.election_id
+        WHERE round.id = sampled_batch_draw.round_id
+        """
```
arsalansufi, line 41: While Arlo asserts in many places that every batch comparison audit is associated with only one contest right now, I ran a test query on the latest prod data just to be sure that that really is the case. Wanted to make sure that there's only one possible contest ID to assign to every existing `SampledBatchDraw` during the backfill step of the migration.

```
SELECT * FROM (
    SELECT
        election.id AS election_id,
        count(*) AS contest_count
    FROM contest
    JOIN election ON contest.election_id = election.id
    WHERE election.audit_type = 'BATCH_COMPARISON'
    GROUP BY election.id
) AS election_contest_counts
WHERE contest_count != 1;
```

```
 election_id | contest_count 
-------------+---------------
(0 rows)
```

jonahkagan (approved):
Super straightforward!

PR #1878, opened 2024-02-13
Update batch tally entry UI to support multiple contests

arsalansufi:
## Overview

This PR updates the batch tally entry UI to support multiple contests. Note that the batch audit creation UI still doesn't support creating audits with multiple contests.

_Two contests_

<table>
two-contests
</table>

_One contest_

I chose to keep the contest header even when there's only one contest, for consistency.

<table>
one-contest
</table>

## Testing

- [x] Updated automated tests
- [x] Cherry-picked changes from my proof-of-concept branch to test manally

### DIFF
--- a/client/src/components/JurisdictionAdmin/BatchRoundSteps/BatchRoundTallyEntry/BatchDetail.tsx
+++ b/client/src/components/JurisdictionAdmin/BatchRoundSteps/BatchRoundTallyEntry/BatchDetail.tsx
@@ -8,6 +8,7 @@ import {
   Classes,
   Colors,
   H4,
+  H5,
   HTMLTable,
   Menu,
   MenuItem,
@@ -62,6 +63,10 @@ const BatchResultTallySheetTable = styled(HTMLTable).attrs({
     width: 100%;
   }
 
+  &.${Classes.HTML_TABLE} thead tr:not(:last-child) {
+    border-bottom: 1px solid ${Colors.LIGHT_GRAY1};
+  }
+
   &.${Classes.HTML_TABLE} tbody tr {
     height: 56px;
   }
@@ -81,6 +86,12 @@ const BatchResultTallySheetTable = styled(HTMLTable).attrs({
   // Firefox
   &.${Classes.HTML_TABLE} input[type='number'] {
     -moz-appearance: textfield;
+  }
+`
+
+const ContestName = styled(H5)`
+  &.${Classes.HEADING} {
+    margin: 0;
   }
 `
 
@@ -148,7 +159,7 @@ function tabsFromSheets(sheets: IBatchResultTallySheetStateEntry[]): ITab[] {
 interface IBatchDetailProps {
   areResultsFinalized: boolean
   batch: IBatch
-  contest: IContest
+  contests: IContest[]
   isEditing: boolean
   saveBatchResults: (
     resultTallySheets: IBatchResultTallySheet[]
@@ -163,7 +174,7 @@ interface IBatchDetailProps {
 const BatchDetail: React.FC<IBatchDetailProps> = ({
   areResultsFinalized,
   batch,
-  contest,
+  contests,
   isEditing,
   saveBatchResults,
   setIsEditing,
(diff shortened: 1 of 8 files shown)


client/src/components/JurisdictionAdmin/BatchRoundSteps/BatchRoundTallyEntry/BatchDetail.tsx
```diff
@@ -494,59 +507,66 @@ const BatchResultTallySheet: React.FC<IBatchResultTallySheetProps> = ({
...
-                    />
-                  ) : (
-                    <span>
-                      {(
-                        selectedSheet?.results[choice.id] ?? ''
-                      ).toLocaleString()}
-                    </span>
-                  )}
-                </td>
+        {contests.map(contest => (
```
arsalansufi, line 510: Easier to review using https://github.com/votingworks/arlo/pull/1878/commits/7f0ad7ed9bbe5ab23a012d02dc5cb416e8b1c049?w=1

client/src/components/JurisdictionAdmin/BatchRoundSteps/BatchRoundTallyEntry/BatchDetail.tsx
```diff
@@ -494,59 +507,66 @@ const BatchResultTallySheet: React.FC<IBatchResultTallySheetProps> = ({
...
+          <BatchResultTallySheetTable key={contest.id}>
```
arsalansufi, line 511: Because choice IDs are unique across contests and tallies are submitted keyed by choice ID alone, e.g.,

```ts
{
  name: 'Sheet 1',
  results: {
    'choice-id-1': 1,
    'choice-id-2': 2,
    'choice-id-3': 3,
    'choice-id-4': 4,
  },
}
```

form submission logic didn't require any updates :tada:

Choice IDs are guaranteed to be unique across contests because `id` is the primary key on the `ContestChoice` model:

https://github.com/votingworks/arlo/blob/237ed24995f5dbb659e1d16f9da1b06911a2deae/server/models.py#L491

jonahkagan: Feels like it would be maybe a bit clearer if we separated out the results by contest (both here and in the batch tally sheets data model). But it's also nice that this change is pretty minimal. Have you thought about this tradeoff at all?

client/src/components/JurisdictionAdmin/BatchRoundSteps/BatchRoundTallyEntry/BatchDetail.tsx
jonahkagan, line 1: I'm wondering if we should move the "Edit Tallies" button up to the top-right of the panel instead of the bottom-right. That way it's always showing when you load a new batch.

jonahkagan (approved):

client/src/components/JurisdictionAdmin/BatchRoundSteps/BatchRoundTallyEntry/BatchDetail.tsx
arsalansufi: I like that idea! Worth noting that, when there are multiple tally sheets, we also have the "Add sheet" button. 

<table>
add-sheet
</table>

I think we'll want to consider the interplay of the two buttons if they're both in the top right. Could group "Add sheet" under the "Edit sheet" button.

Also, my quick exploration of this found that it'll require moving around a decent amount of form state. I'm going to punt on this for now but will add a note to this [existing issue](https://github.com/votingworks/arlo/issues/1655).

client/src/components/JurisdictionAdmin/BatchRoundSteps/BatchRoundTallyEntry/BatchDetail.tsx
jonahkagan: Sounds good! I figured it was on your radar

PR #1881, opened 2024-02-15
Re-enable extra batch selection for Georgia but disable automatic ending after round 1

arsalansufi:
I misinterpreted Ginny's request when I opened: https://github.com/votingworks/arlo/pull/1880 😅

We actually wanted to keep extra batch selection enabled for Georgia but disable automatic audit ending after round 1. Before this PR, those two behaviors were toggled by the same feature flag. This PR decouples the two and sets up the desired behavior for Georgia.


server/tests/batch_comparison/test_sample_extra_batches_by_counting_group.py
```diff
@@ -113,8 +114,17 @@ def batch_tallies(
...
 
 
+@pytest.mark.parametrize(
+    "org_id",
+    [
+        "TEST-ORG/sample-extra-batches-by-counting-group/automatically-end-audit-after-one-round",
+        "TEST-ORG/sample-extra-batches-by-counting-group",
+    ],
+    indirect=True,
+)
```
arsalansufi, line 124: What I used as a reference to figure out how to pass a parameter to a fixture function: https://stackoverflow.com/questions/18011902/how-to-pass-a-parameter-to-a-fixture-function-in-pytest

jonahkagan (approved):

PR #1890, opened 2024-02-28
Clear old CVR ballots from database in background during CVR file upload

jonahkagan:
We've been seeing the database query to delete CVR ballot records be slow in production sometimes. While we're getting to the bottom of it, this change ensures that the query is only run in the `process_cvr_file` background task. It's not as safe as our previous approach of deleting the records on file delete, but should be an ok temporary fix.

### DIFF
--- a/server/tests/ballot_comparison/test_cvrs.py
+++ b/server/tests/ballot_comparison/test_cvrs.py
@@ -427,7 +427,11 @@ def test_cvrs_clear(
         .order_by(CvrBallot.imprinted_id)
         .all()
     )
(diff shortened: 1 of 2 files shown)


server/tests/ballot_comparison/test_cvrs.py
jonahkagan, line 1: Check out the test `test_cvrs_replace_as_audit_admin` that I believe covers the correct behavior we want to keep - namely, that replacing a CVR file clears the previous CVR ballot records

arsalansufi (approved):
LGTM! 🩹

PR #1892, opened 2024-03-01
Move cvr ballot deletion to background task

jonahkagan:
In order to safely delete CvrBallot records, ensure the db query is
always run in a background task. We can't rely on the query planner to
consistently use the appropriate indices for this query, so this is a
safe option to ensure we don't timeout any requests.


arsalansufi (approved):
LGTM!

PR #1893, opened 2024-03-04
Update batch inventory flow to support multi-contest batch audits

arsalansufi:
## Overview

This PR updates the batch inventory flow that we built for Georgia to support multi-contest batch audits. After we deploy this, we can fully close out https://github.com/votingworks/arlo/issues/1815.

## Testing

- [x] Updated automated tests

### DIFF
--- a/server/api/batch_inventory.py
+++ b/server/api/batch_inventory.py
@@ -25,6 +25,7 @@ from ..util.file import (
 from ..worker.tasks import UserError, background_task, create_background_task
 from ..util.csv_download import csv_response, jurisdiction_timestamp_name
 from ..util.isoformat import isoformat
+from .batch_tallies import construct_contest_choice_csv_headers
 
 # (tabulator_id, batch_id)
 BatchKey = Tuple[str, str]
@@ -39,7 +40,7 @@ class ElectionResults(TypedDict):
     ballot_count_by_batch: Dict[BatchKey, int]
     ballot_count_by_group: Dict[str, int]
     batch_to_counting_group: Dict[BatchKey, str]
-    # { batch_key: { choice_name: count } }
+    # { batch_key: { choice_id: count } }
     batch_tallies: Dict[BatchKey, Dict[str, int]]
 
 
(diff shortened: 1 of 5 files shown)


server/api/batch_inventory.py
```diff
@@ -39,7 +40,7 @@ class ElectionResults(TypedDict):
     ballot_count_by_batch: Dict[BatchKey, int]
     ballot_count_by_group: Dict[str, int]
     batch_to_counting_group: Dict[BatchKey, str]
-    # { batch_key: { choice_name: count } }
+    # { batch_key: { choice_id: count } }
     batch_tallies: Dict[BatchKey, Dict[str, int]]
```
arsalansufi, line 44: Note that, if we were to try to generate batch tallies using old batch inventory data persisted before this PR, we'd hit errors. But there isn't any way to do so for previously completed audits. So no need to run a migration.

Similar to an earlier PR, there is a question of whether this `batch_tallies` object should include the contest ID. But I ultimately didn't see a need for them when generating the actual batch-tallies / candidate-totals-by-batch files.

server/api/batch_inventory.py
```diff
@@ -557,21 +579,21 @@ def download_batch_inventory_batch_tallies(
...
 
-    assert len(list(jurisdiction.contests)) == 1
-    contest = list(jurisdiction.contests)[0]
+    contest_choice_csv_headers = construct_contest_choice_csv_headers(jurisdiction)
 
     csv_io = io.StringIO()
     batch_tallies = csv.writer(csv_io)
 
-    batch_tallies.writerow(["Batch Name"] + [choice.name for choice in contest.choices])
+    batch_tallies.writerow(["Batch Name", *contest_choice_csv_headers.values()])
```
arsalansufi, line 587: As of Python 3.7 (we're using 3.8), dictionaries are ordered. So calling `.values()` here and `.keys()` later will return values and keys in a corresponding order

server/api/cvrs.py
```diff
@@ -218,31 +218,37 @@ def get_header_indices(headers_row: List[str]) -> Dict[str, int]:
...
     row_number: int,
-    header_indices: Dict[str, int],
+    header_indices: Union[Dict[str, int], Dict[DominionCsvHeader, int]],
     required: bool = True,
     file_name: str = None,
     remove_leading_equal_sign: bool = False,
+    header_readable_string_override: Union[str, None] = None,
 ):
-    index = header_indices.get(header)
+    header_readable_string: str = header_readable_string_override or str(header)
```
jonahkagan, line 235: Maybe make this function take the header type as a type parameter?

jonahkagan (approved):

PR #1899, opened 2024-03-14
Optimize /jurisdictions endpoint to lower risk of timeouts

jonahkagan:
Small optimizations to the /jurisdictions endpoint to reduce the risk of multiple slow db queries causing a request timeout. Details in commits.

Since I wasn't able to reproduce hypothetical conditions where a request timeout would occur (when the db decided to do a seq scan for the query to count CVR ballots), I can't confirm that this solves the issue.

Planning a follow up PR to move discrepancy calculation to a separate endpoint because the slow query to load sampled ballots is currently blocking the entire page load, which is a pretty bad UX.

### DIFF
--- a/server/api/jurisdictions.py
+++ b/server/api/jurisdictions.py
@@ -193,7 +193,9 @@ def serialize_jurisdiction(
             CvrBallot.query.join(Batch)
             .filter_by(jurisdiction_id=jurisdiction.id)
             .count()
-            if processing and processing["status"] == ProcessingStatus.PROCESSED
+            if (not round_status)
+            and processing
+            and processing["status"] == ProcessingStatus.PROCESSED
             else None
         )
         json_jurisdiction["cvrs"] = {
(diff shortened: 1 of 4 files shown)


server/api/jurisdictions.py
```diff
@@ -193,7 +193,9 @@ def min_num_ballots_for_contest(contest: Contest) -> int:
             CvrBallot.query.join(Batch)
             .filter_by(jurisdiction_id=jurisdiction.id)
             .count()
-            if processing and processing["status"] == ProcessingStatus.PROCESSED
+            if (not round_status)
+            and processing
+            and processing["status"] == ProcessingStatus.PROCESSED
```
arsalansufi, line 198: Double checking: Any client-side query invalidation we need to do to make sure we refetch jurisdiction data after launching an audit (or unlaunching an audit)?

arsalansufi (approved):
Thank you for investigating and putting up this quick fix!

server/api/jurisdictions.py
jonahkagan: Good question! Looks like we do that here for launch: https://github.com/votingworks/arlo/blob/jonah/optimize-jurisdictions-endpoint/client/src/components/AuditAdmin/AuditAdminView.tsx#L50

We don't, however, do it for undoing the launch. Added that here: dfdc677c

arsalansufi: Perfect, thanks for checking/adding!

PR #1900, opened 2024-03-14
Fetch jurisdictions discrepancy counts from separate endpoint

jonahkagan:
Previously, discrepancy counts were included in jurisdiction round status. This can be a slow calculation, since we need to load all the sampled ballot interpretations. Since the jurisdictions endpoint blocks page load, it's not great to have a slow calculation as part of that endpoint. Instead, we load the discrepancy counts separately and show a spinner while they are loading.

https://github.com/votingworks/arlo/assets/530106/d4256457-2774-426d-a0ee-027d670b1f10

### DIFF
--- a/server/api/jurisdictions.py
+++ b/server/api/jurisdictions.py
@@ -424,35 +424,6 @@ def ballot_round_status(election: Election, round: Round) -> Dict[str, JSONDict]
         for jurisdiction in election.jurisdictions
     }
 
-    # Add numDiscrepancies only for ballot comparison
-    if election.audit_type == AuditType.BALLOT_COMPARISON:
-        discrepancy_count_by_jurisdiction: Dict[str, int] = Counter()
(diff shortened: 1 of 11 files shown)


server/api/jurisdictions.py
```diff
@@ -670,3 +604,82 @@ def update_jurisdictions_file(election: Election):
...
+
+@api.route("/election/<election_id>/discrepancy-counts", methods=["GET"])
+@restrict_access([UserType.AUDIT_ADMIN])
+def get_discrepancy_counts_by_jurisdiction(election: Election):
+    discrepancy_count_by_jurisdiction: Dict[str, int] = Counter()
+    round = get_current_round(election)
+    if not round:
+        raise Conflict("Audit not started")
+
+    if election.audit_type == AuditType.BALLOT_COMPARISON:
```
jonahkagan, line 617: The logic to compute discrepancies is copy-pasted from above

PR #1901, opened 2024-03-14
Re-enable formatting on save

arsalansufi:
🙈 Didn't mean to commit setting this to `false` in https://github.com/votingworks/arlo/commit/f09f16983647d8d3cda02e682fe62dea1e90d350


jonahkagan (approved):
thanks!

Issue #1866, opened 2023-12-07
Contest totals by batch from ES&S CVR files (for RI and PA)


carolinemodic:
The process of creating batch totals from ES&S has been a manual process outside of Arlo that has inconsistently been performed. Ultimately, this process could be solved outside of Arlo through support users, but ideally Arlo can auto-generate these files for jurisdictions directly.

Issue automatically created from slack feedback reported by ginny 
https://votingworks.slack.com/archives/C01F2CH62TS/p1701963567937669?thread_ts=1701963567.937669&cid=C01F2CH62TS

arsalansufi:
Per in-person discussion, we may want this to look different for RI and PA than it does for GA, where we generate and provide to users both ballot manifests and candidate-totals-by-batch files. We know that GA does a good job of double checking the generated ballot manifests but don't think that everyone will.

PR #1923, opened 2024-04-04
Update batch inventory tool to support generating candidate-totals-by-batch files from PA and RI ES&S CVR files

arsalansufi:
## Overview

Issue link: https://github.com/votingworks/arlo/issues/1866

Jurisdictions often struggle to prepare their files for batch audits. We've built a "batch inventory" tool for Georgia that helps them prepare these files from their CVR files, while also still making sure that they still physically inventory their ballots (though part of audit setup, an audit of its own, in a way).

For Pennsylvania and Rhode Island's upcoming audits, we want to try a permutation of the above model. We're going to have PA and RI jurisdictions upload their CVR files and auto-generate one of the two batch audit files, the candidate-totals-by-batch file. We'll still have them prepare ballot manifests manually.

Also of note is that PA and RI use ES&S (GA uses Dominion). So I've added a screen to the batch inventory tool to select your system type and updated the backend to support parsing ES&S CVR files, given example files we've received from both states plus our existing knowledge of ES&S CVR files. Some jurisdictions will have a single CSV with all relevant information (this is a new format that we're seeing for the first time). Others will have the multiple files that we're accustomed to (one core CVR file and one or many "ballots" files with additional metadata). For the latter, we'll expect users to bundle all their files into a single ZIP file.

As we generalize this tool, we're rebranding it the "Batch Audit File Preparation Tool". Note that we'll eventually want to replace this tool with something more seamless/integrated into the audit setup UI. It's definitely clunky, uploading files to generate new files which you then have to upload again 😅.

### UI for GA after this update

The "Select System Type" screen is new. We still prepare both ballot manifests and candidate-totals-by-batch files for GA.

GA

### UI for PA and RI

We'll only prepare candidate-totals-by-batch files, not ballot manifests, for PA and RI. The "Inventory Batches" step was only relevant for auto-generated ballot manifests so dropped that for PA and RI.

PA and RI

### PA and RI in action

https://github.com/votingworks/arlo/assets/12616928/302117cc-e4f5-43dc-9319-0723f26ab98e

## Testing

Lots of manual testing with representative files. Did the bare minimum re automated tests. Planning to add more in a follow-up PR as I'm on a bit of a time crunch (RI may audit next week).

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -6,6 +6,7 @@ omit =
     server/config.py
     scripts/*
     fixtures/*
+    server/api/batch_inventory.py
 
 [report]
 fail_under = 100
--- a/client/src/components/JurisdictionAdmin/BatchInventory.tsx
+++ b/client/src/components/JurisdictionAdmin/BatchInventory.tsx
@@ -410,11 +410,11 @@ const DownloadAuditFilesStep: React.FC<{
           alignItems="center"
           style={{ maxWidth: '50%' }}
         >
-          <H4>Batch Inventory Complete</H4>
+          <H4>Batch Audit File Preparation Complete</H4>
           {showBallotManifest ? (
             <p style={{ marginBottom: '15px' }}>
               Next, download the Ballot Manifest and Candidate Totals by Batch
-              files, then upload them on the{' '}
+              files, and upload them on the{' '}
               <Link
                 to={`/election/${electionId}/jurisdiction/${jurisdictionId}`}
               >
@@ -424,7 +424,7 @@ const DownloadAuditFilesStep: React.FC<{
             </p>
           ) : (
             <p style={{ marginBottom: '15px' }}>
-              Next, download this file, then upload it on the{' '}
+              Next, download this file and upload it on the{' '}
               <Link
                 to={`/election/${electionId}/jurisdiction/${jurisdictionId}`}
               >
@@ -524,7 +524,7 @@ const BatchInventorySteps: React.FC<{
     <Wrapper>
       <Inner withTopPadding flexDirection="column">
         <HeadingRow>
-          <H2>Batch Inventory</H2>
+          <H2>Batch Audit File Preparation Tool</H2>
           <LinkButton
             minimal
             to={`/election/${electionId}/jurisdiction/${jurisdictionId}`}
--- a/client/src/components/JurisdictionAdmin/BatchInventory.tsx
+++ b/client/src/components/JurisdictionAdmin/BatchInventory.tsx
@@ -44,7 +44,7 @@ type Step =
   | 'Download Audit Files'
 
 const STEPS: Step[] = [
-  // 'Select System Type',
+  'Select System Type',
   'Upload Election Results',
   'Inventory Batches',
   'Download Audit Files',
--- a/client/src/components/JurisdictionAdmin/BatchInventory.tsx
+++ b/client/src/components/JurisdictionAdmin/BatchInventory.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from 'react'
-import { H2, Button, H4, Checkbox } from '@blueprintjs/core'
+import { H2, Button, H4, Checkbox, HTMLSelect } from '@blueprintjs/core'
 import {
   useQueryClient,
   useMutation,
@@ -13,14 +13,14 @@ import { Wrapper, Inner } from '../Atoms/Wrapper'
 import FileUpload from '../Atoms/FileUpload'
 import { fetchApi } from '../../utils/api'
 import AsyncButton from '../Atoms/AsyncButton'
-import { FileProcessingStatus } from '../useCSV'
+import { CvrFileType, FileProcessingStatus } from '../useCSV'
 import {
   IFileUpload,
   useUploadFiles,
   useDeleteFile,
   useUploadedFile,
 } from '../useFileUpload'
-import { apiDownload } from '../utilities'
+import { apiDownload, assert } from '../utilities'
 import LinkButton from '../Atoms/LinkButton'
 import {
   StepPanel,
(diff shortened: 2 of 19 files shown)


server/migrations/versions/cb8de251c1a5_batchinventorydata_system_type.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def upgrade():
+    op.add_column(
+        "batch_inventory_data",
+        sa.Column("system_type", sa.String(length=200), nullable=True),
+    )
+
+    # Populate the field for existing BatchInventoryData entries
+    op.execute(
```
arsalansufi, line 27: This migration file is auto-generated, other than this one command

client/src/components/useFeatureFlag.ts
```diff
@@ -1,26 +1,55 @@
...
+  'b45800ff-a239-42b3-b285-414cb94d2b6b': {
+    showBallotManifest: true,
+  },
+  // Verified Voting Sandbox
+  'e348fcfd-bd23-4b96-a003-6c3a79abd240': {
+    showBallotManifest: true,
+  },
+  // VotingWorks Internal Sandbox
+  'a67791e3-90a0-4d4e-a5e7-929f82bf4ce6': {
+    showBallotManifest: false,
```
arsalansufi, line 38: Enabling the PA and RI flow (`showBallotManifest: false`) for one of our sandboxes

client/src/components/JurisdictionAdmin/BatchInventory.tsx
```diff
@@ -31,43 +31,119 @@ import {
...
+        }),
+      {
+        onSuccess: () =>
+          Promise.all(
+            ([
+              'systemType',
+              'cvr',
+              'tabulatorStatus',
+              'signOff',
+            ] as const).map(dataType =>
```
arsalansufi, line 117: Changing the system type after uploading files requires re-uploading files. You'll see in the next commit that the backend does in fact clear everything on system type change.

Can add a comment noting this. Dropped some related comments in this file by mistake and will add those back too

server/api/batch_inventory.py
```diff
@@ -301,6 +301,57 @@ def process():
...
+    batch_inventory_data = BatchInventoryData.query.get(jurisdiction.id)
+    if not batch_inventory_data:
+        batch_inventory_data = BatchInventoryData(jurisdiction_id=jurisdiction.id)
+        db_session.add(batch_inventory_data)
+
+    batch_inventory_data.system_type = system_type
+
+    # Clear dependent data
+    if batch_inventory_data.cvr_file_id:
+        File.query.filter_by(id=batch_inventory_data.cvr_file_id).delete()
```
arsalansufi, line 505: This command not only deletes the referenced file record but also clears the `batch_inventory_data.cvr_file_id` field, given the DB schema:

https://github.com/votingworks/arlo/blob/36c7597b8eae01502831ef302d2a6b14321c6f64/server/models.py#L1019

server/api/batch_inventory.py
```diff
@@ -663,11 +663,6 @@ def download_batch_inventory_ballot_manifest(
     batch_inventory_data = get_or_404(BatchInventoryData, jurisdiction.id)
     election_results: ElectionResults = batch_inventory_data.election_results
 
-    if not batch_inventory_data.signed_off_at:
```
arsalansufi, line 616: I could be smart and only perform this check if showing the ballot manifest file (and by extension including the sign-off step). But that configuration isn't stored on the backend right now and just lives in frontend JS (as you saw in https://github.com/votingworks/arlo/pull/1923/commits/1c3aec29ee9d7fdab815a35a364546dca0a4b261). I could have the frontend pass that config up to the backend, but that didn't feel worth the trouble. Long-term, we'll almost certainly be moving the config to the backend

server/api/batch_inventory.py
```diff
@@ -42,15 +42,25 @@
 BatchKey = Tuple[str, str]
 
 
-def batch_key_to_name(batch_key: BatchKey, tabulator_id_to_name: Dict[str, str]) -> str:
+def batch_key_to_name(
```
arsalansufi, line 56: Committing a bit out of order here. This commit takes the parsed data stored by the batch inventory CVR parsing and generates the files that we want from it. A later commit incudes the actual updated ES&S CVR parsing

server/api/batch_inventory.py
```diff
@@ -689,22 +701,29 @@ def download_batch_inventory_ballot_manifest(
     csv_io = io.StringIO()
     ballot_manifest = csv.writer(csv_io)
 
-    ballot_manifest.writerow(["Container", "Batch Name", "Number of Ballots"])
-
-    # We originally didn't have counting group stored, so we make this
-    # optional for backwards compatibility
+    # We originally didn't have a batch_to_counting_group key at all, so we protect against the key
+    # not existing by using .get
```
arsalansufi, line 869: Though the key always exists now, the value can still be `None`. That's why I reworded a bit here

server/api/batch_inventory.py
```diff
@@ -42,15 +42,25 @@
...
+        else f"{tabulator_id} - {batch_id}"
+    )
 
 
 class ElectionResults(TypedDict):
     ballot_count_by_batch: Dict[BatchKey, int]
-    ballot_count_by_group: Dict[str, int]
-    batch_to_counting_group: Dict[BatchKey, str]
+    ballot_count_by_group: Optional[Dict[str, int]]
+    batch_to_counting_group: Optional[Dict[BatchKey, str]]
```
arsalansufi, line 74: These two fields aren't relevant for PA and RI

server/api/cvrs.py
```diff
@@ -504,32 +506,15 @@ def parse_cvr_rows() -> Iterable[CvrBallot]:
...
+class EssCvrFiles(TypedDict):
+    cvr_file_name: str
+    cvr_file: TextIO
+    ballots_files: Dict[str, TextIO]
 
-    zip_file = retrieve_file(jurisdiction.cvr_file.storage_path)
-    file_names = unzip_files(zip_file, working_directory)
-    zip_file.close()
 
+def separate_ess_cvr_and_ballots_files(
```
arsalansufi, line 515: No new logic in this commit. Just some moving around of logic

.coveragerc
```diff
@@ -6,6 +6,7 @@ omit =
     server/config.py
     scripts/*
     fixtures/*
+    server/api/batch_inventory.py
```
arsalansufi, line 9: Punted on full test updating in this PR since I'm on a time crunch. But have tested manually and will do more manual testing before merging. And am planning to add automated tests, ideally before the actual audit next week

server/api/batch_inventory.py
```diff
@@ -223,6 +241,151 @@ def parse_vote(vote: str):
...
+                        header_indices,
+                        required=False,
+                    )
+
+                    if cvr_number not in cvr_number_to_batch:
+                        raise UserError(
+                            f"Unable to find batch for CVR number {cvr_number} in ballots files"
+                        )
+                    batch = cvr_number_to_batch[cvr_number]
+                    batch_key: BatchKey = ("", batch)
```
arsalansufi, line 339: The `BatchKey` type allows for a `(tabulator_id, batch_id)` pair, but the ES&S CVR files that we've seen thus far don't include particularly useful tabulator info. The lack of a tabulator ID is handled just fine when generated ballot manifests and candidate-totals-by-batch files in [this previous commit](https://github.com/votingworks/arlo/pull/1923/commits/6d5f3f8fec5344ba82e9b01bd8c6eb4834f6a600)

client/src/components/JurisdictionAdmin/JurisdictionAdminView.tsx
```diff
@@ -29,7 +29,9 @@ const JurisdictionAdminView: React.FC = () => {
     jurisdictionId: string
   }>()
   const auth = useAuthDataContext()
-  const isBatchInventoryEnabled = useBatchInventoryFeatureFlag(jurisdictionId)
+  const isBatchInventoryEnabled = Boolean(
+    useBatchInventoryFeatureFlag(jurisdictionId)
+  )
```
eventualbuddha: Unnecessary cast?

server/api/batch_inventory.py
```diff
@@ -209,6 +241,151 @@ def parse_vote(vote: str):
...
+                and choice_name != "undervote"
+                and choice_name != "Write-in"
+            ):
+                raise UserError(f"Unrecognized choice in CVR file: {choice_name}")
+
+            return choice_id
+
+        # ZIP file with multiple CSVs
+        if batch_inventory_data.cvr_file.storage_path.endswith(".zip"):
+            entry_names = unzip_files(cvr_file, working_directory)
```
eventualbuddha, line 270: This `unzip_files` function appears to have already been checked in, but I believe it to be unsafe. The documentation for [`zipfile`](https://docs.python.org/3/library/zipfile.html#resources-limitations) indicates that it is vulnerable to zip bombs. Here we extract every file in the `.zip` even though we're only interested in a few of them. The filtering done below should be done before extraction, for example.

I recognize that the risk is not high since you presumably have to have been logged in to be able to upload a malicious zip file and we control the sign up process. Still, I think we should change this and other uses of `unzip_files` to prevent this attack in the future.

server/api/batch_inventory.py
```diff
@@ -68,13 +93,20 @@ def items_list_to_dict(items):
...
 ):
-    jurisdiction = Jurisdiction.query.get(jurisdiction_id)
-    batch_inventory_data = BatchInventoryData.query.get(jurisdiction_id)
+    working_directory = tempfile.mkdtemp()
 
-    def process():
-        contests = list(jurisdiction.contests)
+    def clean_up_file_system():
+        if os.path.exists(working_directory):
+            shutil.rmtree(working_directory)
```
eventualbuddha, line 100: This happens to be safe for our system (and probably for any *nix), but is vulnerable to symlink attacks on non-fd based file systems ([docs](https://docs.python.org/3.12/library/shutil.html#shutil.rmtree.avoids_symlink_attacks)). From our staging console:

```
>>> shutil.rmtree.avoids_symlink_attacks
True
```

server/api/batch_inventory.py
eventualbuddha: If the key always exists, why the `.get`? Backward compatibility?

eventualbuddha (approved):
Nothing obviously blocking from a code review perspective. I didn't get a chance to try it out myself as getting an Arlo dev env set up was being annoying, but I trust that you've manually tested sufficiently. I'd like to see follow-up on the tests (as you said) and also on the potential security issues, though I think those are not urgent due to it being unclear how a threat actor would take advantage of them.

server/api/batch_inventory.py
arsalansufi: Exactly. If we were to open an old audit created before we added this field and try to download a ballot manifest through this tool, the backend would 500.

Though typing this out, I don't foresee a case where this would happen in prod. Since, once an audit is completed, jurisdiction admins no longer see the batch inventory tool. Just a screen indicating that the audit is complete.

Could happen in dev if we were to re-open an old audit for testing, though

arsalansufi: Added back the phrase "backward compatibility" for clarity

server/api/batch_inventory.py
arsalansufi: Wow nice find! Agree that we should update this. Opened an issue to track: https://github.com/votingworks/arlo/issues/1925

server/api/batch_inventory.py
arsalansufi: Thank you for checking that! I'm still having some trouble envisioning how someone would exploit this specific call if we weren't on an fd-based file system. But maybe that's not quite what you're saying - just pointing out the risk that comes with `shutil.rmtree` in general

PR #1930, opened 2024-04-23
Hide batch inventory UI in PA org for now

arsalansufi:
@ginvdr asked that we hide the batch inventory UI in the PA org for now since they're not quite ready to use the feature without our assistance. Some jurisdictions may also have ES&S file versions that we can't yet parse. We can either enable the feature for PA next week once we're actively supporting them or create a clone of their audit in one of our sandboxes to help them prepare files.


arsalansufi:
Pausing to double check that this is the route we want to take

jonahkagan (approved):

PR #1952, opened 2024-08-07
ES&S CVR parsing: Allow "hinting" which file is the CVR file by changing the file name

arsalansufi:
## Overview

ES&S CVRs consist of one CVR file and one or more ballots files. The CVR file contains the actual votes, and the ballots files contain additional useful identifiers.

We use some simple checks to distinguish the CVR file from the ballots files in the upload. But lately, one such check has been misidentifying some CVR files as ballots files. Specifically the check that - if a file has a "Tabulator CVR" column, it's a ballots file. Some jurisdiction's CVR files now include this "Tabulator CVR" column.

Long-term, this actually means that we may not need ballots files from some ES&S jurisdictions. Though even for these jurisdictions that have a "Tabulator CVR" column in their CVR files, we may still sometimes need ballots files. I won't get into the weeds of that right now, but a nice long-term flow that would work for everyone might look something like this:

1. We request your CVR file
2. If that file has everything we need, we move on
3. If that file doesn't have everything we need, we request your ballots files

I've seen the above proposed before, but it's a bigger change that requires more time and consideration.

Washington's current workaround is to open the CVR file and delete the "Tabulator CVR" column. But they understandably don't like having to open and edit the file exported by their system.

This PR allows "hinting" which file is the CVR file when Arlo incorrectly assumes that it's a ballots file by changing the file name to be a specific name (cvr.csv), with error messages that make this workaround clear. Not having to edit the file's contents should be a win for Washington.

<table>
error-message-2
</table>

https://github.com/user-attachments/assets/7d5639ac-aaf1-4ae1-ac48-e0852fd60f0d

## Testing

- [x] Tested manually
- [x] Updated automated tests

### DIFF
--- a/server/api/cvrs.py
+++ b/server/api/cvrs.py
@@ -573,30 +573,47 @@ def separate_ess_cvr_and_ballots_files(
         for file_name in file_names
     }
 
-    def is_ballots_file(file: TextIO):
+    # Allow "hinting" which file is the CVR file when Arlo incorrectly assumes that it's a ballots
+    # file, say, because it has a "Tabulator CVR" column
+    override_cvr_file_name = "cvr.csv"
+
+    def is_ballots_file(file_name: str, file: TextIO):
         first_line = file.readline()
         file.seek(0)
+        if file_name.lower() == override_cvr_file_name:
+            return False
         return first_line.startswith("Ballots") or "Tabulator CVR" in first_line
 
     ballots_files = {
-        name: file for name, file in text_files.items() if is_ballots_file(file)
+        file_name: file
+        for file_name, file in text_files.items()
+        if is_ballots_file(file_name, file)
     }
     cvr_files = {
-        name: file for name, file in text_files.items() if not is_ballots_file(file)
+        file_name: file
+        for file_name, file in text_files.items()
+        if not is_ballots_file(file_name, file)
     }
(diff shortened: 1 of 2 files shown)


server/api/cvrs.py
```diff
@@ -573,30 +573,47 @@ def decode_file(file: IO[bytes], file_name: str) -> TextIO:
...
+        error = "Identified multiple CVR files - please upload only one CVR file containing the cast vote records for each ballot, and at least one ballots file containing the list of tabulated ballots and their corresponding CVR identifiers."
+
+    if error is not None:
+        separator = " / "
+        error_components = [
+            error,
+            f"Identified CVR files: {', '.join(cvr_files.keys()) or 'None'}",
+            f"Identified ballots files: {', '.join(ballots_files.keys()) or 'None'}",
+        ]
+        raise UserError(separator.join(error_components))
```
jonahkagan, line 616: I might just separate these with periods since they read as one paragraph. The slashes looked a bit confusing to me (specifically the one between the error message and "Identified CVR files:")

jonahkagan (approved):

server/api/cvrs.py
arsalansufi: Good call!

PR #1954, opened 2024-08-13
Hotfix: Expect more possible metadata columns when parsing ES&S CVRs

jonahkagan:
Previously, we expected exactly 3 metadata columns. When given a file with more metadata columns, they were treated like contest columns. This caused a bloating of the contest metadata and major performance slowdown.

Manually tested this hotfix, will add regression test separately.


arsalansufi (approved):
Great catch!

PR #1955, opened 2024-08-13
Filter out empty rows and protect against missing values in ES&S CVR parsing

jonahkagan:
We saw a ballots file uploaded with empty rows at the end, which caused an error when trying to access the "Cast Vote Record" column to sort on. Since empty rows are easy to filter out, I added logic to do that. I also changed the sorting code to use `column_value`, which checks for missing values and raises a user-facing error if missing as a backup.


arsalansufi (approved):
LGTM!

PR #1956, opened 2024-08-13
Review screen: Show only one modal for choice name standardization

jonahkagan:
We currently render a modal for each contest that may require choice name standardization. These modals are all controlled by a single boolean flag. That means that showing any one modal will show all of them, which is not what we want to happen.

To fix this, I replaced the boolean flag with a contest ID for the modal to show.

Manually tested that this fixes the issue, and also confirmed that form state resets when closing/reopening the modal.


arsalansufi (approved):
LGTM!

PR #1958, opened 2024-08-14
Ensure logger in worker uses info level

jonahkagan:
In production (but not dev), it looks like the root log level gets changed somewhere before the worker logger is created, so worker logs were not showing up in Heroku. We explicitly force the log level back to info to fix it.

### DIFF
--- a/server/worker/tasks.py
+++ b/server/worker/tasks.py
@@ -13,6 +13,9 @@ from ..util.isoformat import isoformat
 from ..util.jsonschema import JSONDict
 from .. import config
 
+# Something in the imports above is setting the root logging level to WARNING in
+# production, so we need to explicitly set it back to INFO.
+logging.basicConfig(level=logging.INFO, force=True)
 logger = logging.getLogger("arlo.worker")
 
 


server/worker/tasks.py
```diff
@@ -13,6 +13,9 @@
 from ..util.jsonschema import JSONDict
 from .. import config
 
+# Something in the imports above is setting the root logging level to WARNING in
+# production, so we need to explicitly set it back to INFO.
+logging.basicConfig(level=logging.INFO, force=True)
```
arsalansufi, line 18: There's an env var in Heroku called `DD_LOG_LEVEL` that's set to `WARN` for both prod and staging:

<table>
heroku
</table>

Setting it to `INFO` might do the trick and render this change unnecessary?

arsalansufi (approved):
This change LGTM though I do have one idea for us to try before proceeding with this - thanks for investigating!

server/worker/tasks.py
jonahkagan: Interesting idea. I tried making that change, but it looks like that just controls the logging of the Datadog agent itself. Also, I was debugging this issue in a review app that didn't have any Datadog integration set up at all, which further makes me think it's not part of the problem.

PR #1960, opened 2024-08-14
ES&S CVR parsing: Protect against unknown columns in CVR file

jonahkagan:
We added a hotfix in #1954 to allow different sets of metadata columsn in the CVR file. However, we still may see unknown columns we haven't seen before. Since there's no good way to differentiate those columsn from contest columns, our current approach may silently fail in that case. The consequence would be that metadata columns are treated as contests. This may or may not cause downstream issues.

To reduce the likelihood of that happening, we change our method of searching for the dividing line between metadata columns and contest columns to look for the _last_ known metadata header. That way, we'll get it right in every case except the case where the dividing line is a header we haven't seen before, which is much less likely to occur.


arsalansufi (approved):
Nice, good idea!

PR #1982, opened 2024-09-13
make session cookie secure, so it's only sent over https.

benadida:
We're going to have to test this in staging, as many details depend on TLS.


arsalansufi (approved):
🔐

Issue #1962, opened 2024-09-03
Upgrade Python


arsalansufi:
https://votingworks.slack.com/archives/C01F2CH62TS/p1725300762815199

Issue #1872, opened 2024-01-30
Upgrade Poetry and remove buildpack stopgap


arsalansufi:
See https://github.com/votingworks/arlo/pull/1871, PR description and comments

PR #1986, opened 2024-10-14
Upgrade Python to v3.9 and Poetry to 1.8

jonahkagan:
Closes https://github.com/votingworks/arlo/issues/1962
Closes https://github.com/votingworks/arlo/issues/1872

Upgrades Python from 3.8 to 3.9, dealing with various impacts along the way. I also upgraded Poetry to the latest version, since there were build issues involving Poetry that were blocking deploys.

The latest version I could get using apt in my Ubuntu 20 VM was 3.9.5. Heroku and CircleCI needed the latest 3.9 version, which is 3.9.20. This shouldn't cause problems, but you may see that version discrepancy in the diff.

Tested by ensuring that:
- Automated type checking, lint, and tests pass in CI
- Can successfully deploy and access a review app in Heroku


pyproject.toml
```diff
@@ -28,7 +29,7 @@ consistent-sampler = {git = "https://github.com/votingworks/consistent_sampler.g
 scipy = "~1.6.0"
 Flask-SeaSurf = "^0.3.0"
 boto3 = "^1.20.8"
-r2b2 = {git = "https://github.com/gwexploratoryaudits/r2b2", rev = "128874b"}
```
jonahkagan, line 31: Pinning this commit broke for some reason, but there's only one commit after it on `main` in this repo, it isn't actively worked on, and it's only used for some experimental math that we don't use in real audits, so it seemed like a low risk workaround

app.json
```diff
@@ -3,7 +3,7 @@
   "buildpacks": [
     { "url": "heroku/nodejs" },
     {
-      "url": "https://github.com/moneymeets/python-poetry-buildpack.git#0bbaf48423f0caac527e185b1517abac1610dc46"
+      "url": "https://github.com/moneymeets/python-poetry-buildpack.git"
```
jonahkagan, line 6: See https://github.com/votingworks/arlo/pull/1871

.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ orbs:
 executors:
   arlo:
     docker:
-      - image: cimg/python:3.9.5-browsers
```
jonahkagan: This image had an old version of poetry, which didn't work for us since we upgraded the dev env to use the latest poetry

pyproject.toml
```diff
@@ -32,6 +32,9 @@ boto3 = "^1.20.8"
 r2b2 = {git = "https://github.com/gwexploratoryaudits/r2b2.git"}
 MarkupSafe = "==2.0.1" # Related to this problem: https://stackoverflow.com/questions/72191560/importerror-cannot-import-name-soft-unicode-from-markupsafe
 urllib3 = ">=1.26,<2" # https://github.com/python-poetry/poetry-plugin-export/issues/239
+jinja2 = "3.0.3" # Flask breaks with later versions of Jinja2
+itsdangerous = "2.0.1" # Flask breaks with later versions of itsdangerous
+werkzeug = "2.0.3" # Flask breaks with later versions of Werkzeug
```
jonahkagan, line 37: We should probably update Flask, but that felt like a bigger change

.pylintrc
```diff
@@ -60,85 +60,14 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use "--disable=all --enable=classes
 # --disable=W".
-disable=print-statement,
```
jonahkagan, line 63: A bunch of these rules got deprecated from pylint. I think they had to do with the Python 2 -> 3 transition

carolinemodic (approved):

Issue #1909, opened 2024-03-22
Add some total calculations into the final report (audited results, reported results, total number of ballots across batches to start)


arsalansufi:
Need to add some total calculations into the final report (audited results, reported results, total number of ballots across batches to start)

Issue automatically created from Slack feedback reported by ginny: 
https://votingworks.slack.com/archives/C01F2CH62TS/p1711055940985209

Issue #1784, opened 2022-12-24
Support user for batch comparison: create an easy way to figure out how many ballots are sampled


umbernhard:

jonahkagan:
Group with https://github.com/votingworks/arlo/issues/1909

PR #1988, opened 2024-10-16
Improve Batch Comparison Audit Report 

carolinemodic:
https://github.com/votingworks/arlo/issues/1909
https://github.com/votingworks/arlo/issues/1784

Adds the following items to the audit report: 

For the "rounds" section - for ballot comparision audits
- number of batches - distinct number of batches (non distinct if batches are sampled multiple times is already included as the sample size) 
- number of ballots - number of ballots in distinct batches (could be changed to non-distinct or that could be additionally included) 
- reported votes - aggregated reported votes we expect for all sampled batches (deduped) 

I did not add a totals row - felt overly complex with having to thing about how everything should sum for all ballot types and overall fairly low-pri

For the "batches" section
- number of  ballots 

Tested via inspecting snapshot updates and testing manually with the small election fixtures. i had a batch sampled twice in my test to make sure the deduping bits were correct.


server/tests/ballot_comparison/snapshots/snap_test_cvrs.py
```diff
@@ -7,1990 +7,7 @@
...
-            "Choice 2-3": {"column": 4, "num_votes": 3},
-            "Write-In": {"column": 5, "num_votes": 1},
-        },
-        "total_ballots_cast": 14,
-        "votes_allowed": 1,
-    },
-}
-
-snapshots["test_hart_cvr_upload_with_scanned_ballot_information 3"] = [
+snapshots["test_hart_cvr_upload_with_duplicate_batch_names 1"] = [
```
carolinemodic, line 10: I don't really understand why anything in this file changed but I think the snapshots were just out of date ?

server/api/reports.py
```diff
@@ -391,6 +391,11 @@ def round_rows(election: Election):
             ["Audited Votes: CVR", "Audited Votes: Non CVR"]
             if election.audit_type == AuditType.HYBRID
             else []
+        )
+        + (
+            ["# Batches", "# Ballots", "Reported Votes"]
```
jonahkagan, line 396: I think the `#` is a tiny bit confusing. How about "Batches in Sample" and "Ballots in Sample" or "Batches Sampled" and "Ballots Sampled"

server/api/reports.py
```diff
@@ -441,6 +446,36 @@ def round_rows(election: Election):
...
+                    if contest.id in batch.jurisdiction.batch_tallies[batch.name]
+                    else None
+                )
+                if reported_results_for_batch is not None:
+                    for choice in contest.choices:
+                        reported_results[choice.id] += reported_results_for_batch[
+                            choice.id
+                        ]
+
+            reported_results_by_name = reported_results and {
```
jonahkagan, line 475: I think `reported_results` always exists, so this check is unnecessary

server/api/reports.py
```diff
@@ -441,6 +446,36 @@ def round_rows(election: Election):
...
+                .all()
+            )
+            num_distinct_batches = len(distinct_batches)
+            num_distinct_ballots = sum(batch.num_ballots for batch in distinct_batches)
+
+            # get the reported votes aggregated across batches
+            reported_results: dict = {choice.id: 0 for choice in contest.choices}
+            for batch in distinct_batches:
+                reported_results_for_batch = (
+                    batch.jurisdiction.batch_tallies[batch.name][contest.id]
```
jonahkagan, line 465: Accessing `batch.jurisdiction` will make a separate query for each batch's jurisdiction. Could be a bit expensive. Often I will profile API endpoints using Heroku/Sentry tools after we deploy them to identify bottlenecks. (Report generation is already slow and should probably be backgrounded)

We might be able to pull out just the reported vote count we need for each batch at the db level if we decide to optimize it.

server/api/reports.py
```diff
@@ -441,6 +446,36 @@ def round_rows(election: Election):
...
+                    if contest.id in batch.jurisdiction.batch_tallies[batch.name]
+                    else None
+                )
```
jonahkagan, line 468: Consider: `batch.jurisdiction.batch_tallies[batch.name].get(contest.id)`, which will return `None` if the key is missing from the dict

jonahkagan (approved):

server/api/reports.py
carolinemodic: Yeah I was wondering about perf. This data is already accessed/fetched to populate the "batches" section of the csv for batch-comparison audits (where I copied the logic from), but if thats already slow its still not great to add it again.

carolinemodic: What data do you use to profile?

jonahkagan: Oh interesting, if it's already loaded once then maybe we can just reuse it? I guess that might take some finagling

To profile, I use Sentry transaction tracing and Heroku database analytics

carolinemodic: Do you just find some large actual audit that has run to test with though?

jonahkagan: Both of those tools work by sampling queries from actual usage, so I usually just look at those periodically to see what's slow. It's less that I'll test a specific change that I just deployed.

Load testing specific changes I usually do in my dev env with some large audit test data. But then you don't have as useful profiling tools.

Sorry, I realize my initial comment was a bit misleading about how I use those tools.

carolinemodic: Ah got it, I guess I wasn't sure if there was a pattern of testing on fake data at times like right now when audits aren't really actively running to help monitor usage

Issue #1845, opened 2023-11-29
Extra blank lines on batch tally sheets


carolinemodic:
Loudon County, VA would like the extra blank lines on their batch tally sheet so they can record the things they want (like undervote)

Issue automatically created from slack feedback reported by ginny 
https://votingworks.slack.com/archives/C01F2CH62TS/p1700592235185899

PR #1989, opened 2024-10-16
Add extra lines to tally sheey

nbhatia823:
Issue: https://github.com/votingworks/arlo/issues/1845

Simple change to add three extra blank lines to the tally sheet (see snapshot diffs).

Reason (from Issue): "Loudon County, VA would like the extra blank lines on their batch tally sheet so they can record the things they want (like undervote)"


client/src/components/JurisdictionAdmin/generateSheets.ts
```diff
@@ -429,17 +429,22 @@ export const downloadBatchTallySheets = async (
     // longer make that assumption.
     //
 
+    // Add 3 blank lines to contest entries table to allow for additional entries (i.e. overvotes)
+    const blankLines: string[][] = new Array(3).fill(['', ''])
```
jonahkagan, line 433: `new Array` is fine with me, but we also have a `range` helper if you'd like

client/src/components/JurisdictionAdmin/__snapshots__/generateSheets_downloadBatchTallySheets_Handles_more_than_two_contests.png
jonahkagan, line 1: This snapshot reveals a weird case where the Contest 3 table is split across pages but the section on the first page doesn't have any candidate rows in it. Seems like a degenerate case. (Not a new problem introduced by this PR, but might be nice to fix if there's any easy solution)

jonahkagan (approved):

client/src/components/JurisdictionAdmin/__snapshots__/generateSheets_downloadBatchTallySheets_Handles_more_than_two_contests.png
nbhatia823: Oh good call, I'll look into a quick fix for it. Of the path forwards to fix this, I guess I lean towards thinking it's better to have the table just start on the next page? Rather than have it start on page 1 and then continue on page 2, without the headers duplicated. Do you agree?

jonahkagan: agreed

nbhatia823: Will merge this and address in a follow-up PR today as I tend to prefer making more smaller PRs

Issue #1934, opened 2024-04-24
Make it possible to run multiple instances of the Arlo background worker


arsalansufi:
So that one audit's background processing doesn't have to block another's. This past audit season, OC's CVR processing, which can take on the order of an hour of background processing, almost coincided with GA's audit launch (audit launch requires background processing for sample size calculation).

benadida:
I want to call out that this is a risky piece of work that requires thinking carefully about how to prevent performing background jobs multiple times, how to recover from a failed job,etc.

jonahkagan:
Another potential workaround would be to allow Support to cancel a long-running job

PR #1990, opened 2024-10-16
Concurrent background workers

jonahkagan:
Task: #1934 

Updates the background worker so that multiple workers can safely run concurrently. Attempts to ensure that:
- Every job is run exactly once
- If a worker is available, it will immediately start on the next job in the queue
- If a worker is terminated (e.g. by a [Heroku dyno shutdown](https://devcenter.heroku.com/articles/dyno-shutdown-behavior)), any in-progress job will be re-enqueued

Note that there is no logical locking of jobs to prevent conflicts when trying to access the same resources, but we'll need to add that as an additional layer on top of this (since it will require domain-specific logic). Will tackle this in a follow up PR.

For testing, I added some automated testing to simulate multiple concurrent workers competing for jobs. I also created a Heroku review app and tested out running multiple workers simultaneously and restarting workers in the middle of jobs.

### DIFF
--- a/server/tests/test_background_tasks.py
+++ b/server/tests/test_background_tasks.py
@@ -1,10 +1,13 @@
-from datetime import datetime
 import logging
+import multiprocessing
+import random
+import time
 from unittest.mock import patch
 import sqlalchemy
 from sqlalchemy.orm import scoped_session, sessionmaker
 from sqlalchemy_utils import create_database, drop_database
 import pytest
+
 
 from .. import config
 from ..models import *  # pylint: disable=wildcard-import
@@ -16,16 +19,18 @@ from .helpers import (
     find_log,
 )
 from ..worker.tasks import (
+    claim_next_task,
     create_background_task,
     background_task,
+    reset_task,
+    run_task,
     serialize_background_task,
     UserError,
-    run_new_tasks,
 )
+from ..worker.worker import run_worker
 
 
-# Since the worker code assumes that only one worker is running at a time, we
-# give each test case with its own database to work with so there is no
+# We give each test case its own database to work with so there is no
 # interference with tests running concurrently (both among tests in this file
 # and with the other tests).
 @pytest.fixture
@@ -94,7 +99,7 @@ def test_task_happy_path(caplog, db_session):
 
     assert task_ran is False
 
-    run_new_tasks(db_session)
+    run_task(claim_next_task("test_worker", db_session), db_session)
 
     task = db_session.query(BackgroundTask).get(task_id)
     compare_json(
@@ -115,7 +120,8 @@ def test_task_happy_path(caplog, db_session):
         (
             f"TASK_START {{'id': '{task_id}', "
             "'task_name': 'happy_path',"
-            f" 'payload': {{'arg2': 2, 'arg1': 1}}}}"
+            f" 'payload': {{'arg2': 2, 'arg1': 1}},"
+            " 'worker_id': 'test_worker'}"
         ),
     )
     assert find_log(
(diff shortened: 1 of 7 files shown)


server/worker/tasks.py
```diff
@@ -128,28 +124,25 @@ def run_task(task: BackgroundTask, db_session=db_session):
...
+        .order_by(BackgroundTask.created_at)
+        .with_for_update()
+        .limit(1)
+        .one_or_none()
     )
-    for task in stuck_tasks:
-        task.started_at = None
-        logger.info(f"TASK_RESET {task_log_data(task)}")
+    if task:
+        task.started_at = datetime.now(timezone.utc)
```
benadida, line 137: is it worth recording maybe a process ID in addition to the time at which it's started, in case we need to do cleanup later?

server/worker/worker.py
```diff
@@ -1,21 +1,48 @@
...
+    # it can be picked up by another worker.
+    def interrupt_handler(*_args):
+        nonlocal task
+        if task:
+            reset_task(task)
+            task = None
+        sys.exit(1)
+
+    signal.signal(signal.SIGTERM, interrupt_handler)
+    signal.signal(signal.SIGINT, interrupt_handler)
```
benadida, line 32: this looks good in principle, have we tested that these are the signals we get on Heroku and we have time to try to reset the task?

benadida (commented):
I like the architecture. Using the DB with select-for-update to claim a task makes a ton of sense. I would consider augmenting with adding a process identifier to the task, and adding yet another check in the update that says "update start time and process ID where start time and process ID are null". Belt and suspenders kind of thing.

I also like just using Heroku's process based scaling.

server/worker/tasks.py
jonahkagan: Can you say a little more about how you envision the process ID enabling cleanup? What sort of situations would this help with?

server/worker/worker.py
jonahkagan: Haven't tested it yet, just going on the [Heroku docs](https://devcenter.heroku.com/articles/dyno-shutdown-behavior) so far. They claim the dyno will receive one or more SIGTERMs, then get 30s to shutdown before receiving a SIGKILL.

server/worker/tasks.py
benadida: possibly a situation where a process dies and doesn't get a proper chance to clean up, then there's a trace of which process it was that grabbed it. Partly I'm nervous about the state of who owns a task being only stored in RAM.

jonahkagan: I see. If we could differentiate the currently running workers from past workers, then I see how this could be useful. I bet there's some ID assigned to each dyno that we could access.

That being said, I think it's also technically solvable if we know how many workers are currently running. E.g. if there are 2 running workers and 3 active tasks, the active task with the oldest start time is not actually being run by a worker and should be cleaned up.

I'm not opposed to the process ID though, I agree it could come in handy hypothetically.

arsalansufi:
Planning to properly review right around/after planning!

server/tests/test_background_tasks.py
```diff
@@ -367,6 +389,72 @@ def interrupted(num):
...
+        time.sleep(0.1)
+
+    for worker in workers:
+        worker.terminate()
+
+    expected_sorted_results = list(range(num_tasks))
+    # Each task should have run exactly once
+    assert sorted(results) == expected_sorted_results
+    # Tasks should not have run in order, since some got interrupted and reset
+    assert results != expected_sorted_results
```
arsalansufi, line 460: Nice, good check!

arsalansufi (approved):
LGTM!

PR #1991, opened 2024-10-16
chore: re-sync `.node-version` file

eventualbuddha:
It says it wants NodeJS v12, but we install NodeJS v16. This sets the version to the latest v16.


jonahkagan (approved):
Good catch!

PR #1992, opened 2024-10-16
Update batch comparison audit reports to better handle extra batches

carolinemodic:
Where there are "extra" batches included we want to ignore those in the summary round-level data. This change updates the logic introduced in https://github.com/votingworks/arlo/pull/1988 to filter out "extra" batches when computing those fields. 

It also adds a "totals" row to the end of the batches section to still capture the information with the extra batches included. This row just computes totals for number of ballots, audited results, and reported results. Note that if there are multiple rounds occurring in the audit the totals will the be total of the batches from all rounds combined. This can be seen in some of the snapshot test updates.

### DIFF
--- a/server/api/reports.py
+++ b/server/api/reports.py
@@ -981,7 +981,7 @@ def sampled_batch_rows(election: Election, jurisdiction: Jurisdiction = None):
             pretty_choice_votes(total_audit_results_by_name),
             pretty_choice_votes(total_reported_results_by_name),
             "",  # change in results not calculated for totals
-            "",  # change in margin not calculcated for totals
+            "",  # change in margin not calculated for totals
         ]
     totals_row += ""  # last edited col
     rows.append(totals_row)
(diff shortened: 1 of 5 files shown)


server/api/reports.py
```diff
@@ -947,6 +965,26 @@ def sampled_batch_rows(election: Election, jurisdiction: Jurisdiction = None):
...
+        }
+        total_audit_results_by_name = {
+            choice.name: total_audit_results[contest.id][choice.id]
+            for choice in contest.choices
+        }
+        totals_row += [
+            pretty_choice_votes(total_audit_results_by_name),
+            pretty_choice_votes(total_reported_results_by_name),
+            "",  # change in results not calculated for totals
+            "",  # change in margin not calculcated for totals
```
jonahkagan, line 984: fix typo: calculcated -> calculated

jonahkagan (approved):
Nice catch. I like this fix - seems like it will probably satisfy most use cases

PR #1993, opened 2024-10-16
Update pytest-cov to 3.0.0

nbhatia823:
Seeing an issue where `coverage` is being installed without the C extension, resulting in a ModuleNotFoundError when looking for `coverage.tracer`'s `CTracer`. This package comes from `pytest-cov 2.12.1` and results in `coverage 5.5 without the C extension`. 

I was able to resolve this by upgrading `pytest-cov` which requires `coverage` under the hood. `v3.0.0` seems to fix this, as this installs `coverage 6.5.0 with the C extension` for me. Why? Well, I've done some digging and this is my current understanding...

The C extension is automatically installed if the environment supports it
`Most of the time, the C extension will be installed without any special action on your part.`

And then it says:
`If you are installing on Linux, you may need to install the python-dev and gcc support files before installing coverage via pip`

I've pulled the recommended packages and I'm still not seeing the C extension get downloaded when on 2.12.1. Looking at the changelog for 3.0.0, it's not clear what might cause this, though it does say they `updated various test dependencies`, which might have an effect, and separately 3.0.0 removed a hack they were using which was related to the coverage package, so maybe one of those was the root cause here. 

The main difference is that older versions of Python are dropped, so it should be safe for us to upgrade to 3.0.0.


poetry.lock
```diff
@@ -2,20 +2,22 @@
 
 [[package]]
 name = "alembic"
-version = "1.6.5"
+version = "1.13.3"
```
jonahkagan, line 5: It looks like this also updated a bunch of other packages. In theory, that should be fine if they are all semantically versioning correctly, but it's a bit risky.

I'd prefer if you undo the changes to poetry.lock and then use `poetry lock --no-update` to update the lockfile with just the `pytest-cov` changes.

I think separately we should do a PR to update all the dependencies so we don't need to worry about this in the future, but I'd like to separate it just in case something goes wrong and we need to revert it.

jonahkagan (changes requested):

poetry.lock
nbhatia823: Hm, just went to go do that but realized that I actually need `coverage` to update to `6.5` which does not happen if I include `no-update` as it only updates `pytest-cov` then 🤔

jonahkagan: Oh, it doesn't update the direct dependencies when you do that? That's weird.

In the case maybe we can make a PR to do the general update first and base this one off of it? Or I could be convinced to just let this slide and hope it works out lol

nbhatia823: I know right, I thought it would as well but chatgpt cleared is up - 

```
The command poetry lock --no-update does not update direct dependencies. It regenerates the poetry.lock file without modifying the currently installed versions of the direct dependencies listed in the pyproject.toml file.

With --no-update: It regenerates the poetry.lock file, ensuring it's in sync with the current pyproject.toml but does not go out to fetch newer versions of any packages. It uses the existing versions defined in the lock file for both direct and transitive dependencies.
```

`In the case maybe we can make a PR to do the general update first and base this one off of it?` - sorry I'm a bit unsure what that means 😅 

I wonder if we should wait till at least one other person sees this issue before we make the change, and I can just keep the change locally for now. If we see someone else have the same issue, we might feel more confident that the change is worth the risk

nbhatia823: Updating deps in this [PR](https://github.com/votingworks/arlo/pull/1997). That PR also pulls in `coverage 6.5` so jonah was spot on with recommending we do that isolated update first. Closing this PR now.

PR #1994, opened 2024-10-16
Improve tally sheet page break for tables

nbhatia823:
Follow up to this [thread](https://github.com/votingworks/arlo/pull/1989#discussion_r1803500062)

This PR:
1. Updates table page break behavior to avoid situations where a table only prints header cells before the page ends, as it then reprints the headers on the next page
2. Adds text to indicate when a table has continued from a previous page


nbhatia823:
I'm not feeling strongly that commit 1 or 2 are better than what we currently have. Curious what you think @jonahkagan?

client/src/components/JurisdictionAdmin/generateSheets.ts
```diff
@@ -452,6 +452,7 @@ export const downloadBatchTallySheets = async (
           right: pageMargin,
           top: pageMargin,
         },
+        pageBreak: 'avoid',
```
nbhatia823: Considering only adding this for contests 2+, as we probably want contest 1 printing on the first page always

client/src/components/JurisdictionAdmin/__snapshots__/generateSheets_downloadBatchTallySheets_Handles_more_than_two_contests.png
nbhatia823, line 1: This actually doesn't look too bad, with the table page breaks but without the headers reprinted

jonahkagan:
I think we need the headers on both pages for tables that are split across pages to make sure there's clear context when using the backside of the page.

I was hoping there would be some way to avoid splitting tables that were starting so close to the bottom that we'd only see the headers on the starting page. Since there could theoretically be more than 2 pages, we'd need some logic that could apply on any page. Maybe checking `startY` could do it?

client/src/components/JurisdictionAdmin/generateSheets.ts
```diff
@@ -474,6 +487,17 @@ export const downloadBatchTallySheets = async (
...
         },
+        willDrawCell(data) {
+          // Indicate if a table is continuing from a previous page
+          if (
+            data.cell.section === 'head' &&
+            data.row.index === 0 &&
+            data.pageNumber > 1
+          ) {
+            // eslint-disable-next-line no-param-reassign
+            data.cell.text[0] += ' (continued)'
```
jonahkagan, line 498: Love this addition!

jonahkagan (approved):

PR #1997, opened 2024-10-18
Run poetry lock

nbhatia823:
This PR runs `poetry lock` to update server dependencies according to our existing semantic versioning for deps as listed in `pyproject.toml`. Discussion [here](https://votingworks.slack.com/archives/CKCVA0F9S/p1729187464644569).

Testing: Went through a batch comparison audit to test the normal flow.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2,20 +2,22 @@
 
 [[package]]
 name = "alembic"
-version = "1.6.5"
+version = "1.13.3"
 description = "A database migration tool for SQLAlchemy."
 optional = false
-python-versions = "!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,!=3.5.*,>=2.7"
+python-versions = ">=3.8"
 files = [
-    {file = "alembic-1.6.5-py2.py3-none-any.whl", hash = "sha256:e78be5b919f5bb184e3e0e2dd1ca986f2362e29a2bc933c446fe89f39dbe4e9c"},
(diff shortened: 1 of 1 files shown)


poetry.lock
```diff
@@ -337,189 +476,198 @@ resolved_reference = "62d87b3a64160600825b8d818ba89b519288517b"
...
-test-no-codebase = ["Pillow", "matplotlib", "pytest"]
+docs = ["furo", "sphinx (>=7.2)", "sphinx-copybutton"]
+mypy = ["contourpy[bokeh,docs]", "docutils-stubs", "mypy (==1.11.1)", "types-Pillow"]
+test = ["Pillow", "contourpy[test-no-images]", "matplotlib"]
+test-no-images = ["pytest", "pytest-cov", "pytest-rerunfailures", "pytest-xdist", "wurlitzer"]
 
 [[package]]
 name = "coverage"
-version = "5.5"
+version = "6.5.0"
```
nbhatia823, line 563: This was the change that I was wanting from upgrading `pytest-cov`, so after this I am able to pull `Coverage` with the `C Extension`, so it looks like we won't also need to upgrade `pytest-cov`. Good call!

jonahkagan (approved):
Looks great! Thanks for taking this on. As a final test to make sure it also works on Heroku, I recommend creating a Review App before merging and verifying that it builds and deploys.

nbhatia823:
Successfully deployed and created an audit on the review app [here](https://arlo-update-server-deps-6rkrqj.herokuapp.com/)

PR #1999, opened 2024-10-21
Update missing snapshots

jonahkagan:
Somehow it seems like a bunch of snapshots for test_cvrs.py got deleted. When I run the tests on `main`, they added a bunch of snapshot data as if the snapshots had never been created at all. I'm not sure what happened, but since we haven't actually modified code related to these tests, I'm ok with just updating the snapshots to the current result of running the tests.

### DIFF
--- a/server/tests/ballot_comparison/snapshots/snap_test_cvrs.py
+++ b/server/tests/ballot_comparison/snapshots/snap_test_cvrs.py
@@ -7,6 +7,982 @@ from snapshottest import Snapshot
 
 snapshots = Snapshot()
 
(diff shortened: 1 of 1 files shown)


carolinemodic (approved):
hmm I think I may have accidently deleted these... now thinking about this I think I was running update-snapshot when I was filtering to a subset of tests so maybe that caused this test to delete the snapshots related to tests not matching the filter?

Bizarre it wouldn't fail on CI

jonahkagan:
> Bizarre it wouldn't fail on CI

I think the default behavior is to create any missing snapshots and pass the tests (because it assumes that it's the first time you're running those tests)

Issue #1932, opened 2024-04-24
Provide an in-product solution for the case that a physically combined batch is selected


arsalansufi:
I don't fully understand this problem or the workaround that Mark came up with in the field but at least wanted to document it.

https://votingworks.slack.com/archives/C013V6CT9H9/p1712704071287439?thread_ts=1712701354.438949&cid=C013V6CT9H9

https://docs.google.com/document/d/1eio3nKzbfHID9sjlG0I92NJ0YaLswVcDp2EFo4uIuds/edit?disco=AAAAq0ybV3s

ginvdr:
More context: https://votingworks.slack.com/archives/CKCVA0F9S/p1717596026680309

PR #2000, opened 2024-10-21
Combine commingled batches during audit

jonahkagan:
Closes: https://github.com/votingworks/arlo/issues/1932

In batch comparison audits, sometimes auditors discover after launching the audit the some of the sampled batches have been physically commingled such that it's impossible to know which ballots belonged to which batch. This PR adds the ability for Support Users to combine batches after the audit has started to mitigate this issue. A combined batch may consist of both sampled and unsampled batches.

Combined batches will then appear to auditors like any other batch, and the original batches that were combined will be hidden. The audit math is updated to conservatively incorporate these combined batches into the risk measurement calculation. Finally, the discrepancy report and audit reports are updated to include the combined batch (while also showing which original batches were combined to make it).

On the technical side, since this is an edge case, I wanted to implement it in a lightweight way that didn't fundamentally alter the data model and require widespread changes. While it's important to get the math and main flows right, if there are certain places in Arlo that don't take into account combined batches, it's ok.

Thus, I came up with the following approach:
- Tag batches that are combined with a common `combined_batch_name` field
- Pick one representative sampled batch for each combined batch. This batch will be used to store the combined batch audit results. The rest of the batches in the combined batch are given tallies with 0 votes for each candidate. (This means we don't need to create some new entity to represent the "combined batch")

Support Tools interface

https://github.com/user-attachments/assets/23a63617-a2e0-4fad-aec2-88b1e1b35dc2

Jurisdiction auditing interface (unchanged, just shows combined batch)
Screenshot 2024-10-23 at 11 14 04 AM

### DIFF
--- a/server/api/batches.py
+++ b/server/api/batches.py
@@ -7,7 +7,6 @@ from werkzeug.exceptions import BadRequest, Conflict
 from sqlalchemy.orm import Query, joinedload
 from sqlalchemy import func
 
-
 from . import api
 from ..auth import get_loggedin_user, get_support_user, restrict_access, UserType
 from ..database import db_session
--- a/server/api/batches.py
+++ b/server/api/batches.py
@@ -7,11 +7,12 @@ from werkzeug.exceptions import BadRequest, Conflict
 from sqlalchemy.orm import Query, joinedload
 from sqlalchemy import func
 
+
 from . import api
 from ..auth import get_loggedin_user, get_support_user, restrict_access, UserType
 from ..database import db_session
 from ..models import *  # pylint: disable=wildcard-import
-from .shared import get_current_round
+from .shared import get_current_round, group_combined_batches
 from ..util.csv_download import csv_response, jurisdiction_timestamp_name
 from ..util.jsonschema import JSONDict, validate
 from ..util.isoformat import isoformat
@@ -22,6 +23,27 @@ from ..activity_log.activity_log import (
     record_activity,
 )
 from ..util.get_json import safe_get_json_list
+
+
+def replace_combined_batches_with_representative_batches(
+    batches: List[Batch],
+) -> List[Batch]:
+    regular_batches = []
+    all_sub_batches = []
+    for batch in batches:
+        if batch.combined_batch_name is None:
+            regular_batches.append(batch)
+        else:
+            all_sub_batches.append(batch)
+
+    combined_batches = group_combined_batches(all_sub_batches)
+    representative_batches = [
+        combined_batch["representative_batch"] for combined_batch in combined_batches
+    ]
+    for representative_batch in representative_batches:
+        representative_batch.name = representative_batch.combined_batch_name  # type: ignore
+
+    return regular_batches + representative_batches
 
 
 def already_audited_batches(jurisdiction: Jurisdiction, round: Round) -> Query:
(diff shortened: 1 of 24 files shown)


server/api/shared.py
```diff
@@ -92,6 +92,31 @@ def batch_tallies(contest: Contest) -> BatchTallies:
...
 
 
+class CombinedBatch(TypedDict):
+    name: str
+    representative_batch: Batch
+    sub_batches: List[Batch]
+
+
+def combined_batch_representative(sub_batches: List[Batch]) -> Batch:
+    assert len(sub_batches) > 0
```
nbhatia823, line 102: Any additional need to assert that the sub_batches are unique?

server/api/batches.py
```diff
@@ -7,11 +7,12 @@
 from sqlalchemy.orm import Query, joinedload
 from sqlalchemy import func
 
+
```
nbhatia823, line 10: nit: extra line

server/tests/audit_math/test_macro.py
jonahkagan, line 1: @MarkVVaudits could you review these tests?

nbhatia823 (commented):
Went commit by commit and it lgtm! Appreciate the thorough tests

nbhatia823 (approved):

arsalansufi:
Will review tomorrow AM!

server/tests/audit_math/test_macro.py
MarkVVaudits: Looked at the latest version of the file, focusing on the calculations. They look great -- it's really nice how the calculations are self-documenting.

server/api/reports.py
```diff
@@ -891,17 +923,22 @@ def sampled_batch_rows(election: Election, jurisdiction: Jurisdiction = None):
...
             )
             if reported_results is not None:
-                for choice in contest.choices:
-                    total_reported_results[contest.id][choice.id] += reported_results[
-                        choice.id
-                    ]
+                # For combined batches, we need to add the combined batch reported
+                # results to the total, since it may include some unsampled batches.
+                # However, for audit results, we want to use only the sampled
+                # batches to avoid double counting.
```
arsalansufi, line 936: Hmm, I'm having some trouble following what the double counting here would be

arsalansufi (commented):
Looks great! Some quick thoughts and clarifying questions

server/api/shared.py
jonahkagan: It's not something I'm super worried about given the context (since we're never dealing with lists of batches that have duplicates)

server/api/shared.py
```diff
@@ -92,6 +92,31 @@ def batch_tallies(contest: Contest) -> BatchTallies:
...
+    sampled_sub_batches = [sub_batch for sub_batch in sub_batches if sub_batch.draws]
+    return sorted(sampled_sub_batches, key=lambda batch: batch.id)[0]
```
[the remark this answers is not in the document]
jonahkagan: Good question, I think when I wrote it I decided that crashing was fine, since there would be no motivation for a support user to try to combine batches without including a sampled batch lol. But being more explicit couldn't hurt. I'll add it

server/api/reports.py
jonahkagan: Yeah, I got confused reading this comment again lol. Edited to hopefully clarify 3c69ed2f

server/api/reports.py
```diff
@@ -456,10 +457,28 @@ def round_rows(election: Election):
...
+            num_distinct_batches = (
+                len(distinct_batches)
+                - len(combined_sub_batches)
+                + len(combined_batches)
+            )
+
+            for combined_batch in combined_batches:
+                for sub_batch in combined_batch["sub_batches"]:
+                    if not any(sub_batch.id == batch.id for batch in distinct_batches):
+                        distinct_batches.append(sub_batch)
```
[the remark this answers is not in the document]
jonahkagan: Yeah, it is a little confusing. Added some comments to try to clarify - let me know if that helps 3c69ed2f

server/api/reports.py
jonahkagan: After rebase, that's actually: 153f72029c11dc6fa562f6bd1e86b227a8009eec

server/api/reports.py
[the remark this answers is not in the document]
jonahkagan: After rebase, that's: 153f72029c11dc6fa562f6bd1e86b227a8009eec

PR #2001, opened 2024-10-21
Pin sentry sdk version

nbhatia823:
Seeing `main` fail because `ImportError: cannot import name 'OP' from 'sentry_sdk.consts'`. Oddly, it works on branches, like [this run ](https://app.circleci.com/pipelines/github/votingworks/arlo/7180/workflows/3d7d0d49-92fe-43d8-a7a7-3a49ddd66a0a)that was just branched off of `main` with no changes. That makes me think it is a caching issue on main, but I cleared the DLC cache and am still seeing the same issue. So, this PR pins `sentry-sdk` to the version before [this PR](https://github.com/votingworks/arlo/pull/1997), as if this fixes the issue, it might not be worth investigating further as I have no other leads.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -180,17 +180,17 @@ files = [
 
 [[package]]
 name = "boto3"
-version = "1.35.43"
+version = "1.35.45"
 description = "The AWS SDK for Python"
 optional = false
 python-versions = ">=3.8"
 files = [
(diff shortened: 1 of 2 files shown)


carolinemodic (approved):

nbhatia823:
Closing as it seems we found the [root cause as a bug](https://github.com/votingworks/arlo/pull/2003) in the `.circleci/config` caching behavior that was preventing new python deps from being pulled

PR #2003, opened 2024-10-21
Bump cache-version

nbhatia823:
Follow up to https://github.com/votingworks/arlo/pull/2002 which didn't fix the issue. Maybe since we are caching all of ~./cache, that is also caching the poetry deps even though that ideally should be separated out.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,11 +103,10 @@ jobs:
           docker_layer_caching: true
       - python/install-packages:
           pkg-manager: poetry
-          cache-version: v2
       - node/install-packages:
           pkg-manager: yarn
           app-dir: client
-          cache-version: v3
+          cache-version: v4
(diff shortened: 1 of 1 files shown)


jonahkagan (approved):

.circleci/config.yml
```diff
@@ -103,11 +103,10 @@ jobs:
           docker_layer_caching: true
       - python/install-packages:
           pkg-manager: poetry
-          cache-version: v2
```
jonahkagan, line 106: I think we should maybe leave this in just to be safe

nbhatia823: Oh okay. Thinking it might have a side effect if we go down a cache-version?

jonahkagan: I just don't know what's going on so refreshing all the caches seems safest. Blunt tool 🤷

PR #2006, opened 2024-10-22
Add dependencies that are required to build psycog2-binary on some VMs

nbhatia823:
We've now seen multiple cases where installing `gcc` and `python-dev` allow for the `psycog2-binary` to properly build, so it seems worth adding to our `deps` command.

To test, I created a fresh VM and was able to run `make deps` successfully without any additional commands, while previously I would run into the `psycog2-binary` issue on each VM I created.

As a follow up task, I'm thinking the `Makefile` might benefit from a bit of a reorganization, but will limit this PR to this change since it is impeding some people's setup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ PATH := $(PATH):$(HOME)/.local/bin
 
 deps:
 	sudo apt update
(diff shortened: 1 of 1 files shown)


jonahkagan (approved):
Thanks for adding this!

Issue #1857, opened 2023-11-30
Second round escalation in batch comparison needs review


carolinemodic:
Review of Arlo behavior is needed of the https://votingworks.slack.com/archives/C013V6CT9H9/p1700517973088139situation in Loudon Co, VA> where a .102 risk measurement escalated to a full hand tally for two reasons:
1. Because all remaining batches were immediately selected vs. a random selection which may have selected batches multiple times
2. 17 new batches being selected when the risk measurement was just below the risk limit is extreme

Issue automatically created from slack feedback reported by ginny 
https://votingworks.slack.com/archives/C01F2CH62TS/p1701354091136549

PR #2008, opened 2024-10-23
Improve batch computations

MarkVVaudits:
Primarily this addresses #1857. Sample size calculations after round 1 should take the cumulative measured risk (p-value) into account; U does not change. 

Secondarily, compute_risk was set to terminate as soon as the risk limit was reached. That's valid, but tends to produce inflated p-values in many use cases. In discussion with @ginvdr we agreed that a good conservative approach is to use all the data from the RLA sample for the contest. (In principle it is legitimate to use the sequential minimum, which might be less if a batch late in the sample sequence has large discrepancies -- but that seems too clever.)

### DIFF
--- a/server/audit_math/macro.py
+++ b/server/audit_math/macro.py
@@ -53,9 +53,6 @@ def compute_error(
         the maximum across-contest relative overstatement for batch p
     """
 
-    if contest.name not in batch_results:
-        return None
-
     def error_for_candidate_pair(winner, loser) -> Optional[BatchError]:
         v_wp = batch_results[contest.name][winner]
         v_lp = batch_results[contest.name][loser]
@@ -171,9 +168,10 @@ def get_sample_sizes(
     ticket_numbers: Dict[str, BatchKey],
 ) -> int:
     """
-    Computes initial sample sizes parameterized by likelihood that the
-    initial sample will confirm the election result, assuming no
-    discrepancies.
+    Computes a sample size expected to confirm the election result
+    (attain the risk limit). The base computation assumes no discrepancies;
+    1 is added to accommodate small discrepancies. For rounds after the first,
+    the sample size depends on the measured risk so far.
 
     Inputs:
         risk_limit       - the risk-limit for this audit
@@ -196,15 +194,7 @@ def get_sample_sizes(
                          reported_results
 
     Outputs:
-        samples - dictionary mapping confirmation likelihood to sample size:
-                {
-                   contest1:  {
-                        likelihood1: sample_size,
-                        likelihood2: sample_size,
-                        ...
-                    },
-                    ...
-                }
+        sample_size - sample size (currently a single integer value)
     """
     alpha = Decimal(risk_limit) / 100
     assert alpha < 1, "The risk-limit must be less than one!"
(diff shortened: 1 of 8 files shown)


server/tests/batch_comparison/snapshots/snap_test_batch_comparison.py
jonahkagan, line 1: These snapshots seem to be removing some of the fields that were recently added to the audit report in https://github.com/votingworks/arlo/pull/1988. I think you can fix that by rebasing your branch on `main` and then rerunning the tests. Let me know if you need help figuring that out.

server/audit_math/macro.py
jonahkagan, line 1: Is there a paper I could check out to get an understanding of the changes to the sample size logic to validate them? I trust you but want to do my diligence as a reviewer (and want to learn!)

jonahkagan (commented):
Looks great so far! I am holding off on approving for now until I can review the snapshot updates to make sure they look right

server/audit_math/macro.py
MarkVVaudits: I don't think anyone has written on how to choose batch sample sizes after the first round. But the logic is pretty straightforward. As shown in compute_risk, the p value is a repeated product of `(1 - 1/U) / (1 - taint)`, which reduces to 1 - 1/U if all taints are 0. For batches already audited, we use the observed taints to compute the measured risk or p value; suppose that after round 1, the measured risk = p, which is greater than the risk limit _alpha_. To simplify notation, let's abbreviate 1 - 1/U as _mult_ (the multiplier when taint is 0), So we're looking for n such that `p * mult^n <= alpha`. Taking logs [any base, but here I'll use math.log() aka ln] , we have `log(p) + n * log(mult) <= log(alpha)`, so `n >= (log(alpha) - log(p)) / log(mult)`. (The inequality reverses because mult < 1, so log(mult) < 0.)  

For instance, suppose the margin is 5%, so U = 21, mult = 0.952..., log(mult) = -0.0488 or so. Further suppose that alpha is 10% = 0.1, so log(alpha) is about -2.30; and that p is 0.15 ("close, but not that close"), so log(p) is about -1.90. Then `(log(alpha) - log(p)) / log(mult)` is about 8.31, and 9 would probably be an adequate sample size, but the code conservatively returns 10. It's definitely helpful to have a mental model around this log math, in order to understand how discrepancies of various sizes affect the measured risk.

jonahkagan: Thanks! The insight around looking for a number of batches to sample (that we assume will have 0 taint) to create a risk-limit-meeting p-value is very clarifying. It might be nice to explain some of this in the code with comments (though not required).

server/audit_math/macro.py
```diff
@@ -159,14 +158,7 @@ def compute_U(
     """
     U = Decimal(0.0)
     for batch in reported_results:
-        if batch in sample_results:
-            error = compute_error(
-                reported_results[batch], sample_results[batch], contest
-            )
```
jonahkagan, line 165: It looks like L57 of this file is now not covered by any test, so the test coverage check is failing. My guess is that it's a result of removing this line (or maybe removing a test). In any case, the check seems unnecessary to me - we should never be computing the error for a batch without having its reported results, so I think it's probably safe to just remove that line.

server/tests/audit_math/test_macro.py
```diff
@@ -374,8 +399,12 @@ def test_compute_risk(contests, batches) -> None:
     sample = {}
     sample_ticket_numbers = {}
 
+    # Contest A: margin = 0.05, U = 21
+    # Contest B: margin = 0.1, U = 11
+    # Contest C: margin = 0.15, U = 23/3 = 7.666...
+
     # Draws with taint of 0
-    for i in range(31):
+    for i in range(30):
```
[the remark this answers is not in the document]
MarkVVaudits: Sure.

server/audit_math/macro.py
MarkVVaudits: Yes, I broadly favor that direction.

server/audit_math/macro.py
MarkVVaudits: Agreed.

MarkVVaudits:
This looks clean except for a CircleCI Cypress error that baffles me.

jonahkagan (approved):
Thanks for adding all those comments! I reran the Cypress test because it seemed like there was no reason for the failure, hopefully it passes and it was just a weird blip

Issue #1547, opened 2022-07-13
Print audit board signs for sort-and-stack auditing


jonahkagan:
![image](https://user-images.githubusercontent.com/530106/178799704-389cf0e9-aef6-45c1-b51d-5c39277ffb72.png)

This applies in batch comparison audits specifically.

ginvdr:
https://docs.google.com/document/d/1f0jQ2h_T4LkpEkzWpIYoWiNPkNLiNowbJiinx6sk41E/edit?usp=sharing

PR #2009, opened 2024-10-23
Stack labels for batch comparison audits

nbhatia823:
Issue: https://github.com/votingworks/arlo/issues/1547
Slack convo: https://votingworks.slack.com/archives/CKCVA0F9S/p1729531494392819

Allow for users to download stack labels to assist with sorting votes during batch comparison audits.

Design decisions:
- If last names are long, use a smaller font size. If they are really long, truncate the name
- Keep last name suffixes. Other than than, the last name parsing is simplistic, and won't currently handle cases like "John Von Dutch" well as it will just be "Dutch". I don't have a better solution for this for now.
- Rotate through a set of 10 colors
- Truncate long contest names 

Via JurisdictionAdmin:

https://github.com/user-attachments/assets/83ba79c1-307b-47fd-8844-6896380e715f

Via AuditAdmin:

https://github.com/user-attachments/assets/0fd330cf-25af-43b9-b0e6-9d9d3097d956

**NOTE - renamed Audit Board Signs to Stack Labels**


client/src/components/JurisdictionAdmin/generateSheets.ts
```diff
@@ -795,3 +797,128 @@ export const downloadTallyEntryLoginLinkPrintout = async (
...
+  Colors.GREEN1,
+  Colors.VIOLET1,
+  Colors.ORANGE1,
+  Colors.VERMILION1,
+  Colors.BLUE1,
+  Colors.COBALT1,
+  Colors.GOLD1,
+  Colors.SEPIA1,
+  Colors.DARK_GRAY1,
+  Colors.INDIGO1,
```
jonahkagan, line 815: I noticed some of the colors looked pretty similar in the sample (specifically the bluish ones and the orangish ones). How do they look when printed? Are they distinguishable?

I like using the dark variants for high contrast readability - good choice there. Maybe going one variant lighter would help with differentiation?

client/src/components/JurisdictionAdmin/generateSheets.ts
```diff
@@ -795,3 +797,128 @@ export const downloadTallyEntryLoginLinkPrintout = async (
...
+]
+const DEFAULT_AUDIT_BOARD_SIGNS = [
+  { title: 'No Vote', subtitle: '(undervote, overvote, blank, etc.)' },
+  { title: 'For Review', subtitle: '(undetermined, duplicated, etc.)' },
+]
```
jonahkagan, line 820: I think we can remove parens from subtitles since they are already styled in a way that makes it clear they are less important. Also we might want to capitalize the first letter of the subtitle if we remove the parens

client/src/components/JurisdictionAdmin/generateSheets.ts
```diff
@@ -795,3 +797,128 @@ export const downloadTallyEntryLoginLinkPrintout = async (
...
+
+function parseSurname(fullName: string, maxLength: number) {
```
jonahkagan, line 822: I think this approach is pretty good as a starting point!

client/src/components/JurisdictionAdmin/generateSheets.ts
```diff
@@ -795,3 +797,128 @@ export const downloadTallyEntryLoginLinkPrintout = async (
...
+          )}...`
+        : contest.name
+    contest.choices.forEach(choice => {
+      const candidate = {
+        title: parseSurname(choice.name, AUDIT_BOARD_SIGN_MAX_SURNAME_LENGTH),
+        subtitle: `(Contest: ${contestName})`,
+      }
+      candidateSigns.push(candidate)
+    })
+  })
```
jonahkagan, line 856: Nit: consider using `contests.flatMap` and `contest.choices.map`

client/src/components/JurisdictionAdmin/generateSheets.ts
```diff
@@ -795,3 +797,128 @@ export const downloadTallyEntryLoginLinkPrintout = async (
...
+      contest.name.length > AUDIT_BOARD_SIGN_MAX_CONTEST_NAME_LENGTH
+        ? `${contest.name.slice(
+            0,
+            AUDIT_BOARD_SIGN_MAX_CONTEST_NAME_LENGTH
+          )}...`
+        : contest.name
+    contest.choices.forEach(choice => {
+      const candidate = {
+        title: parseSurname(choice.name, AUDIT_BOARD_SIGN_MAX_SURNAME_LENGTH),
+        subtitle: `(Contest: ${contestName})`,
```
jonahkagan, line 852: I think parens and "Contest:" could be removed to simplify

client/src/components/JurisdictionAdmin/generateSheets.ts
```diff
@@ -795,3 +797,128 @@ export const downloadTallyEntryLoginLinkPrintout = async (
   )
   return doc.output() // returned for test snapshots
 }
+
+const AUDIT_BOARD_SIGN_MAX_CONTEST_NAME_LENGTH = 40
+const AUDIT_BOARD_SIGN_MAX_SURNAME_LENGTH = 20
+const AUDIT_BOARD_SIGN_TITLE_LENGTH_XL_FONT_CUTOFF = 12
+const AUDIT_BOARD_SIGN_TEXT_COLOR = Colors.GRAY3
```
jonahkagan, line 804: I think this could be slightly darker - it's a little hard to read on screen. Though if it's better printed, I defer to your judgement

client/src/components/JurisdictionAdmin/generateSheets.ts
```diff
@@ -795,3 +797,128 @@ export const downloadTallyEntryLoginLinkPrintout = async (
...
+    const titleColor =
+      AUDIT_BOARD_SIGN_TITLE_COLORS[i % AUDIT_BOARD_SIGN_TITLE_COLORS.length]
+
+    // Mid-page fold
+    doc.setFont('Helvetica', 'normal').setFontSize(defaultFontSize)
+    doc.setTextColor(AUDIT_BOARD_SIGN_TEXT_COLOR)
+    const foldCaption = 'Fold here and place at edge or table or tray'
+    let x = pageMargin / 2
+    let y = pageHeight / 2
+    doc.text(foldCaption, x, y)
```
jonahkagan, line 883: Should we add a dotted line? It's a classic fold signifier

client/src/components/JurisdictionAdmin/generateSheets.test.tsx
```diff
@@ -562,4 +563,74 @@ describe('generateSheets', () => {
...
+
+  const names = [
+    'Aaron Adams',
+    'Barry Batts',
+    'Crazy Cabbs',
+    'Danny Durnham Sr.',
+    'Elliot Ezekiel III',
+    'Farrih Fallahahahah ii',
+    'Farrih Fallahahahahah Jr',
+    'Hubert Blaine Wolfeschlegelsteinhausenbergerdorff Sr',
```
jonahkagan, line 575: 😆

jonahkagan (approved):
Looks great! A few small design notes. I think we should consider changing the name from "Audit Board Signs" (raised this question in Slack)

nbhatia823:
Renamed `Audit Board Signs` to `Stack Labels`

client/src/components/JurisdictionAdmin/generateSheets.ts
nbhatia823: Good call!

client/src/components/JurisdictionAdmin/generateSheets.ts
nbhatia823: Good points, I'll make some edits there. But, I don't actually have a color printer with me, so I'll post in the group and ask for someone to validate

client/src/components/JurisdictionAdmin/generateSheets.ts
nbhatia823: I was trying to mimic what Ginny had in her doc, but agreed that slightly darker is better. Shifting to `Gray2`

client/src/components/JurisdictionAdmin/BatchRoundSteps/DownloadStackLabelsButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+}: DownloadStackLabelsButtonProps): JSX.Element | null => {
+  const contestsQuery = useContestsJurisdictionAdmin(electionId, jurisdictionId)
+
+  const onClick = async () => {
+    /* istanbul ignore next */
+    while (!contestsQuery.isSuccess) {
+      if (contestsQuery.isError) return
+      // eslint-disable-next-line no-await-in-loop
+      await sleep(100)
+    }
```
kshen0, line 32: The query implementation uses `fetch` which should have a default timeout, but what do you think about explicitly exiting this loop after a certain duration? The combination of while loop + sleep seems like it could leave us open to accidentally blocking for a long time.

kshen0 (approved):

kshen0 (approved):

client/src/components/JurisdictionAdmin/generateSheets.ts
```diff
@@ -795,3 +797,125 @@ export const downloadTallyEntryLoginLinkPrintout = async (
...
+  const suffixes = ['sr.', 'sr', 'jr.', 'jr', 'ii', 'iii', 'iv', 'v']
+  const nameParts = fullName.trim().split(' ')
+  const lastNameIndex = suffixes.includes(
+    nameParts[nameParts.length - 1].toLowerCase()
+  )
+    ? nameParts.length - 2
+    : nameParts.length - 1
+  let surname = nameParts.slice(lastNameIndex).join(' ')
+  if (surname.length > maxLength) {
+    surname = `${surname.slice(0, maxLength)}...`
```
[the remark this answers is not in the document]
nbhatia823: Ah yeah, I actually tried using it as well as the unicode character for it and each resulted in a weird formatting issue. I assumed something with how `jspdf` handles encoding didn't work well with it and returned to this

client/src/components/JurisdictionAdmin/BatchRoundSteps/DownloadStackLabelsButton.tsx
nbhatia823: Hm yeah good point. I followed the pattern set in this [other button](https://github.com/votingworks/arlo/blob/main/client/src/components/JurisdictionAdmin/BatchRoundSteps/DownloadBatchTallySheetsButton.tsx#L34), but what you're saying does make sense and it does seem nice to be explicit about exiting after a certain duration. Though I don't have a good sense of what a more reasonable timeout might be, so I feel partially pulled to rely on the default `fetch` timeout and the existing pattern that we are relying on elsewhere we fetch the contests query to ensure behavior remains consistent

jonahkagan: It is a slightly scary looking pattern, I agree (I think I originally made it). However, the usual pattern would be to show some loading state instead of rendering the component until the data dependencies are loaded, which has equivalent risk of blocking for a long time. However, we don't usually worry about timeouts for something like that, since we're pretty confident the data will load just fine. So I'm not too worried here.

PR #2010, opened 2024-10-23
CircleCI: Cache Cypress separately

jonahkagan:
Previously, we were caching Cypress by including all of `~/.cache` in the node cache, which only updates when `client/yarn.lock` changes. We hypothesize that since `~/.cache` also included cached Python dependencies, changes to the Python dependencies would be overriden by this cache. To fix this, we cache Cypress separately and restore the original behavior of the node cache (only caching node deps).

To be safe, I updated the cache version for all the existing caches.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,15 +103,19 @@ jobs:
           docker_layer_caching: true
       - python/install-packages:
           pkg-manager: poetry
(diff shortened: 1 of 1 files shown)


nbhatia823 (approved):
Yessss! LGTM

PR #2012, opened 2024-10-23
Support large(r) file uploads by directly uploading to S3

carolinemodic:
https://github.com/votingworks/arlo/issues/1674

## Suppor large(r) file uploads by directly uploading to S3 

In order to enable larger file uploads in arlo directly upload from the client to s3 rather then doing "passthrough" file uploads through our server. This is the heroku recommended approach and generally recommended by most other heroku/s3 users. There will still be a 5GB max limit above which we need to chunk the transfer to s3. The process involves
1. Making a GET to <url>/upload-url to get the url to POST the file to for the actual upload. If configured to save files to the local file system this url will be a local server endpoint, otherwise it will be a presigned AWS url. 
2. Make a POST to the url returned in step 1 with the file attached. Again this will either be a server endpoint (/file) or a AWS endpoint. 
3. Make a POST to <url>/upload-complete to mark the file as uploaded and do any post-upload steps like updating DB elements and kicking off background jobs. 

I landed on keeping the upload-url endpoint for step 1 rather then the other suggested approaches as different files have different authentication which directly relates to the file path that that endpoint needs to construct (it needs to know about the election_id and in some cases jurisdiction_id) so it seemed simplest to leave it like this. 

This PR also changes CVR uploads to be 1 file and always be the "wrapper" zip we previously created for Hart/ESS. In the case of HART we now check if there were not any zip files inside of the top-level zip after unzipping to handle the situation where someone just uploads the zip of cvrs and does not "wrap" it in anything. There was some validation logic for hart around not having images or other file types in the zip that is now moves to happen in the processing step. 

I recommend reviewing by commit, I tried to isolate the bigger test refactors but let me know if you want me to break up any commits further. 

 If we wanted to support multiple file uploads in the future to not change this I think we would need to:
- Upload all selected files to the s3 bucket (this part is easy enough we can even call the presigned POST url multiple times )
- Change the way it references a single CVR file in the DB to either reference a path to the s3 bucket containing all the files, or allow a list of cvr files 
- Change the hart/ess background tasks to fetch all the files in that bucket instead of fetching the zip and unzipping.

### DIFF
--- a/client/src/components/JurisdictionAdmin/BatchInventory.tsx
+++ b/client/src/components/JurisdictionAdmin/BatchInventory.tsx
@@ -147,9 +147,7 @@ const useBatchInventoryCVR = (
         ),
     }),
     uploadFiles: files => {
-      const formData = new FormData()
-      formData.append('cvr', files[0], files[0].name)
-      return uploadFiles.mutateAsync(formData)
+      return uploadFiles.mutateAsync({ file: files[0] })
     },
     uploadProgress: uploadFiles.progress,
     deleteFile: () => deleteFile.mutateAsync(),
@@ -175,9 +173,7 @@ const useBatchInventoryTabulatorStatus = (
         ),
     }),
     uploadFiles: files => {
-      const formData = new FormData()
-      formData.append('tabulatorStatus', files[0], files[0].name)
-      return uploadFiles.mutateAsync(formData)
+      return uploadFiles.mutateAsync({ file: files[0] })
     },
     uploadProgress: uploadFiles.progress,
     deleteFile: () => deleteFile.mutateAsync(),
@@ -237,7 +233,7 @@ const SelectSystemStep: React.FC<{
           <HTMLSelect
             large
             onChange={setSystemType}
-            value={systemType ?? undefined}
+            value={systemType || undefined}
           >
             {!systemType && <option value={undefined}></option>}
             <option value={CvrFileType.DOMINION}>Dominion</option>
--- a/server/api/cvrs.py
+++ b/server/api/cvrs.py
@@ -33,6 +33,8 @@ from werkzeug.exceptions import BadRequest, NotFound, Conflict
 from sqlalchemy import func, and_
 from sqlalchemy.orm import Session
 
+from server.util.isoformat import isoformat
+
 from . import api
 from ..database import db_session, engine as db_engine
 from ..models import *  # pylint: disable=wildcard-import
@@ -44,23 +46,23 @@ from ..worker.tasks import (
     create_background_task,
 )
 from ..util.file import (
+    get_file_upload_url,
+    get_standard_file_upload_request_params,
     retrieve_file,
     serialize_file,
     serialize_file_processing,
-    store_file,
     timestamp_filename,
     unzip_files,
-    zip_files,
+    validate_zip_mimetype,
 )
 from ..util.csv_download import csv_response
 from ..util.csv_parse import (
     CSVIterator,
     decode_csv,
-    does_file_have_csv_mimetype,
     reject_no_rows,
     validate_comma_delimited,
-    validate_csv_mimetype,
     validate_not_empty,
+    validate_csv_mimetype,
 )
 from ..util.collections import find_first_duplicate
 from ..audit_math.suite import HybridPair
@@ -413,6 +415,7 @@ def parse_dominion_cvrs(
     jurisdiction: Jurisdiction,
 ) -> Tuple[CVR_CONTESTS_METADATA, Iterable[CvrBallot]]:
     cvr_file = retrieve_file(jurisdiction.cvr_file.storage_path)
+    print(cvr_file)
     cvrs = csv_reader_for_cvr(cvr_file)
 
     # Parse out all the initial metadata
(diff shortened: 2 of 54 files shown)


server/util/file.py
```diff
@@ -41,6 +41,7 @@ def s3():  # pylint: disable=invalid-name
         "s3",
         aws_access_key_id=config.AWS_ACCESS_KEY_ID,
         aws_secret_access_key=config.AWS_SECRET_ACCESS_KEY,
+        region_name="us-west-1",
```
carolinemodic, line 44: I had to add this or the url returned in the get presigned post call will not be domain specific which causes errors. I assume this should be an env var?

client/src/components/JurisdictionAdmin/BatchInventory.tsx
```diff
@@ -237,7 +233,7 @@ const SelectSystemStep: React.FC<{
           <HTMLSelect
             large
             onChange={setSystemType}
-            value={systemType ?? undefined}
+            value={systemType || undefined}
```
carolinemodic, line 236: the pre commit hook yelled about this line and made me change it

jonahkagan:
Overall feedback:
- I think the three step approach is an elegant solution to supporting both s3 and local file uploads.
- As you mentioned, I think we can take the idea a little further by making the local file upload interface more generic. It could even just mimic the S3 interface to simplify the rest of the code even further. Then it could definitely just be a single endpoint.
- If we mimic the S3 interface for local uploads, then step 1 could become simpler: we'd just need logic somewhere that tells us the S3 path for each file upload. This could potentially be a backend endpoint for each file upload like you have currently, or it could be a single backend endpoint. It could even just be a function on the frontend or hardcoded into each file upload on the frontend.
- It makes total sense to me to have all of the clearing logic and such in the step 3 endpoint handler. I think we could go further and do all the validation there as well. We don't really need to validate anything before uploading a file to S3, do we?

Python notes:
- Python type-checking is not great. I wouldn't waste too much time on trying to get it to work.

arsalansufi:
> The endpoint naming convention I've adopted, the idea of calling back to the server on local filesystem uploads.

I like the endpoint names! Very clear. I also can't think of a way to make this fewer than 3 calls.

And I like the idea of keeping the 3 calls for local file system uploads, too. To keep branching between local dev and prod as minimal as possible.

> The endpoint implementations in the server for Step 2 are basically identical, other then a few cases where the file type is different, should I have a generic endpoint for this or keep it duplicated for each file option?

I like the idea of consolidating if easy to do so, again just to make local dev as similar to prod as possible.

> I moved any sort of "clearing" step from occuring BEFORE the new uploaded file is stored to after in Step 3. I did this so that all of the db updates happen in the same endpoint, but I'm not sure if that matters and if this should happen somewhere else.

This makes sense to me!

> Validation logic - There are a few things like file type that get validated in multiple steps, is that necessary? Are there places where I should add more validation? I think some of my step 2 refactors may need more file type validation.

I like the idea of consolidating validation logic, probably in step 3. The only validation that I'd want to happen sooner is preventing wrong file types, and I think we already cover that on the frontend by virtue of the allowed file types that we specify when we instantiate the file input.

> Is it "filename" or "file_name" lol

I've always used `file_name` for consistency with `file_path`, `file_type`, etc. haha, but don't feel strongly

> I noticed in general that the validation in these endpoints was often refactored to its own function, however I would then hit issues where I needed to re-do it or add "type: ignore" markers because it wouldn't know in the main server endpoint body that some field was already validated to not be None. Is there a reason why we usually do this? Is it ok to not do this? Any suggestions?

I have in some cases used `typing.cast` to provide the Python type checker knowledge it's missing, e.g., https://github.com/votingworks/arlo/blob/ab2a9729f9d23038bc4ec0422c67b8d5d7eca574/server/api/standardized_contests.py#L132-L135

Might help in some of the cases you're describing? But also okay with punting as Jonah suggested.

-----

To the questions re wrapper ZIPs, assumptions around Hart/ES&S, etc., I believe we covered all those in [Slack](https://votingworks.slack.com/archives/CKCVA0F9S/p1729612395205649) but am happy to answer any follow up questions!

server/api/standardized_contests.py
```diff
@@ -177,43 +178,57 @@ def process_standardized_contests_file(election_id: str):
...
 
+
+@api.route(
+    "/election/<election_id>/standardized-contests/file/upload-complete",
+    methods=["POST"],
+)
+@restrict_access([UserType.AUDIT_ADMIN])
+def complete_upload_for_standardized_contests_file(election: Election):
+    if election.audit_type not in [AuditType.BALLOT_COMPARISON, AuditType.HYBRID]:
+        raise Conflict("Can't upload CVR file for this audit type.")
```
jonahkagan, line 217: Weird, I think this error message should say "Can't upload standardized contests file for this audit type." Looks like an existing bug. Also doesn't really matter since the UI doesn't allow it

server/api/cvrs.py
```diff
@@ -1035,19 +1034,31 @@ def parse_hart_cvrs(
 
     cvr_zip_files: Dict[str, BinaryIO] = {}  # { file_name: file }
     scanned_ballot_information_files: List[BinaryIO] = []
+    hasNonCsvZipFiles = []
```
jonahkagan, line 1037: Nit: remove `has` since this isn't a boolean

server/api/cvrs.py
```diff
@@ -1035,19 +1034,31 @@ def parse_hart_cvrs(
...
             )
+        else:
+            hasNonCsvZipFiles.append(file_name)
 
-    assert len(cvr_zip_files) != 0  # Validated during file upload
+    # If there are no zip files inside the "wrapper" we assume it was not a wrapper and there was only one cvr zip file uploaded, unwrapped.
+    if len(cvr_zip_files) == 0 and len(scanned_ballot_information_files) == 0:
+        cvr_zip_files[jurisdiction.cvr_file.name] = retrieve_file(
+            jurisdiction.cvr_file.storage_path
+        )
```
jonahkagan, line 1056: Do we need to download the file again in this case, or is there a way to reuse it since we already have it?

server/tests/ballot_comparison/test_cvrs.py
```diff
@@ -377,11 +365,12 @@ def test_dominion_cvrs_with_leading_equal_signs(
     rv = client.get(
         f"/api/election/{election_id}/jurisdiction/{jurisdiction_ids[0]}/cvrs"
     )
+    print(json.loads(rv.data))
```
jonahkagan, line 368: remove print

server/tests/conftest.py
```diff
@@ -71,30 +71,28 @@ def election_id(client: FlaskClient, org_id: str, request) -> str:
 
 @pytest.fixture
 def jurisdiction_ids(client: FlaskClient, election_id: str) -> List[str]:
-    rv = client.put(
-        f"/api/election/{election_id}/jurisdiction/file",
-        data={
-            "jurisdictions": (
-                # We expect the API to order the jurisdictions by name, so we
-                # upload them out of order.
```
jonahkagan, line 79: Maybe keep this comment

server/tests/helpers.py
```diff
@@ -413,3 +419,175 @@ def string_to_bytes_io(string: str) -> io.BytesIO:
...
+        data={
+            "storagePathKey": f"test_dir/{filename}",
+            "fileName": filename,
+            "fileType": file_type,
+            "cvrFileType": cvr_file_type,
+        },
+    )
+
+
+def setup_ballot_manifest_upload(
```
jonahkagan, line 455: These helpers are great! One nit: they are doing more than just set up the upload, they are doing it, so might be slightly clearer to name them: `upload_ballot_manifest`, etc.

jonahkagan (approved):
Thanks for splitting this up into easy to read commits! It was pretty straightforward to review.

I've got a bunch of small notes but nothing major (hopefully)

server/api/public.py
```diff
@@ -151,3 +152,22 @@ def compute_batch_comparison_sample_size(
         + 2
     )
     return min(sample_size, contest.ballots)
+
+@api.route(
+    "/file-upload",
+    methods=["POST"],
+)
+@allow_public_access
```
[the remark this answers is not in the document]
carolinemodic: Yeah I was starting to worry about this as well and then realized that since this endpoint fails when the config env var is set for s3 we should be ok. Although I'm not sure if there is a similar attack one could do against s3 with the presigned urls if they got them 🤔 

But I will add a any logged in user decorator anyways

server/tests/conftest.py
carolinemodic: good catch

server/api/cvrs.py
carolinemodic: since the wrapper zip is immediately closed I didn't see a obvious way and wanted to err on the side of touching this code as little as possible haha

jonahkagan: Maybe we could just do `wrapper_zip.seek(0)` and move the closing of the file to later?

carolinemodic: filing a followup for this https://github.com/votingworks/arlo/issues/2025

carolinemodic:
was hoping to merge tonight but there is some security issue I'm not understanding in my test deployment so will pick up tomorrow

Issue #1983, opened 2024-09-17
Other miscellaneous dependency updates


arsalansufi:
https://votingworks.slack.com/archives/CKCVA0F9S/p1726541786829819

PR #2013, opened 2024-10-23
Update deps

nbhatia823:
Update deps according to this [issue](https://github.com/votingworks/arlo/issues/1983).

`requests` and `sentry-sdk` are just being bumped to the versions currently used in the `lock` file. `scipy` is actually being updated to a new version, currently causing a slight difference in [calculation for a test](https://app.circleci.com/pipelines/github/votingworks/arlo/7235/workflows/7e235a3c-805b-4810-b7f4-83c31647104e/jobs/21858) that now has the test failing

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -180,17 +180,17 @@ files = [
 
 [[package]]
 name = "boto3"
-version = "1.35.43"
+version = "1.35.47"
 description = "The AWS SDK for Python"
 optional = false
 python-versions = ">=3.8"
 files = [
-    {file = "boto3-1.35.43-py3-none-any.whl", hash = "sha256:e6a50a0599f75b21de0de1a551a0564793d25b304fa623e4052e527b268de734"},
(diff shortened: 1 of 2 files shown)


nbhatia823:
@MarkVVaudits @jonahkagan We are upgrading `scipy` as the version we had been using is a bit too outdated. But, it seems the newer version has some dependency on the hardware it's running on, as this [test](https://github.com/votingworks/arlo/blob/main/server/tests/audit_math/test_minerva.py#L249) is [failing now](https://app.circleci.com/pipelines/github/votingworks/arlo/7235/workflows/7e235a3c-805b-4810-b7f4-83c31647104e/jobs/21858), and gets slightly different values when run on CI (`0.16896200607859224`) vs on my local (`0.1689620060785905`) vs the previous value (`0.16896200607848647`). It looks like they are all the same for the first ~12 significant figures, and then the new CI and local values are the same for all but the last 3 digits.

A) Does this change in calculation/variability seem acceptable?
B) If so, it seems we could add in a tolerance threshold to the tests, where if the values are within say 1^-10 of each other, than we accept. Does that sound acceptable?

jonahkagan (approved):

PR #2014, opened 2024-10-24
Temporarily disable batch audit file prep tool for Ginny

arsalansufi:
So that Ginny can prep some screenshots for orgs that won't be using batch inventory


nbhatia823 (approved):

arsalansufi:
Actually, thinking we're just going to have Ginny create a new sandbox for this purpose! So may just close this out, TBD

PR #2015, opened 2024-10-24
Fix flaky auth code expiration test

jonahkagan:
The current test only allowed one second until the code expired and could be regenerated. It was being flaky and sometimes returning a new code when we attempted to check that you couldn't generate a new code inside that window. We need a little bit longer window to make that check safely.


server/tests/test_auth.py
```diff
@@ -513,7 +513,7 @@ def test_jurisdiction_admin_too_many_attempts(
         ]
     }
 
-    time.sleep(1)
+    time.sleep(3)
```
jonahkagan, line 516: The request right before this one was succeeding when it should fail.

nbhatia823 (approved):
Thanks for looking into this, was gonna raise it as well!

PR #2018, opened 2024-10-28
support MS Excel tabulator status file

kshen0:


server/tests/batch_comparison/test_batch_inventory.py
```diff
@@ -1141,13 +1155,50 @@ def test_batch_inventory_wrong_tabulator_status_file(
...
+        {
+            "file": {"name": "tabulator-status.xml", "uploadedAt": assert_is_date},
+            "processing": {
+                "status": ProcessingStatus.PROCESSED,
+                "startedAt": assert_is_date,
+                "completedAt": assert_is_date,
+                "error": None,
+            },
+        },
+    )
```
jonahkagan, line 1169: Should we test that the data was parsed correctly? I think in the happy path test above we confirm this by downloading the batch inventory worksheet

server/api/batch_inventory.py
```diff
@@ -442,19 +443,81 @@ def process_batch_inventory_tabulator_status_file(
...
+            raise UserError(TABULATOR_STATUS_PARSE_ERROR)
+
+        # Ignore rows containing data we don't need:
+        # 1. "Tabulator Status"
+        # 2. Election name
+        # 3. "Official"/"Unofficial" status
+        # 4. Timestamp
+        # 5. Empty row for spacing
+        row_offset = 5
+        rows = rows[row_offset:]
```
jonahkagan, line 475: I wonder if it would be a bit more reliable to search for the headers row rather than assume its in row 6.

server/api/batch_inventory.py
```diff
@@ -442,19 +443,81 @@ def process_batch_inventory_tabulator_status_file(
...
+        if worksheet is None:
+            raise UserError(TABULATOR_STATUS_PARSE_ERROR)
+
+        tabulator_status_table = worksheet.find("ss:Table", namespaces)
+        if tabulator_status_table is None:
+            raise UserError(TABULATOR_STATUS_PARSE_ERROR)
+
+        columns = tabulator_status_table.findall("Column")
+        if columns is None:
+            raise UserError(TABULATOR_STATUS_PARSE_ERROR)
```
jonahkagan, line 462: Do we use these? If not, why look for them?

server/api/batch_inventory.py
```diff
@@ -442,19 +443,81 @@ def process_batch_inventory_tabulator_status_file(
...
+        namespaces = {"ss": "urn:schemas-microsoft-com:office:spreadsheet"}
+        worksheet = cvr_xml.find(
+            ".//ss:Worksheet[@ss:Name='Tabulator Status']", namespaces
+        )
+        if worksheet is None:
+            raise UserError(TABULATOR_STATUS_PARSE_ERROR)
+
+        tabulator_status_table = worksheet.find("ss:Table", namespaces)
+        if tabulator_status_table is None:
+            raise UserError(TABULATOR_STATUS_PARSE_ERROR)
```
jonahkagan, line 458: Consider collapsing these into one `find` using a nested path

server/api/batch_inventory.py
```diff
@@ -442,19 +443,81 @@ def process_batch_inventory_tabulator_status_file(
...
+        name_key = "Name"
+
+        return {
+            (values := row.findall("./ss:Cell/ss:Data", namespaces))[
+                header_indices[tabulator_id_key]
+            ]
+            .text: values[header_indices[name_key]]
+            .text
+            for row in rows
+        }
```
jonahkagan, line 494: Nit: this is kind of hard to read

jonahkagan (approved):
Nice work!

server/tests/batch_comparison/test_batch_inventory.py
kshen0: Good call: https://github.com/votingworks/arlo/pull/2018/commits/abddab989b1569c29fd6d0404c4905b926c11fea#diff-6c78d492863efe745b0b5db8818551b44a2e7d9367ca6ee77b0b527be3a3e038R1172

server/api/batch_inventory.py
kshen0: https://github.com/votingworks/arlo/pull/2018/commits/abddab989b1569c29fd6d0404c4905b926c11fea#diff-6c612f15bc7de123abfc44b5aead6ce7e0c3a7765f3a0b08543553d068f0e233R464-R473 searches for the headers instead of assuming row 6

server/api/batch_inventory.py
kshen0: I used to validate the number of columns was > 0 but removed that check because reading column header values serves the same purpose. Deleted this leftover in https://github.com/votingworks/arlo/pull/2018/commits/abddab989b1569c29fd6d0404c4905b926c11fea#diff-6c612f15bc7de123abfc44b5aead6ce7e0c3a7765f3a0b08543553d068f0e233L460-L462

server/api/batch_inventory.py
kshen0: https://github.com/votingworks/arlo/pull/2018/commits/abddab989b1569c29fd6d0404c4905b926c11fea#diff-6c612f15bc7de123abfc44b5aead6ce7e0c3a7765f3a0b08543553d068f0e233R458

server/api/batch_inventory.py
kshen0: Ah yeah I was debating between a dict comprehension and a more standard construction like https://github.com/votingworks/arlo/pull/2018/commits/abddab989b1569c29fd6d0404c4905b926c11fea#diff-6c612f15bc7de123abfc44b5aead6ce7e0c3a7765f3a0b08543553d068f0e233R487

Agree that the comprehension is a little hard to comprehend so I'm fine with being more verbose

server/api/batch_inventory.py
```diff
@@ -447,51 +453,46 @@ def get_tabulator_id_to_name_dict_for_excel_file(
         cvr_xml: ElementTree.ElementTree,
     ) -> Dict[Optional[str], Optional[str]]:
         namespaces = {"ss": "urn:schemas-microsoft-com:office:spreadsheet"}
-        worksheet = cvr_xml.find(
-            ".//ss:Worksheet[@ss:Name='Tabulator Status']", namespaces
+        rows = iter(
```
kshen0, line 456: I actually think using indices and a `list[Element]` is easier to read than the current `iterator` implementation, but using an `iterator` is marginally more performant and probably worth it for files with many tabulators.

jonahkagan: I wouldn't worry too much about loading all of the file into memory. There can't be _that_ many tabulators given real world constraints.

You could also do something like this to stick with the iterator approach throughout

```py
row_contents = (row.find("ss:Cell/ss:Data[@ss:Type='String']", namespaces) for row in rows)
column_header_rows = next(
  (
     row_content
     for row_content in row_contents
     if row_content and strip_optional_string(row_content.text) == TABULATOR_ID,
  ),
  None
)
```

server/api/batch_inventory.py
```diff
@@ -442,19 +449,76 @@ def process_batch_inventory_tabulator_status_file(
...
+        # Headers in the order they appear in the file eg. ["Tabulator Id", "Name", "Load Status", "Total Ballots Cast"]
+        headers = [(data.text or "").strip() for data in column_header_values]
+        # The indices of values of interest, according to the headers above
+        tabulator_id_index = headers.index(TABULATOR_ID)
+        tabulator_name_index = headers.index(NAME)
+
+        tabulator_id_to_name = {}
+        tabulator_data_row = next(rows, None)
+        # Iterate over and parse tabulator data rows until exhausted
+        while tabulator_data_row is not None:
```
jonahkagan, line 487: I think you can just do `for tabulator_data_row in rows:`

jonahkagan (approved):

server/api/batch_inventory.py
kshen0: Given expected filesize I've refactored to do all tree/string handling first and business logic second. I think this is easier to follow at the (now negligible) cost of not being exactly o(1n)

[44c4e17](https://github.com/votingworks/arlo/pull/2018/commits/44c4e172336afcd96ee0fc9c31388ad5dc74e868)

PR #2019, opened 2024-10-28
Rollback db transaction when background task is interrupted

jonahkagan:
Bugfix for #1990 

Previously, when a task got interrupted, we let the process die immediately, which would rollback the transaction automatically.

Now, we try to reset the task, which involves clearing the task started_at timestamp and then commiting the current transaction to save that change to the db. So in order to ensure that any in-progress changes from the task itself are not committed alongside those changes, we need to rollback the current transaction and start a new one.

I updated the tests for task resetting to cover this behavior, which they previously didn't cover because they didn't use the db.

### DIFF
--- a/server/tests/test_background_tasks.py
+++ b/server/tests/test_background_tasks.py
@@ -342,25 +342,42 @@ def test_task_multiple_run_in_order(db_session):
 
 
 def test_task_interrupted(caplog, db_session):
-    results = []
+    db_session.execute(
+        """
+        CREATE TABLE IF NOT EXISTS task_to_interrupt_results (
+            num INT,
+            inserted_at TIMESTAMPTZ DEFAULT NOW()
+        )
+        """
+    )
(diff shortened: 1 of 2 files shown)


server/tests/test_background_tasks.py
```diff
@@ -454,7 +474,8 @@ def num_completed_tasks():
         worker.terminate()
 
     expected_sorted_results = list(range(num_tasks))
+    results = [
+        num for num, in db_session.execute("SELECT num FROM count_results").fetchall()
+    ]
     # Each task should have run exactly once
     assert sorted(results) == expected_sorted_results
-    # Tasks should not have run in order, since some got interrupted and reset
-    assert results != expected_sorted_results
```
jonahkagan, line 460: While I like what this assertion is trying to test, it relies on non-deterministic timing for the tasks to get executed out of order. I couldn't think of an easy way to make this work deterministically, so to prevent flaky tests I'm removing it.

arsalansufi (approved):
Good catch!

PR #2020, opened 2024-10-29
Only allow one background task per audit to run at a time

jonahkagan:
Task: #1934 

Many background tasks modify audit data in the database that is relied on by other background tasks. We don't want to allow background tasks that touch the same data to run concurrently, as this may result in inconsistent behavior due to race conditions.

The simplest guard to prevent that is to ensure that only one background task per audit runs at a time. Background tasks don't ever modify data across multiple audits, so this is a safe level of granularity. We'll still get the benefit of having multiple concurrent workers when there are multiple audits running at the same time.

### DIFF
--- a/server/tests/test_background_tasks.py
+++ b/server/tests/test_background_tasks.py
@@ -1,4 +1,6 @@
+from collections import defaultdict
 import logging
+import math
 import multiprocessing
 import random
 import time
@@ -483,22 +485,21 @@ def test_multiple_workers(db_session):
     for worker in workers:
         worker.start()
 
-    def num_completed_tasks():
+    def num_incomplete_tasks():
         return (
             db_session.query(BackgroundTask)
-            .filter_by(task_name="count")
-            .filter(BackgroundTask.completed_at.isnot(None))
+            .filter_by(task_name="count", completed_at=None)
             .count()
         )
 
-    while num_completed_tasks() < num_tasks / 2:
+    while num_incomplete_tasks() > num_tasks / 2:
         time.sleep(0.1)
 
     # Terminate some workers to make sure their tasks are reset and picked up by others
     workers[0].terminate()
     workers[1].terminate()
 
-    while num_completed_tasks() < num_tasks:
+    while num_incomplete_tasks() > 0:
         time.sleep(0.1)
 
     for worker in workers:
@@ -555,14 +556,13 @@ def test_multiple_workers_lock_on_election(db_session):
     num_tasks_per_election = 20
     num_workers = 4
 
-    election_id_1 = "election-1"
-    election_id_2 = "election-2"
+    election_ids = [f"election-{i}" for i in range(3)]
 
     db_session.execute(
         "CREATE TABLE IF NOT EXISTS election_count (election_id TEXT, count INT)"
     )
     db_session.execute("TRUNCATE TABLE election_count")
-    for election_id in [election_id_1, election_id_2]:
+    for election_id in election_ids:
         db_session.execute(
             "INSERT INTO election_count (election_id, count) VALUES (:election_id, 0)",
             dict(election_id=election_id),
@@ -582,10 +582,13 @@ def test_multiple_workers_lock_on_election(db_session):
         )
         db_session.commit()
 
-    # Enqueue tasks
-    for _ in range(num_tasks_per_election):
-        for election_id in [election_id_1, election_id_2]:
+    created_tasks_per_election = defaultdict(int)
+
+    # Enqueue some tasks to start
+    for _ in range(math.floor(num_tasks_per_election / 2)):
+        for election_id in election_ids:
             create_background_task(add1, dict(election_id=election_id), db_session)
+            created_tasks_per_election[election_id] += 1
     db_session.commit()
 
     db_url = db_session.bind.url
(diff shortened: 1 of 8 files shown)


server/tests/test_background_tasks.py
```diff
@@ -488,3 +510,139 @@ def num_completed_tasks():
...
+        )
+    db_session.commit()
+
+    @background_task
+    def add1(election_id, db_session):
+        (current_count,) = db_session.execute(
+            "SELECT count FROM election_count WHERE election_id = :election_id",
+            dict(election_id=election_id),
+        ).fetchone()
+        time.sleep(random.randint(0, 2) / 10)
```
benadida, line 578: love it.

benadida (commented):
I *think* there is still a way you end up not locking out on election ID, as follows:

- process A locks all pending tasks
- process A reads all lock keys, and gets [1,2,3]
- a new pending task is added for election ID 4
- process B locks that new pending task
- process B reads all lock keys, and gets [1,2,3]
- process A selects a task for election ID 4
- process B selects thew one pending task with election ID 4
- process A and B are now working concurrently on election ID 4.

I think the safer way to do this (and probably simpler to read) is to `LOCK TABLE background_tasks` while you're claiming a task, and once you've updated that task and its lock key, then release the lock before you execute the task.

Selecting a task should be relatively fast, so locking the task table for that long should have minimal performance impact and is probably the right thing to do logically.

jonahkagan:
Thanks @benadida. I think you're right. I had originally considered locking the whole table but didn't want to block any in-progress tasks from completing. But I think you're right that it's quick enough that it shouldn't matter.

benadida:
> Thanks @benadida. I think you're right. I had originally considered locking the whole table but didn't want to block any in-progress tasks from completing. But I think you're right that it's quick enough that it shouldn't matter.

you are quite right to call out that it will prevent writes to the task table and thus block task completion, but as you say, that's probably fine.

benadida:
> @benadida I've been trying to create a test to reproduce the case you outlined but haven't been able to do it. I'm wondering if, in your example, the step "process B locks that new pending task" won't happen because process B will be trying to lock _all_ unstarted tasks (most of which are already locked by process A), so will block until they are unlocked.

Oh! Yeah good point! That's likely right. Might be good to verify this by trying this in `psql` just to see if the locking really behaves the way you think it does... but I believe it. In which case, I think we're good.

jonahkagan:
Ran the commands manually and confirmed that process B would block while waiting to lock all queued tasks, even if new ones are added that weren't locked by process A.

jonahkagan:
Updated with improved tests and comments to make the safety of this locking approach clearer.

benadida (approved):

server/worker/tasks.py
```diff
@@ -133,22 +149,60 @@ def run_task(task: BackgroundTask, db_session):
...
+    # them (since it will also be trying to lock _all_ queued tasks, including
+    # the ones we already locked).
+    #
+    # Importantly, this also ensures that we get a safe view of which lock_keys
+    # are currently in use by in-progress tasks. If we're the only worker that
+    # can claim a task right now, then no additional lock_keys can be locked. It
+    # is possible for another worker to finish an in-progress task and release a
+    # lock_key while we're running this query, but that's fine because even if
+    # we don't see that change and wanted to claim that lock_key, we'll just fail
+    # and try again later.
```
arsalansufi, line 169: I'm following the high-level, just struggling a bit to parse this third paragraph, specifically this sentence:

> If we're the only worker that can claim a task right now, then no additional lock_keys can be locked.

server/worker/tasks.py
```diff
@@ -133,22 +149,60 @@ def run_task(task: BackgroundTask, db_session):
...
+            .with_entities(BackgroundTask.id)
+        )
+    )
+    in_progress_task_lock_keys = (
+        db_session.query(BackgroundTask)
+        .filter(BackgroundTask.started_at.isnot(None))
+        .filter_by(completed_at=None)
+        .with_entities(BackgroundTask.lock_key)
+        .subquery()
+    )
```
arsalansufi, line 190: Even if we've reasoned through the fact that this system works and covers all edge cases, I am wondering if it's still preferable to just lock the entire tasks table when we look for a task.

I'm finding myself re-reading the above explanatory comment and code a few times to understand the mechanism.

And as noted elsewhere in this PR, it seems like there isn't really much downside to locking the entire tasks table. Yes, task completion might be blocked for just a moment, but that should be inconsequential relative to overall task processing time. And the mental model really is much simpler.

I'm ultimately okay with the current approach but just wanted to flag!

arsalansufi (approved):
:tada:

server/worker/tasks.py
jonahkagan: I had the same thought yesterday during that discussion. I experimented with locking the entire tasks table but, surprisingly, I couldn't get it working. All workers ended up hanging when trying to lock the table. There's likely some explanation for this, but I'm not sure whether it's worth spending more time debugging to try to figure out why.

Do you think it's worth spending some more time on it? Or just going with what I have in the PR already (which seems to be working)?

server/worker/tasks.py
jonahkagan: Is this any clearer?

> Importantly, this also ensures that we get a safe view of which lock_keys are currently in use by in-progress tasks. If this worker has successfully locked all queued tasks, then it's the only worker that can claim a task right now, so no additional tasks can be started. That means the current set of in-progress lock_keys won't change until this worker releases the lock on the queued tasks.

arsalansufi: Yes that does read clearer, thanks for editing!

server/worker/tasks.py
arsalansufi: Oh interesting 🤔 Yeah in that case, I'm fine with merging this to have something working in place, so long as we're confident in the approach and not concerned about tricky to debug issues during audit season itself (which I believe is the case!)

PR #2021, opened 2024-10-29
validate Batch Inventory Worksheet uploaded to ballot manifest

kshen0:
Returns a more specific error message when a Batch Inventory Worksheet is uploaded instead of ballot manifest.

Note that because CSV validation currently happens in upload before `process()`, it's easiest to trigger a toast notification rather than permanent job error under the upload field.

https://github.com/user-attachments/assets/2af2afea-46f7-43ef-994e-4f4d6eea2d6f

Screenshot 2024-10-29 at 2 51 46 PM

### DIFF
--- a/server/api/ballot_manifest.py
+++ b/server/api/ballot_manifest.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import BinaryIO, Optional
 import uuid
 import logging
 from datetime import datetime
@@ -12,6 +12,7 @@ from ..database import db_session, engine
 from ..models import *  # pylint: disable=wildcard-import
 from ..auth import restrict_access, UserType, get_loggedin_user, get_support_user
 from ..worker.tasks import (
+    UserError,
     background_task,
     create_background_task,
 )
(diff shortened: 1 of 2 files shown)


server/api/ballot_manifest.py
```diff
@@ -228,12 +231,25 @@ def process() -> None:
...
+
 # Raises if invalid
 def validate_ballot_manifest_upload(request: Request):
     if "manifest" not in request.files:
         raise BadRequest("Missing required file parameter 'manifest'")
 
-    validate_csv_mimetype(request.files["manifest"])
+    file = request.files["manifest"]
+    validate_is_not_batch_inventory_worksheet(file)
+    validate_csv_mimetype(file)
```
jonahkagan, line 252: > Note that because CSV validation currently happens in upload before process(), it's easiest to trigger a toast notification rather than permanent job error under the upload field.

This doesn't match my reading. I see CSV mimetype validation happening during the initial file upload, but the actual parsing of the CSV file seems to happen within the background task.

If possible, making this a file processing error would be best since it will be persistent in the UI.

kshen0: You are absolutely right, sorry! Updated

jonahkagan (approved):

PR #2024, opened 2024-10-29
Ensure that completed tasks cannot be reset

jonahkagan:
There was a race condition that allowed for tasks to be reset if the worker was interrupted immediately after the task completed (but before the worker set `task = None`). Instead of clearing the `task` variable, we protect against this case by reloading the task from the db before resetting it and ensuring that it is not completed. We also check that it has in fact started, for good measure, though that's not strictly necessary since resetting an unstarted task wouldn't do anything.

To test this, I ran the tests repeatedly. Before the change, this bug would cause a failure with 10 test runs. Afterwards, it didn't fail at all in 20 test runs.

I did catch one other small bug, which is that sometimes when exiting the worker after resetting a task, its possible for the work loop to try to commit a transaction even though the db connection has already been closed. I'm not totally sure why this happens, and it's pretty inconsequential since the worker is already exiting, but I added a guard for it just to keep things tidy.

### DIFF
--- a/server/worker/worker.py
+++ b/server/worker/worker.py
@@ -28,20 +28,17 @@ def run_worker(worker_id: str, db_session, pause_between_tasks_seconds):
         nonlocal task
         if task:
             reset_task(task, db_session)
-            task = None  # Guard against multiple interrupts
         sys.exit(1)
 
     signal.signal(signal.SIGTERM, interrupt_handler)
     # Also handle SIGINT for local development
     signal.signal(signal.SIGINT, interrupt_handler)
 
-    while True:
+    # If the database connection is closed due to sys.exit above, stop working.
+    while db_session.is_active:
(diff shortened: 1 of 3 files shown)


server/worker/worker.py
```diff
@@ -28,20 +28,17 @@ def interrupt_handler(*_args):
...
 
-    while True:
+    # If the database connection is closed due to sys.exit above, stop working.
+    while db_session.is_active:
         task = claim_next_task(worker_id, db_session)
         if task:
             run_task(task, db_session)
-            # Ensure we don't reset the task on interrupt once it completes
-            # successfully
-            task = None
```
jonahkagan, line 44: The existing guards were shoddy, so removing to avoid confusion

arsalansufi:
Thanks for the thorough testing/validation!

PR #2026, opened 2024-10-30
Server endpoint for surfacing discrepancies

nbhatia823:
This PR adds a server endpoint for getting discrepancies, and for now only adds the implementation for batch comparison audits.

Next, I plan to wire together the rest of the flow to the client, and then add implementation for ballot comparison audits. My plan is to build a bit more incrementally, so the PR doesn't become too much for me to reason about as I go.

For now, thought it might be helpful to agree upon the API and models I set up, of course subject to some minor changes if needed when the rest of the feature is built out.

Sample response received by the Client from this endpoint -
Screenshot 2024-10-31 at 9 43 38 AM

### DIFF
--- a/server/api/jurisdictions.py
+++ b/server/api/jurisdictions.py
@@ -1,5 +1,4 @@
 from collections import Counter, defaultdict
-from dataclasses import dataclass
 import logging
 from typing import Dict, List, Optional, Mapping, cast as typing_cast
 import enum
@@ -730,19 +729,22 @@ def get_discrepancy_counts_by_jurisdiction(election: Election):
     )
 
 
-@dataclass
-class Discrepancy:
-    audit_count: int
-    reported_count: int
-    candidate_name: str
+DiscrepanciesByJurisdiction = Dict[str, Dict[str, Dict[str, Dict[str, int]]]]
+# DiscrepanciesByJurisdiction = {
+#     jurisdictionID: {
+#         batchName: {
+#             contestID: {
+#                 reportedVotes:  {choiceID: int},
+#                 auditedVotes:   {choiceID: int},
+#                 discrepancies:  {choiceID: int},
+#     }
+# }
 
 
-@dataclass
-class ContestDiscrepancies:
-    contest_id: str
-    jurisdiction_id: str
-    batch_name: str
-    discrepancies: List[Discrepancy]
+def create_nested_discrepancies_dict():
+    return defaultdict(
+        lambda: defaultdict(lambda: defaultdict(lambda: defaultdict(int)))
+    )
 
 
 @api.route("/election/<election_id>/discrepancy", methods=["GET"])
@@ -752,7 +754,7 @@ def get_discrepancies_by_jurisdiction(election: Election):
     if not current_round:
         raise Conflict("Audit not started")
 
-    discrepancies_by_jurisdiction: Dict[str, List[ContestDiscrepancies]] = defaultdict()
+    discrepancies_by_jurisdiction: DiscrepanciesByJurisdiction = {}
 
     if election.audit_type == AuditType.BATCH_COMPARISON:
         discrepancies_by_jurisdiction = (
@@ -769,8 +771,9 @@ def get_discrepancies_by_jurisdiction(election: Election):
 def get_batch_comparison_audit_discrepancies_by_jurisdiction(
     election: Election, round_id: str
 ):
-    discrepancies_by_jurisdiction: Dict[str, List[ContestDiscrepancies]] = defaultdict()
+    discrepancies_by_jurisdiction = create_nested_discrepancies_dict()
 
+    # TODO: Add support for combined batches
     batch_keys_in_round = set(
         SampledBatchDraw.query.filter_by(round_id=round_id)
         .join(Batch)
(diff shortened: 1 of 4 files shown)


nbhatia823:
Note - Tests will be added soon, but a review of the endpoint would still be appreciated

server/api/jurisdictions.py
```diff
@@ -719,3 +720,97 @@ def get_discrepancy_counts_by_jurisdiction(election: Election):
...
+    if election.audit_type == AuditType.BATCH_COMPARISON:
+        discrepancies_by_jurisdiction = (
+            get_batch_comparison_audit_discrepancies_by_jurisdiction(
+                election, current_round.id
+            )
+        )
+    else:
+        raise Conflict("Discrepancies are only implemented for batch comparison audits")
+
+    return jsonify(discrepancies_by_jurisdiction)
```
jonahkagan, line 758: A few thoughts on the data format here:
- I like that you use a dictionary with jurisdiction keys - this makes sense since they are unique
- I'd suggest using dictionaries more liberally to help indicate where there are no duplicates expected

So that might lead me to some of these ideas:

```py
# Nesting by contest
{
    jurisdiction_id: {
        contest_id: {
            batch_name: {
                choice_id: {reported_votes: int, audited_votes: int, discrepancy: int}
            }
        }
    }
}
# Nesting by batch (maybe slightly more applicable to the UI you mocked up)
{
    jurisdiction_id: {
        batch_name: {
            contest_id: {
                choice_id: {reported_votes: int, audited_votes: int, discrepancy: int}
            }
        }
    }
}
# Could also consider flipping the final level of the hierarchy, since we already have the reported_votes and audited_votes constructed
{
    jurisdiction_id: {
        batch_name: {
            contest_id: {
                reported_votes: {choice_id: int},
                audited_votes: {choice_id: int},
                discrepancies: {choice_id: int},
            }
        }
    }
}
```

server/api/jurisdictions.py
```diff
@@ -719,3 +720,97 @@ def get_discrepancy_counts_by_jurisdiction(election: Election):
...
+                if audit_count == reported_count:
+                    continue
+                discrepancy = Discrepancy(audit_count, reported_count, choice.name)
+                discrepancies.append(discrepancy)
+
+            contest_discrepancies = ContestDiscrepancies(
+                contest.id, jurisdiction_id, batch_name, discrepancies
+            )
+            discrepancies_by_jurisdiction[jurisdiction_id].append(contest_discrepancies)
+    return discrepancies_by_jurisdiction
```
jonahkagan, line 816: Unfortunately, this logic will need to take into account combined batches. See the discrepancy counting endpoint above and/or `sampled_batch_rows` function in reports.py (which computes the discrepancy report). May be best to add this as a follow up once we get the rest of the feature working.

server/api/jurisdictions.py
nbhatia823: Chatted on Slack, key points: 
```
Nikhil - [This comment ](https://github.com/votingworks/arlo/pull/2026#discussion_r1823147709)in particular helped point out the question I was unsure about - I shied away from more nested dictionaries, because I felt the types might get a bit harder to follow with more nesting. But, I notice that in most places we don't define explicit return types like what I have for ContestDiscrepancies and Discrepancies, and we just notate the shape of the object that we send to the frontend via something like a comment. With the approach of a more nested dictionary, would you recommend I drop the classes I defined?

Jonah - I don’t have a strong opinion. In Arlo, since there’s a Python backend and TS frontend, there’s not as much value in defining types for everything, since we can’t share them. I usually make typed data structures on the server when I’m reusing them across functions within the backend. If it’s just to send data to the frontend, I just write a comment.
But that’s not set in stone, it’s just what I’ve been doing
```

Updating types to follow third option above, removing dataclass models

server/api/jurisdictions.py
```diff
@@ -727,3 +727,90 @@ def get_discrepancy_counts_by_jurisdiction(election: Election):
...
+#                 auditedVotes:   {choiceID: int},
+#                 discrepancies:  {choiceID: int},
+#     }
+# }
+
+
+def create_nested_discrepancies_dict():
+    return defaultdict(
+        lambda: defaultdict(lambda: defaultdict(lambda: defaultdict(int)))
+    )
```
jonahkagan, line 747: Based on the way you're using this, I don't think you need the last two levels of `defaultdict`. I think this will suffice:

```py
defaultdict(lambda: defaultdict(dict))
```

Also maybe just inline it since it's only called in one place?

server/api/jurisdictions.py
```diff
@@ -727,3 +727,90 @@ def get_discrepancy_counts_by_jurisdiction(election: Election):
...
+        )
+    else:
+        raise Conflict("Discrepancies are only implemented for batch comparison audits")
+
+    return jsonify(discrepancies_by_jurisdiction)
+
+
+def get_batch_comparison_audit_discrepancies_by_jurisdiction(
+    election: Election, round_id: str
+):
```
jonahkagan, line 773: Maybe annotate the return type of this function

jonahkagan (approved):
Looks great, thanks for the thorough testing!

server/api/jurisdictions.py
nbhatia823: Ah yeah, thanks. I initially separated it out anticipating I would also need to use it for the future ballot comparison discrepancy function, but with the shorter init then inline seems fine.

PR #2027, opened 2024-10-30
Make extra batch sampling deterministic by using sortable batch keys

jonahkagan:
Previously, we were choosing extra batches from a nondeterministically ordered set of batch ids. Instead, we use batch keys (instead of db ids, which are random) and sort them before choosing batches.

### DIFF
--- a/server/api/shared.py
+++ b/server/api/shared.py
@@ -665,24 +665,16 @@ def compute_sample_batches_for_contest(
     if is_enabled_sample_extra_batches_by_counting_group(election) and round_num == 1:
         rand = random.Random(str(election.random_seed))
         for jurisdiction in contest.jurisdictions:
-            batch_ids_with_container_and_num_ballots = (
-                Batch.query.filter_by(jurisdiction_id=jurisdiction.id)
-                .with_entities(Batch.id, Batch.container, Batch.num_ballots)
-                .all()
-            )
-            batch_ids = [
-                batch_id for batch_id, _, _ in batch_ids_with_container_and_num_ballots
-            ]
-            batch_id_to_num_ballots = {
-                batch_id: num_ballots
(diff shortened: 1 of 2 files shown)


server/api/shared.py
```diff
@@ -665,61 +665,60 @@ def compute_sample_batches_for_contest(
...
+                for batch in jurisdiction.batches
             }
             # To simplify this experiment, we specify the counting group in the
             # container column of the ballot manifest
-            bmd_batch_ids = {
-                batch_id
-                for batch_id, container, _ in batch_ids_with_container_and_num_ballots
-                if container
+            bmd_batch_keys = {
+                (jurisdiction.name, batch.name)
```
kshen0, line 675: OOC are batch keys a construct outside of this?

kshen0 (commented):
Is there a business-logic reason we would want nondeterministic behavior in this function?

server/api/shared.py
jonahkagan: Yeah, we use them in a bunch of places (including earlier in this function). It's a unique identifier for the batch that doesn't depend on the database. This ensures that if you upload the same batch data to Arlo and enter the same random seed, you'll always get the same behavior, which is important both for testing and for audit integrity.

To answer:
> Is there a business-logic reason we would want nondeterministic behavior in this function?

We only want non-determinism modulo the random seed.

kshen0 (approved):

server/api/shared.py
kshen0: In that case I wonder if it might be useful to codify the batch key construct as a type or in a helper function to generate batch key from inputs rather than use a regular tuple + convention. But that would be outside the scope of this PR.

jonahkagan: Yeah, I've been thinking that too as we start using it in more places

Issue #1914, opened 2024-03-22
Order audits on JM screen by audit date (most recent first) instead of alphabetically


arsalansufi:
Current election needs to be first button for jurisdiction managers with multiple audits. -Tabitha 

Issue automatically created from Slack feedback reported by ginny: 
https://votingworks.slack.com/archives/C01F2CH62TS/p1702310679633499

jonahkagan:
Consider visually separating active vs inactive audits

PR #2029, opened 2024-10-30
Prioritize active audits on home screen

jonahkagan:
Closes https://github.com/votingworks/arlo/issues/1914

For both jurisdiction managers and audit admins, groups the audits on their home screen into two groups: active and completed. Within each group, orders the audits by creation date, putting the most recent at the top.

Audit admin
(Note that most audit admins are part of one org, but some internal users are part of multiple, so we still support that case)
Screenshot 2024-10-30 at 3 18 00 PM

Jurisdiction manager
Screenshot 2024-10-30 at 3 20 51 PM


server/api/rounds.py
```diff
@@ -175,8 +176,8 @@ def is_round_ready_to_finish(election: Election, round: Round) -> bool:
         return num_jurisdictions_without_results == 0
 
 
-def is_audit_complete(round: Round):
-    if not round.ended_at:
+def is_audit_complete(round: Optional[Round]):
+    if not (round and round.ended_at):
```
kshen0, line 180: In what cases will a `round` not exist? If an audit has no rounds, does that mean it hasn't started yet and should be considered incomplete?

jonahkagan: No rounds means the audit hasn't started yet (is still in the setup phase). So yes, I'm considering that incomplete.

kshen0 (approved):

Issue #1865, opened 2023-12-05
Parse CVRs for all systems to create contest totals by batch files - Hart next


carolinemodic:
## Context

The batch audit file preparation tool (referred to as the batch inventory tool in the code) is a tool that helps jurisdictions prepare the files necessary for a batch audit (a ballot manifest and a candidate totals by batch file) from their CVR files.

| 1 | 2 |
| -- | -- |
| ![Image](https://github.com/user-attachments/assets/23bfbbe6-1484-43d1-8dfa-71f23456caec) | ![Image](https://github.com/user-attachments/assets/f40ff075-c0e6-4004-88e5-bbafc5bd3a99) |

The tool is controlled by a frontend feature flag. It's turned off by default, turned on completely for GA and some sandboxes, and partially turned on for others - partially meaning that we display an option to download a candidate-totals-by-batch file but not a ballot manifest. For the partial case, we can still internally access the generated ballot manifest by navigating to the correct URL. We want to encourage users in this bucket to prepare their ballot manifest from scratch, but it's nice to have the ability to cross-check against the likely correct end result internally.

https://github.com/votingworks/arlo/blob/f48b67822232d40e266a3d91909db1b26be42ec9/client/src/components/useFeatureFlag.ts#L8-L44

Georgia is the main user of this tool, but we've considered using the tool to help others prep their files, in various capacities. In some cases, we've even considered leaving the tool completely turned off so as not to confuse end users but using one of our sandboxes with the tool enabled and CVR files from a jurisdiction to generate batch audit files.

## Supporting Hart

We currently support generating batch audit files from Dominion and ES&S CVRs. And we'd next like to support Hart CVRs. The most likely user of this tool with Hart files is Lancaster County in PA, sample files available here: https://votingworks.slack.com/archives/C013V6CT9H9/p1724676881810049.

### Ballot comparison audit CVR parsing vs. batch audit file preparation tool CVR parsing

A reminder that CVR parsing for ballot comparison audits is more complete and rigorous than the CVR parsing we need here. The batch audit file preparation tool uses fewer fields than ballot comparison audit CVR parsing does. For example, I don't think that we need to worry about the multiple ZIPs, one per tabulator case, that we use for Orange County right off the bat. (Orange County will not be using this tool.)

[...]

PR #2030, opened 2024-10-31
Support hart cvr files in batch inventory tool

carolinemodic:
https://github.com/votingworks/arlo/issues/1865
Adds basic support for hart cvr uploads in the batch inventory tool. Extracts BatchNumber, when provided, from Hart CVRs to use to group batches, and falls back to the precinct (split) name. The hart documentation suggests that BatchNumber is only provided on central system drives, and in sample hart cvr files we have there are some without BatchNumbers. If the CVRs with a missing BatchNumber are precinct cvrs then grouping by precinct is a logical fallback. We may want to tweak this implementation as we have real use cases for it and understand the data better.

### DIFF
--- a/server/api/batch_inventory.py
+++ b/server/api/batch_inventory.py
@@ -46,8 +46,10 @@ from ..util.file import (
     timestamp_filename,
     unzip_files,
     validate_csv_or_zip_mimetype,
+    validate_zip_mimetype,
     validate_xml_mimetype,
 )
+from ..util.hart_parse import find_xml, parse_contest_results, find_text_xml
 from ..worker.tasks import UserError, background_task, create_background_task
 from ..util.csv_download import csv_response, jurisdiction_timestamp_name
 from ..util.isoformat import isoformat
@@ -76,6 +78,26 @@ def batch_key_to_name(
         if tabulator_id_to_name
         else f"{tabulator_id} - {batch_id}"
     )
+
+
+def validate_choice_name_and_get_choice_id(
+    contest: Contest, choice_name: str
+) -> Optional[str]:
+    choice_id = None
+    for choice in contest.choices:
+        if choice.name == choice_name:
+            choice_id = choice.id
+            break
+
+    if (
+        not choice_id
+        and choice_name
+        and choice_name.lower() != "overvote"
+        and choice_name.lower() != "undervote"
+        and choice_name.lower() != "write-in"
+    ):
+        raise UserError(f"Unrecognized choice in CVR file: {choice_name}")
+    return choice_id
 
 
 class ElectionResults(TypedDict):
@@ -261,24 +283,6 @@ def process_batch_inventory_cvr_file(
             lambda: defaultdict(int)
         )
 
-        def validate_choice_name_and_get_choice_id(choice_name: str) -> Optional[str]:
-            choice_id = None
-            for choice in contest.choices:
-                if choice.name == choice_name:
-                    choice_id = choice.id
-                    break
-
-            if (
-                not choice_id
-                and choice_name
-                and choice_name != "overvote"
-                and choice_name != "undervote"
-                and choice_name != "Write-in"
-            ):
-                raise UserError(f"Unrecognized choice in CVR file: {choice_name}")
-
-            return choice_id
-
         # ZIP file with multiple CSVs
         if batch_inventory_data.cvr_file.storage_path.endswith(".zip"):
             entry_names = unzip_files(cvr_file, working_directory)
@@ -351,7 +355,9 @@ def process_batch_inventory_cvr_file(
                         )
                     batch = cvr_number_to_batch[cvr_number]
                     batch_key: BatchKey = ("", batch)
-                    choice_id = validate_choice_name_and_get_choice_id(choice_name)
+                    choice_id = validate_choice_name_and_get_choice_id(
+                        contest, choice_name
+                    )
 
                     ballot_count_by_batch[batch_key] += 1
                     if choice_id:
@@ -380,7 +386,9 @@ def process_batch_inventory_cvr_file(
                     )
 
                     batch_key: BatchKey = ("", batch)
-                    choice_id = validate_choice_name_and_get_choice_id(choice_name)
+                    choice_id = validate_choice_name_and_get_choice_id(
+                        contest, choice_name
+                    )
 
                     ballot_count_by_batch[batch_key] += 1
                     if choice_id:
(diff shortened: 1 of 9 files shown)


server/api/batch_inventory.py
```diff
@@ -378,14 +384,82 @@ def validate_choice_name_and_get_choice_id(choice_name: str) -> Optional[str]:
...
+        )
+
+        # cvr_file is a ZIP file with multiple XMLs
+        zip_entry_names = unzip_files(cvr_file, working_directory)
+        cvr_file_names = [
+            entry_name
+            for entry_name in zip_entry_names
+            if entry_name.lower().endswith(".xml") and not entry_name.startswith(".")
+            # ZIP files created on Macs include a hidden __MACOSX folder
+            and not entry_name.startswith("__")
```
jonahkagan, line 427: I don't see this check in our regular Hart CVR parsing. I wonder if it's unnecessary or if we should add it there as well

server/util/hart_parse.py
```diff
@@ -0,0 +1,40 @@
...
+from xml.etree import ElementTree as ET
+
+NAMESPACE = "http://tempuri.org/CVRDesign.xsd"
+
+
+def find_text_xml(xml: Optional[Union[ET.ElementTree, ET.Element]], tag: str):
+    if xml is None:
+        return None
+    result = find_xml(xml, tag)
+    return None if result is None else result.text
```
jonahkagan, line 12: Is the text ever an empty string? If so, should we convert to `None`?

server/api/batch_inventory.py
```diff
@@ -378,14 +384,82 @@ def validate_choice_name_and_get_choice_id(choice_name: str) -> Optional[str]:
...
+                )
+            batch_key: BatchKey = (
+                "",
+                batch_key_value,
+            )  # Tabulator ID is not present in Hart CVRs
+
+            ballot_count_by_batch[batch_key] += 1
+            contest_results = parse_contest_results(cvr_xml)
+            for contest in contests:
+                choices = contest_results.get(contest.name) or set()
```
jonahkagan, line 450: Nit: `get` takes a default as a second arg: `contest_results.get(contest.name, set())`

server/api/batch_inventory.py
```diff
@@ -78,6 +80,26 @@ def batch_key_to_name(
...
+        if choice.name == choice_name:
+            choice_id = choice.id
+            break
+
+    if (
+        not choice_id
+        and choice_name
+        and choice_name.lower() != "overvote"
+        and choice_name.lower() != "undervote"
+        and choice_name.lower() != "write-in"
```
jonahkagan, line 97: Are these special values the same in Hart CVRs as they are in ES&S CVRs?

server/api/batch_inventory.py
```diff
@@ -378,14 +384,82 @@ def validate_choice_name_and_get_choice_id(choice_name: str) -> Optional[str]:
                     )
 
                     batch_key: BatchKey = ("", batch)
-                    choice_id = validate_choice_name_and_get_choice_id(choice_name)
+                    choice_id = validate_choice_name_and_get_choice_id(
+                        contest, choice_name
+                    )
 
                     ballot_count_by_batch[batch_key] += 1
```
jonahkagan, line 391: I know this is existing code, but is it supposed to count a ballot once for each contest (seems like double counting potentially)? I noticed the Hart code you wrote doesn't do this - it counts before looping over contests.

jonahkagan: Maybe we need some tests for the ES&S code. I know it's not really in scope but I'm worried about it now.

server/tests/batch_comparison/snapshots/snap_test_batch_inventory.py
```diff
@@ -156,3 +156,21 @@
...
+BATCH4,1\r
+"""
+
+snapshots[
+    "test_batch_inventory_hart_cvr_upload 2"
+] = """Batch Name,Choice 1-1,Choice 1-2\r
+BATCH1,1,2\r
+BATCH2,2,1\r
+BATCH3,0,0\r
+BATCH4,0,0\r
```
jonahkagan, line 175: It seems like BATCH4 has votes for both Choice 1-1 and Choice 1-2 based on this line from the tests:

```py
build_hart_cvr("BATCH4", "1", "1-4-1", "1,1,0,0,1", add_write_in=True)
```

Do you know why they aren't showing up here? If it's because it's an overvote, I can't figure out where that's getting filtered out.

Also, how are write-ins handled?

server/api/batch_inventory.py
carolinemodic: overvotes and undervotes are not returned as choices in the parse_contest_xml helper I'm using and Write-In was capitlized differently which is why I changed these to check the lower() version

server/api/batch_inventory.py
carolinemodic: yeah the es&s code seemed pretty questionable to me lol, don't want to put adding a test for that in scope here but probably a good followup

jonahkagan: Sounds good, added a high pri task: https://github.com/votingworks/arlo/issues/2037

server/api/batch_inventory.py
jonahkagan: Ah, so is this filtering out write-ins? Should we be including them?

server/api/batch_inventory.py
carolinemodic: will add it there

server/util/hart_parse.py
carolinemodic: Not in my testing but that is probably a good idea

server/tests/batch_comparison/snapshots/snap_test_batch_inventory.py
carolinemodic: Its getting filtered because of both being write-ins, going to update the test a bit to make things clearer (and ask about what to do with write-ins in slack)

carolinemodic:
back to you @jonahkagan, added support for write-ins matching dominions logic.

jonahkagan (approved):

server/api/batch_inventory.py
arsalansufi: I claim full responsibility for the questionable code written in a rush 🙈 I should've surfaced earlier that this was worth revisiting and long had on my to-do to write tests for this but ultimately never found the time.

I think you're totally right that this is a bug, one that was masked by the fact that we only expected it to be used for a single-contest audit and manually tested as such. One more relevant piece of context is that it the batch audit file prep tool ultimately wasn't used in the audit that we built ES&S support for. So this code hasn't actually been used.

PR #2032, opened 2024-10-31
Presentation mode for random seed input

### LINKED ISSUE #1777: Full screen modal for seed input
After visually sharing the data field for the random seed and then the review screen.... these both need to look better on a monitor.  And the random seed needs it's own "view" so people aren't asking questions about the other selections.

Issue automatically created from slack feedback reported by ginny 
https://votingworks.slack.com/archives/C01F2CH62TS/p1670983234762699

jonahkagan:
Closes #1777 

Adds a presentation mode for inputting the random seed. To begin an audit, state officials often have a ceremony where they publicly roll dice to generate the random seed and then type it into Arlo. They like to project Arlo onto a screen during this ceremony. This PR adds a presentation mode for the random seed input to be used during this ceremony.

https://github.com/user-attachments/assets/635e0c5a-0384-4737-b7a3-89cdf3cc50e4

### DIFF
--- a/client/src/components/AuditAdmin/Setup/Settings/Settings.tsx
+++ b/client/src/components/AuditAdmin/Setup/Settings/Settings.tsx
@@ -1,6 +1,14 @@
 import React from 'react'
 import { Formik, FormikProps, Field, ErrorMessage } from 'formik'
-import { RadioGroup, Radio, HTMLSelect, Button } from '@blueprintjs/core'
+import {
+  RadioGroup,
+  Radio,
+  HTMLSelect,
+  Button,
+  Dialog,
+  InputGroup,
+  Classes,
+} from '@blueprintjs/core'
 import FormButtonBar from '../../../Atoms/Form/FormButtonBar'
 import FormWrapper from '../../../Atoms/Form/FormWrapper'
 import FormSection from '../../../Atoms/Form/FormSection'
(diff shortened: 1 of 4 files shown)


client/src/components/AuditAdmin/Setup/Settings/Settings.tsx
```diff
@@ -115,7 +190,7 @@ const Settings: React.FC<IProps> = ({
             </FormSection>
             {auditType === 'BALLOT_POLLING' && (
               <FormSection label="Audit Board Data Entry">
-                <label htmlFor="online">
+                <label>
```
jonahkagan, line 193: Resolving a browser warning since this `for` attribute didn't have a corresponding input

client/src/components/AuditAdmin/Setup/Settings/Settings.tsx
```diff
@@ -15,6 +23,70 @@ import {
...
+      </div>
+      <div className={Classes.DIALOG_FOOTER}>
+        <div className={Classes.DIALOG_FOOTER_ACTIONS}>
+          <Button onClick={onClose}>Cancel</Button>
+          <AsyncButton
+            intent="success"
+            icon="tick"
+            disabled={!value}
+            onClick={async () => {
+              await onSave(value)
```
nbhatia823, line 79: Intentional to not have any input validation here? Wondering what would happen if someone input "   "

nbhatia823 (approved):
One question but lgtm! And interesting that this is part of the ceremonial process :D

client/src/components/AuditAdmin/Setup/Settings/Settings.tsx
jonahkagan: I don't think there was any validation in place here. This form is quite old (before my time) and could probably use some attention in that regard. There is some server-side validation to ensure it's a string, but no length requirements. Technically, Arlo can use any string as a random seed, so there's not a high risk, but it might be worth thinking through if there's any validation that would help users. Due to time constraints I don't want to block this PR on this, but happy to discuss more and see if we think there's a follow up that would make sense.

Issue #1912, opened 2024-03-22
Discrepancy report improvement - allow easier view of discrepancy


arsalansufi:
Discrepancy report - downloading the report after each new one pops up is a few steps (download report, expand coloumns, scroll down to correct county, etc.) , being able to click on the discrepancy and see just that counties difference would be much nicer.

Issue automatically created from Slack feedback reported by ginny: 
https://votingworks.slack.com/archives/C01F2CH62TS/p1710943103900869

arsalansufi:
Simple idea: Make it possible to click to the flag and view the discrepancy for that jurisdiction in product

PR #2033, opened 2024-10-31
Batch comparison discrepancies in UI batch comparison

nbhatia823:
Commit by commit review will probably be easiest

This PR adds the ability to see batch comparison audit discrepancies in the UI from [this issue](https://github.com/votingworks/arlo/issues/1912).

Screenshot 2024-10-31 at 4 08 19 PM

https://github.com/user-attachments/assets/faf2b240-0a5f-41fe-b0b0-946d4c3fa58e

Follow ups:
- Add support for ballot comparison
- Add support for combined batches

### DIFF
--- a/client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
+++ b/client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
@@ -1,0 +1,126 @@
+import React from 'react'
+import { Classes, Colors, Dialog, H6, HTMLTable } from '@blueprintjs/core'
+import styled from 'styled-components'
+import {
+  IJurisdiction,
+  useDiscrepanciesByJurisdiction,
+} from '../../useJurisdictions'
+import useContestsJurisdictionAdmin from '../../JurisdictionAdmin/useContestsJurisdictionAdmin'
+import { IContest } from '../../../types'
+
+const ContestDiscrepanciesTable = styled(HTMLTable).attrs({
+  bordered: true,
+  striped: true,
+})`
+  &.${Classes.HTML_TABLE} {
+    background: #ffffff;
+    border: 1px solid ${Colors.LIGHT_GRAY1};
+    margin-bottom: 32px;
+    table-layout: fixed;
+    width: 100%;
+  }
+
+  &.${Classes.HTML_TABLE} td {
+    vertical-align: middle;
+  }
+`
+
+/* istanbul ignore next */
+function getContestName(contests: IContest[], contestId: string) {
+  const contest = contests.find(c => c.id === contestId)
+  return contest ? contest.name : 'Contest not found'
+}
+
+/* istanbul ignore next */
+function getChoiceName(
+  contests: IContest[],
+  contestId: string,
+  choiceId: string
+) {
+  const contest = contests.find(c => c.id === contestId)
+  if (!contest) return 'Choice not found'
+  const choice = contest.choices.find(c => c.id === choiceId)
+  return choice ? choice.name : 'Choice not found'
+}
+
+export interface IJurisdictionDiscrepanciesProps {
+  electionId: string
+  handleClose: () => void
+  jurisdiction: IJurisdiction
+}
+
+const JurisdictionDiscrepancies: React.FC<IJurisdictionDiscrepanciesProps> = ({
+  handleClose,
+  jurisdiction,
+  electionId,
+}) => {
+  const discrepancyQuery = useDiscrepanciesByJurisdiction(electionId, {})
+  const contestsQuery = useContestsJurisdictionAdmin(
+    electionId,
+    jurisdiction.id
+  )
+
+  if (!discrepancyQuery.isSuccess || !contestsQuery.isSuccess) {
+    return null
+  }
+
+  const discrepanciesByBatch = discrepancyQuery.data[jurisdiction.id]
+  const contests = contestsQuery.data
+
+  return (
+    <Dialog
+      isOpen
+      onClose={handleClose}
+      style={{ width: '600px' }}
+      title={`${jurisdiction.name} Discrepancies`}
+    >
+      <div className={Classes.DIALOG_BODY} style={{ marginBottom: 0 }}>
+        {Object.entries(discrepanciesByBatch).map(
+          ([batchName, discrepanciesByContest]) => {
+            return Object.entries(discrepanciesByContest).map(
+              ([contestId, contestDiscrepancies], idx) => (
+                <div key={contestId}>
+                  <H6 style={idx === 0 ? { marginTop: '8px' } : {}}>
+                    {batchName} - {getContestName(contests, contestId)}
+                  </H6>
+                  <ContestDiscrepanciesTable>
+                    <thead>
+                      <tr>
+                        <th>Choice</th>
+                        <th>Reported Votes</th>
+                        <th>Audited Votes</th>
+                        <th>Discrepancy</th>
+                      </tr>
+                    </thead>
+                    <tbody>
+                      {Object.keys(contestDiscrepancies.discrepancies).map(
+                        choiceId => (
+                          <tr key={choiceId}>
+                            <td>
+                              {getChoiceName(contests, contestId, choiceId)}
+                            </td>
+                            <td>
+                              {contestDiscrepancies.reportedVotes[choiceId]}
+                            </td>
+                            <td>
+                              {contestDiscrepancies.auditedVotes[choiceId]}
+                            </td>
+                            <td>
+                              {contestDiscrepancies.discrepancies[choiceId]}
+                            </td>
+                          </tr>
+                        )
+                      )}
+                    </tbody>
+                  </ContestDiscrepanciesTable>
+                </div>
+              )
+            )
+          }
+        )}
+      </div>
+    </Dialog>
+  )
+}
+
+export { JurisdictionDiscrepancies }
--- a/client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
+++ b/client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
@@ -7,20 +7,19 @@ import {
 } from '../../useJurisdictions'
 import useContestsJurisdictionAdmin from '../../JurisdictionAdmin/useContestsJurisdictionAdmin'
 import { IContest } from '../../../types'
+import { assert } from '../../utilities'
 
 const ContestDiscrepanciesTable = styled(HTMLTable).attrs({
   bordered: true,
   striped: true,
 })`
-  &.${Classes.HTML_TABLE} {
-    background: #ffffff;
-    border: 1px solid ${Colors.LIGHT_GRAY1};
-    margin-bottom: 32px;
-    table-layout: fixed;
-    width: 100%;
-  }
+  background: #ffffff;
+  border: 1px solid ${Colors.LIGHT_GRAY1};
+  margin-bottom: 32px;
+  table-layout: fixed;
+  width: 100%;
 
-  &.${Classes.HTML_TABLE} td {
+  td {
     vertical-align: middle;
   }
 `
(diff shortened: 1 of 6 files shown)


client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+function getChoiceName(
+  contests: IContest[],
+  contestId: string,
+  choiceId: string
+) {
+  const contest = contests.find(c => c.id === contestId)
+  if (!contest) return 'Choice not found'
+  const choice = contest.choices.find(c => c.id === choiceId)
+  return choice ? choice.name : 'Choice not found'
```
nbhatia823, line 47: Any suggestion on how I should handle these cases **that we should not see** where the contest choice isn't found?

client/src/components/AuditAdmin/Progress/Progress.tsx
```diff
@@ -90,12 +91,16 @@ const Progress: React.FC<IProgressProps> = ({
...
+    jurisdictionDiscrepanciesId,
+    setJurisdictionDiscrepanciesId,
+  ] = useState<string | null>(null)
 
   const ballotsOrBatches =
     auditType === 'BATCH_COMPARISON' ? 'Batches' : 'Ballots'
 
   const someJurisdictionHasDiscrepancies = Object.values(
-    discrepancyCountsQuery.data ?? {}
+    discrepancyCountsQuery.data ? discrepancyCountsQuery.data : {}
```
nbhatia823, line 103: For some reason eslint wouldn't allow me to commit this file because it kept giving the following:
`103:32  ✖  This experimental syntax requires enabling the parser plugin: 'nullishCoalescingOperator' (103:32)`

client/src/components/AuditAdmin/Progress/Progress.test.tsx
```diff
@@ -325,14 +326,29 @@ describe('Progress screen', () => {
     })
   })
 
-  it('shows round status for batch comparison', async () => {
+  it('shows round and discrepancy status for batch comparison', async () => {
```
nbhatia823, line 329: This test was already checking discrepancy counts and downloading the discrepancy report, so it seemed like a good place to also test the discrepancy modal

client/src/components/AuditBoard/__snapshots__/AuditBoardView.test.tsx.snap
```diff
@@ -61,19 +61,19 @@ exports[`AuditBoardView ballot interaction renders ballot route 1`] = `
         class="sc-bwzfXH ijTqJq"
       >
         <div
-          class="sc-bMvGRv eoqrva"
+          class="sc-CtfFt dyWieC"
```
nbhatia823, line 64: I didn't make any change related to this file, and I see that @jonahkagan also had a similar change to this snapshot in [this PR that didn't touch the associated file](https://github.com/votingworks/arlo/pull/2000/files#diff-77684cd11a1fe7b0251766ae865668179aa663969064792892f030cf467c0f98R10), so I assume this is safe.

nbhatia823: @arsalansufi any insight on what might be triggering this change? I see Jonah's [commit message](https://github.com/votingworks/arlo/pull/2000/commits/dc789f1c931b89c74c65a430b4e968dfe13678a9) referenced styled components and all the classes changed are prefixed with `sc`, so maybe this is some behavior we have seen before

client/src/components/AuditAdmin/Progress/Progress.tsx
kshen0: You can try upgrading to `ecmaVersion: 2021` [here](https://github.com/votingworks/arlo/blob/main/client/.eslintrc.js#L43) per this post: https://stackoverflow.com/questions/76092858/how-can-i-get-eslint-to-accept-the-nullish-coalescing-assignment

client/src/components/AuditAdmin/Progress/Progress.tsx
```diff
@@ -324,10 +334,23 @@ const Progress: React.FC<IProgressProps> = ({
             )
           }
           if (!value) return null
+          if (!showDiscrepanciesReview)
+            return (
+              <>
+                <Icon icon="flag" intent="danger" /> {value.toLocaleString()}
+              </>
```
kshen0, line 341: Delete the wrapping brackets?

client/src/components/AuditAdmin/Progress/Progress.tsx
```diff
@@ -501,6 +524,17 @@ const Progress: React.FC<IProgressProps> = ({
           auditSettings={auditSettings}
         />
       )}
+      {jurisdictionDiscrepanciesId && round && (
```
kshen0, line 527: Why is `round` needed in this check?

client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
nbhatia823: Good points. I'm thinking we could say "Unknown" and then include the relevant ID, so there is at least something to reference. I lean towards thinking it's better than having the page crash, as it doesn't seem dangerous or nonsensical, just unclear. But happy to be overridden on that, still developing the intuition

client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
```diff
@@ -0,0 +1,126 @@
...
+      style={{ width: '600px' }}
+      title={`${jurisdiction.name} Discrepancies`}
+    >
+      <div className={Classes.DIALOG_BODY} style={{ marginBottom: 0 }}>
+        {Object.entries(discrepanciesByBatch).map(
+          ([batchName, discrepanciesByContest]) => {
+            return Object.entries(discrepanciesByContest).map(
+              ([contestId, contestDiscrepancies], idx) => (
+                <div key={contestId}>
+                  <H6 style={idx === 0 ? { marginTop: '8px' } : {}}>
```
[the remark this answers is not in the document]
nbhatia823: Updating! For my understanding, does that tend to be preferable to using the `style` tag?

client/src/components/AuditAdmin/Progress/Progress.tsx
nbhatia823: Ah yeah it looks confusing but they are actually needed as they wrap the Icon and a string

client/src/components/AuditAdmin/Progress/Progress.tsx
nbhatia823: Hm slipped in, not needed as we already hide discrepancies and thus the ability to select one if `round` is not set [here](https://github.com/votingworks/arlo/pull/2033/files#diff-e6c71ff26929b295d562cf958683bb2cd9259f1810dd15ee35cd7bf8b0222363R80). Removing

client/src/components/AuditAdmin/Progress/Progress.tsx
nbhatia823: Screenshot 2024-11-04 at 12 06 14 PM

Still getting the same error from the pre-commit hook. I did some quick debugging last week and wasn't able to fix it, and it's bit a strange since this code clearly was already committed. 

I'm wondering if it makes sense to just move forward for now, given the tight timeline, and then working on fixing this as a follow up ticket sometime soon.

client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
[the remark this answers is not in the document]
kshen0: I think you could argue either way - for something simple like this, it's nice to have the styles located exactly where the the element is. It also keeps the code streamlined to inline the styles.

On the other hand, using styled components makes it easier to extend or reuse styles later. For applying a single style I think it comes down to personal preference. For multiple styles I'd recommend using styled components.

client/src/components/AuditAdmin/Progress/Progress.tsx
kshen0: Oh absolutely right, I missed the string in the right half of the line

kshen0 (approved):

client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
```diff
@@ -0,0 +1,126 @@
...
+  useDiscrepanciesByJurisdiction,
+} from '../../useJurisdictions'
+import useContestsJurisdictionAdmin from '../../JurisdictionAdmin/useContestsJurisdictionAdmin'
+import { IContest } from '../../../types'
+
+const ContestDiscrepanciesTable = styled(HTMLTable).attrs({
+  bordered: true,
+  striped: true,
+})`
+  &.${Classes.HTML_TABLE} {
```
jonahkagan, line 15: Nit: I don't think you need the `${Classes.HTML_TABLE}` in the selectors here, since styled-components will be creating its own class that applies only to this component.

client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
jonahkagan: I think these are impossible cases and would recommend using assertions to communicate that. In order for a contest or choice to not be found, there'd need to be a bug in the backend or a caching error. We should fail fast in either case so we can fix it.

client/src/components/AuditAdmin/Progress/Progress.tsx
jonahkagan: Weird, yeah I don't think it's worth figuring out right now. Also we can replace the discrepancy counts query by using the discrepancies query soon anyway.

client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
```diff
@@ -0,0 +1,132 @@
...
+    background: #ffffff;
+    border: 1px solid ${Colors.LIGHT_GRAY1};
+    margin-bottom: 32px;
+    table-layout: fixed;
+    width: 100%;
+  }
+
+  &.${Classes.HTML_TABLE} td {
+    vertical-align: middle;
+  }
```
jonahkagan, line 25: Did you look at the various styled table options in components/Atoms/Table.tsx? There might be one that works well here, but if not I'm fine with keeping what you have

client/src/components/AuditBoard/__snapshots__/AuditBoardView.test.tsx.snap
jonahkagan: I think when you add a new styled component (or update one, I can't quite remember) it regenerates all the class names, which are included in the snapshots. We found a [plugin](https://github.com/styled-components/jest-styled-components) that fixes this issue and added it to VxSuite, but haven't gotten around to doing it in Arlo.

jonahkagan (approved):
Looks great!

jonahkagan:
Small nit: I think we can remove the danger intent from the "Download Discrepancy Report" flag icon now that we have the buttons to review on screen. I think we should focus users on the on-screen review now that we have that.

client/src/components/AuditBoard/__snapshots__/AuditBoardView.test.tsx.snap
nbhatia823: Ah okay, thanks for the explanation

client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
nbhatia823: StyledTable
Screenshot 2024-11-04 at 3 35 32 PM

FlexTable
Screenshot 2024-11-04 at 3 29 53 PM

HTMLTable
Screenshot 2024-11-04 at 3 34 08 PM

I hadn't, so just mocked the three up. I don't have a strong leaning (though the new options could definitely be improved with some borders etc.). I'm probably in between the current and the StyledTable (with border added) due to the row striping.

jonahkagan: Thanks for investigating. I'm now realizing we haven't been that consistent in using these, and I actually like the current design the best haha. Sorry for the detour!

nbhatia823: No worries, was good to see those options! Will continue as is

jonahkagan:
@nbhatia823 I think we had a slight miscommunication - you removed the flag entirely, but I was suggesting just to remove the red color from the flag icon. Now it has no icon and is a bit out-of-pattern with the other buttons:

Screenshot 2024-11-04 at 4 15 01 PM

I think we could potentially switch to the download button that the other buttons have, or put the flag back in, but gray

nbhatia823:
Ah sorry about that. I will make a PR to add it back in

PR #2034, opened 2024-11-01
Show discrepancies in UI for ballot comparison audits

nbhatia823:
Recommend to review commit by commit

This PR adds discrepancies in UI for ballot comparison audits.

[Issue](https://github.com/votingworks/arlo/issues/1912)

Screenshot 2024-11-06 at 1 25 56 PM

https://github.com/user-attachments/assets/ab0b007c-cf93-4312-9005-808ac2705667

### DIFF
--- a/client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
+++ b/client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
@@ -48,16 +48,16 @@ function getChoiceName(
   return choice.name
 }
 
-function displayDiscrepancy(val: string | number | undefined): string | number {
+function formatVoteCount(val: string | number | undefined): string | number {
   switch (val) {
     case 'o':
       return 'Overvote'
     case 'u':
       return 'Undervote'
-    case undefined:
+    case undefined: // Seen on 'reportedVotes' when a candidate did not have a vote originally
       return 0
     default:
-      return val
+      return val.toLocaleString()
   }
 }
 
@@ -122,17 +122,17 @@ const JurisdictionDiscrepancies: React.FC<IJurisdictionDiscrepanciesProps> = ({
                               {getChoiceName(contests, contestId, choiceId)}
                             </td>
                             <td>
-                              {displayDiscrepancy(
+                              {formatVoteCount(
                                 contestDiscrepancies.reportedVotes[choiceId]
                               )}
                             </td>
                             <td>
-                              {displayDiscrepancy(
+                              {formatVoteCount(
                                 contestDiscrepancies.auditedVotes[choiceId]
                               )}
                             </td>
                             <td>
-                              {displayDiscrepancy(
+                              {formatVoteCount(
                                 contestDiscrepancies.discrepancies[choiceId]
                               )}
                             </td>
--- a/client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
+++ b/client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
@@ -48,15 +48,18 @@ function getChoiceName(
   return choice.name
 }
 
-function formatVoteCount(val: string | number | undefined): string | number {
+function formatVoteCount(val: string | number | undefined): string {
   switch (val) {
     case 'o':
       return 'Overvote'
     case 'u':
       return 'Undervote'
     case undefined: // Seen on 'reportedVotes' when a candidate did not have a vote originally
-      return 0
+      return '0'
     default:
+      if (typeof val === 'string') {
+        return val
+      }
       return val.toLocaleString()
   }
 }
--- a/client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
+++ b/client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
@@ -48,6 +48,19 @@ function getChoiceName(
   return choice.name
 }
 
+function displayDiscrepancy(val: string | number | undefined): string | number {
+  switch (val) {
+    case 'o':
+      return 'Overvote'
+    case 'u':
+      return 'Undervote'
+    case undefined:
+      return 0
+    default:
+      return val
+  }
+}
+
 export interface IJurisdictionDiscrepanciesProps {
   electionId: string
   handleClose: () => void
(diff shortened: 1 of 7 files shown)


server/api/jurisdictions.py
```diff
@@ -804,9 +814,110 @@ def get_batch_comparison_audit_discrepancies_by_jurisdiction(
...
+    )
+    sampled_ballot_id_to_jurisdiction_id = dict(
+        SampledBallot.query.filter(SampledBallot.id.in_(ballots_in_round))
+        .join(Batch)
+        .with_entities(SampledBallot.id, Batch.jurisdiction_id)
+    )
+    # make a readable identifier - it is the same default format as the ImprintedID
+    # but some sampled ballots have ImprintedIDs of a different format that are
+    # very long and unreadable, so here we ensure they all have the same format
+    # Ex. "Tabulator X, Batch Y, Ballot Position Z, Container 0"
```
nbhatia823, line 851: It might be more idiomatic if the frontend created this readable ID, but it makes the typing much more consistent with the batch comparison approach to create the ID here. Happy to get input on this design choice

server/api/jurisdictions.py
```diff
@@ -804,9 +814,110 @@ def get_batch_comparison_audit_discrepancies_by_jurisdiction(
...
+                }
+
+    return discrepancies_by_jurisdiction
+
+
+# normalize_vals_for_discrepancies normalizes the return type for batch and ballot comparison discrepanices.
+# It converts string numbers to int numbers, leaving non-numbers as is (i.e. 'o', 'u'). This is useful for
+# ballot CVRs to ensure that numbers are always returned as numbers, to match the behavior of batch discrepancies
+# and of vote deltas
+def normalize_vals_for_discrepancies(
```
nbhatia823, line 911: Ballot comparison cvrs are different than their batch comparison equivalent in that the values are all strings vs ints, to support 'o' and 'u'. I thought it would be better to be consistent in returning ints where it is a number, so that is the purpose of this function. It allows us to have a single return type for discrepancies that is relatively simple. Feel free to let me know if it doesn't seem right

server/api/jurisdictions.py
```diff
@@ -804,9 +814,110 @@ def get_batch_comparison_audit_discrepancies_by_jurisdiction(
...
+            if reported_cvr and contest.id in reported_cvr:
+                reported_votes = reported_cvr[contest.id]
+
+            audited_votes = {}
+            if audited_cvr and contest.id in audited_cvr:
+                audited_votes = audited_cvr[contest.id]
+
+            if ballot_id in sampled_ballot_id_to_jurisdiction_id:
+                if not reported_cvr:
+                    reported_cvr = {}
```
jonahkagan, line 891: What is this case for? I think it could be removed

server/api/jurisdictions.py
```diff
@@ -804,9 +814,110 @@ def get_batch_comparison_audit_discrepancies_by_jurisdiction(
...
+            if not vote_deltas or isinstance(vote_deltas, str):
+                continue
+
+            reported_votes = {}
+            if reported_cvr and contest.id in reported_cvr:
+                reported_votes = reported_cvr[contest.id]
+
+            audited_votes = {}
+            if audited_cvr and contest.id in audited_cvr:
+                audited_votes = audited_cvr[contest.id]
```
jonahkagan, line 887: I think `reported_cvr` and `audited_cvr` are guaranteed to be non-null at this point due to the checks in `ballot_vote_deltas`, so these could probably be condensed to:

```py
reported_votes = reported_cvr.get(contest.id, {})
audited_votes = audited_cvr.get(contest.id, {})
```

server/api/jurisdictions.py
jonahkagan: What happens if we just leave them as their internal representation (ints for batch comparison, strings for ballot comparison)? Might be simpler to just keep the same representation until there's somewhere we actually need to use the strings as ints, that way you don't need to think about whether it's an int in one place or a string in other places

client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
```diff
@@ -48,6 +48,19 @@ function getChoiceName(
...
 }
 
+function displayDiscrepancy(val: string | number | undefined): string | number {
+  switch (val) {
+    case 'o':
+      return 'Overvote'
+    case 'u':
+      return 'Undervote'
+    case undefined:
+      return 0
```
jonahkagan, line 58: When is this undefined? Could maybe use a comment explaining this case

client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
```diff
@@ -48,6 +48,19 @@ function getChoiceName(
...
+    default:
+      return val
```
jonahkagan, line 60: Might be nice to add a `.toLocaleString()` for longer numbers

client/src/components/AuditAdmin/Progress/JurisdictionDiscrepancies.tsx
```diff
@@ -48,6 +48,19 @@ function getChoiceName(
   return choice.name
 }
 
+function displayDiscrepancy(val: string | number | undefined): string | number {
```
jonahkagan, line 51: Nit: this function is used to format both votes and discrepancies, so the name isn't quite accurate. Maybe `formatVoteCount`?

jonahkagan (approved):
Thanks for the explanatory comments!

server/api/jurisdictions.py
nbhatia823: Hm that's true. I guess I thought it might be a bit odd if the frontend is getting strings on some API calls and ints on others for the same field values, when they might both be representing a 1, so that is what motivated me to normalize it. But, leaving them as is will work for sure too.

server/api/jurisdictions.py
nbhatia823: Removing. I think it slipped through a previous iteration, as `reported_cvr` isn't even accessed after this

server/api/jurisdictions.py
jonahkagan: I meant they are logically guaranteed to not be None. I know the type system can't figure that out, since the checks to handle the None cases are in a different function. Up to you whether you prefer handling an impossible case to satisfy the type-system or overriding with a type assertion (I usually go for the latter, but that's just my style). Totally fine to leave it as is

server/api/jurisdictions.py
nbhatia823: Removed normalization [here](https://github.com/votingworks/arlo/pull/2034/commits/808cfeb3afd121a738256ffd1682da7e6430f754)

server/api/jurisdictions.py
nbhatia823: Oh interesting. Thanks for the clarification. I'll add in the type assertion

PR #2035, opened 2024-11-04
provide ability to allow-list specific email addresses for support users

benadida:


benadida:
switching to Auth0-based controls instead.

PR #2041, opened 2024-11-05
Add icon to download discrepancy report button

nbhatia823:
This PR follows up [this](https://github.com/votingworks/arlo/pull/2033) to add the icon that was removed from the download button. @jonahkagan [suggested](https://github.com/votingworks/arlo/pull/2033#issuecomment-2455501439) removing the danger intent from the flag, and I mistakenly removed the icon altogether.

I currently lean towards the Flag instead of the Download icon to match the flag used for discrepancies elsewhere on the page, and because it is a slightly different sort of report in that it is flagging issues rather than being a data dump, but happy to go with either.

Screenshot 2024-11-04 at 8 21 51 PM


jonahkagan (approved):
Thanks for fixing!

Issue #1928, opened 2024-04-23
Incorporate ballots that haven't been counted until after audit launches


arsalansufi:
For batch comparison, Arlo isn't looking at total number of ballot discrepancies which could be inflating the p-value.  This is a Mark L flag, we should start collecting the number of "other" ballots essentially.  And doing so, may result in Arlo saying there are more ballots than expected b/c +1-3 happens and now allow the entry of the totals.

Issue automatically created from Slack feedback reported by ginny: 
https://votingworks.slack.com/archives/C01F2CH62TS/p1712758560619089

ginvdr:
More context: https://votingworks.slack.com/archives/CKCVA0F9S/p1728670752890309

ginvdr:
And more: https://votingworks.slack.com/archives/CKCVA0F9S/p1728671574682849

PR #2042, opened 2024-11-05
Incorporate pending ballots in batch comparison audits

jonahkagan:
Task: #1928 

Problem as described by @MarkVVaudits:

> The underlying challenge is that batch data, state-level vote counts at audit launch, and the final certified totals all can be different. This can occur **because some ballots are still pending at audit launch (they have not yet been tallied, and their eligibility may be undetermined)** and/or because some ballots are treated as unauditable and not included in batch uploads.

This feature provides a way for audit admins to tell Arlo about any ballots that have not been tallied before audit launch (aka pending ballots). The batch comparison audit math is then updated to treat these ballots conservatively, assuming they are votes for the loser and modifying the reported margin as such. This increases the resulting sample size and p-value.

This should have minimal, but important, impact in real world audits, according to  @MarkVVaudits:
> I foresee this being used in contexts where pending/unauditable ballots are not a large fraction of any margin.

Note that I will be following up with a second PR to handle "unauditable ballots," which require a different approach.

https://github.com/user-attachments/assets/6bb3d026-0ac5-406a-b6e5-e23bda704e9a

### DIFF
--- a/client/src/components/AuditAdmin/Setup/Contests/ContestForm.tsx
+++ b/client/src/components/AuditAdmin/Setup/Contests/ContestForm.tsx
@@ -171,6 +171,7 @@ export interface IContestValues {
   votesAllowed: string
   choices: IChoiceValues[]
   totalBallotsCast?: string
+  pendingBallots?: string
   jurisdictionIds: string[]
 }
 
@@ -183,6 +184,7 @@ const contestToValues = (contest: IContest): IContestValues => ({
     numVotes: choice.numVotes.toString(),
   })),
   totalBallotsCast: contest.totalBallotsCast?.toString(),
+  pendingBallots: contest.pendingBallots?.toString() || '',
 })
 
 const contestFromValues = (contest: IContestValues): IContest => ({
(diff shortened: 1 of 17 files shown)


client/src/components/AuditAdmin/Setup/Contests/ContestForm.tsx
```diff
@@ -481,6 +486,18 @@ const ContestForm: React.FC<IProps> = ({
                             </label>
                           </FormSection>
                         )}
+                        {isBatchComparison && isTargeted && (
```
carolinemodic, line 489: So pending ballots is defined on a per-contest basis right? I suppose that seems fine, is there a reason why we chose that over having it set at the audit level? I suppose it is better to have more flexibility but is there a known case where it could be different by contest?

carolinemodic (approved):
I did not carefully review the math in the test you added, but all of the logic LGTM and the high level idea makes sense to me.

client/src/components/AuditAdmin/Setup/Contests/ContestForm.tsx
jonahkagan: I assume it could be different if there are different ballot styles that include different contests. But this approach of defining it for each contest was specified by @MarkVVaudits, so he'd be the best person to answer.

MarkVVaudits: @carolinemodic, great question. Yes, what Jonah said -- applicable not only once ballot styles are explicitly supported, but in the current design where contests can span different sets of jurisdictions. Imagine auditing a U.S. House contest nested in a statewide audit: the number of pending ballots statewide might be completely immaterial to the House contest, depending on the underlying facts.

Issue #1532, opened 2022-07-07
Prevent race conditions between dependent file uploads


jonahkagan:
If a jurisdiction reuploads its ballot manifest while its CVR file is processing, this can cause a race condition since the batches will be deleted and recreated in the database. In general, file uploads for a jurisdiction should not be allowed to occur simultaneously. This includes automatic reprocessing of files (e.g. when CVR automatically reprocessed on manifest upload).

jonahkagan:
Another related race condition: In the batch inventory flow, if a user uploads a CVR file and then deletes a tabulator status file before reprocessing starts, the worker will fail to find the tabulator status file in the db and error unexpectedly. (This could also happen with any other dependent files.)

jonahkagan:
Another example that occurred: A user deleted the batch tallies file while it was queued to be reprocessed, and when the task reached the worker, it failed due to a missing batch tallies file in the db.

eventualbuddha:
Noting that this is only partially fixed. In particular, deleting files that a currently-processing file needs to reference is not blocked in most cases. cc @arsalansufi to re-prioritize

jonahkagan:
Given that we've been switching most of the file upload UI to only have a delete button, rather than a replace button, I think it's important to handle the delete cases as well. I'd like to reopen and make sure we handle those cases too.

PR #2043, opened 2024-11-05
fix: prevent some updates that would affect incomplete jobs

eventualbuddha:
Refs https://github.com/votingworks/arlo/issues/1532

When making updates we'd like to check that no incomplete background jobs are dependent on the data we're about to change. This change takes steps in that direction, but does not fully fix the issue. For example, this change does not prevent deleting files where an incomplete job references that file.

I'd appreciate a more thorough review of this since it's the first non-trivial change I've made to Arlo in a while. See [this Slack thread](https://votingworks.slack.com/archives/CKCVA0F9S/p1730750352776399) for context.

### DIFF
--- a/server/api/batch_tallies.py
+++ b/server/api/batch_tallies.py
@@ -199,6 +199,9 @@ def validate_batch_tallies_upload(election: Election, jurisdiction: Jurisdiction
     if not jurisdiction.manifest_file_id:
         raise Conflict("Must upload ballot manifest before uploading batch tallies.")
 
+    if jurisdiction.manifest_file.is_processing():
+        raise Conflict("Cannot update batch tallies while manifest file is processing.")
+
 
 def clear_batch_tallies_data(jurisdiction: Jurisdiction):
     jurisdiction.batch_tallies = None
--- a/server/api/batch_tallies.py
+++ b/server/api/batch_tallies.py
@@ -197,10 +197,14 @@ def validate_batch_tallies_upload(election: Election, jurisdiction: Jurisdiction
         raise Conflict("Jurisdiction does not have any contests assigned")
 
     if not jurisdiction.manifest_file_id:
-        raise Conflict("Must upload ballot manifest before uploading candidate totals by batch.")
+        raise Conflict(
+            "Must upload ballot manifest before uploading candidate totals by batch."
+        )
 
     if jurisdiction.manifest_file.is_processing():
-        raise Conflict("Cannot update candidate totals by batch while ballot manifest is processing.")
+        raise Conflict(
+            "Cannot update candidate totals by batch while ballot manifest is processing."
+        )
 
 
 def clear_batch_tallies_data(jurisdiction: Jurisdiction):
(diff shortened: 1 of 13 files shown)


server/tests/batch_comparison/test_multi_contest_batch_comparison.py
```diff
@@ -1197,3 +1197,82 @@ def test_multi_contest_batch_comparison_editing_contests_after_uploads(
...
         assert batch_tallies_status["error"] is None
+
+def test_updating_contests_while_jurisdictions_file_is_being_processed(
+    client: FlaskClient,
+    election_id: str,
+    jurisdiction_ids,  # pylint: disable=unused-argument
+    contest_ids,  # pylint: disable=unused-argument
+    election_settings,  # pylint: disable=unused-argument
+    manifests,  # pylint: disable=unused-argument
+    batch_tallies,  # pylint: disable=unused-argument
```
jonahkagan, line 1208: I don't think you need any of these fixtures to execute for this test. Might be good to review all the test cases you added and remove unneeded fixtures. But not a huge issue if you miss a spot

server/tests/helpers.py
```diff
@@ -595,3 +597,11 @@ def zip_hart_cvrs(cvrs: List[str]):
...
     files["WriteIns"] = io.BytesIO()
     return io.BytesIO(zip_files(files).read())
+
+@contextmanager
+def no_automatic_task_execution():
+    config.RUN_BACKGROUND_TASKS_IMMEDIATELY = False
+    try:
+        yield
+    finally:
+        config.RUN_BACKGROUND_TASKS_IMMEDIATELY = True
```
jonahkagan, line 607: 🆒

server/api/batch_tallies.py
```diff
@@ -199,6 +199,9 @@ def validate_batch_tallies_upload(election: Election, jurisdiction: Jurisdiction
     if not jurisdiction.manifest_file_id:
         raise Conflict("Must upload ballot manifest before uploading batch tallies.")
 
+    if jurisdiction.manifest_file.is_processing():
+        raise Conflict("Cannot update batch tallies while manifest file is processing.")
```
jonahkagan, line 203: The user-facing name for batch tallies is "candidate totals by batch" so we should update all error messages to use that language. And I would also replace "manifest file" with "ballot manifest"

jonahkagan (approved):
Looks great!

carolinemodic (approved):
nothing to add on top of jonah's comments

server/api/batch_tallies.py
eventualbuddha: [8ac01cb](https://github.com/votingworks/arlo/pull/2043/commits/8ac01cb173a94a26a6a4d3b10e73304bb9eabcf4)

server/tests/batch_comparison/test_multi_contest_batch_comparison.py
eventualbuddha: [9fed584](https://github.com/votingworks/arlo/pull/2043/commits/9fed584c13723d2859bc8c78d250d7b5f5336e93)

eventualbuddha:
Clearly I had not yet set up the `husky` git commit hooks. Now I have 🤣

PR #2044, opened 2024-11-06
Upgrade stylelint-processor-styled-components to fix error

jonahkagan:
Fixes the error about using the nullish coalescing operator. Based on https://github.com/stylelint/stylelint/issues/4473#issuecomment-675555699


nbhatia823 (approved):
Thank you 🙏

PR #2045, opened 2024-11-06
Incorporate unauditable ballots in batch comparison audits

jonahkagan:
Task: #1928 

Second step after #2042. When the vote totals entered by the audit admin during audit setup are greater than the vote totals from the batch tallies uploaded by jurisdictions, that means there are some ballots that were counted but are unauditable.

This changes the audit math to calculate how many ballots were left out of the batch tallies based on the vote difference, and then treats them as votes for the losing candidate.

It also includes the aggregated vote totals from the batch tallies in the audit report so that it's possible to reconstruct the p-value later.


MarkVVaudits:
This looks good to me. I suggest editing the docstring for compute_unauditable_ballots to note that the calculation doesn't take account of undervoted ballots (which can be conservatively estimated and incorporated in the pending count).

MarkVVaudits (approved):
I put my one comment in conversation, but substantively this looks fine to me.

carolinemodic (approved):

Issue #2039, opened 2024-11-04
Add more validation of storagePathKey in upload-complete endpoints


carolinemodic:
See discussion [here](https://votingworks.slack.com/archives/CKCVA0F9S/p1730745761088699)

We should add validation that this path matches what we expect to avoid saving a reference to an arbitrary file location.

PR #2047, opened 2024-11-06
Add stronger validation for storage path on upload complete endpoints

carolinemodic:
https://github.com/votingworks/arlo/issues/2039 

Validates that the storage path passed to upload-complete is the expected path before saving.

### DIFF
--- a/server/api/batch_inventory.py
+++ b/server/api/batch_inventory.py
@@ -35,20 +35,18 @@ from ..models import *  # pylint: disable=wildcard-import
 from ..util.csv_parse import (
     validate_comma_delimited,
     is_filetype_csv_mimetype,
-    validate_csv_mimetype,
 )
 from ..util.file import (
+    FileType,
     get_file_upload_url,
-    get_standard_file_upload_request_params,
+    get_jurisdiction_folder_path,
+    validate_and_get_standard_file_upload_request_params,
     retrieve_file,
     retrieve_file_to_buffer,
     serialize_file,
     serialize_file_processing,
     timestamp_filename,
     unzip_files,
-    validate_csv_or_zip_mimetype,
-    validate_zip_mimetype,
-    validate_xml_mimetype,
 )
 from ..util.hart_parse import find_xml, parse_contest_results, find_text_xml
 from ..worker.tasks import UserError, background_task, create_background_task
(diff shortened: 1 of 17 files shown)


server/util/file.py
```diff
@@ -163,7 +164,22 @@ def get_file_upload_url(
...
+    election_id: str,
+    jurisdiction_id: str,
+) -> str:
+    return f"{get_audit_folder_path(election_id)}/jurisdictions/{jurisdiction_id}"
+
+
+def validate_and_get_standard_file_upload_request_params(
+    request: Request,
+    validate_storage_path: Callable[[str], None],
+    validate_file_type: Callable[[str], None],
```
jonahkagan, line 181: As a reader, it's a little complicated to grok the call sites of this function. It seems like they are mostly doing the same thing, so I wonder if this function could have the following signature:

```
def validate_and_get_standard_file_upload_request_params(
  request: Request,
  expected_file_directory_path: str,
  expected_file_name_prefix: str,
  expected_file_extension: str | List[str]
  expected_file_type: str | List[str]
)
```

Also, I'm realizing that file types and file extensions are in a predefined mapping, so maybe we could pull that out into a constant and just pass in one of them.

server/api/batch_inventory.py
```diff
@@ -747,21 +755,23 @@ def complete_upload_for_batch_inventory_cvr(
...
-    elif batch_inventory_data.system_type == CvrFileType.ESS:
-        validate_csv_or_zip_mimetype(file_type)
+    expected_file_types = [FileType.CSV]
+    if batch_inventory_data.system_type == CvrFileType.ESS:
+        expected_file_types = [FileType.CSV, FileType.ZIP]
     elif batch_inventory_data.system_type == CvrFileType.HART:
-        validate_zip_mimetype(file_type)
-    else:
+        expected_file_types = [FileType.ZIP]
+    elif batch_inventory_data.system_type != CvrFileType.DOMINION:
```
jonahkagan, line 763: It's a little confusing that the dominion case is set as a default at the top and then checked as the last case in the if/else chain. Now that this is just a mapping from `CvrFileType` to `List[FileType]`, maybe we could just use a `dict`?

server/tests/batch_comparison/test_batch_inventory.py
```diff
@@ -316,7 +316,7 @@ def test_batch_inventory_happy_path(
         json.loads(rv.data),
         {
             "file": {
-                "name": asserts_startswith("batchTallies"),
+                "name": asserts_startswith("batch_tallies"),
```
jonahkagan, line 319: Nit: should we be consistent with file prefixes using either hyphen or underscore word separators? I noticed there's a mix

jonahkagan (approved):
Looks great! Excited to see where we can use `FileType` further!

PR #2049, opened 2024-11-07
Support combined batches when calculating discrepancies

nbhatia823:
This PR closes out https://github.com/votingworks/arlo/issues/1912.

It adds support for combined batches when calculating discrepancies.

Screenshot 2024-11-07 at 2 07 26 PM

### DIFF
--- a/server/api/jurisdictions.py
+++ b/server/api/jurisdictions.py
@@ -785,7 +785,6 @@ def get_batch_comparison_audit_discrepancies_by_jurisdiction(
         lambda: defaultdict(dict)
     )
 
-    # TODO: Add support for combined batches
     batch_keys_in_round = set(
         SampledBatchDraw.query.filter_by(round_id=round_id)
         .join(Batch)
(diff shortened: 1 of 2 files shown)


server/api/jurisdictions.py
```diff
@@ -805,6 +826,28 @@ def get_batch_comparison_audit_discrepancies_by_jurisdiction(
...
+                    for sub_batch in sub_batches
+                )
+                reported_contest_result = {
+                    choice.id: sum(
+                        reported_results[choice.id]
+                        for reported_results in sub_batch_reported_results
+                        if reported_results is not None
+                    )
+                    for choice in contest.choices
+                }
```
jonahkagan, line 849: After reading this code, I went back and checked out how we compute the discrepancy counts for combined batches, and I think there's a bug there, since we only use the reported results from the representative batch rather than summing all the sub-batches. The test didn't catch it because the test data still resulted in a discrepancy, just not the correct one.

I think we should remove that endpoint anyways and just count the discrepancies sent to the frontend, so it's probably moot, but that's a motivating reason to do that refactor sooner rather than later.

jonahkagan (approved):

server/api/jurisdictions.py
nbhatia823: Hm right, good catch. I'm off tomorrow as well, but happy to take it on early next week if we'd like

jonahkagan: I'll just knock it out right now. Should be quick

jonahkagan: https://github.com/votingworks/arlo/pull/2051

PR #2051, opened 2024-11-07
Refactor Audit Progress to use discrepancy query instead of discrepancy counts endpoint

jonahkagan:
Now that we load the discrepancies themselves, we don't need to make a separate query to load the counts - we can just count them.

### DIFF
--- a/client/src/components/AuditAdmin/Progress/Progress.test.tsx
+++ b/client/src/components/AuditAdmin/Progress/Progress.test.tsx
@@ -242,11 +242,6 @@ describe('Progress screen', () => {
   it('shows round and discrepancy status for ballot comparison', async () => {
     const expectedCalls = [
       aaApiCalls.getMapData,
-      aaApiCalls.getDiscrepancyCounts({
-        [jurisdictionMocks.allComplete[0].id]: 0,
-        [jurisdictionMocks.allComplete[1].id]: 2,
-        [jurisdictionMocks.allComplete[2].id]: 1,
-      }),
       aaApiCalls.getDiscrepancies({
         [jurisdictionMocks.oneComplete[1].id]: {
           ballot1: {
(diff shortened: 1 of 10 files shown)


client/src/components/AuditAdmin/Progress/Progress.test.tsx
```diff
@@ -369,6 +373,11 @@ describe('Progress screen', () => {
       expect(footers[4]).toHaveTextContent('60')
       expect(footers[5]).toHaveTextContent('0')
 
+      userEvent.click(within(dialog).getByRole('button', { name: 'Close' }))
+      await waitFor(() => {
+        expect(dialog).not.toBeInTheDocument()
+      })
```
nbhatia823, line 379: Ah, right

client/src/components/AuditAdmin/Progress/Progress.tsx
```diff
@@ -60,6 +61,19 @@ const totalFooter = <T extends object>(headerName: string) => (
...
+// We count the number of batch-contest pairs with discrepancies, not the number
+// of batches with discrepancies.
+const countDiscrepanciesForJurisdiction = (
+  discrepancies: DiscrepanciesByJurisdiction,
+  jurisdictionId: string
+) => {
+  return sum(
+    Object.entries(discrepancies[jurisdictionId] ?? {}).map(
+      ([_, contestDiscrepancies]) => Object.keys(contestDiscrepancies).length
+    )
```
nbhatia823, line 73: nit: maybe `Object.values()` since the key isn't used?

```
Object.values(discrepancies[jurisdictionId] ?? {}).map(
      contestDiscrepancies => Object.keys(contestDiscrepancies).length
    )
```

client/src/components/AuditAdmin/Progress/Progress.tsx
```diff
@@ -505,6 +519,7 @@ const Progress: React.FC<IProgressProps> = ({
       )}
       {jurisdictionDiscrepanciesId && (
         <JurisdictionDiscrepancies
+          discrepancies={discrepancyQuery.data!}
```
nbhatia823, line 522: What does the `!` do?

nbhatia823 (approved):
LGTM!

client/src/components/AuditAdmin/Progress/Progress.tsx
jonahkagan: It's a non-null assertion: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

Our lint rules outlaw it in VxSuite (I don't quite know the reasoning behind this, because we built a helper, `assertDefined`, to do the same thing)

PR #2052, opened 2024-11-07
Make list of audits for support users easier to navigate

carolinemodic:
Tweaks the list of audits on the support users home screen and the per-organization view to include a "status" tag with one of the following statuses: 

- Not Started - Round 1 has not started
- Round X In Progress
- Completed

The active audits list on the home page is already querying for audits updated in the last 2 weeks, I tweaked this to order the results by the last update date as I have personally found this confusing behavior. 

The organization page now orders audits by creation date (most recent first) to match the audit admin view 

New Screens 
Screenshot 2024-11-07 at 3 05 54 PM
Screenshot 2024-11-07 at 3 05 45 PM

### DIFF
--- a/client/src/components/SupportTools/SupportTools.tsx
+++ b/client/src/components/SupportTools/SupportTools.tsx
@@ -48,6 +48,7 @@ import { List, LinkItem } from './List'
 import Breadcrumbs from './Breadcrumbs'
 import { stateOptions, states } from '../AuditAdmin/Setup/Settings/states'
 import StatusTag from '../Atoms/StatusTag'
+import { sortBy } from '../../utils/array'
 
 const Table = styled(HTMLTable)`
   margin: 10px 0;
@@ -239,8 +240,8 @@ const Organization = ({ organizationId }: { organizationId: string }) => {
 
   const { name, defaultState, elections, auditAdmins } = organization.data
 
-  const sortedElections = elections.sort(
-    (a, b) => new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime()
+  const sortedElections = sortBy(elections, a =>
+    new Date(a.createdAt).getTime()
   )
 
   const onClickRemoveAuditAdmin = (auditAdmin: IAuditAdmin) =>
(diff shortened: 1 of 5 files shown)


server/api/support.py
```diff
@@ -89,7 +100,11 @@ def list_active_elections():
             )
         )
         .join(Organization)
-        .order_by(Organization.name, Election.audit_name)
+        .join(
+            ActivityLogRecord,
+            ActivityLogRecord.info["base"]["election_id"].as_string() == Election.id,
+        )
+        .order_by(ActivityLogRecord.timestamp.desc())
```
carolinemodic, line 107: I don't love that I order the results for active audits in the server endpoint, and order the audits on the organization level in the client. However it seemed to make way more sense to order this specific endpoint here and not have the ActivityLogRecord knowledge need to leave this function. Otherwise for the organization view sorting by creation date felt more like a choice of that particular page in the UI to sort that way. But I could be swayed to make it consistent either way.

client/src/components/SupportTools/SupportTools.tsx
```diff
@@ -222,6 +239,10 @@ const Organization = ({ organizationId }: { organizationId: string }) => {
 
   const { name, defaultState, elections, auditAdmins } = organization.data
 
+  const sortedElections = elections.sort(
+    (a, b) => new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime()
+  )
```
jonahkagan, line 244: We have `sortBy` util function that's an option, if you'd like

server/api/support.py
jonahkagan: This feels fine to me!

I do think you could potentially simplify this query now that we're joining on `ActivityLogRecord` to also do the filtering after the join, rather than in a subquery. I don't think it will make a huge difference but might just end up a tad simpler

client/src/components/SupportTools/support-api.tsx
```diff
@@ -24,6 +24,8 @@ export interface IElectionBase {
     | 'HYBRID'
   online: boolean
   deletedAt: string | null
+  createdAt: string
+  currentRound: IRound | null
```
jonahkagan, line 28: I think `IElectionBase` is used in a few different types, but you didn't add these fields to all of them (e.g. `IJurisdiction`), so I think this will end up being a bit inaccurate. Instead, I think you should just add these fields to the types that represent the response from the appropriate API endpoints

jonahkagan (approved):
Nice, looks great and should be really useful!

Issue #2054, opened 2024-11-12
Update no container message here for GA to point to the tool like shown in the second message


arsalansufi:
Update no container message here for GA to point to the tool like shown in the second message.

Issue automatically created from Slack feedback reported by ginny: 
https://votingworks.slack.com/archives/C01F2CH62TS/p1731426367622819?thread_ts=1731426367.622819&cid=C01F2CH62TS

PR #2056, opened 2024-11-12
Improve error message when using batch inventory but missing container

jonahkagan:
Fixes: https://github.com/votingworks/arlo/issues/2054

Follow up to https://github.com/votingworks/arlo/pull/2011 to also show a similar error message when the Container column is missing, not only when it contains an invalid value.

The previous message was just `Missing required column "Container"`, which wasn't clear enough.

### DIFF
--- a/server/api/ballot_manifest.py
+++ b/server/api/ballot_manifest.py
@@ -120,12 +120,14 @@ def process_ballot_manifest_file(
             AuditType.HYBRID,
         ]
 
-        is_counting_group_required = is_enabled_sample_extra_batches_by_counting_group(
-            jurisdiction.election
-        )
-
         columns = [
-            CSVColumnType(CONTAINER, CSVValueType.TEXT, required_column=False),
+            CSVColumnType(
+                CONTAINER,
+                CSVValueType.TEXT,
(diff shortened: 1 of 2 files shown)


server/api/ballot_manifest.py
```diff
@@ -125,11 +125,7 @@ def process() -> None:
         )
 
         columns = [
-            CSVColumnType(
-                CONTAINER,
-                CSVValueType.TEXT,
-                required_column=is_counting_group_required,
-            ),
+            CSVColumnType(CONTAINER, CSVValueType.TEXT, required_column=False),
```
kshen0, line 128: It might be slightly confusing to readers that we indicate the column isn't required, but later on enforce it in L158. Thoughts on allowing an error message override to be specified in the `CSVColumnType`? I could see it being overkill, so if you agree, maybe just add a comment here on why we mark `required_column=False` and enforce with special logic later.

kshen0 (approved):

server/api/ballot_manifest.py
jonahkagan: Good point, I'll add a comment!

PR #2057, opened 2024-11-12
Expect zipped CVR file uploads in audit admin view

jonahkagan:
We already changed the jurisdiction manager UI for uploading ES&S CVR files to accept zips rather than CSVs, but didn't also change this audit admin interface. (See #2012)

Had to manual test since there's not an easy way to automated test this that I know of.


carolinemodic (approved):
ugh good catch thanks

PR #2058, opened 2024-11-13
build(node): update from v16 to v22

eventualbuddha:
Updates NodeJS to the latest LTS release. This shouldn't have any effect on the production behavior of the application as NodeJS is only used for a dev server and bundling the JS assets in Arlo. Updating the `circleci/node` orb is primarily to include [the change to inspect `.node-version` files to determine which NodeJS to use](https://github.com/CircleCI-Public/node-orb/pull/218).


jonahkagan (approved):
Awesome!

PR #2060, opened 2024-11-13
reverse order of elections on support org page

carolinemodic:
Accidently reversed the order here when I updated the PR to use sortBy, this fixes it and makes most recently created at the top of the list


jonahkagan (approved):
If there's an easy test to cover this, might be nice

PR #2061, opened 2024-11-13
Show last login time on audit progress

kshen0:
Addresses https://github.com/votingworks/arlo/issues/1512

* Adds an endpoint to query for last login time by jurisdiction
* Updates status badges on audit progress screen
* Updates jurisdiction color on map

**Login status badges**

At least one round has started:
Screenshot 2024-11-14 at 2 17 32 PM

No rounds started:
Screenshot 2024-11-14 at 2 17 08 PM

**Login detail view**
Last login data available (including but not limited to `LOGGED IN` status):
Screenshot 2024-11-13 at 4 39 19 PM

No last login data available:
Screenshot 2024-11-13 at 4 39 26 PM


client/src/components/AuditAdmin/ActivityLog.tsx
```diff
@@ -73,12 +74,8 @@ const prettyAction = (activity: IActivity) => {
 const ActivityLog: React.FC = () => {
   const auth = useAuthDataContext()
   const user = auth && (auth.user as IAuditAdmin)
-  const organizations = useQuery<IOrganization[]>(
-    'orgs',
-    () => fetchApi(`/api/audit_admins/${user!.id}/organizations`),
-    { enabled: !!user }
-  )
-  if (!organizations.isSuccess) return null
```
kshen0, line 81: Factored out to use in `AuditAdminView` but later moved organization lookup to the backend. Still, seems like we prefer to organize query hooks outside components so I'm inclined to keep.

server/api/jurisdictions.py
```diff
@@ -535,6 +537,64 @@ def status(jurisdiction_id: str) -> JurisdictionStatus:
...
+        .all()
+    )
+
+    # Get the most recent login activity for each jurisdiction
+    last_login_by_jurisdiction: Dict[str, ActivityLogRecord] = {}
+    for activity, jurisdiction_id in activities:
+        if (jurisdiction_id not in last_login_by_jurisdiction) or (
+            activity.timestamp > last_login_by_jurisdiction[jurisdiction_id].timestamp
+        ):
+            last_login_by_jurisdiction[jurisdiction_id] = activity
```
kshen0, line 587: It's possible we could bake this postprocessing into the query, but I'm not sure it would be more readable

client/src/components/AuditAdmin/Progress/JurisdictionDetail.tsx
```diff
@@ -125,6 +128,16 @@ const JurisdictionDetail: React.FC<IJurisdictionDetailProps> = ({
             auditSettings={auditSettings}
           />
         )}
+        {lastLoginActivity && (
+          <>
+            <H5>Last Login</H5>
+            <span>
+              {`${lastLoginActivity.user?.key || 'Unknown'} at ${new Date(
```
jonahkagan, line 135: Is it possible for `user` to be undefined? It feels like if a user logged in, there should be a user defined...

client/src/components/AuditAdmin/Progress/Progress.tsx
```diff
@@ -172,20 +178,18 @@ const Progress: React.FC<IProgressProps> = ({
...
-          case JurisdictionProgressStatus.UPLOADS_NOT_STARTED:
-            return (
-              <Status>
-                {auditType === 'BALLOT_POLLING'
-                  ? 'No manifest uploaded'
-                  : filesUploadedText}
-              </Status>
-            )
+          case JurisdictionProgressStatus.UPLOADS_NOT_STARTED_LOGGED_IN:
+            return <Status intent="warning">Logged in</Status>
```
jonahkagan, line 182: I wonder if we should make this a different color to differentiate from in-progress status (specifically thinking about the map, where color is more important). I don't think it's a blocking concern though, so happy to ship this and then reconsider it later

client/src/components/useJurisdictions.ts
```diff
@@ -144,6 +157,24 @@ export const useJurisdictions = (
   })
 }
 
+export const jurisdictionsWithLastLoginQueryKey = (
+  electionId: string
+): string[] => ['elections', electionId, 'jurisdictions', 'last-login']
```
jonahkagan, line 162: I think we should consider using the pattern established below where we concat to the end of ` jurisdictionsQueryKey(electionId)`. That way we won't need to explicitly invalidate this query, since it will be invalidated whenever the jurisdictions query is invalidated (since invalidating a query key will invalidate any query whose key starts with the invalidated key)

server/api/jurisdictions.py
```diff
@@ -10,6 +10,8 @@
 from sqlalchemy.orm import joinedload
 from werkzeug.exceptions import Conflict, BadRequest
 
+from server.api.activity import serialize_activity
```
jonahkagan, line 13: Use relative imports for consistency

server/api/jurisdictions.py
```diff
@@ -535,6 +537,64 @@ def status(jurisdiction_id: str) -> JurisdictionStatus:
...
+    # Look for events after most recent round started, or election creation time
+    # if no rounds exist yet
+    query_timestamp_after = election.created_at
+    round = (
+        Round.query.filter_by(election_id=election.id)
+        .order_by(Round.created_at.desc())
+        .one_or_none()
+    )
+    if round is not None:
+        query_timestamp_after = round.created_at
```
jonahkagan, line 563: Consider using `get_current_round(election)`

```py
current_round = get_current_round(election)
query_timestamp_after = current_round.created_at if current_round else election.created_at
```

server/api/jurisdictions.py
```diff
@@ -535,6 +537,64 @@ def status(jurisdiction_id: str) -> JurisdictionStatus:
...
+
+    # Query for login activities from users who are related to jurisdictions we found earlier
+    activities = (
+        ActivityLogRecord.query.join(
+            User, ActivityLogRecord.info["base"]["user_key"].astext == User.email
+        )
+        .join(User.jurisdictions)
+        .filter(
+            Jurisdiction.id.in_(jurisdiction_ids),
```
jonahkagan, line 572: It might be a bit more efficient to just filter by the jurisdiction's `election_id` field

```py
.join(User.jurisdictions)
.filter(Jurisdiction.election_id == election.id)
```

server/api/jurisdictions.py
jonahkagan: I think you can do it like so:
```py
 .order_by(Jurisdiction.id, ActivityLogRecord.timestamp.desc())
 .distinct(Jurisdiction.id)
 .with_entities(Jurisdiction.id, ActivityLogRecord)
```

This is nice because then you can just wrap the whole query expression with `dict()` and that will be `last_login_by_jurisdiction`

server/tests/api/test_jurisdictions.py
```diff
@@ -1,8 +1,15 @@
...
+from datetime import datetime, timedelta
 from typing import List
 from flask.testing import FlaskClient
 
+from server.activity_log.activity_log import (
+    ActivityBase,
+    JurisdictionAdminLogin,
+    record_activity,
+)
+from server.auth.auth_routes import record_login
```
jonahkagan, line 11: Use relative imports

client/src/components/AuditAdmin/Progress/Progress.tsx
```diff
@@ -172,20 +178,18 @@ const Progress: React.FC<IProgressProps> = ({
...
           case JurisdictionProgressStatus.AUDIT_IN_PROGRESS:
             return <Status intent="warning">In progress</Status>
           case JurisdictionProgressStatus.AUDIT_COMPLETE:
             return <Status intent="success">Complete</Status>
-          case JurisdictionProgressStatus.AUDIT_NOT_STARTED:
-            return <Status>Not started</Status>
+          case JurisdictionProgressStatus.AUDIT_NOT_STARTED_LOGGED_IN:
+            return <Status intent="warning">Logged in</Status>
+          case JurisdictionProgressStatus.AUDIT_NOT_STARTED_NO_LOGIN:
+            return <Status>Not logged in</Status>
```
jonahkagan, line 192: Need to make sure the new added statuses sort correctly by updating `sortType` below

server/api/jurisdictions.py
kshen0: Sorry, I need to update my IDE autocomplete config to not do this

jonahkagan: mine does it too lol. let me know if you figure out how to change it

client/src/components/AuditAdmin/Progress/JurisdictionDetail.tsx
kshen0: `IActivity` ([src](https://github.com/votingworks/arlo/pull/2061/files#diff-79e5cec27ff845c276566983488e2e282ca8a908829e0575d3fad9f14ec48ef1R10)) doesn't have subtypes for different activities so `activity.user` is nullable, but you're right we'll never hit this fallback in practice. Removed the fallback in [ec442b4](https://github.com/votingworks/arlo/pull/2061/commits/ec442b44d0d0bde29574ee57ebbaf9a5209aa770)

We could extend these types in a follow up PR if that level of granularity is helpful in other places.

client/jest.config.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
     // this number should only be getting closer to 0).
     global: {
       lines: -161,
-      branches: -194,
+      branches: -195,
```
kshen0, line 15: Looking at the detailed coverage report, I couldn't find a single uncovered branch in any lines included in this diff.

server/api/jurisdictions.py
kshen0: https://github.com/votingworks/arlo/pull/2061/commits/1552349d670efb58c45844de06fda8cea005718d

client/src/components/AuditAdmin/ActivityLog.tsx
```diff
@@ -73,12 +74,8 @@ const prettyAction = (activity: IActivity) => {
...
+  const organizations = useAuditAdminsOrganizations(user)
+  if (!(organizations && organizations.isSuccess)) return null
```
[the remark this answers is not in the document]
kshen0: [36b6db1](https://github.com/votingworks/arlo/pull/2061/commits/36b6db12a193d21f519a8d6da7d0b5ee601cf983)

server/api/jurisdictions.py
kshen0: https://github.com/votingworks/arlo/pull/2061/commits/1552349d670efb58c45844de06fda8cea005718d#diff-fc2514bfe998fc28ba94dbd7860ae951626cf2ddb9a6aeb3bbd29f7e67cd4cdaR547

server/api/jurisdictions.py
kshen0: https://github.com/votingworks/arlo/pull/2061/commits/1552349d670efb58c45844de06fda8cea005718d#diff-fc2514bfe998fc28ba94dbd7860ae951626cf2ddb9a6aeb3bbd29f7e67cd4cdaR559

server/tests/api/test_jurisdictions.py
kshen0: https://github.com/votingworks/arlo/pull/2061/commits/8e891e40947f71caadcb91be095ba59cf83ac4e4

client/src/components/AuditAdmin/Progress/Progress.tsx
```diff
@@ -195,30 +195,48 @@ const Progress: React.FC<IProgressProps> = ({
...
+          cvrs,
+        } = jurisdiction
+        const progressStatus = getJurisdictionStatus(
+          jurisdiction,
+          lastLoginQuery.data![jurisdiction.id]
+        )
+        const hasLoggedIn = ![
+          JurisdictionProgressStatus.UPLOADS_NOT_STARTED_NO_LOGIN,
+          JurisdictionProgressStatus.AUDIT_NOT_STARTED_NO_LOGIN,
+        ].includes(progressStatus)
```
kshen0, line 212: We could check login status directly but that results in a weaker tie between `JurisdictionProgressStatus` and the sort order. Since we ultimately use `JurisdictionProgressStatus` to decide what status badge text to render it seemed important that `JurisdictionProgressStatus` is directly responsible for sort order of status badges.

client/src/components/AuditAdmin/Progress/Progress.tsx
```diff
@@ -195,30 +195,48 @@ const Progress: React.FC<IProgressProps> = ({
...
+          )
+          if (anyFailed) return hasLoggedIn ? 0 : -1
+          if (numComplete === 0) return hasLoggedIn ? -1 : -2
+          return numComplete
         }
-      ),
+        return {
+          [JurisdictionRoundStatus.NOT_STARTED]: () => {
+            if (hasLoggedIn) {
+              return 0
```
kshen0, line 232: This preserves the behavior that the following cases are weighted equally when a login exists:

1. Round not started, uploads attempted but failed
2. Round started but no action taken in audit

IMO case 2 should have higher sort weight than case 1, but I don't have context on the previous decision to weight them equally.

client/src/components/useAuditAdminsOrganizations.ts
```diff
@@ -0,0 +1,15 @@
+import { useQuery, UseQueryResult } from 'react-query'
+import { fetchApi, ApiError } from '../utils/api'
+import { IAuditAdmin, IOrganization } from './UserContext'
+
+function useAuditAdminsOrganizations(
+  user: IAuditAdmin | null
+): UseQueryResult<IOrganization[], ApiError> {
+  return useQuery<IOrganization[], ApiError>(
+    'orgs',
+    () => fetchApi(`/api/audit_admins/${user?.id}/organizations`),
```
arsalansufi, line 10: nit: I see that the original code used to assert that the `user` is defined, e.g., `user!.id`, I presume because of the `{ enabled: !!user }` gate. Slight preference for the `!` over the `?` for clarity since user should never be undefined, unless I'm missing something

server/api/jurisdictions.py
```diff
@@ -535,6 +536,45 @@ def status(jurisdiction_id: str) -> JurisdictionStatus:
...
+    "/election/<election_id>/jurisdictions/last-login",
+    methods=["GET"],
+)
+@restrict_access([UserType.AUDIT_ADMIN])
+def get_last_login_by_jurisdiction(election: Election):
+    # Look for events after most recent round started, or election creation time
+    # if no rounds exist yet
+    current_round = get_current_round(election)
+    query_timestamp_after = (
+        current_round.created_at if current_round else election.created_at
```
arsalansufi, line 549: Nice! This covers one of the main things that I wanted to check, that we reset this for every phase of the audit

server/api/jurisdictions.py
```diff
@@ -535,6 +536,45 @@ def status(jurisdiction_id: str) -> JurisdictionStatus:
...
+            Jurisdiction.election_id == election.id,
+            ActivityLogRecord.organization_id == election.organization_id,
+            ActivityLogRecord.timestamp > query_timestamp_after,
+            ActivityLogRecord.activity_name == "JurisdictionAdminLogin",
+        )
+        .order_by(Jurisdiction.id, ActivityLogRecord.timestamp.desc())
+        .distinct(Jurisdiction.id)
+        .with_entities(Jurisdiction.id, ActivityLogRecord)
+        .all()
+    )
```
arsalansufi, line 568: Will be curious to see how performant this query is in prod - I can't quite recall which indexes exist on the ActivtyLogRecord table

client/src/components/AuditAdmin/Progress/JurisdictionDetail.tsx
```diff
@@ -125,6 +128,16 @@ const JurisdictionDetail: React.FC<IJurisdictionDetailProps> = ({
             auditSettings={auditSettings}
           />
         )}
+        {lastLoginActivity && (
```
arsalansufi, line 131: What happens if the only login thus far was by a support user logging in as someone else? Will we still render a last login in that case? My gut feeling is that we should filter those out. I think that has to happen on the backend as the frontend only gets the very last login, which may have been a support user but there may have been a real login just before

I'm okay pushing that to a follow-up next week if need be!

client/src/components/AuditAdmin/Progress/Progress.tsx
arsalansufi: Good q! I believe the context here is that round starts are audit-wide and apply across jurisdictions. It isn't possible for a round to be started for just a subset of jurisdictions

I might be misunderstanding what you're pointing out though

A comment describing the complete hierarchy here could be helpful! I'm having a bit of a hard time parsing (already was with the prior code and we've understandably had to add even more logic)

arsalansufi (approved):
LGTM! One follow-up (I think) that we don't need to block on but does feel relevant to address while context is fresh

client/src/components/AuditAdmin/Progress/Progress.tsx
kshen0: That answers my question, thanks. Comment and tests for this behavior added in https://github.com/votingworks/arlo/pull/2061/commits/2e277b6fc66f4f04ebbc77e33738718ae43e9caa

client/jest.config.js
jonahkagan: This has been happening to me lately as well. Can't figure it out!

client/src/components/AuditAdmin/Progress/JurisdictionDetail.tsx
jonahkagan: IIRC, we don't record an activity when a support user logs in as a JM or AA. I can't recall why, but my guess is that we thought it would be too distracting to see in the activity log.

server/api/jurisdictions.py
jonahkagan: Only the primary key index on ID, so this is likely going to be a seq scan

client/src/components/AuditAdmin/AuditAdminView.tsx
```diff
@@ -48,16 +52,19 @@ const AuditAdminView: React.FC = () => {
         isDrawSampleComplete(rounds)
       ) {
         queryClient.invalidateQueries(jurisdictionsQueryKey(electionId))
+        queryClient.invalidateQueries(
+          jurisdictionsWithLastLoginQueryKey(electionId)
+        )
```
jonahkagan, line 57: I wasn't totally clear on this in my last comment, but we can remove all of the invalidations for `jurisdictionsWithLastLoginQueryKey` because they are covered by `jurisdictionsQueryKey`.

kshen0: Sorry that was clear. It just got lost in the other updates. I'll open a PR.

PR #2070, opened 2024-11-18
Truncate long error messages in Slack activity notifications

jonahkagan:
Otherwise we get a `msg_too_long` error. I based the limit of 4000 characters on the Slack API docs: https://api.slack.com/methods/chat.postMessage#truncating


arsalansufi (approved):
Good find!

PR #2076, opened 2024-11-20
Fix audit report sampled batch accounting for combined extra batches

jonahkagan:
When using combined batches and the sample extra batches feature flag, there were some bugs with how we counted the sampled batches (the extra non-RLA batches weren't being properly filtered out).


server/tests/batch_comparison/snapshots/snap_test_sample_extra_batches_by_counting_group.py
```diff
@@ -37,3 +37,32 @@
...
+Contest Name,Targeted?,Number of Winners,Votes Allowed,Total Ballots Cast,Vote Totals,Vote Totals from Batches,Pending Ballots\r
+Contest 1,Targeted,1,2,5000,candidate 1: 5000; candidate 2: 2500; candidate 3: 2500,candidate 1: 5000; candidate 2: 2500; candidate 3: 2500,0\r
+\r
+######## AUDIT SETTINGS ########\r
+Audit Name,Audit Type,Audit Math Type,Risk Limit,Random Seed,Online Data Entry?\r
+Test Audit test_sample_extra_batches_with_combined_batches[TEST-ORG/sample-extra-batches-by-counting-group],BATCH_COMPARISON,MACRO,10%,1234567890,No\r
+\r
+######## ROUNDS ########\r
+Round Number,Contest Name,Targeted?,Sample Size,Risk Limit Met?,P-Value,Start Time,End Time,Audited Votes,Batches Sampled,Ballots Sampled,Reported Votes\r
+1,Contest 1,Targeted,1,No,0.6666666667,DATETIME,DATETIME,candidate 1: 1200; candidate 2: 650; candidate 3: 650,1,1250,candidate 1: 1200; candidate 2: 650; candidate 3: 650\r
```
jonahkagan, line 57: This line is the important part of the test snapshot. In particular, the value `1` for the `Batches Sampled` column, since there was only one sampled batch and we made it part of a combined batch. This column had value `-2` with the previous code before this change.

server/api/reports.py
```diff
@@ -1082,7 +1090,7 @@ def sampled_batch_rows(election: Election, jurisdiction: Jurisdiction = None):
                         sampler_contest.from_db_contest(contest),
                         0,
                     )
-                    if is_audited and audit_results and not is_combined
+                    if audit_results
```
jonahkagan, line 1093: Unrelated to the change, but adding a test caught the fact that `is_combined` isn't actually defined on this line. It was a copy/paste error that never manifested due to short-circuiting (I assume). Any reasonable type-checking/linting should have caught this, so I'm a bit annoyed

arsalansufi: Dang good catch - I've been wondering whether some version upgrades would give us better Python typechecking

server/api/reports.py
```diff
@@ -479,27 +479,35 @@ def round_rows(election: Election):
...
                 .all()
             )
-            combined_batches = group_combined_batches(combined_sub_batches)
-
-            num_sampled_batches = (
-                len(sampled_batches)
-                # Instead of counting sub batches that got combined, count the
-                # combined batches
-                - len(combined_sub_batches)
-                + len(combined_batches)
```
jonahkagan, line 495: This was wrong in a few ways:
- It counted combined batches that had no RLA sampled batches in them (e.g. combined batches with only extra sampled batches and unsampled batches)
- It subtracted unsampled sub batches (that are included in `len(combined_sub_batches)`) rather than only the sampled sub batches

You can see the impact of the latter in the test snapshot update for test_batch_comparison.py, which has nothing to do with extra batch sampling.

arsalansufi (approved):

server/api/reports.py
arsalansufi: Thanks for the clarification!

arsalansufi (approved):
Definitely worth exploring after audit season whether there's any way to simplify here, since the interplay of all these situations is getting hard to reason about

Maybe unavoidable but worth revisiting!

Issue #1854, opened 2023-11-29
Support tool button that allows the download of a list of audit admins/jurisdiction managers by org


carolinemodic:
I need a support tool button that allows me to download a list of audit admins by org.

Issue automatically created from slack feedback reported by ginny 
https://votingworks.slack.com/archives/C01F2CH62TS/p1694529745833549?thread_ts=1694529745.833549&cid=C01F2CH62TS

ginvdr:
Sending communications to audit admins or customer service surveys to JMs would be the primary use for this list. JMs would also need some information with a date stamp.  For example, all JM's with audits from x date to x date.

PR #2077, opened 2024-11-21
Add tool for support users to export Audit Admins and Jurisdiction Managers

kshen0:
https://github.com/votingworks/arlo/issues/1854

Adds a tool to the bottom of the Organizations column on the support page to export a csv of Audit Admins and Jurisdiction Managers.

Screenshot 2024-11-21 at 5 27 13 PM

### DIFF
--- a/client/src/components/SupportTools/SupportTools.tsx
+++ b/client/src/components/SupportTools/SupportTools.tsx
@@ -81,6 +81,7 @@ const SupportTools: React.FC = () => {
               <Row>
                 <ActiveAudits />
                 <Organizations />
+                <SupportUserTools />
               </Row>
             </Route>
             <Route path="/support/orgs/:organizationId">
@@ -111,9 +112,6 @@ const Column = styled.div`
 const Row = styled.div`
   display: flex;
   width: 100%;
-`
-const DownloadUserListButton = styled(AnchorButton)`
-  margin: 10px 0;
 `
 
 const AuditStatusTag = ({ currentRound }: { currentRound: IRound | null }) => {
@@ -206,15 +204,27 @@ const Organizations = () => {
           </LinkItem>
         ))}
       </List>
-      <DownloadUserListButton
+    </Column>
+  )
+}
+
+const DownloadUsersButton = styled(AnchorButton)`
+  margin-bottom: 10px;
+`
+
+const SupportUserTools = () => {
+  return (
+    <Column>
+      <H2>Downloads</H2>
+      <DownloadUsersButton
         icon="download"
         href="/api/support/organizations/users"
       >
         Download User List
-      </DownloadUserListButton>
+      </DownloadUsersButton>
       <p>
-        Export a list of Audit Admins and Jurisdiction Managers for recently
-        completed audits for all organizations.
+        Export a list of Audit Admins and Jurisdiction Managers for all audits
+        completed in the last 12 weeks.
       </p>
     </Column>
   )
--- a/client/src/components/SupportTools/SupportTools.tsx
+++ b/client/src/components/SupportTools/SupportTools.tsx
@@ -112,6 +112,9 @@ const Row = styled.div`
   display: flex;
   width: 100%;
 `
+const DownloadUserListButton = styled(AnchorButton)`
+  margin: 10px 0;
+`
 
 const AuditStatusTag = ({ currentRound }: { currentRound: IRound | null }) => {
   if (!currentRound) {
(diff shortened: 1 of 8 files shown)


server/api/support.py
```diff
@@ -133,6 +141,68 @@ def list_organizations():
...
+            # pylint: disable=no-member
+            .subqueryload(Jurisdiction.jurisdiction_administrations).subqueryload(
+                JurisdictionAdministration.user
+            ),
+            subqueryload(Election.organization)
+            # pylint: disable=no-member
+            .subqueryload(Organization.audit_administrations).subqueryload(
+                AuditAdministration.user
+            ),
+        )
```
kshen0, line 168: This query is fairly unoptimized but runs locally in < 250ms for all audits in the last year.

client/src/components/SupportTools/SupportTools.tsx
```diff
@@ -203,6 +206,16 @@ const Organizations = () => {
...
+      <DownloadUserListButton
+        icon="download"
+        href="/api/support/organizations/users"
+      >
+        Download User List
+      </DownloadUserListButton>
+      <p>
+        Export a list of Audit Admins and Jurisdiction Managers for recently
+        completed audits for all organizations.
+      </p>
```
kshen0, line 218: the `Organizations` list seemed like the best place to put this since the users do not strictly relate to active audits. If we keep adding miscellaneous support tools that operate at the scope of all audits/organizations we may want to break out a section specifically for those tools.

server/api/support.py
```diff
@@ -133,6 +141,68 @@ def list_organizations():
...
+        .all()
+    )
+
+    for election in elections_with_users:
+        if is_audit_complete(get_current_round(election)):
```
kshen0, line 173: Felt risky to encode this logic in the query when it might change. Of the 9 audits found, here are the runtimes for this check (which involves 1 db query):

0.0057 seconds
0.0015 seconds
0.0016 seconds
0.0037 seconds
0.0016 seconds
0.0050 seconds
0.0015 seconds
0.0016 seconds
0.0014 seconds

client/src/components/SupportTools/SupportTools.tsx
```diff
@@ -203,6 +206,16 @@ const Organizations = () => {
...
         ))}
       </List>
+      <DownloadUserListButton
+        icon="download"
+        href="/api/support/organizations/users"
+      >
+        Download User List
+      </DownloadUserListButton>
+      <p>
+        Export a list of Audit Admins and Jurisdiction Managers for recently
```
nikhilb4a, line 216: I wonder if instead of `recently` it should explicitly say `within the last 12 weeks`, so there is no ambiguity if someone is confused why an audit from 3.5 months back isn't showing

client/src/components/SupportTools/SupportTools.tsx
nikhilb4a: Yeah I agree. It feels a bit odd having it be at the bottom of that list, but in the current product `Organizations` does seem to be the closest association. That being said, I wonder if we should just add a new section now, as it looks slightly out of place, and we run the risk of the next person adding a support tool without creating a new section in the name of following the existing pattern. But, just a gentle nudge, I don't feel strongly

server/api/support.py
nikhilb4a: What do you mean when you say `it might change`?

nikhilb4a: Also, just checking - how come we only want completed audits and not active audits?

nikhilb4a (approved):
Looks great! Just a few clarification questions

client/src/components/SupportTools/SupportTools.tsx
kshen0: Good point, updated in [0be290c](https://github.com/votingworks/arlo/pull/2077/commits/0be290c47651edf8421bdd57b53177bee6ac8d46)

client/src/components/SupportTools/SupportTools.tsx
kshen0: Added its own section in [0be290c](https://github.com/votingworks/arlo/pull/2077/commits/0be290c47651edf8421bdd57b53177bee6ac8d46). Given this is expected to be used on a computer screen (not tablet) and the potential for the existing columns to stretch pretty far down the page, this seemed like a better place to put it than in another row after "Active Audits" and "Organizations".

This probably deserves a redesign at some point eg. a separate tab

server/api/support.py
kshen0: The use case for this tool is to reach out for feedback from jurisdiction managers and audit admins after audits have completed

kshen0: By "it might change" I mean, as with any function, the logic that determines whether an audit is complete might be updated in the future. I mention this because I don't like doing one db call per election in post-processing, but the alternative is to reimplement the logic in the query above. In the event that `is_audit_complete` is updated later, the query in this endpoint will behave differently than expected.

client/src/components/SupportTools/SupportTools.tsx
nikhilb4a: Ooo nice, yeah I like that this takes advantage of the screen width!

server/api/support.py
nikhilb4a: Ah I see, right. That makes sense to me

PR #2079, opened 2024-11-26
Modify parse-xml-cvrs to support ask from Clark County

carolinemodic:
Updates a script used to parse xml cvrs into dominion csv form to
- If provided parse scanned ballot information files and adds "Workstation" column to the outputted csv, in the future if we want to add other columns from the scanned ballot info that would be easy. This option also assumes the xml cvr files are in a subdirectory within an outer directory containing that folder and the scanned ballot information files. 
- If run with --include-votes-cast-per-contest will add a column for every contest with the total votes cast in each contest for that cvr row. This will allow a human interpretting the csv to determine which cvrs represent undervotes or overvotes without needing to understand the number of votes allowed to be cast per contest in the context of the script 

In order to filter cvrs to rows that represent undervotes in a specific contest I have a hack on top of this script on the following branch: [caro/script_clark_co_hart_parsing_filter_contest_undervote](https://github.com/votingworks/arlo/compare/caro/script_clark_co_hart_parsing...caro/script_clark_co_hart_parsing_filter_contest_undervote?expand=1) . That feels like a specific enough part of this ask to not merge to main.

### DIFF
--- a/scripts/parse-xml-cvrs.py
+++ b/scripts/parse-xml-cvrs.py
@@ -5,6 +5,14 @@ import csv
 from xml.etree import ElementTree
 from collections import defaultdict
 
+from server.util.csv_parse import get_header_indices, column_value
+from server.api.cvrs import read_ess_ballots_file
+
+# This script that parses hart CVRS and outputs a CSV file similar to the dominion format.
+# Run with:
+# FLASK_ENV=development poetry run python -m scripts.parse-xml-cvrs <path/to/hart/cvrs> <output-file.csv> [--include-votes-cast-per-contest] [--cvrs-include-scanned-ballot-info] [--cvrs-exported-by-tabulator]
+##
+
 # Annoyingly, ElementTree requires that you specify the namespace in all tag
 # searches, so we make some wrapper functions
 ns = "http://tempuri.org/CVRDesign.xsd"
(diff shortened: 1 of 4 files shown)


scripts/parse-xml-cvrs.py
```diff
@@ -18,13 +26,37 @@ def findall(xml, tag):
...
 def get_directory_name(file_path):
     directory_path = os.path.dirname(file_path)
     directory_name = os.path.basename(directory_path)
     return directory_name
 
 
-def parse_cvr_file(file_path, use_directory_name_as_tabulator=False):
+def parse_scanned_ballot_file(file_path, cvr_workstation_mapping):
+    with open(file_path, "r", encoding="utf-8") as ballots_file:
+        headers, rows = read_ess_ballots_file(ballots_file)
```
arsalansufi, line 40: Oh interesting that this ESS ballots file parsing function works on this file! This file is technically not an ESS ballots file but a Hart scanned ballot information file. Looks like they're similar enough in structure that the same function works. But to avoid confusion, we should probably use this function instead:

https://github.com/votingworks/arlo/blob/b00d3f64ef627ee2ea9b6b63de8ec903fd99eab6/server/api/cvrs.py#L1079-L1081

arsalansufi (commented):
Thanks for tackling this! Just one thought

scripts/parse-xml-cvrs.py
carolinemodic: facepalm lol

PR #2080, opened 2024-11-26
Consolidate discrepancy visibility logic on server

nikhilb4a:
We recently saw an issue where discrepancies were surfaced differently in the report vs in UI due to the visibility logic being split between client and server. This PR consolidates the visibility logic on the server.

Ballot comparison:
https://github.com/user-attachments/assets/081fdb6c-8a7a-438a-9af1-7d525146280a

Batch comparison:
https://github.com/user-attachments/assets/79e9db46-6bca-4855-adf2-e8b5445a5a6b


server/api/discrepancies.py
```diff
@@ -195,32 +229,61 @@ def get_ballot_comparison_audit_discrepancies_by_jurisdiction(
...
-            vote_deltas = ballot_vote_deltas(contest, reported_cvr, audited_cvr)
-            if not vote_deltas or isinstance(vote_deltas, str):
-                continue
-
-            # CVRs are guaranteed to be non-null due to ballot_vote_deltas() checks
-            assert isinstance(audited_cvr, dict)
-            audited_votes = audited_cvr.get(contest.id, {})
-            assert isinstance(reported_cvr, dict)
-            reported_votes = reported_cvr.get(contest.id, {})
             if ballot_id in sampled_ballot_id_to_jurisdiction_id:
```
nikhilb4a, line 240: The core diff for this block is really just lines 240 to 242 of the updated code. The `if ballot_id in sampled_ballot_id_to_jurisdiction_id:` was moved higher up as I needed `jurisdiction_id` to index `show_discrepancies_by_jurisdiction`.

That being said, it seems like it would be safe to drop that check, as I think `jurisdiction_id` not being present in `sampled_ballot_id_to_jurisdiction_id` would be a data integrity issue. Does that sound right to you?

jonahkagan: Hmm, it looks like `ballot_id` is coming from `sampled_ballot_interpretations_to_cvrs`, which by my read doesn't filter by round, so it could include ballots from earlier rounds that aren't in this round. `sampled_ballot_id_to_jurisdiction_id` only includes ballots sampled in this round. So that might create a case where `ballot_id` is not in `sampled_ballot_id_to_jurisdiction_id`.

But I could be misreading - could try removing it to see if a test fails (if it's a necessary guard, we should cover it with a test)

server/api/discrepancies.py
```diff
@@ -195,32 +229,61 @@ def get_ballot_comparison_audit_discrepancies_by_jurisdiction(
...
+) -> Dict[str, bool]:
+    jurisdictions = list(election.jurisdictions)
+    is_single_jurisdiction_election = len(jurisdictions) == 1
+    if is_single_jurisdiction_election:
+        jurisidiction_id = jurisdictions[0].id
+        return {jurisidiction_id: True}
+
+    rounds = list(election.rounds)
+    audit_board_status_by_jurisdiction = jurisdiction_audit_board_status(
+        jurisdictions, rounds[-1]
```
jonahkagan, line 281: Convention is to use `get_current_round` rather than manipulate the rounds list manually

jonahkagan (approved):
Thanks for cleaning this up!

Given that no tests needed to be changed, I think that means we're missing test assertions that ensure that discrepancies are hidden in the /discrepancy endpoint until the jurisdiction is finalized/signed-off. So we should probably add those.

server/api/discrepancies.py
nikhilb4a: Oh right. I forgot that `sampled_ballot_interpretations_to_cvrs` includes ballots from other rounds. Thank you

nikhilb4a: Added another call to `/discrepancies` at the end of `test_ballot_comparison_two_rounds` to test that code path. Now when we hit the endpoint after round 2 is completed, we skip over the round 1 audit ballots, as expected.

PR #2083, opened 2024-12-02
update node cache

carolinemodic:


jonahkagan:
I think this may be mitigated by #2084, so closing for now

Issue #2065, opened 2024-11-15
Audit progress last login time polish


kshen0:
* Differentiate "Logged in" from "In progress" color https://github.com/votingworks/arlo/pull/2061/files#r1842991530
* Query for and show failed logins as well

PR #2089, opened 2024-12-04
differentiate logged-in status from in-progress status

kshen0:
https://github.com/votingworks/arlo/issues/2065

Updates `Logged in` status to be a different color than the various `In progress` statuses. The latter is now `GOLD1`/`GOLD4` in [Blueprint](https://blueprintjs.com/docs/#core/colors) upon which Arlo's design system is based. `Logged in` used to use the `warning` `Intent` that `In progress` also uses. `warning` uses orange per [Blueprint docs](https://blueprintjs.com/docs/#core/colors).

I didn't want to get into changing the vendored design system so I added a non-standard `alert` intent on the Arlo component that needs it. The non-standard intent is filtered out so we don't pass it as an invalid prop to the underlying Blueprint components.

Screenshot 2024-12-03 at 3 57 40 PM

### DIFF
--- a/client/src/components/Atoms/StatusTag.tsx
+++ b/client/src/components/Atoms/StatusTag.tsx
@@ -13,10 +13,10 @@ const StyledTag = styled(({ hasProgressBar: _, ...props }) => (
   text-transform: uppercase;
   font-weight: 500;
   ${p =>
-    p.intent === 'alert' &&
-    // Gold 4 in RGBA
-    `background-color: rgba(240, 183, 38, 0.2);
-     color: ${Colors.GOLD1}`}
+    p.intent === 'in-progress' &&
+    // Cobalt 4 in RGBA
+    `background-color: rgba(69, 128, 230, 0.2);
+     color: ${Colors.COBALT1}`}
   ${props =>
     props.hasProgressBar &&
     `border-bottom-left-radius: 0;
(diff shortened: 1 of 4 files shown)


jonahkagan:
I think the gold and orange are too similar to be reliably distinguished.

kshen0:
> I think the gold and orange are too similar to be reliably distinguished.

The [options](https://blueprintjs.com/docs/#core/colors) if staying within the design system are somewhat slim. There's a learned pattern that some colors don't come before orange when denoting status (eg. green). IMO yellow is the natural choice to precede orange but I see how it could be difficult to tell them apart.

I ruled out blue and red for any status because of their connotations with electoral college maps. We could do something like purple or indigo for "Logged in", which would make the progression:
`white > grey > purple > orange > green`

Purple feels distinct enough from the other colors and not traditionally associated with status that it wouldn't be confusing to put purple before orange, though it might not be very intuitive either.

We could also make `Logged in == orange` and change `In progress` to cerulean, turqouise, or lime as an "almost green but not quite". I didn't want to change the meaning of the colors too much during peak audit season but it's probably safer to do that now.

kshen0:
Playing around with this more, the most logical thing seems to be changing `Logged in` to be orange and `In progress` to be one of blue or purple. The progress of `orange > blue > green` felt most understandable (purple felt like it was out of left field). Since we already use red for `Upload error`, I don't feel too bad about using blue for another status.

Screenshot 2024-12-03 at 6 11 46 PM

PR #2090, opened 2024-12-04
Attempt to fix Cypress CI caching issue

nikhilb4a:
We've been seeing issues the last couple weeks with the cached Cypress binary being found in the cache, supposedly restored, but then it is missing when Cypress runs.

We implemented a safe fail last week, where we check if Cypress is there, and then install it if not. This has unblocked development, but recent builds have been needing to use the failsafe, reinstalling Cypress on [main](https://app.circleci.com/jobs/github/votingworks/arlo/23455) and this recent [branch](https://app.circleci.com/pipelines/github/votingworks/arlo/7630/workflows/2936d6fa-1f68-49c2-978c-7dfacd1dbac2/jobs/23519) during the `Verify Cypress` step. 

1. This PR changes the cache path from using a `~` to indicate the home path, to explicitly using the home path of `home/circleci`. In a somewhat unrelated post, it was suggested that circle can have issues with the `~`, even though it is what they use in their documentation. I gave it a try, and I've seen Cypress consistently restore from cache now.
2. Maintains the `Verify Cypress` step, but now puts it before the `Save Cache` step, so that way when we do have to generate a new binary when the `package.json` changes, we automatically generate the new binary and cache it
3. Small cleanup of the CI file, including moving Workflows to the top. This is just stylistic preference, treating it kind of like a "table of contents" so it's clear what the main steps in the workflow are.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,7 +112,7 @@ jobs:
           name: "Saving Cypress cache"
           key: v1-cypress-cache-{{ checksum "client/yarn.lock" }}
           paths:
-            - ~/.cache/Cypress
+            - /.cache/Cypress
       - browser-tools/install-chrome
       - create-data-model
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,12 +42,10 @@ jobs:
       - python/install-packages:
           pkg-manager: poetry
       - create-data-model
-
       - run:
           name: typecheck server
           command: |
             make typecheck-server
-
       - run:
           name: format server
           command: |
@@ -88,7 +86,6 @@ jobs:
           command: |
             sudo apt-get update
             sudo apt-get install graphicsmagick
-
       - run:
           name: test client
           command: |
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -110,19 +116,16 @@ jobs:
...
-          key: v1-cypress-cache-{{ checksum "client/yarn.lock" }}
-      - save_cache:
-          name: "Saving Cypress cache"
-          key: v1-cypress-cache-{{ checksum "client/yarn.lock" }}
-          paths:
-            - ~/.cache/Cypress
-      # We've seen a weird issue where the Cypress cache appears to restore
-      # successfully but the Cypress binary is still missing. So we just check
-      # for it and reinstall if necessary.
+          key: v4-cypress-cache-{{ checksum "client/yarn.lock" }}
```
nikhilb4a, line 119: I borked v2 and v3 of the cache as I tested ;p

jonahkagan (approved):
So weird, hope this works!

PR #2092, opened 2024-12-09
Update Support Tools UI Homepage

nikhilb4a:
Updated with Fixed Padding

https://github.com/user-attachments/assets/d0cb0ea6-be43-46af-bd72-8fc75a91c2b8

Updated with Filter Input:

https://github.com/user-attachments/assets/36cd9445-4d3b-4a41-b033-2ca4347c00fb

Updated after PR feedback:

https://github.com/user-attachments/assets/95cd42f5-a993-4315-a2fa-aed159ac88d9

Outdated:

https://github.com/user-attachments/assets/9399c5ad-a0c3-44c6-b35a-d62016c980ce

Main changes:

1. Increase width of support tools pages. Next, I plan to improve the spacing of the Organization, Audit, and Jurisdiction sub-pages.
2. Make Active Audits the widest column. This is just preference for how the page looks with it. Also move the status tag to the right side.
3. Rename Downloads to Tools, move Create Org button there, disable button until input is dirty, add success toast. Add tooltip to Download User List.
4. Misc changes: remove extra wrapper divs, extra margin/padding, reduce H2s to H3s

### DIFF
--- a/client/src/components/Atoms/Wrapper.tsx
+++ b/client/src/components/Atoms/Wrapper.tsx
@@ -28,8 +28,8 @@ export const SupportToolsInner = styled.div`
   justify-content: center;
   margin-left: auto;
   margin-right: auto;
-  padding: 30px 0;
-  width: 80%;
+  padding: 30px 150px;
+  width: 100%;
 `
 
 export default Wrapper
--- a/client/src/components/Atoms/Wrapper.tsx
+++ b/client/src/components/Atoms/Wrapper.tsx
@@ -3,7 +3,7 @@ import styled from 'styled-components'
 export const Wrapper = styled.div`
   display: flex;
   flex-direction: column;
-  width: 85%;
+  width: 100%;
   padding-bottom: 30px;
 `
 
@@ -17,9 +17,19 @@ export const Inner = styled.div<IInnerProps>`
   margin-right: auto;
   margin-left: auto;
   width: 100%;
+  max-width: 1020px;
   padding: 0 30px;
   padding-top: ${props => (props.withTopPadding ? '30px' : undefined)};
   flex-direction: ${props => props.flexDirection || undefined};
+`
+
+export const SupportToolsInner = styled.div`
+  display: flex;
+  justify-content: center;
+  margin-left: auto;
+  margin-right: auto;
+  padding: 30px 0;
+  width: 80%;
 `
 
 export default Wrapper
--- a/client/src/components/Atoms/Wrapper.tsx
+++ b/client/src/components/Atoms/Wrapper.tsx
@@ -3,7 +3,7 @@ import styled from 'styled-components'
 export const Wrapper = styled.div`
   display: flex;
   flex-direction: column;
-  width: 100%;
+  width: 85%;
   padding-bottom: 30px;
 `
 
@@ -17,7 +17,6 @@ export const Inner = styled.div<IInnerProps>`
   margin-right: auto;
   margin-left: auto;
   width: 100%;
-  max-width: 1020px;
   padding: 0 30px;
   padding-top: ${props => (props.withTopPadding ? '30px' : undefined)};
   flex-direction: ${props => props.flexDirection || undefined};
--- a/client/src/components/Header.tsx
+++ b/client/src/components/Header.tsx
@@ -22,7 +22,7 @@ import LinkButton from './Atoms/LinkButton'
 const SupportBar = styled(Navbar)`
   background-color: ${Colors.ROSE3};
   height: 35px;
-  padding: 0 10%;
+  padding: 0 150px;
   color: ${Colors.WHITE};
   font-weight: 500;
   .bp3-navbar-group {
--- a/client/src/components/Header.tsx
+++ b/client/src/components/Header.tsx
@@ -22,7 +22,7 @@ import LinkButton from './Atoms/LinkButton'
 const SupportBar = styled(Navbar)`
   background-color: ${Colors.ROSE3};
   height: 35px;
-  padding: 0;
+  padding: 0 10%;
   color: ${Colors.WHITE};
   font-weight: 500;
   .bp3-navbar-group {
@@ -109,19 +109,17 @@ const Header: React.FC = () => {
     <>
       {auth && auth.supportUser && (
         <SupportBar>
-          <InnerBar>
-            <NavbarGroup align={Alignment.LEFT}>
-              <a href="/support">
-                <Icon icon="eye-open" />
-                <span>Arlo Support Tools</span>
-              </a>
-            </NavbarGroup>
-            <NavbarGroup align={Alignment.RIGHT}>
-              <span>{auth.supportUser.email}</span>
-              <NavbarDivider />
-              <a href="/auth/support/logout">Log out</a>
-            </NavbarGroup>
-          </InnerBar>
+          <NavbarGroup align={Alignment.LEFT}>
+            <a href="/support">
+              <Icon icon="eye-open" />
+              <span style={{ fontWeight: 600 }}>Arlo Support Tools</span>
+            </a>
+          </NavbarGroup>
+          <NavbarGroup align={Alignment.RIGHT}>
+            <span>{auth.supportUser.email}</span>
+            <NavbarDivider />
+            <a href="/auth/support/logout">Log out</a>
+          </NavbarGroup>
         </SupportBar>
       )}
       {!supportMatch && (
--- a/client/src/components/SupportTools/SupportTools.tsx
+++ b/client/src/components/SupportTools/SupportTools.tsx
@@ -1,4 +1,4 @@
-import React, { ChangeEvent, useState } from 'react'
+import React, { useState } from 'react'
 import { Redirect, Route, Switch, Link } from 'react-router-dom'
 import { toast } from 'react-toastify'
 import styled from 'styled-components'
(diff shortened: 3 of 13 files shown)


jonahkagan:
Nice, I like the bigger width and the focus on active audits.

Some thoughts on the UI before I dive into the code:
- I think it's kind of nice having the form to create a new org colocated with the list of orgs since they are relating to the same entity (orgs). We use that pattern elsewhere in Support Tools (e.g. creating an audit admin). "Tools" feels a little random and maybe would make it harder to have a mental model of where things live.
- I get that it's awkward that we have this "Download User List" button that doesn't really go with anything else. I'd probably just stick it at the bottom or top of the org list for now. It's rarely used so I don't think it matters, and it's a bit weird to have its own column.

Neither of these are blockers for me, so feel free to disagree, in which case I'll move forward with code review.

nikhilb4a:
Sweet, thanks for the feedback.

- Ah yeah. I do see the tradeoff there. I found that having the "Create Org" button as the top of the Organizations list makes it the first place the eyes goes, when from my understanding it's not the most common action. Most of the time I assumed people go to an existing Organization or to an Active Audit. When I was new to Arlo, I felt a bit confused why Org creation was so high up on the visual hierarchy. So, I was thinking it would be nice to take the space where the Create Org button was at the top of the list and replace it with a Input Filter to allow people to more easily find an existing org. The design principle here being prioritizing the most common actions, but I may be wrong on what the most common actions are. 
- Fair. I can move it to the bottom of the list. On the initial PR I advised against having it at the bottom of the Org list because I tend to not be a fan of tucking things in a category they don't quite belong in, as it can set a precedent that gets followed and it would be hard to find for someone who didn't already know it existed. I thought Tools (or maybe Actions) provided a nice way to take advantage of the screen width while aligning the visually similar buttons and homepage actions.

Because I spent a little while starting at the Support Tools page and landed on this feeling clear, I am currently a bit biased towards this breakdown. But, to be honest I do tend to default to your opinion on this since you have much more experience with Arlo and how it is used.

Mind just letting me know once you read those comments if they sway you at all, or if you still do lean to your original feedback? Glad I could at least discuss this option but either path I think is a definite step forward.

Btw here is a screenshot with the change you suggested, in case it confirms your preference for that.

Screenshot 2024-12-11 at 12 02 38 PM

nikhilb4a:
Removed Tools column - 

https://github.com/user-attachments/assets/95cd42f5-a993-4315-a2fa-aed159ac88d9

jonahkagan:
>  I was thinking it would be nice to take the space where the Create Org button was at the top of the list and replace it with a Input Filter to allow people to more easily find an existing org

I like this idea!

Looking at the updated screenshot, I am realizing that the active audit list items are quite wide, so the audit status tag is far away from the audit name. So maybe sticking with three columns makes sense now that the screen is wider.

Anyway, I see you moved forward with removing the third column since I was slow to respond, and I'm ok with that too.

I think I'd say go with your preference, since there are pros/cons each way. Happy to review code once you've decided.

nikhilb4a:
Cool, thanks!

If you see some benefit with the three column as well, I think I'd prefer to go with that approach for now. And I'm happy to revert in later if we decide it wasn't helpful. 

Will add the input filter and then will request your code review!

client/src/components/SupportTools/SupportTools.tsx
```diff
@@ -213,19 +177,74 @@ const DownloadUsersButton = styled(AnchorButton)`
...
-        Download User List
-      </DownloadUsersButton>
-      <p>
-        Export a list of Audit Admins and Jurisdiction Managers for all audits
-        completed in the last 12 weeks.
-      </p>
+        <form
+          style={{
+            display: 'flex',
+            // flexDirection: 'column',
```
jonahkagan, line 264: Remove commented line

client/src/components/SupportTools/SupportTools.tsx
```diff
@@ -213,19 +177,74 @@ const DownloadUsersButton = styled(AnchorButton)`
...
+            gap: '8px',
+          }}
+          onSubmit={handleSubmit(onSubmitCreateOrganization)}
+        >
+          <Button
+            type="submit"
+            // icon="insert"
+            style={{}}
+            loading={formState.isSubmitting}
+            className={Classes.BUTTON}
```
jonahkagan, line 273: Is the class name needed? I would expect the component to already have its styles baked in

client/src/components/SupportTools/SupportTools.tsx
```diff
@@ -213,19 +177,74 @@ const DownloadUsersButton = styled(AnchorButton)`
...
+            disabled={!formState.isDirty || formState.isSubmitting}
```
jonahkagan, line 274: I think this partially regresses the previous behavior which used the `loading` prop for when submitting (which shows a loading spinner IIRC)

client/src/components/Atoms/Wrapper.tsx
```diff
@@ -23,4 +23,13 @@ export const Inner = styled.div<IInnerProps>`
   flex-direction: ${props => props.flexDirection || undefined};
 `
 
+export const SupportToolsInner = styled.div`
+  display: flex;
+  justify-content: center;
+  margin-left: auto;
+  margin-right: auto;
+  padding: 30px 0;
+  width: 80%;
```
jonahkagan, line 32: Should we just go full-width (with some padding)? I feel like using 80% width will make things very dependent on the window width, which could lead to unexpected results on different screens.

client/src/components/Header.tsx
```diff
@@ -22,7 +22,7 @@ import LinkButton from './Atoms/LinkButton'
 const SupportBar = styled(Navbar)`
   background-color: ${Colors.ROSE3};
   height: 35px;
-  padding: 0;
+  padding: 0 10%;
```
jonahkagan, line 25: What's the thinking behind using a percentage here? I generally think it's safer to work with absolute values for padding

client/src/components/SupportTools/SupportTools.tsx
```diff
@@ -109,6 +114,20 @@ const Column = styled.div`
   padding-right: 30px;
 `
 
+const Column1 = styled.div`
+  flex: 0 0 25%;
+  padding-right: 30px;
```
jonahkagan, line 119: Instead of using padding here, I'd add a `gap` to the parent `Row`. A little easier to manage. Also with that change you could probably just get away with inline styles for the flex basis instead of defining styled components for each column

client/src/components/SupportTools/SupportTools.tsx
```diff
@@ -134,99 +153,157 @@ const ActiveAudits = () => {
...
+        className={Classes.INPUT}
+        onChange={handleInputChange}
+        placeholder="Filter organizations..."
+        style={{
+          marginBottom: '20px',
+          width: '100%',
+        }}
+        type="text"
+        value={filterText}
+      />
```
jonahkagan, line 207: We have a nice `FilterInput` component with a little icon in Atoms/Table.tsx. Could use if you like it

client/src/components/SupportTools/SupportTools.tsx
```diff
@@ -213,19 +177,74 @@ const DownloadUsersButton = styled(AnchorButton)`
...
+            display: 'flex',
+            // flexDirection: 'column',
+            gap: '8px',
+          }}
+          onSubmit={handleSubmit(onSubmitCreateOrganization)}
+        >
+          <Button
+            type="submit"
+            // icon="insert"
+            style={{}}
```
jonahkagan: Remove commented line and empty style

client/src/components/SupportTools/SupportTools.tsx
```diff
@@ -213,19 +177,74 @@ const DownloadUsersButton = styled(AnchorButton)`
...
+  const onSubmitCreateOrganization = async ({ name }: { name: string }) => {
+    try {
+      await createOrganization.mutateAsync({ name })
+      reset()
+    } catch (error) {
+      // Do nothing - errors toasted by queryClient
+    }
+  }
+
+  console.log(getValues('name'), formState)
```
jonahkagan: Remove console.log

client/src/components/Header.tsx
nikhilb4a: It was to match the 80%, 10% on both sides

client/src/components/SupportTools/SupportTools.tsx
nikhilb4a: Part of an outdated commit

client/src/components/SupportTools/SupportTools.tsx
nikhilb4a: Part of an outdated commit

client/src/components/SupportTools/SupportTools.tsx
nikhilb4a: Finding that using `gap` is getting a bit off because it isn't accounted for when the flex basis is calculated, so the total width goes over 100% when using them together. But agreed on making the styles inline instead of styled components

client/src/components/SupportTools/SupportTools.tsx
```diff
@@ -114,22 +115,9 @@ const Column = styled.div`
...
-  padding-right: 30px;
-`
-
-const Column3 = styled.div`
-  flex: 0 0 25%;
-`
-
 const Row = styled.div`
   display: flex;
+  gap: 30px;
```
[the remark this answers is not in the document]
nikhilb4a: It was intentional, yeah - in the subsequent PR that you reviewed I use the `Row` component on the sub-pages, with Rows having Columns as children similar to this example. So now they consistently have the same gap across pages. I think it should work well, but I'll keep it mind if it ends up actually making styling more difficult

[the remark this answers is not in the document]
jonahkagan: Great, I like standardizing that bit of layout!

PR #2095, opened 2024-12-14
Only surface discrepancies in single jurisdiction for finalized batches

nikhilb4a:
We recently introduced a special case where we surface discrepancies for single jurisdiction batch comparison audits even before the round is finalized. But, this meant that immediately upon the round opening, we were showing discrepancies.

Screenshot 2024-12-13 at 3 54 54 PM

We only want to show discrepancies for batches that have been audited. There may be a more correct way to do this using a db query, but this PR uses the approach of checking the `audited_batch_result` object that we are already using, and looks for a single non-zero entry indicating the result is valid.

Now we no longer see the discrepancies upon launching. 

Screenshot 2024-12-13 at 3 54 09 PM

But once a single batch is audited, we see only those discrepancies.

Screenshot 2024-12-13 at 3 59 21 PM

1. The discrepancy report already handled this by checking if a batch is `audited` using [this check ](https://github.com/votingworks/arlo/blob/main/server/api/reports.py#L950) which uses a [set of queries](https://github.com/votingworks/arlo/blob/main/server/api/reports.py#L858-L886) to build the `audited_results_by_batch` versus the `discrepancies` endpoint which uses the `sampled_batch_results` function. This way seems _more_ correct, but is more expensive
2. Ballot comparison audits did not have this issue in my test. Only audited ballots have discrepancies visible

### DIFF
--- a/server/api/discrepancies.py
+++ b/server/api/discrepancies.py
@@ -109,8 +109,6 @@ def get_batch_comparison_discrepancies_by_jurisdiction(
         show_batch_comparison_discrepancies_by_jurisdiction(election)
     )
 
-    is_single_jurisdiction_election = len(list(election.jurisdictions)) == 1
-
     for contest in list(election.contests):
         audited_batch_results = sampled_batch_results(
             contest, include_non_rla_batches=True
@@ -124,13 +122,6 @@ def get_batch_comparison_discrepancies_by_jurisdiction(
                 continue
 
             if batch_key not in batch_keys_in_round:
-                continue
-
-            # Special case: for single jurisdiction elections we show discrepancies even if the round isn't closed,
-            # but we only want to show a batch's discrepancies if the batch is finalized
-            if is_single_jurisdiction_election and not is_audited_batch_finalized(
-                audited_batch_result
-            ):
                 continue
 
             audited_contest_result = audited_batch_result[contest.id]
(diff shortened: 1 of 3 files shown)


server/api/discrepancies.py
```diff
@@ -187,6 +197,15 @@ def show_batch_comparison_discrepancies_by_jurisdiction(
     }
 
 
+# Check if any value is non-zero for the audited batch result, indicating the batch has actually been audited
+def is_audited_batch_finalized(audited_batch_result: Dict[str, Dict[str, int]]) -> bool:
+    return any(
+        value != 0
+        for contest in audited_batch_result.values()
+        for value in contest.values()
+    )
```
jonahkagan, line 206: `sampled_batch_results` was originally written to be used only in cases where all the batches had been audited (p-value calculation, sample sizes calculation), so it assumes that's the case. Now that we're using it during the round, that assumption breaks. I think we should fix this bug there by filtering out batches that haven't been audited yet (i.e. have no recorded tally sheets).

The solution you have here will work in all cases except the case when a batch is actually audited as having 0 votes for each candidate. This is probably never going to happen, but still would rather use a fully safe solution.

jonahkagan (changes requested):
Thank you for taking the initiative to fix this bug!

server/api/discrepancies.py
nikhilb4a: Thanks for that context, I've [added a filter](https://github.com/votingworks/arlo/pull/2095/commits/f037d934dc9e6d3310f5095c6eea5827b7364971) to `sampled_batch_results` instead that seems to work as intended. 

Screenshot 2024-12-16 at 1 35 22 PM

jonahkagan (approved):
Looks great!

PR #2100, opened 2025-03-16
Fixing ESS overvote discrepancy bug

### LINKED ISSUE #2099: ES&S ballot comparison overvotes always reported as discrepancies
When performing a ballot comparison audit of an ES&S election, overvotes are recorded in the CVR as `overvote` instead of votes for each candidate summing to greater than the number of votes allowed. If this is indeed what happened on the ballot and the audit board records votes for more than the allowable number of candidates, this will always be reported as a discrepancy, even when the original tabulation matches the audit board's determination. 

This appears to be due to the way discrepancy accounting happens when generating the audit report/discrepancy report, introduced in #1831 and carried forward in #1832. The specific issue appears to be with the [`ballot_vote_deltas`](https://github.com/votingworks/arlo/blob/00c762241746fefbeb464fab6085e355bffbe1fa/server/api/shared.py#L509) function in /server/api/shared.py, specifically the following code block:

```python
 for choice in contest.choices:
        reported_vote = (
            0 if reported[choice.id] in ["o", "u"] else int(reported[choice.id])
        )
        audited_vote = (
            0 if audited[choice.id] in ["o", "u"] else int(audited[choice.id])
        )
        deltas[choice.id] = reported_vote - audited_vote

    if all(delta == 0 for delta in deltas.values()):
        return None
```

This code assumes that if an overvote or an undervote is recorded by the audit board that the audited CVR will contain a `o` or `u`. However, audit boards to not have an option to mark an overvote; rather they just mark all votes for each candidate as they see them appear. Because in ES&S elections the report CVR _does_ contain an `o` for over votes, however, this leads to the delta always equaling `-1`, even when the audit board determines that the ballot contains an overvote as originally reported. 

This issue is visible in the test snapshot for ES&S ballot comparison audits server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison.py, specifically [line 59](https://github.com/votingworks/arlo/blob/00c762241746fefbeb464fab6085e355bffbe1fa/server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison.py#L59). This appears to also impact undervotes, but because a vote in an undervote is always 0, the discrepancy math doesn't result in any different behavior. 

Tl;dr: Code for reporting discrepancies assumes that audited overvotes will be recorded as `o`, but they are actually recorded as votes for every candidate. This results in discrepancies getting reported even when there aren't any.

umbernhard:
Fixes #2099. 

I'm pretty sure y'all can write a more elegant fix here, but this at least fixes the issues in my testing. 

Another way to fix this might be to actually create the `u` and `o` notation when the audit board marks an undervote/overvote. In some ways that's more elegant. There is a weird case here where non-ES&S CVR overvotes can still lead to discrepancies, i.e. if the reported votes are {A:1, B:1, C:0}, and the audited votes are {A:0, B:1, C:1}, I think Arlo will still output a discrepancy, even though both cases are overvotes.


umbernhard:
Another potential workaround here is to just use the discrepancy logic in the math module. I'm not sure how much extra plumbing that requires, but it doesn't seem necessary to have two separate ways of computing discrepancies?

arsalansufi:
Thanks for flagging @umbernhard and sorry for the extreme delay here! We'll pick this up and make sure it gets merged

server/tests/ballot_comparison/snapshots/snap_test_ballot_comparison.py
```diff
@@ -55,7 +55,7 @@
 J1,0002,BATCH2,5,0002000175,"Round 1: 0.128937575131137250, 0.240487859312182291",AUDITED,Choice 1-2,Choice 1-2,,,Choice 2-2,Choice 2-2,,\r
 J2,0001,BATCH1,1,0001013415,Round 1: 0.228946820159681463,AUDITED,"Choice 1-1, INVALID_WRITE_IN",Choice 1-1,,,"Choice 2-1, INVALID_WRITE_IN",Choice 2-1,,\r
 J2,0001,BATCH1,3,0001013417,Round 1: 0.457121710197159606,AUDITED,Choice 1-1,Choice 1-1,,,Choice 2-1,Choice 2-1,,\r
-J2,0001,BATCH2,3,0001000417,Round 1: 0.269793733438455805,AUDITED,"Choice 1-1, Choice 1-2",Overvote,Choice 1-1: -1; Choice 1-2: -1,,"Choice 2-1, Choice 2-3",Overvote,Choice 2-1: -1; Choice 2-3: -1,\r
+J2,0001,BATCH2,3,0001000417,Round 1: 0.269793733438455805,AUDITED,"Choice 1-1, Choice 1-2",Overvote,Choice 1-2: -1,,"Choice 2-1, Choice 2-3",Overvote,,\r
```
arsalansufi, line 58: I'm looking at this snapshot diff more closely and noticing something off.

The following

```
Audit result: Choice 1-1, Choice 1-2
CVR result:   Overvote
```

is still resulting in a discrepancy of `Choice 1-2: -1` (previously `Choice 1-1: -1; Choice 1-2: -1`).

In contrast, the following

```
Audit result: Choice 2-1, Choice 2-3
CVR result:   Overvote
```

is expectedly resulting in no discrepancy (prevoiusly `Choice 2-1: -1; Choice 2-3: -1`).

I have a rough idea here what's going on and am prepping an edit to account for this. Because I can't push to this branch from a fork, I'll open a new PR.

jonahkagan: As far as I can tell, when a contest is overvoted, we do save an "o" for every candidate: https://github.com/votingworks/arlo/blob/main/server/api/cvrs.py#L885-L900

arsalansufi:
@umbernhard I know you flagged this related edge case in your PR description:

> There is a weird case here where non-ES&S CVR overvotes can still lead to discrepancies, i.e. if the reported votes are {A:1, B:1, C:0}, and the audited votes are {A:0, B:1, C:1}, I think Arlo will still output a discrepancy, even though both cases are overvotes.

That comment plus our own investigation has gotten us thinking about overvote handling in ballot comparison audits, and we're chatting with Mark L to check our understanding. Based on that conversation, we're going to work on a more holistic fix, not just for ES&S overvotes but for overvotes in general.

We're accordingly going to pause on this PR (as well as the PRs that @jonahkagan and I opened) and open something more holistic, definitely before the next big round of ballot comparison audits.

arsalansufi:
No worries! We've got it from here and will just make sure to tag you on the eventual fix we land on

PR #2147, opened 2025-07-11
Add record ID to ballot retrieval lists

arsalansufi:
We've received a request from Nevada to add CVR record IDs to our ballot retrieval lists for ballot comparison audits. They've upgraded some machines, and those machines will now emit slightly different CVRs. So they'll now need this field to properly locate ballots after loading CVRs into Arlo.


nikhilb4a (approved):
✅

PR #2148, opened 2025-07-22
Don't surface total row error when a CSV has only two rows

arsalansufi:
We have a valid ballot manifest in Georgia that's getting flagged as having a total row, i.e., a row where the last row is simply the sum of all rows above, and thereby rejected.

| Container | Batch Name | Number of Ballots |
| :- | :- | :- |
| Advanced Voting | AV-Elections Office ICP 1 - 0 | 6 |
| Election Day | Election Day Central ICP 1 - 0 | 6 |

> :x: It looks like the last row in the CSV might be a total row. Please remove this row from the CSV.

To get this file through, I'm disabling the total row check when a CSV has only two rows, which generally seems like a good idea.


kofi-q (approved):
✅

PR #2149, opened 2025-07-23
Ballot polling: Account for not-found ballots during sample size calculation

arsalansufi:
## Overview

This PR addresses an edge case in the ballot polling audit math.

We recently encountered a ballot polling audit that required escalation to a round 2 but for which sample size calculation was finding that the audit should be considered complete. This surfaced in the UI confusingly as such:

<table>
before
</table>

The issue was in our accounting of ballots that could not be found. When a ballot cannot be found in an audit, we assume the "worst" and count it as a vote for the losing candidate(s). We properly make this assumption when calculating whether the risk limit has been met but are forgetting to do so when calculating round 2 sample sizes. In this audit, there was 1 ballot that could not be found, and that 1 ballot was the decisive factor in determining whether the risk limit was met.

The fix was simple, applying the not-found logic not only during risk limit calculation but also during sample size calculation.

<table>
after
</table>

Note that, because the risk limit calculation has always been correct, we can be confident that no past audits were considered done when they shouldn't have been :relieved:

## Testing

- [x] Updated automated tests
- [x] Tested manually using a production DB snapshot

### DIFF
--- a/server/audit_math/ballot_polling.py
+++ b/server/audit_math/ballot_polling.py
@@ -19,6 +19,7 @@ def get_sample_size(
     risk_limit: int,
     contest: Contest,
     sample_results: Optional[BALLOT_POLLING_SAMPLE_RESULTS],
+    samples_not_found: Dict[str, int],
     math_type: AuditMathType,
     round_sizes: Optional[BALLOT_POLLING_ROUND_SIZES],
 ) -> Dict[str, SampleSizeOption]:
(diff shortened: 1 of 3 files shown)


server/audit_math/ballot_polling.py
```diff
@@ -34,6 +35,11 @@ def get_sample_size(
         - A sample size dictionary containing sample sizes for different
           finishing probabilities
     """
+    # When a sampled ballot can't be found, count it as a vote for every loser
+    if sample_results:
+        for round_id, num_not_found in samples_not_found.items():
+            for loser in contest.losers:
+                sample_results[round_id][loser] += num_not_found
```
arsalansufi, line 42: Logic copied from the risk limit logic:

https://github.com/votingworks/arlo/blob/30eb4ee5ab68d18b70f730c0644952f6b2074bdd/server/audit_math/ballot_polling.py#L61-L64

Could ultimately place this somewhere shared but sufficient for now to unblock the Texas round 2 tomorrow

nikhilb4a (approved):
:nice: I read about this requirement for not-found ballots in the paper, so it's interesting to see its effect in play

arsalansufi:
^ Oh very cool!

PR #2152, opened 2025-08-12
chore(deps): move packages used for development into `devDependencies`

eventualbuddha:
This makes it easier to know which packages to focus on for updates and vulnerability audits.


jonahkagan (approved):
✨

PR #2156, opened 2025-08-12
fix(security): prevent path traversal in `/api/file-upload`

eventualbuddha:
Prevents a user-controllable string (`key` in the request) from being used to write data to an arbitrary path on the local file system.


eventualbuddha:
> I forget exactly but I think this flow is only in local development where its not storing in s3

Yeah, it should only be reachable when not using an s3-backed file storage. For us, that's only development.

PR #2161, opened 2025-08-20
Ballot comparison: Fix ES&S overvotes getting always flagged as discrepancies

arsalansufi:
Closes https://github.com/votingworks/arlo/issues/2099

A modified version of https://github.com/votingworks/arlo/pull/2100 that corrects a bug within that patch


jonahkagan:
While I was reloading context to review this PR, I tried out a different solution that is based on the audit math itself, let me know what you think: https://github.com/votingworks/arlo/pull/2162

PR #2165, opened 2025-08-26
Use open instead of retrieve_file for local file system read

arsalansufi:
Addressing a crash that doesn't occur locally but occurs when deployed.

TLDR: `retrieve_file` only pulls from S3 in deployed environments. While the ZIP file that we're interacting with is uploading to S3, it's downloaded and then unzipped on the local file system. So reading the unzipped files requires a standard `open`.

The `retrieve_file` abstraction is a bit leaky. That's probably the root issue here that we should address at a later time.

### DIFF
--- a/server/tests/api/test_jurisdictions.py
+++ b/server/tests/api/test_jurisdictions.py
@@ -2,6 +2,7 @@ import json, io
 from datetime import datetime, timedelta
 from typing import List
 from flask.testing import FlaskClient
+import pytest
 
 from ...activity_log.activity_log import (
     ActivityBase,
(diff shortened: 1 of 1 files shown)


server/api/batch_inventory.py
```diff
@@ -518,7 +518,11 @@ def process_ess_md():
         cvr_file.close()
 
         cvr_file_readers = [
-            csv_reader_for_cvr(retrieve_file(join(working_directory, file_name)))
+            csv_reader_for_cvr(
+                # TODO: Revisit file cleanup holistically
```
arsalansufi, line 522: Even when we're using `retrieve_file` we aren't calling `close` on files as far as I can tell so no regression here as far as I can tell. Python garbage collection should eventually take care of things but might be worth revisiting file cleanup holistically.

arsalansufi:
@eventualbuddha I'm gonna use my admin privileges to merge this just so I can get it deployed in time for Ginny's training. But I've confirmed that this change works by deploying to staging, and I know we also discussed the change in Slack. Feel free to leave comments after though!

eventualbuddha:
LGTM!

PR #2166, opened 2025-08-26
refactor: improve `retrieve_file` abstraction

eventualbuddha:
Rather than taking an arbitrary path, take a `File` to make it very clear that this API is only meant to be used with uploaded files whose entries are stored in the database and not with arbitrary files.


eventualbuddha:
> Looks good! I wonder if it also makes sense to update the `store_file` and `delete_file` functions since they are part of the same interface

I updated `delete_file`, but left `store_file` alone since it's used in a few places with a path that doesn't come from a `File`, such as the `/file-upload` API route.

Issue #2210, opened 2025-10-15
Oregon security audit updates - high pri odds and ends


arsalansufi:
We've wrapped up nearly everything that we laid out in [Oregon Arlo Security Assessment Response - Latest.pdf](https://github.com/user-attachments/files/22930002/Oregon.Arlo.Security.Assessment.Response.-.Latest.pdf)

Just two high pri odds and ends remain:
- [ ] One last Python package update - pyjwt up to 2.10.1
- [ ] Switching our Python XML parsing library from the default to a safer alternative with protection against XML External Entity (XXE) Injection, e.g., defusedxml

Original Oregon security audit reports for reference:
- https://votingworks.slack.com/files/U038Z7CU3PU/F09AAHMT197/polaris_developer_detail_sca_report.pdf
- https://votingworks.slack.com/files/U038Z7CU3PU/F099X9WQ1GA/polaris_developer_detail_static_report_-arlo_github.pdf

PR #2216, opened 2025-10-23
fix: replace `xml` with `defusedxml` for XML parsing

eventualbuddha:
Refs #2210 

Prevents various attacks against the standard Python XML parser.


server/api/batch_inventory.py
```diff
@@ -11,7 +11,8 @@
 import io
 from typing import TypedDict
 import uuid
-from xml.etree import ElementTree
+from defusedxml.ElementTree import parse as parse_xml
+from xml.etree.ElementTree import ElementTree
```
eventualbuddha: I still need to import from `xml` since `defusedxml` does not seem to export the `xml.etree.ElementTree.ElementTree` and `xml.etree.ElementTree.Element` values. I don't love that because it means we need to be pretty careful to not accidentally reintroduce an `ElementTree.parse` call somewhere.

Issue #2191, opened 2025-10-08
Audit board UI: Order Yes/No options in that order rather than alphabetically


arsalansufi:
Washington has asked if we can display candidates in the audit board UI in the same order as physical ballots. This is obviously complicated because candidate orders can vary across ballot styles / precincts. We _might_ be able to infer candidate order from CVRs but there isn't a guarantee there either.

The simplest thing we can do is order Yes/No contests to display the Yes before the No (instead of the No before the Yes, which is what alphabetical order results in).

Issue automatically created from Slack feedback reported by ginny: 
https://votingworks.slack.com/archives/C01F2CH62TS/p1756383456403229

PR #2220, opened 2025-10-31
Audit board UI: Sort yes before no for propositions

arsalansufi:
## Overview

Closes https://github.com/votingworks/arlo/issues/2191

The bigger long-term goal here is to have the audit board UI display candidates in the same order as the retrieved ballot, which as a reminder can vary across ballot style given rotation rules. But TBD which CVR types include the granular order info needed to make that possible.

For the time being, we're tackling the one easy case of a proposition, always displaying the yes option before no option. Before this PR, that order wasn't guaranteed and the no option could sometimes end up ahead of the yes option.

audit-board-ui

## Testing

- [x] Updated automated tests
- [x] Tested manually

### DIFF
--- a/client/src/components/AuditBoard/BallotAudit.tsx
+++ b/client/src/components/AuditBoard/BallotAudit.tsx
@@ -267,12 +267,19 @@ const BallotAuditContest = ({
     }
   }
 
+  // Sort yes before no for propositions
+  const yesOption = contest.choices.find(c => c.name.toLowerCase() === 'yes')
+  const noOption = contest.choices.find(c => c.name.toLowerCase() === 'no')
+  const sortedContestChoices =
+    contest.choices.length === 2 && yesOption && noOption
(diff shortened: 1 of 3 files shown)


client/src/components/AuditBoard/BallotAudit.tsx
```diff
@@ -267,12 +267,19 @@ const BallotAuditContest = ({
     }
   }
 
+  // Sort yes before no for propositions
+  const yesOption = contest.choices.find(c => c.name.toLowerCase() === 'yes')
+  const noOption = contest.choices.find(c => c.name.toLowerCase() === 'no')
```
arsalansufi: We can refine this logic if we encounter other situations that are clearly propositions but don't use yes or no as the options, e.g., affirm, reject, etc.

nikhilb4a (approved):
Smart low hanging fruit!

PR #2222, opened 2025-11-03
Log failed JM log-ins for future analysis

### LINKED ISSUE #2022: Log failed JM log-ins for future analysis
Jurisdiction managers often struggle to log into Arlo. Let's make sure we log as much info as we can about the various failures during this upcoming audit season so that we can analyze which ones happen most often and try to make improvements to prevent common failures in the future.

Ideas of potential failures to capture:
- JM missing from participants file
- mistyped email
- entered the wrong code
- code expired

More discussion: https://votingworks.slack.com/archives/CKCVA0F9S/p1730223661738059

eventualbuddha:
Closes #2022 

As mentioned in the issue, we want to get some better logging as to why Jurisdiction Manager logins are failing sometimes. Is it human error? Some bug on our side? Something we should be doing better? Hopefully this new logging will tell us.

### DIFF
--- a/client/src/components/AuditAdmin/Setup/Contests/ContestForm.tsx
+++ b/client/src/components/AuditAdmin/Setup/Contests/ContestForm.tsx
@@ -74,6 +74,16 @@ interface IDropdownCheckboxListProps {
   contestIndex: number
 }
 
+const DropdownCheckboxListContainer = styled.div`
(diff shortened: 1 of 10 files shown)


jonahkagan (approved):
Looks great. Very thorough!

PR #2312, opened 2026-05-14
Audit majority claim for runoff-subject contests

nikhilb4a:
Georgia is Arlo's only customer that conducts batch comparison audits (MACRO) of primary contests subject to **majority-or-runoff** law: if a candidate receives a majority of valid votes they win the primary outright; otherwise the top two advance to a runoff. 

Arlo today only audits the pairwise margins (e.g. with `num_winners=2`, that each reported top-2 candidate beats every other candidate). It does _not_ statistically verify the majority claim — i.e., whether the reported leader's share actually does or does not cross 50%.

This PR adds:
1. The missing mathematical assertions that the winning candidate received a majority, or that the top two vote-getters did not.
2. Updates the setup form to indicate the Runoff Law applies
3. Updates the audit report to indicate the runoff rules applied to the audit
4. Tests

Testing:
- Automated tests added
- I've done one end-to-end test, but I plan to do more to test the different permutations.

**Form updates:**

https://github.com/user-attachments/assets/4af008a3-a8ef-4300-82bb-8c07b0b2f5ad

**Audit report updates:**

Screenshot 2026-05-19 at 5 03 49 PM

### DIFF
--- a/client/src/components/AuditAdmin/Setup/Contests/ContestForm.tsx
+++ b/client/src/components/AuditAdmin/Setup/Contests/ContestForm.tsx
@@ -217,18 +217,21 @@ const computeRunoffPreview = (choices: IChoiceValues[]): string => {
     name: c.name,
     votes: parseNumber(c.numVotes) || 0,
   }))
-  const totalValid = sum(parsed.map(c => c.votes))
-  if (totalValid <= 0) return ''
+  const totalVotes = sum(parsed.map(c => c.votes))
+  if (totalVotes <= 0) return ''
   const sorted = sortBy(parsed, c => -c.votes)
   const leader = sorted[0]
-  if (leader.votes > totalValid - leader.votes) {
+  if (leader.votes > totalVotes - leader.votes) {
     return `Reported results: ${leader.name} received a majority, no runoff required.`
   }
   const runnerUp = sorted[1]
   return `Reported results: ${leader.name} and ${runnerUp.name} are the top two vote-getters and neither received a majority, runoff required.`
 }
 
-const contestFromValues = (contest: IContestValues): IContest => ({
+const contestFromValues = (
+  contest: IContestValues,
+  auditType: AuditType
+): IContest => ({
   ...contest,
   id: contest.id || uuidv4(), // preserve given id if present, generate new one if empty string
   totalBallotsCast: parseNumber(contest.totalBallotsCast),
@@ -240,10 +243,10 @@ const contestFromValues = (contest: IContestValues): IContest => ({
     numVotes: parseNumber(choice.numVotes),
   })),
   pendingBallots: parseNumber(contest.pendingBallots),
-  // Only include when true; the API defaults missing values for batch-comparison
-  // to false and non-batch-comparison schemas will reject
   isSubjectToRunoff:
-    contest.numWinners === '1' && contest.isSubjectToRunoff ? true : undefined,
+    auditType === 'BATCH_COMPARISON'
+      ? contest.numWinners === '1' && !!contest.isSubjectToRunoff
+      : undefined,
 })
 
 const ContestForm: React.FC<IProps> = ({
@@ -329,7 +332,9 @@ const ContestForm: React.FC<IProps> = ({
         }))
       : values.contests
     await updateContestsMutation.mutateAsync(
-      contestsToUpdate.map(contestFromValues).concat(restContests)
+      contestsToUpdate
+        .map(c => contestFromValues(c, auditType))
+        .concat(restContests)
     )
     goToNextStage()
   }
--- a/client/src/components/AuditAdmin/Setup/Contests/ContestForm.tsx
+++ b/client/src/components/AuditAdmin/Setup/Contests/ContestForm.tsx
@@ -43,8 +43,9 @@ import { testNumber } from '../../../utilities'
 import { isObjectEmpty } from '../../../../utils/objects'
 import useStandardizedContests from '../../../useStandardizedContests'
 import { ErrorLabel } from '../../../Atoms/Form/_helpers'
-import { partition } from '../../../../utils/array'
-import { AuditType } from '../../../useAuditSettings'
+import { partition, sortBy } from '../../../../utils/array'
+import { sum } from '../../../../utils/number'
+import { AuditType, IAuditSettings } from '../../../useAuditSettings'
 import { parse as parseNumber } from '../../../../utils/number-schema'
 
 const CustomMenuItem = styled.li`
@@ -171,6 +172,7 @@ interface IProps {
   electionId: string
   isTargeted: boolean
   auditType: AuditType
+  electionState: IAuditSettings['state']
   goToPrevStage: () => void
   goToNextStage: () => void
 }
@@ -192,6 +194,7 @@ export interface IContestValues {
   choices: IChoiceValues[]
   totalBallotsCast?: string
   pendingBallots?: string
+  isSubjectToRunoff?: boolean
   jurisdictionIds: string[]
 }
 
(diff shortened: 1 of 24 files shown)


server/audit_math/sampler_contest.py
```diff
@@ -44,13 +45,17 @@ class Contest:
...
+    is_subject_to_runoff: bool
     name: str  # The name of the contest
 
     winners: dict[str, int]  # list of all the winners
     losers: dict[str, int]  # list of all the losers
 
     diluted_margin: float  # The smallest diluted margin in this contest
     margins: dict[str, dict]  # dict of the margins for this contest
+    # Threshold pairs for runoff-subject contests (real candidate vs aggregate __not_X)
+    runoff_pairs: list[dict]
```
jonahkagan: Personally I'm not a big fan of the pattern here of pre-computing stuff and attaching it to the contest object. I'd rather just have the math functions compute what they need. In this case, since this logic is specific only to macro, I think it would be better encapsulated to that module

server/audit_math/macro.py
```diff
@@ -27,6 +27,21 @@ class BatchError(TypedDict):
...
+    dict with `__not_<X>` aggregate entries (one per entry in `contest.winners`),
+    where each aggregate's batch tally is `sum_of_real_candidates_in_batch −
+    X_votes_in_batch`. Idempotent. No-op if the contest isn't runoff-subject.
+    """
+    if not contest.is_subject_to_runoff:
+        return
+    contest_batch = batch_results[contest.name]
+    real_total = sum(contest_batch.get(c, 0) for c in contest.candidates)
+    for w_id in contest.winners:
+        contest_batch[f"__not_{w_id}"] = real_total - contest_batch.get(w_id, 0)
```
jonahkagan: This approach of creating a synthetic `__not_{w_id}` candidate and adding it to the results seems a bit confusing/roundabout. I get that the idea is to take advantage of the existing math for computing error between two candidates. What if we factored it out a bit differently:

If a contest is subject to runoff, after computing the max error for the batch for all of the candidates, call out to a function to compute the max error from the runoff assertions. Then compare that error to the max error from candidates, and return whichever one is greater. Inside that function, we can do the summing for the "not-X" candidate. We can then either replicate the logic for computing error or factor it out of `error_for_candidate_pair` so it takes the actual values (`v_wp`, `v_lp`, etc.) instead of keys to look up the values.

server/api/contests.py
```diff
@@ -273,6 +277,18 @@ def validate_contests(contests: list[JSONDict], election: Election):
                     f" ({total_votes} votes, {total_allowed_votes} allowed)"
                 )
 
+    for contest in contests:
+        if not contest.get("isSubjectToRunoff", False):
+            continue
+        if election.audit_type != AuditType.BATCH_COMPARISON:
+            raise BadRequest(
+                "Runoff-subject contests are only supported for batch comparison audits"
```
jonahkagan: Nit: "Runoff-subject contests" is kind of a weird phrase

server/api/contests.py
```diff
@@ -317,6 +333,10 @@ def normalize_contest(contest):
         contest["choices"] = sorted(
             contest["choices"], key=lambda choice: str(choice["id"])
         )
+        # Client conditionally omits this when false; serialize_contest always
+        # emits it. Normalize missing → False so a no-op save doesn't look like
+        # a flag change.
+        contest.setdefault("isSubjectToRunoff", False)
```
jonahkagan: Should we just line up the behavior of the client and serialize_contest to make this unnecessary?

client/src/components/AuditAdmin/Setup/Contests/ContestForm.tsx
```diff
@@ -205,8 +208,26 @@ const contestToValues = (contest: IContest): IContestValues => ({
...
+  isSubjectToRunoff: contest.isSubjectToRunoff ?? false,
 })
 
+const computeRunoffPreview = (choices: IChoiceValues[]): string => {
+  if (!choices.every(c => c.name && c.numVotes !== '')) return ''
+  const parsed = choices.map(c => ({
+    name: c.name,
+    votes: parseNumber(c.numVotes) || 0,
+  }))
+  const totalValid = sum(parsed.map(c => c.votes))
```
jonahkagan: Nit: `totalValid` -> `totalVotes`

client/src/components/AuditAdmin/Setup/Contests/ContestForm.tsx
```diff
@@ -269,6 +296,8 @@ const ContestForm: React.FC<IProps> = ({
   )
     return null // Still loading
 
+  const showRunoffOption = isBatchComparison && electionState === 'GA'
```
jonahkagan, line 302: I don't know why I didn't think of this earlier - do we want to flag this feature by the election state or by the organization ID? We have some prior art of doing feature flagging by org: https://github.com/votingworks/arlo/blob/main/server/feature_flags.py

.basedpyright/baseline.json
jonahkagan, line 1: Do you know why this file is being updated every commit? I haven't really worked with based pyright yet but my basic understanding is that this is a baseline of known type issues from before we added this type checker. It seems like we probably should try to not introduce new type issues.

server/api/reports.py
```diff
@@ -321,6 +326,18 @@ def contest_rows(election: Election):
...
+        if show_runoff_columns:
+            if contest.is_subject_to_runoff:
+                row += [
+                    "Subject to runoff law",
+                    reported_runoff_results_label(
+                        [choice.num_votes for choice in contest.choices]
+                    ),
+                ]
+            else:
+                row += ["Not subject to runoff law", ""]
```
jonahkagan: Weak opinion: Might just leave this empty to make the contests that are subject to runoff law stand out more

jonahkagan (approved):
Logic looks good overall. I left a bunch of suggestions that are mostly about code style, so feel free to take those as you will.

Also, I've generally tried to add an end-to-end test on the backend for new audit functionality to make sure it all works together. That might be useful here

.basedpyright/baseline.json
arsalansufi: I can share some context here. I know @eventualbuddha added this some time back as a potential improvement to the typechecking experience in Arlo.

In practice, I haven't been the biggest fan as it requires updating any time existing code is even moved around. So even if you haven't added a type issue but have added some code that pushes some existing code with a type issue down, the file needs an update. This means that the file needs an update on just about every change and can make it hard to distinguish actual regressions from existing type issues.

I think it's worth exploring alternatives again (separate from this PR and after current audit prep)

jonahkagan: Ah gotcha, makes sense, thanks

server/audit_math/macro.py
nikhilb4a: Agreed - updated in this https://github.com/votingworks/arlo/pull/2312/changes/138626439ec5c9f1dcefb4ae4bd2c78057a38a38. Dropped `runoff_pairs` and added a local helper function `error_for_threshold` to both `compute_error` and `compute_max_error`. Seems more straightforward. 

I didn't separate out finding the max error of candidates pairs vs. thresholds because I felt the code read pretty cleanly with just grouping them together before finding the max. Feel free to add additional comments!

server/audit_math/sampler_contest.py
nikhilb4a: Agreed, encapsulated in macro https://github.com/votingworks/arlo/pull/2312/changes/138626439ec5c9f1dcefb4ae4bd2c78057a38a38

server/audit_math/macro.py
jonahkagan: Nice, lgtm!

server/api/reports.py
nikhilb4a: Good point. Updated [here](https://github.com/votingworks/arlo/pull/2312/changes/06220e67d2885c59662d585ff355f1351a3600c2#diff-cd1c4f9d2c75d55bb6d1dd7846640971e8109e0e6ad706b955c395c09eae94bfR339)

server/api/contests.py
nikhilb4a: Agreed. Updated client to only include `isSubjectToRunoff` for batch comparisons, matching `serialize_contest` https://github.com/votingworks/arlo/pull/2312/changes/06220e67d2885c59662d585ff355f1351a3600c2

server/api/contests.py
nikhilb4a: Just used the variable name instead `isSubjectToRunoff` https://github.com/votingworks/arlo/pull/2312/changes/06220e67d2885c59662d585ff355f1351a3600c2

client/src/components/AuditAdmin/Setup/Contests/ContestForm.tsx
nikhilb4a: Makes sense to keep it consistent with the server's use of `organizationId`. Given it's not too consequential of change, thinking I'll make that change as a follow up PR so I can first do some end to end testing, since now it looks like Georgia will do their audit setup tomorrow instead of Friday

nikhilb4a:
Added [two end to end tests](https://github.com/votingworks/arlo/pull/2312/commits/8ec0e24938ce4ecd8fbf49aa6bacd3a7ec4d4ac4) - one straightforward case for no majority, one that fails the first round because of discrepancies that cause the threshold assertion to have an error too high to meet the risk limit

PR #2315, opened 2026-05-20
Fix spacing of candidate labels/buttons

nikhilb4a:
Low lift change to align the spacing of the "Remove.." buttons with their candidate rows. I just overrode styling so I didn't need to consider side effects of editing the shared components, as the initial styling is coming from the `bp3` package.

Before:
Screenshot 2026-05-19 at 5 40 09 PM

After:
Screenshot 2026-05-19 at 5 33 03 PM


jonahkagan (approved):
Thank you!

PR #2318, opened 2026-05-21
Set GA default system type to Liberty Vote; rename Dominion

### LINKED ISSUE #2192: For Georgia is there a way we can set the equipment to Dominion for the entire state in the batch inventory tool?
For Georgia is there a way we can set the equipment to Dominion for the entire state in the batch inventory tool?

Issue automatically created from Slack feedback reported by lectusconsulting: 
https://votingworks.slack.com/archives/C01F2CH62TS/p1753111560240689

### LINKED ISSUE #2292: *Problem:* 
Liberty Vote is still listed as Dominion in Arlo

*From:* Tabitha

*Proposed solution:* 
We should update Dominion to Liberty Vote

_Collected by @lectusconsulting_
*Problem:* 
Liberty Vote is still listed as Dominion in Arlo

*From:* Tabitha

*Proposed solution:* 
We should update Dominion to Liberty Vote

_Collected by @lectusconsulting_

Issue automatically created from Slack feedback reported by Product Feedback: 
https://votingworks.slack.com/archives/C01F2CH62TS/p1762982965640609

nikhilb4a:
Closes https://github.com/votingworks/arlo/issues/2292
Closes https://github.com/votingworks/arlo/issues/2192

This PR:
1. Renames **Domion** to **Libery Vote (Dominion)** on user facing screens.
2. Sets Liberty Vote to be the default system type for GA in the batch inventory tool

https://github.com/user-attachments/assets/8325d160-928e-4f6a-987e-6b653cd91fa4

### DIFF
--- a/client/src/components/JurisdictionAdmin/BatchInventory.tsx
+++ b/client/src/components/JurisdictionAdmin/BatchInventory.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
 import { H2, Button, H4, Checkbox, HTMLSelect } from '@blueprintjs/core'
 import {
   useQueryClient,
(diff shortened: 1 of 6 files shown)


client/src/components/JurisdictionAdmin/BatchInventory.tsx
```diff
@@ -212,14 +212,22 @@ const useBatchInventorySignOff = (
...
-}> = ({ systemTypeQueries, nextStep }) => {
+}> = ({ systemTypeQueries, defaultSystemType, nextStep }) => {
+  const systemType = systemTypeQueries.systemType.data
+  const setSystemTypeMutate = systemTypeQueries.setSystemType.mutate
+
+  useEffect(() => {
+    if (systemType === null && defaultSystemType) {
+      setSystemTypeMutate(defaultSystemType)
+    }
+  }, [systemType, defaultSystemType, setSystemTypeMutate])
```
arsalansufi, line 225: I'd usually prefer setting defaults like this on the backend to avoid a flash and a useEffect on the frontend, but because these batch inventory configs are defined on the frontend for now and Georgia is kicking off soon, this is reasonable

arsalansufi (approved):
LGTM!

client/src/components/JurisdictionAdmin/BatchInventory.tsx
nikhilb4a: > but because these batch inventory configs are defined on the frontend for now

Agreed with it being a better fit for the backend, but also that the reasoning you gave was why I just went with the client side for now. Also the `initialStep` gets [determined](https://github.com/votingworks/arlo/blob/main/client/src/components/JurisdictionAdmin/BatchInventory.tsx#L534) based on the value provided by the backend, so unless I made a bit more of an involved change, we would skip to the second step if the backend provided the default.

PR #2319, opened 2026-05-23
WIP Fix pytest-xdist race condition in test org creation with retry logic

kshen0:
Possibly fixes flaky test in https://app.circleci.com/pipelines/github/votingworks/arlo/8236/workflows/8810e198-aa50-480a-9fb4-c367479935c8/jobs/26039

Claude's analysis:

> The tests used hardcoded org IDs (e.g. "TEST-ORG/sample-extra-batches-by-counting-group") shared across multiple parametrized test cases. When pytest-xdist distributed those tests to parallel workers, two workers could simultaneously check whether the org existed, both find it absent, and both attempt to insert it — causing one to fail with a UniqueViolation on the primary key.

> The fix adds get_or_create_org() in helpers.py: it attempts the insert and catches IntegrityError, then retries. On retry, the org now exists (created by the other worker) and is returned successfully.

The fix itself seems low-risk enough to try without concretely reproing the issue, but I'm not sure why we needed so many changes to .basedpyright/baseline.json to pass commit hook linting.

### DIFF
--- a/server/tests/helpers.py
+++ b/server/tests/helpers.py
@@ -116,27 +116,13 @@ def add_admin_to_org(org_id: str, user_email: str):
     return audit_admin.id
 
 
-def get_or_create_org(org_id: str, user_email: str = DEFAULT_AA_EMAIL) -> str:
-    """Get or create an organization, handling race conditions in concurrent tests.
-
-    With pytest-xdist, multiple workers can try to create the same org simultaneously.
-    This function is idempotent and retries if another worker creates the org first.
-    """
-    for attempt in range(3):
-        if Organization.query.get(org_id):  # pyright: ignore[reportUnknownMemberType]
-            return org_id
-
-        org: Organization = Organization(id=org_id, name=org_id)  # pyright: ignore[reportCallIssue]
-        db_session.add(org)  # pyright: ignore[reportUnknownMemberType]
-        try:
-            _ = add_admin_to_org(org.id, user_email)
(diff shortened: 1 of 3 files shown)


server/tests/helpers.py
```diff
@@ -116,6 +116,15 @@ def add_admin_to_org(org_id: str, user_email: str):
     return audit_admin.id
 
 
+def create_org(org_id: str, user_email: str = DEFAULT_AA_EMAIL) -> None:
+    try:
+        with db_session.begin_nested():
```
jonahkagan, line 121: what's the nested session for here? why does it only apply to adding the org, not also adding the admin to the org?

jonahkagan (approved):

server/tests/helpers.py
nikhilb4a: Good question. My understanding is that is how we are able to catch the error immediately, otherwise it would only be caught when the next db.flush happens, making our catch case here useless. And then the error either wouldn't be handled properly wherever that next db.flush is happening, or if that handler did catch an `IntegrityError` as well it would be catching an error for a db update it wasn't designed for. Adding the admin actually has the same thing already with the `create_user` [function](https://github.com/votingworks/arlo/blob/8396aa9e79e627ed30031329a74a03ce1dda36f9/server/tests/helpers.py#L89) that it calls

jonahkagan: Ah makes sense, thanks for explaining. I've forgotten how these things work

PR #2321, opened 2026-05-27
Normalize names for batch prep tool

nikhilb4a:
https://votingworks.slack.com/archives/C0B5C56T15G/p1779916628631129?thread_ts=1779912698.727599&cid=C0B5C56T15G

Screenshot 2026-05-27 at 2 08 08 PM

Noticed with a PA county that their CVR upload was failing to produce CTBB that had any counts, and it was because the target contest for the audit was not matching any contests in the CVR. This is because the CVRs have newlines in the middle of the contest names. The same issue also affects choice names. 

This PR adds normalization of the contest names when getting contest/choice values from the CVR, and additionally adds normalization of choice names when assigning vote counts to choice IDs.

### DIFF
--- a/server/api/batch_inventory.py
+++ b/server/api/batch_inventory.py
@@ -121,12 +121,16 @@ def items_list_to_dict(items):
     }
 
 
+def collapse_whitespace(name: str) -> str:
+    return re.sub(r"\s+", " ", name).strip()
+
+
 def validate_choice_name_and_get_choice_id(
     contest: Contest, choice_name: str
 ) -> str | None:
     choice_id = None
     for choice in contest.choices:
-        if choice.name == choice_name:
+        if collapse_whitespace(choice.name) == collapse_whitespace(choice_name):
             choice_id = choice.id
             break
         # handle capitalization mismatches for the write in column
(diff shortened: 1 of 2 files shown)


server/api/batch_inventory.py
```diff
@@ -121,12 +121,17 @@ def items_list_to_dict(items):
...
+
+
 def validate_choice_name_and_get_choice_id(
     contest: Contest, choice_name: str
 ) -> str | None:
     choice_id = None
+    normalized_choice_name = collapse_whitespace(choice_name)
     for choice in contest.choices:
-        if choice.name == choice_name:
+        if collapse_whitespace(choice.name) == normalized_choice_name:
```
arsalansufi: nit: I might prefer

```python
collapse_whitespace(choice.name) == collapse_whitespace(choice_name)
```

instead of using the `normalized_choice_name` intermediate since `collapse_whitespace(choice.name)` and `collapse_whitespace(choice_name)` are both normalized choice names in a sense, so it becomes a bit arbitrary which one is the `normalized_choice_name`

server/tests/ballot_comparison/test_cvrs.py
```diff
@@ -2021,8 +2021,9 @@ def build_contest(contest_name: str, choice_names: list[str]):
     contests = "\n".join(
         [
             build_contest(
-                "Contest 1",
-                (["Choice 1-1"] if interpretations[0] == "1" else [])
+                # Intentional extra whitespace to test collapsing whitespace from contest/choice names
+                "Contest \n 1",
+                (["Choice   1-1"] if interpretations[0] == "1" else [])
```
arsalansufi: Since these helpers are also used in ballot comparison tests, this may result in a test failure as ballot comparison doesn't use the above batch audit file prep tool normalization code. I'm okay with removing this test update and getting this change deployed to unblock Verified Voting for now, with a follow-up to add a test that applies only to the batch audit file prep tool.

arsalansufi (approved):
Approving to keep you unblocked!

server/tests/ballot_comparison/test_cvrs.py
nikhilb4a: You were right it did break the ballot comparison test! I just added a new simple [test](https://github.com/votingworks/arlo/pull/2321/changes/8674661d5cf64728785e51c678b802a83ca0286e) instead.

Issue #2291, opened 2026-04-16
*Problem:* 
The audit report totals are calculated inconsistently across the reported/audited columns vs


arsalansufi:
*Problem:* 
The audit report totals are calculated inconsistently across the reported/audited columns vs. the ballots in batch column when there's a combined batch.

The reported/audited columns account for combined batches whereas the ballots in batch column does not. For example, if Batch 1 and Batch 2 are combined but only Batch 1 is sampled, the ballots in batch total will count Batch 1 but not Batch 2. The reported/audited totals will, on the other hand, count both Batch 1 and Batch 2.

*From:* Arsalan / Ginny

*Proposed solution:* 
Update the ballots in batch column to mimic the handling of the reported/audited columns, using the ballot count of the larger combined batch rather than the sampled batch alone when computing totals. Additionally consider consolidating the sampled batch rows and combined batch rows in the audit report, e.g., right now we have a row for Batch 1 and a row for the combined batch consisting of Batch 1 and Batch 2. We could get by with only the combined batch row. The totals row would in this case make more intuitive sense.

_Collected by @arsalan_

Issue automatically created from Slack feedback reported by Product Feedback: 
https://votingworks.slack.com/archives/C01F2CH62TS/p1763413513621449?thread_ts=1763413513.621449&cid=C01F2CH62TS

PR #2327, opened 2026-05-29
Audit report: Account for combined batches in the Ballots in Batch total

nikhilb4a:
https://github.com/votingworks/arlo/issues/2291

The Ballots in Batch total was not considering additional batches that may have been added by being included in a combined batch. These additional batches may not have been included in the original batch selection, but once combined with a batch that was selected, they need to be accounted for in the totals.

Follow up:
- As noted in the ticket, should we remove the row entries for the batches that were selected but then combined from the audit report? Wanted to just get this merged, since it's actually an error. Thought it could be worth checking with Jonah/success if the original batch rows were maintained for a reason, though seems like they aren't worth keeping.

**After - Ballots in Batches total matches with Ballots Sampled**

Screenshot 2026-05-29 at 1 20 11 PM

**Before - Ballots in Batches total mismatches with Ballots Sampled**

Screenshot 2026-05-29 at 1 19 54 PM

### DIFF
--- a/server/api/reports.py
+++ b/server/api/reports.py
@@ -1138,8 +1138,16 @@ def sampled_batch_rows(election: Election, jurisdiction: Jurisdiction | None = N
                 f"Combines {', '.join(sub_batch.name for sub_batch in sorted(sub_batches, key=lambda batch: batch.name))}",
             ]
             rows.append(combined_batch_row)
-
-    totals_row = ["Totals", "", sum(batch.num_ballots for batch in batches)]
+    total_ballots = sum(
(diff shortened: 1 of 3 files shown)


arsalansufi (approved):
LGTM!

> As noted in the ticket, should we remove the row entries for the batches that were selected but then combined from the audit report?

Overall, I do still think that this would make sense, just so that the total really is the sum of the numbers above. Can of course confirm with others and open a follow-up PR next week. At least the number is accurate now!

nikhilb4a:
Agreed, I'll keep the parent ticket open for now to make sure it doesn't slip by

PR #2328, opened 2026-05-29
Combined batches: Fix updating combined batch tallies

nikhilb4a:
image (1)

We saw an error when updating combined batch tallies, because the zero tally sheets we create for the non-representative batches were not deleted before we attempted to recreate them. This PR fixes this by deleting them first.

### DIFF
--- a/server/api/batches.py
+++ b/server/api/batches.py
@@ -315,6 +315,11 @@ def record_batch_results(
             Batch.jurisdiction_id == jurisdiction.id,
             Batch.id != batch.id,
         )
+        BatchResultTallySheet.query.filter(
+            BatchResultTallySheet.batch_id.in_(
+                non_representative_sub_batches.with_entities(Batch.id)
+            )
+        ).delete(synchronize_session="fetch")
         for sub_batch in non_representative_sub_batches:
             sub_batch.result_tally_sheets = [
                 BatchResultTallySheet(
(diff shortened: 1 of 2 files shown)


server/api/batches.py
```diff
@@ -315,6 +315,11 @@ def record_batch_results(
             Batch.jurisdiction_id == jurisdiction.id,
             Batch.id != batch.id,
         )
+        BatchResultTallySheet.query.filter(
```
arsalansufi, line 319: A comment along the lines of `# Delete existing sub batch tallies if any` could be useful here

server/tests/batch_comparison/test_batch_comparison.py
```diff
@@ -1020,7 +1020,15 @@ def test_batch_comparison_combined_batches(
         ]
     }
 
-    # Audit the combined batch
+    # Make two requests: a first put request without discrepancies, and a second
```
arsalansufi, line 1023: I think the word we want here isn't "discrepancies" but "tallies"

arsalansufi (approved):
Some quick nits, otherwise LGTM!

server/tests/batch_comparison/test_batch_comparison.py
nikhilb4a: Both requests have tallies but the difference in the tallies is that the discrepancies are added (lines 1040-1045)

arsalansufi: Oh gotcha, makes sense!

PR #2330, opened 2026-06-01
Batch comparison: Count all extra batches when counting sampled batches

nikhilb4a:
Ginny noticed recently that the unique sampled batch count was higher than when the toggle was off for some jurisdictions. A good example was Appling, which has only two Extra batches selected. Count was 2 when unique, and 1 when total. The non-unique/total case was off because the query was counting distinct ticket numbers, but the extra batches share a ticket number of "EXTRA".

Toggle on -
Screenshot 2026-06-01 at 4 41 31 PM

Before patch - 
Screenshot 2026-06-01 at 4 41 25 PM

After patch - 
Screenshot 2026-06-01 at 4 41 11 PM

### DIFF
--- a/server/api/jurisdictions.py
+++ b/server/api/jurisdictions.py
@@ -6,7 +6,7 @@ import uuid
 import datetime
 import math
 from flask import jsonify, request
-from sqlalchemy import func
+from sqlalchemy import distinct, func, tuple_
 from sqlalchemy.orm import joinedload
 from werkzeug.exceptions import Conflict, BadRequest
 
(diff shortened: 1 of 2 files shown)


server/api/jurisdictions.py
```diff
@@ -454,7 +454,16 @@ def batch_round_status(election: Election, round: Round) -> dict[str, JSONDict]:
         .join(Batch)
         .group_by(Batch.jurisdiction_id)
         .values(
-            Batch.jurisdiction_id, func.count(SampledBatchDraw.ticket_number.distinct())
+            Batch.jurisdiction_id,
+            func.count(
+                distinct(
+                    tuple_(
+                        SampledBatchDraw.batch_id,
+                        SampledBatchDraw.contest_id,
```
nikhilb4a, line 462: I was not certain if I should include `contestId` in here. I don't think it's so important, only for the case where the same extra batch was chosen for multiple contests. I _think_ that's a real potential edge case, but could use a sanity check.

jonahkagan: I think I agree with your analysis, so we should include it

jonahkagan (approved):

PR #2336, opened 2026-06-04
Batch comparison: Sort combined batches within sampled batches in reports

### LINKED ISSUE #2335: Batch comparison: Put combined batches in same sort order as the rest of the batches (by jurisdiction name, etc.)
**Problem**

Combined batches are all listed at the bottom of the report

**Solution**

Intersperse them properly with sampled batches

Issue automatically created from Slack feedback reported by nikhil: 
https://votingworks.slack.com/archives/C0B5C56T15G/p1780440748390289?thread_ts=1780438602.166029&cid=C0B5C56T15G

nikhilb4a:
Closes https://github.com/votingworks/arlo/issues/2335

[Ginny](https://votingworks.slack.com/archives/C0B5C56T15G/p1780440748390289?thread_ts=1780438602.166029&cid=C0B5C56T15G) had to manually move combined batch rows in the audit report to sort them within sampled batch rows. This PR aims to automate that.

Screenshot showing combined batch interspersed with sampled batch rows:
Screenshot 2026-06-03 at 4 54 23 PM

### DIFF
--- a/server/api/reports.py
+++ b/server/api/reports.py
@@ -1,4 +1,5 @@
 import io
+import re
 import csv
 from typing import cast as typing_cast
 from collections import defaultdict, Counter
@@ -201,6 +202,15 @@ def pretty_cvr_interpretation(
         for choice_id, interpretation in cvrs_by_choice.items()
         if interpretation == "1"
     )
+
+
+def natural_sort_key(name: str) -> list:
+    # Mirror the human_sort Postgres function used to order batches in queries,
+    # so e.g. "Batch 2" sorts before "Batch 10".
+    return [
+        int(chunk) if chunk.isdigit() else chunk.lower()
+        for chunk in re.split(r"(\d+)", name)
+    ]
 
 
 def add_sign(value: int) -> str:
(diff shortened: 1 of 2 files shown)


nikhilb4a:
Added @arsalansufi as a reviewer since I probably got this out too late for @jonahkagan to review with him ooo tomorrow

server/api/reports.py
```diff
@@ -203,6 +204,15 @@ def pretty_cvr_interpretation(
     )
 
 
+def natural_sort_key(name: str) -> list:
```
nikhilb4a: There is also a library that has a natural sort function `natsorted from natsort` if you'd recommend that over defining this here. 

Claude created this function and ran it through some tests + it keeps ordering consistent with our existing snapshots

jonahkagan: Is there any way we could do this at the database level so we don't have to duplicate the logic? Maybe when ordering the batches in the query we could use the combined batch name for sub-batches instead of their original batch names, then when we process the batch list, when we come across sub-batches, merge them into a combined batch.

nikhilb4a: That's a good push. I'll give that a try

nikhilb4a: Followed that [approach](https://github.com/votingworks/arlo/pull/2336/commits/86d0e8ad0463e6a7de5006ae0a81945e3bd034ce) instead, good call. Future improvement could be extracting some of the shared logic between the non-combined and combined batch rows

jonahkagan (approved):

PR #2337, opened 2026-06-09
Fix combined batch selector

nikhilb4a:
When there was a really long list, the select would take the full browser height, covering up the input area.

Future improvement under consideration:
- allow for selecting multiple batches based on input (don't clear input after selection)

Before:
Screenshot 2026-06-08 at 9 50 51 PM

After:

https://github.com/user-attachments/assets/3dd28874-1f99-4172-b38c-ddcb118f7dc1

### DIFF
--- a/client/src/components/SupportTools/Jurisdiction.tsx
+++ b/client/src/components/SupportTools/Jurisdiction.tsx
@@ -12,7 +12,7 @@ import {
   MenuItem,
   Menu,
 } from '@blueprintjs/core'
-import { MultiSelect } from '@blueprintjs/select'
+import { MultiSelect, renderFilteredItems } from '@blueprintjs/select'
 import { useForm, Controller } from 'react-hook-form'
 import {
   useJurisdiction,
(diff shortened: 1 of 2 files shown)


client/src/components/SupportTools/Jurisdiction.tsx
```diff
@@ -216,6 +217,22 @@ const Jurisdiction = ({ jurisdictionId }: { jurisdictionId: string }) => {
                               value.filter((id: string) => id !== item.id)
                             )
                           }}
+                          itemListRenderer={({
+                            filteredItems,
+                            itemsParentRef,
+                            renderItem,
+                          }) => (
+                            <Menu
+                              ulRef={
```
nikhilb4a, line 222: Without this ref attached, keyboard navigation wouldn't scroll the menu once the item was no longer visible

client/src/components/SupportTools/Jurisdiction.tsx
```diff
@@ -216,6 +217,22 @@ const Jurisdiction = ({ jurisdictionId }: { jurisdictionId: string }) => {
                               value.filter((id: string) => id !== item.id)
                             )
                           }}
+                          itemListRenderer={({
```
jonahkagan: Looks like there might be an undocumented `menuProps` prop that you could use to pass the style overrides to the underlying `Menu`. See https://github.com/palantir/blueprint/blob/develop/packages/docs-app/src/examples/select-examples/multiSelectExample.tsx#L290 (I'm assuming the only goal is to add a max height)

jonahkagan (approved):

client/src/components/SupportTools/Jurisdiction.tsx
nikhilb4a: Wasn't able to get that to work. Looks like that prop is on newer versions of `blueprint` than Arlo currently uses

nikhilb4a:
Screenshot 2026-06-09 at 12 23 29 PM

Updated the `itemListRendered` prop to make sure no results state was properly handled. Without this update, only a few pixel height `MenuItem` was rendered when there were no results.

Issue #2194, opened 2025-10-08
Error when creating a CSD elections


arsalansufi:
Error when creating a CSD elections.

1. I select all of the targeted contests
2. Select Save and Next
3. Select Save and Next on opportunistic contests (not adding any)
4. This toast pops up
Everything seems to save just fine though and when I go back and repeat 2-3 (no need to select targeted contests, they already are), no toast.

Issue automatically created from Slack feedback reported by ginny: 
https://votingworks.slack.com/archives/C01F2CH62TS/p1755088230354989?thread_ts=1755088230.354989&cid=C01F2CH62TS

nikhilb4a:
Pretty certain it was solved with: https://github.com/votingworks/arlo/pull/2338

PR #2338, opened 2026-06-09
Audit setup: Await putContests response

nikhilb4a:
I couldn't reproduce the issue from this [ticket](https://github.com/votingworks/arlo/issues/2194), but I did notice a bug that might be the same cause. When updating target contests, and then navigating to opportunistic contests, the data showing on the opportunistic contests is always stale. 

It looks like there is a race condition from the `PutContests` that happens when setting target contests, so that's why it's hard to recreate on PUT creates. But on PUT updates, a bit more happens on the backend, so maybe that's why I can reliably recreate it.

Old

https://github.com/user-attachments/assets/8ee5618c-03d5-4340-890a-22e2164ec4af

New

https://github.com/user-attachments/assets/fcfa4e7e-58be-42a0-969f-87235eeeec26

### DIFF
--- a/client/cypress/end-to-end/ballot-comparison.cy.js
+++ b/client/cypress/end-to-end/ballot-comparison.cy.js
@@ -94,8 +94,6 @@ describe('Ballot Comparison Test Cases', () => {
     cy.findByText('Save & Next').click()
     cy.findAndCloseToast('Must have at least one targeted contest')
 
-    cy.findByText('Back').click()
-
     // select targeted contest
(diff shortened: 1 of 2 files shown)


jonahkagan (approved):
Nice catch. Might be worth doing a check for similar errors in other functions

nikhilb4a:
Good call. Had Claude _Fable_ help with a review of other `api()` calls and other steps in the audit setup flow, and found no other similar bugs

Issue #2324, opened 2026-05-28
Batch comparison: Handle interaction between extra batch selection and multi-contest batch audits


arsalansufi:
**Problem:** 

I'm wondering if the GA Extra batch selection is being done by contest and therefore there are a few more batches than necessary selected (Blake and I both think it's negligible and not something to correct today).

That is very likely the case

A good example of the tricky ways in which Arlo features interact (in this case, extra batch selection and multi-contest batch audits)

**Solution:**

Need to understand the issue better, but it seems like we should only assign extra batches once per audit rather than per audit-contest.

- We select batches for contest 1
- We select batches for contest 2
- We then select batches on top of that for extra batch selection to meet the required heuristics. That might mean that some jurisdictions need no extra batches because the combo of contest 1 and contest 2 batches has the required heuristics covered.

Issue automatically created from Slack feedback reported by ginny: 
https://votingworks.slack.com/archives/C013V6CT9H9/p1779991822966629?thread_ts=1779991822.966629&cid=C013V6CT9H9

PR #2346, opened 2026-06-16
Batch comparison: Add extra batches once per round

nikhilb4a:
https://github.com/votingworks/arlo/issues/2324

Georgia noticed in their recent audit that more extra batches were selected than needed, and this was because extra batches where being added once per contest instead of once per round. Batches audited for one contest are already audited for the other contest, so we don't need extra batches added for each.

This PR makes a low-lift effort to refactor the extra batch selection to make it once per round instead of per contest.

This [subsequent PR](https://github.com/votingworks/arlo/pull/2344) suggests data model changes to move extra batches out of the experimental state and into official support, but given we are close to the audit day (Thursday), I think it's safer to just go with the lower lift change for now. 

Testing: 

Will do a manual test with the GA extra batch flag turned on for multi-contest.
Will add a multi-contest extra batch test.

### DIFF
--- a/server/api/shared.py
+++ b/server/api/shared.py
@@ -217,8 +217,16 @@ def sampled_batch_results(
     # and audit reports
     if not include_non_rla_batches:
         extra_batch_keys = set(
-            SampledBatchDraw.query.filter_by(
-                contest_id=contest.id, ticket_number=EXTRA_TICKET_NUMBER
+            SampledBatchDraw.query.filter_by(ticket_number=EXTRA_TICKET_NUMBER)
+            .join(Batch)
+            .join(Jurisdiction)
+            .filter(Jurisdiction.election_id == contest.election_id)
+            .values(Jurisdiction.name, Batch.name)
+        )
+        rla_sampled_batch_keys = set(
+            SampledBatchDraw.query.filter(
+                SampledBatchDraw.contest_id == contest.id,
+                SampledBatchDraw.ticket_number != EXTRA_TICKET_NUMBER,
             )
             .join(Batch)
             .join(Jurisdiction)
@@ -227,7 +235,7 @@ def sampled_batch_results(
         results = {
             batch_key: result
             for batch_key, result in results.items()
-            if batch_key not in extra_batch_keys
+            if batch_key not in (extra_batch_keys - rla_sampled_batch_keys)
         }
 
     return results
--- a/server/api/shared.py
+++ b/server/api/shared.py
@@ -635,7 +635,6 @@ class BatchDraw(TypedDict):
 
 def compute_sample_batches_for_contest(
     election: Election,
-    round_num: int,
     contest: Contest,
     contest_sample_size: SampleSize,
 ) -> list[BatchDraw]:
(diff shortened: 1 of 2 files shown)


nikhilb4a:
@jonahkagan I put together this simpler fix for this ticket, as my other PR (mentioned in PR description) ended up being having more surface area than I expected, and I don't feel as confident about getting it done by Thursday. Can you take a look if this approach seems reasonable for now, and then I'll finalize the logic/testing?

jonahkagan (approved):
Approach seems reasonable to me. The comment from copilot seems worth looking into

server/api/shared.py
```diff
@@ -681,12 +680,48 @@ def compute_sample_batches_for_contest(
...
+
+    extra_batches: list[BatchDraw] = []
+
     if is_enabled_sample_extra_batches_by_counting_group(election) and round_num == 1:
         rand = random.Random(str(election.random_seed))
-        for jurisdiction in contest.jurisdictions:
+        for jurisdiction in jurisdictions:
+            representative_contest_id = jurisdiction_id_to_contest_id[jurisdiction.id]
             batch_key_to_num_ballots = {
                 (jurisdiction.name, batch.name): batch.num_ballots
```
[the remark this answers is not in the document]
nikhilb4a: @jonahkagan I think the comment is correct, though it seems to predate this PR. sampled_batch_results() gets all batches which have batch tally sheets for this contest, and then when excluding non-rla batches, it only filters out extra batches that were tagged for this contest. But if contest B pulls in an extra batch, there will be a tally sheet with contest A for that batch, which ends up included in the sampled_rows. 

I think this [fixes](https://github.com/votingworks/arlo/pull/2346/commits/e2d6e1bf098671c9b57cf24f1d08fbf833b4164b) it, by removing all extra batches except any that also live in the rla set (if a batch gets chosen as extra in first round, but part of sample in second round)

server/api/shared.py
```diff
@@ -227,7 +235,7 @@ def sampled_batch_results(
         results = {
             batch_key: result
             for batch_key, result in results.items()
-            if batch_key not in extra_batch_keys
+            if batch_key not in (extra_batch_keys - rla_sampled_batch_keys)
```
jonahkagan, line 238: Why do we need to subtract out `rla_sampled_batch_keys` here? I didn't think it was possible for an RLA-sampled batch to also be used as an extra batch

nikhilb4a: That was considering a multi-round audit that has a batch selected as an extra in round 1, but then part of the sample in round 2, as I think that sequencing can happen

jonahkagan: Oh, interesting edge case. Cool

nikhilb4a:
Did a manual test with the GA election from May and saw intended results. In my local snapshot, I cleared the batch tallies and undid the round start. Upon redrawing, every jurisdiction saw their number of extra batches stay the same or go down. Looked into some jurisdictions with the largest changes (along with some Claude investigation) and they were correctly sampled.

Feeling confident enough to merge.

PR #2358, opened 2026-07-14
Client: Wrap sheet names in batch tally entry

### LINKED ISSUE #2106: From a GA user's feedback survey: When you are dealing with batch totals in the thousands, we break up the batch into multiple tally sheets for the ease of the audit teams
From a GA user's feedback survey: When you are dealing with batch totals in the thousands, we break up the batch into multiple tally sheets for the ease of the audit teams. Adding more tally sheets to arlo is fine, but the more you add, the harder it is to visually operate the system as the table expands horizontally. If the table expanded vertically instead, as you add more rows, it would be much easier to operate with batch totals that are above 1000.

Issue automatically created from Slack feedback reported by ginny: 
https://votingworks.slack.com/archives/C01F2CH62TS/p1734116618845899?thread_ts=1734116618.845899&cid=C01F2CH62TS

nikhilb4a:
Closes https://github.com/votingworks/arlo/issues/2106

Jurisdictions with a large batch split across many tally sheets are finding the horizontal scrolling hard to deal with.

Since we have an audit ongoing, this UI change is minimal. It starts wrapping sheet names to new rows. The screen recording looks a bit jumpy because the sheet names are consistently shorter than the input width.

**Updated**

https://github.com/user-attachments/assets/0083e7e6-2dd6-48ac-906b-0bf426c150c9

**Before**

https://github.com/user-attachments/assets/8d1bb2e6-3a23-4b14-b3a5-8a3a21cdbba6

### DIFF
--- a/client/src/components/JurisdictionAdmin/BatchRoundSteps/BatchRoundTallyEntry/BatchDetail.tsx
+++ b/client/src/components/JurisdictionAdmin/BatchRoundSteps/BatchRoundTallyEntry/BatchDetail.tsx
@@ -27,10 +27,16 @@ import { sum } from '../../../../utils/number'
 
 const BatchName = styled(H4)`
   &.${Classes.HEADING} {
-    // Match the height of the batch search bar such that the batch name and search bar are
-    // vertically middle aligned
+    /* Match the height of the batch search bar such that the batch name and search bar are
+       vertically middle aligned */
     line-height: 30px;
   }
+`
+
+const BatchNameRow = styled.div`
+  align-items: flex-start;
+  display: flex;
+  justify-content: space-between;
 `
 
 interface ITabsWrapperProps {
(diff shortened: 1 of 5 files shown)


client/src/components/JurisdictionAdmin/BatchRoundSteps/BatchRoundTallyEntry/BatchDetail.tsx
```diff
@@ -27,21 +27,32 @@ import { sum } from '../../../../utils/number'
 
 const BatchName = styled(H4)`
   &.${Classes.HEADING} {
-    // Match the height of the batch search bar such that the batch name and search bar are
-    // vertically middle aligned
+    /* Match the height of the batch search bar such that the batch name and search bar are
```
nikhilb4a, line 30: Stylelint required some changes to the css in this file

client/src/components/JurisdictionAdmin/BatchRoundSteps/BatchRoundTallyEntry/BatchDetail.tsx
```diff
@@ -312,7 +328,19 @@ const BatchDetail: React.FC<IBatchDetailProps> = ({
 
   return (
     <Detail>
-      <BatchName>{batch.name}</BatchName>
+      <BatchNameRow>
+        <BatchName>{batch.name}</BatchName>
+        {sheets.length > 1 && (
+          <AddSheetButton
```
nikhilb4a, line 334: I moved it from the the tab name row because it felt odd having it wrap. I kept it top-right aligned so support doc screenshots would have this in a generally similar location. Could see an argument for moving it elsewhere though

jonahkagan (approved):
Looks great! Much better than the horizontal scroll

PR #2359, opened 2026-07-14
Server: Reject ballot manifest if 1+ batch has 0 ballots

### LINKED ISSUE #2299: *Problem:* 
A PA county (at least one) included many batches in their manifest with 0 ballots in them
*Problem:* 
A PA county (at least one) included many batches in their manifest with 0 ballots in them.

*From:* Ginny

*Proposed solution:* 
Have Arlo check for 0 ballot batches and tell the user to remove.

_Collected by @ginny_

Issue automatically created from Slack feedback reported by Product Feedback: 
https://votingworks.slack.com/archives/C01F2CH62TS/p1763065623559299

nikhilb4a:
Closes https://github.com/votingworks/arlo/issues/2299

Some of the tools states use to produce their ballot manifests generate batches with 0 ballots. These batches could be selected during extra batch selection, though not RLA sample selection. We could ignore them during extra batch selection, but Ginny advised we should reject ballot manifests with these rows. 

**0 ballot batch**
Screenshot 2026-07-14 at 3 05 53 PM

**Negative ballot batch**
Screenshot 2026-07-14 at 3 56 12 PM

### DIFF
--- a/server/api/ballot_manifest.py
+++ b/server/api/ballot_manifest.py
@@ -178,7 +178,7 @@ def process_ballot_manifest_file(
                         f'Invalid value for column "Container", row {row_index + 2}: "{counting_group}". Use the Batch Audit File Preparation Tool to create your ballot manifest, or correct this value to one of the following: {", ".join(counting_group_allowlist)}.'
                     )
 
-            if row[NUMBER_OF_BALLOTS] < 1:
+            if row[NUMBER_OF_BALLOTS] == 0:
                 rows_without_ballots.append(row_index + 2)
                 continue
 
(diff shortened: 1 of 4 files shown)


server/api/ballot_manifest.py
```diff
@@ -178,7 +178,7 @@ def process() -> None:
                         f'Invalid value for column "Container", row {row_index + 2}: "{counting_group}". Use the Batch Audit File Preparation Tool to create your ballot manifest, or correct this value to one of the following: {", ".join(counting_group_allowlist)}.'
                     )
 
-            if row[NUMBER_OF_BALLOTS] < 1:
+            if row[NUMBER_OF_BALLOTS] == 0:
```
nikhilb4a, line 181: I initially checked for `< 1` here, but it was making the error message a little awkward. Claude suggested rejecting negative numbers at the parser level as well since none of our files want negative numbers, so I went with that, but open to input

jonahkagan (approved):

PR #2368, opened 2026-08-05
Client: Add Falls Church and Virginia Beach to Virginia map corrections

### LINKED ISSUE #1810: Virginia map missing two small cities
From Virginia folks:

> When I was looking at the Virginia map in Arlo, I noticed it was missing Falls Church City and Manassas Park City, which is different from Manassas City. They are both small cities in the northern part of the Commonwealth.

nikhilb4a:
The map topology has polygons for both cities, but without corrections entries their names never match jurisdictions named with the City suffix.

Closes #1810


arsalansufi (approved):
LGTM!

PR #2380, opened 2026-08-11
Server: Require election/jurisdiction authorization for file uploads and cap upload size

nikhilb4a:
### Summary

1. Updates the `file-upload` api to validate the provided storage key and ensure the logged in user has access to the election/jurisdiction. An alternative implementation could rely on the `restrict_access` decorator but that change was ballooning a bit, so I went with this simpler version that calls `check_access` directly.
2. Adds a 5gb file size upload limit, matching the s3 client default.

### Details

On filesystem-storage deployments, `POST /api/file-upload` (step 2 of the 3 step upload process) — previously required only a logged-in session of any type, and wrote the client-supplied storage key as provided. S3-based deployments are unaffected (uploads go browser → S3 via presigned POST, and S3 itself caps single uploads at 5 GB). This brings the endpoint in line with the access control that the upload-URL (step 1) and upload-complete (step 3) steps already enforce:

- Key validation — the storage key must have exactly the shape issued by the upload-URL endpoints, `audits/<election_id>[/jurisdictions/<jurisdiction_id>]/<file_name>`: with ./.. segments and nested paths rejected.
- Authorization — the caller must pass the same `check_access` used by `restrict_access` for the election/jurisdiction named in the key: audit admins (of the election's org) for election-level files; audit admins or that jurisdiction's admins for jurisdiction files. The jurisdiction must belong to the keyed election, so an admin of a jurisdiction in one election can't write under another election's path.
- Size cap — request bodies are capped at 5 GB (configurable via `ARLO_MAX_UPLOAD_FILE_SIZE_BYTES`), matching the ceiling S3 already imposes.

### Test plan

Tests that did not pass the storage key requirements are updated.
New tests added to cover authorization, upload size cap.

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -24354,16 +24354,16 @@
             {
                 "code": "reportUnknownVariableType",
                 "range": {
-                    "startColumn": 32,
-                    "endColumn": 51,
+                    "startColumn": 4,
+                    "endColumn": 23,
                     "lineCount": 1
                 }
             },
             {
                 "code": "reportUnknownVariableType",
                 "range": {
-                    "startColumn": 53,
-                    "endColumn": 84,
+                    "startColumn": 4,
+                    "endColumn": 35,
                     "lineCount": 1
                 }
             },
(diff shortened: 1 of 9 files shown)


benadida (approved):

arsalansufi (approved):
LGTM!

PR #2382, opened 2026-08-13
Server: Improve behavior of zip filename handling

nikhilb4a:
### Summary

Return actual extracted file paths from ZIP upload helpers

### Description

`unzip_files()` extracted uploaded ZIPs with `extractall()` but returned the original archive's raw entry names rather than the names of the files extraction actually wrote. The two can differ, since extraction normalizes entry names before writing. Callers rebuild paths from the returned names to re-open the extracted files, so a mismatch leaves them pointing at paths that were never written.

`unzip_files()` now extracts entry-by-entry and returns the paths that were actually written, relative to the extraction directory.

`read_zip_filenames()`, which lists entry names without extracting, now rejects names that wouldn't correspond to a file inside the extraction directory.

### Testing
Added tests covering nested folders and irregular entry names in both helpers.

### DIFF
--- a/server/tests/util/test_file_storage.py
+++ b/server/tests/util/test_file_storage.py
@@ -5,6 +5,7 @@ import shutil
 import tempfile
 import io
 from unittest.mock import patch
+from zipfile import ZipFile
 from werkzeug.exceptions import BadRequest
 import pytest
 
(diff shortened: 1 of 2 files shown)


benadida (approved):
review of high level logic and tests looks good!

arsalansufi (approved):
LGTM!

PR #2383, opened 2026-08-13
Server: Pin math dependencies to specific commits

nikhilb4a:
### Summary

1. Pin math dependencies to specific commits - `athena`, `consistent-sampler`, and `r2b2`
2. Drop unused `cryptorandom` dependency
3. `noauth` was the last dependency that wasn't pinned to a specific commit, so I decided to pin that too, even though we maintain it, since this PR already pinned `consistent-sampler`. Can drop this if preferred.


benadida (approved):
I like pinning things to either version numbers or github commits, just good defense in depth.

arsalansufi (approved):
LGTM!

PR #2388, opened 2026-08-25
Server: Sort batch keys before sampling

nikhilb4a:
## Overview

Sort batch keys before sampling so sampling is independent of uploaded file batch ordering.

## Testing

Snapshots updated as now samples are slightly different.


arsalansufi (approved):
LGTM!

PR #2403, opened 2026-09-09
Server: Update ppeb sampling to support nested contest samples for improved multi-contest efficiency 

nikhilb4a:
### Overview

This PR adds a sampler function that allows nesting of a child sample within a parent sample, maintaining batch selection probabilities while achieving maximal overlap.

### Commits

I recommend reviewing commit by commit

1. Rewrite ppeb sampling to not use the generator.choice method, so we can eventually keep track of offsets that are required for contest nesting. This commit extracts the logic from the `numpy` method and calls it inline, as the `numpy` method hides the offset which we need to achieve nesting. Code should look more or less identical.
2. Return each draw's offset within its batch from draw_ppeb_positions to then be used within nested sampling. This commit actually exposes the offset made available from commit 1.
3. Add draw_nested_ppeb_positions to derive one contest's draws from another's. This commit is the heavy lifting for this PR. It is the updated math sampling code that actually achieves nesting.
4. Tests for 3
5. Add draw_nested_ppeb_samples to draw every contest's sample together. This is the second significant commit of this PR. It uses the new function from 3 in the sampling path, though with each contest sample built without any nesting relationship. All existing batch comparison snapshots have the same output, confirming preexisting behavior is maintainted. 
6. Tests for 5

The next PR will introduce the ability to set nesting relationships, and then will actually nest samples when the relationship is set.


server/audit_math/sampler.py
```diff
@@ -122,6 +122,94 @@ def draw_ppeb_positions(
     return sampled_batch_indexes.tolist(), offsets.tolist()
 
 
+def draw_nested_ppeb_positions(
+    parent_sampled_batch_indexes: list[int],
+    parent_offsets: list[float],
+    parent_weights: list[float],
+    child_weights: list[float],
+) -> tuple[list[int], list[float]]:
```
kshen0, line 133: Discussed during sync review: wonder if it's beneficial to use `decimal` here to avoid float arithmetic imprecision. See https://docs.python.org/3/library/decimal.html

kshen0 (approved):

MarkVVaudits (approved):
This looks good to go.

server/audit_math/sampler.py
nikhilb4a: Discussed with Claude and I am leaning towards leaning as is as I tend to agree with it's points. Lmk what you think 

> The border case can't be made to go away, only moved. The random draws themselves are float64 values from the generator, so there is always some representable number that sits right at a boundary. Whichever arithmetic you compare it with, some draw lands within one unit of rounding of an edge. Decimal changes which draws those are; it doesn't eliminate them.

> And a border case isn't an error. If a draw sits within about 1e-16 of the boundary between batch A and batch B, either answer is a correct sample. The random number is uniform, so the chance of landing there is on the order of 1e-16 per draw, and whichever batch gets picked was picked with probability that differs from its true weight by that same 1e-16. No auditor could measure that. The one thing that matters is that the outcome is deterministic and reproducible from the seed, which floats give us: same seed, same doubles, same comparisons, every time, on every machine, because IEEE 754 arithmetic on these operations is exact-rounding and platform-independent.

> Where imprecision could actually bite, we already guard. In the nesting function, redirect_p could round up to exactly total_excess and index one past the last batch. That's the min(..., len(excess) - 1) clamp with its comment. That's the honest response to float rounding here: catch the one case where it could crash, accept the immeasurable cases where it picks a neighbor.

Issue #2394, opened 2026-09-03
GA: Add pair-nesting support to increase contest batch selection overlap


nikhilb4a:
From Mark:
> I think it should be pretty easy to refine and port the sampling "guts" of my nesting code to Arlo and to provide nesting as an option [behind the scenes, not in UI]. Being able to optimize the "reuse" of audit batches over multiple contests could save some counties substantial pain, especially if two statewide target contests are rather close.

Latest update: 
https://votingworks.slack.com/archives/C07PT9ZF1PF/p1788466713824179

PR #2406, opened 2026-09-18
Server: Extract batch comparison sampling code into smaller functions

nikhilb4a:
This is first of a few PRs that will come for https://github.com/votingworks/arlo/issues/2394. I'm splitting them up in hopes of making it easier to review.

This PR has no behavior change, and no real code change, aside from extracting code from a large function that does sampling into three smaller functions. Eventually, the purpose of this will become clearer, because the functions will no longer always be called with the same parameters when sample nesting is introduced. As of this PR, there is no nesting code added, so they are all called with the same parameters as before.

As a preface, sample nesting is where we get as much overlap as possible between a parent sample for contest A and child contest for sample B, while maintaining their independent batch sampling probabilities. This will be helpful for GA's 20 contest audit, to allow them to sample fewer total number of batches while still auditing all contests.


kshen0 (approved):
Simple code move LGTM

PR #2407, opened 2026-09-18
Server: Update edge case handling for when no batches have a targeted contest

nikhilb4a:
When working on the nested sampling code with Claude, I noticed this edge case that was leading to some unnecessarily complex defensive handling. 

The edge case in question is when we are trying to generate a sample for a contest that has no uploaded results. We have defensively handled this case by returning an empty sample, but it starts to become more complicated when we have to prevent against this in nested samples, because then we need to check if either a parent or child meets this case. Instead, let's just assert against it in this sampling function. Claude also confirmed that this input can't reach the sampler in practice, so an assert is accurate here.

> 
> That input can't reach the sampler in practice: sample size computation divides by U first, and batch tallies are validated before a round starts. Failing loudly matches the sampler's other preconditions.
> 
> The old behavior wasn't actually safe: a round with an empty sample for a targeted contest is a stuck audit, since the contest can never make progress. The change swaps a quiet failure for a loud one on an input nothing produces.
>


nikhilb4a:
Whoops, didn't mean to merge this. Will reopen

kshen0 (commented):
LGTM

Just making sure I understand - the motivation is that the caller of `draw_ppeb_sample` must now handle the case where a parent or child's sample value is `[]`? Before nested samples we would only need to check for `[]` but now we have to traverse and check for `[]`, so it's not worth it considering we never expect to hit this assertion?

